PR #264, opened 2014-07-26
Sorts the QM files into seperate folders by method

pierrelb:
Currently, if you keep a standard folder and change the method used from pm3 to pm6 (as an example), you may be using old pm3 calculations when you intend to use pm6. These changes create a new subfolder, named after the method, within your QM folder. In the case that someone would want to use old pm3 calculations but run new ones at pm6, they could copy the pm3 files to a pm6 folder. This avoids overwriting possibly useful files that used a different method.

Additionally this also contains some changes to the logging messages and storing of the qm thermo data. We weren't storing comments about where the thermo data came from when using Gaussian, but we also weren't logging some thing on the MOPAC side. Things should now be consistent.


connie:
Can you please rebase to official/master when you can? And check my pull request to your branch.  Thanks!

connie:
These changes seem to be only creating a new folder with the method name under the QMscratch folder, not the `fileStore` folder, where the actual storage of old files are... So I don't think it is doing what we want.  The QMscratch folder appears to be empty most of the time also. What is it being used for?

pierrelb:
The `fileStore` has been corrected. Problem was in the `setDefaultOutputDirectory` method (main.py).

The `QMscratch` folder has never been used, but should be. We should be running the qm jobs from the `scratchDirectory` and storing the result (`.thermo` file) in `fileStore`.

pierrelb:
I think this should address #355, #358, and #360. Any thoughts?

connie:
In terms of https://github.com/GreenGroup/RMG-Py/issues/360, do we want to retain the scratch files in the folder by default?

connie:
In terms of https://github.com/GreenGroup/RMG-Py/issues/355, the sorting of QM files by method.  I don't know if the unit tests are checking this correctly.  I just tried to run the unit tests.  The first run is completely error free.  However the second time around I get the following errror:

FAIL: Test that getThermoData() works correctly.
  File "/files/RMG-Py/rmgpy/qm/mainTest.py", line 250, in testGetThermoDataMopac
    self.assertAlmostEqual(thermo2.H298.value_si, 167704.4270, 1) # to 1 decimal place
AssertionError: 167704.34319999997 != 167704.427 within 1 places
-------------------- >> begin captured logging << --------------------
root: INFO: Setting the quantum mechanics scratchDirectory to /files/RMG-Py/testing/qm/QMscratch
root: INFO: Setting the quantum mechanics fileStore to /files/RMG-Py/testing/qm/QMfiles
root: INFO: Setting the quantum mechanics scratchDirectory to /files/RMG-Py/testing/qm/QMscratch
root: INFO: Setting the quantum mechanics scratchDirectory to /files/RMG-Py/testing/qm/QMscratch
root: INFO: Reading existing thermo file /files/RMG-Py/testing/qm/QMfiles/Species/c1ccc2ccccc2c1/pm3/c1ccc2ccccc2c1.thermo
root: INFO: Reading existing thermo file /files/RMG-Py/testing/qm/QMfiles/Species/c1ccc2ccccc2c1/pm6/c1ccc2ccccc2c1.thermo
root: INFO: Reading existing thermo file /files/RMG-Py/testing/qm/QMfiles/Species/c1ccc2ccccc2c1/pm7/c1ccc2ccccc2c1.thermo

connie:
Just discussed issue https://github.com/GreenGroup/RMG-Py/issues/358 with Richard, I think we agreed that we need to actually do HBI method on the saturated library value even for cases that aren't associated with QM.  Basically, noncyclics or jobs that don't run QM should also default to library saturated value, rather than group additivity saturated value.  Because you could end up with the case where you have a very good library saturated value, and a completely uncorrelated value for the radical based off an inaccurate group additivity value.

connie:
We should also store the attempt information (or perhaps even the keywords inputted to the program), when we find thermo from a previous run.  Currently it says " QM MopacMolPM3 calculation\
 found from previous run." but we lose the attempt information

connie:
Also should we distinguish between different results that come from different programs??

PR #323, opened 2015-01-05
Adapting kinetics for with multiple transition states

connie:
Putting up this pull request so that we don't forget about this issue: https://github.com/GreenGroup/RMG-Py/issues/142

In certain families, there are multiple reaction paths that can occur, sometimes with different kinetics.  The example found in the issue is that of isobutyl radical + tertbutyl radical reacting to form isobutane and isobutene.  There are two possible paths in the Disproportionation family: (C_rad/H2/Cs,Cmethyl_Csrad) or (C_rad/Cs3,C/H/NdNd_Csrad).  What is currently found in Java is a randomly chosen version of the kinetics (Sometimes the first template appears, other times the 2nd template appears).

What these commits attempt to do is to calculate degeneracy by allowing identical reactions IF their reaction templates are different (https://github.com/GreenGroup/RMG-Py/commit/3aea3e5a5fe228f23166ec12e50cc4fbace2f847).  Then degeneracy is calculated separately for each 'transition state' based on bookkeeping of the reaction pairs.  

Remaining issues:
- Is matching reaction pairs sufficient?  It works in the case of bimolecular reactions, but unimolecular reactions or recombination reactions  (these probably don't have multiple TS's to begin with) won't really be distinguished (see https://github.com/GreenGroup/RMG-Py/issues/140).  
- If kinetics are identical, they should probably be combined through degeneracy for simplicity's sake in the chemkin file and then have their diff templates written as comments. In this commit they remain as duplicates.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1221,16 +1221,27 @@ class KineticsFamily(Database):
             # for each reaction, make its reverse reaction and store in a 'reverse' attribute
             for rxn in reactionList:
                 reactions = self.__generateReactions(rxn.products, products=rxn.reactants, forward=True, failsSpeciesConstraints=failsSpeciesConstraints)
+                #assert len(reactions) == 1, "Expecting one matching reverse reaction, not {0}. Forward reaction {1!s} : {1!r}".format(len(reactions), rxn)
                 if len(reactions) != 1:
-                    logging.error("Expecting one matching reverse reaction, not {0} in reaction family {1} for forward reaction {2}.\n".format(len(reactions), self.label, str(rxn)))
-                    for reactant in rxn.reactants:
-                        logging.info("Reactant")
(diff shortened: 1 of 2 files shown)


connie:
Updated so that RMG uses the template of the reaction to determine whether the rxn has a different TS.  However, it is difficult to determine which reaction to pick for families that are their own reverse.  When there are two reverse reactions, which one should be picked?  Probably need to use some mapping of the atomlabels with hardcoding for those families..

The specific case that might be looked into is the reaction between these two species reacting in Intra_H_migration.  This is for future reference. 

species(
    label='A',
    reactive=True,
    structure=SMILES("[O]c1c[c]oc1"),
)

species(
    label='B',
    reactive=True,
    structure = SMILES("[O]c1[c]coc1"),
)

goldmanm:
This is on hold likely till #1055 gets merged

mliu49:
Replaced by #1055.

PR #324, opened 2015-01-06
Fix duplicate reaction handling in reaction libraries and seed mechanisms

connie:
These commits address the issue https://github.com/GreenGroup/RMG-Py/issues/276 where some libraries have duplicate reactions containing both a pdep reaction and a non-pdep reaction, the final mechanism from RMG will end up excluding one of them.  It also works when there are two duplicates of pdep reactions (for instance 2 chebyshevs, or literally any combination of reactions, as long as they are marked duplicate.)

This allows reactions to enter the core by changing the way checkForExistingReaction function works in rmgpy/rmg/model.py
It ignores marked duplicate reactions from reaction libraries and seed mechanisms.  

I've tested this with duplicate reactions in a library as a seed mechanism, a library, and in both library + seed mech.   I hope the pdep handling in the reaction library is ok.

### DIFF
--- a/rmgpy/rmg/model.py
+++ b/rmgpy/rmg/model.py
@@ -629,12 +629,12 @@ class CoreEdgeReactionModel:
         reactionList = []
         if speciesB is None:
             for moleculeA in speciesA.molecule:
-                reactionList.extend(database.kinetics.generateReactions([moleculeA], failsSpeciesConstraints=self.failsSpeciesConstraints))
+                reactionList.extend(database.kinetics.generateReactionsFromFamilies([moleculeA], products=None, failsSpeciesConstraints=self.failsSpeciesConstraints))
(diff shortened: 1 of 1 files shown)


connie:
Does anyone know the behavior of pdep reaction library rates and how they work with RMG-generated pdep networks?   Currently I've tested reaction libraries with some pdep rates in them when pdep is OFF, they seem to override RMG estimates.  What happens when pdep is ON?

rwest:
I think our aim was that the PDep rate from a Reaction Library would take precedence, would be assumed to be complete and trusted, and would not be added to by RMG.  How RMG would treat the rest of the unimolecular network that it is part of, I am not sure we ever resolved - I can't remember off the top of my head.  I'm not sure what's implemented.

PR #330, opened 2015-01-31
Fix up for reading in chemkin dup reactions

enochd:
importChemkinLibrary.py was failing for a chemkin file with duplicate
reactions that are written in opposite directions (i.e., A=B and B=A).


connie:
Closed via 160ad1b

PR #353, opened 2015-03-10
Sensitivity analysis

connie:
This pull request adds sensitivity analysis to reaction coefficients and thermo and saves them to csv files in the 'solver' folder.  The sensitivity analysis has been verified against simulations performed in CHEMKIN and RMG-Java

Notable changes:
- Switch to flexible DASx class which can default to DASSL if DASPK in PyDAS is not compiled.  However, sensitivity analysis cannot be performed if DASPK is not present.
- Automatic compilation in Makefile depending on whether DASPK is present
- Analytical jacobian fixes for 3 reactants or 3 products.
- Saving of csv files for sensitivities.  
- Standalone sensitivity module and example file
- Updates to documentation


rwest:
Looking good. Should we update the Mac / Windows instructions too?

connie:
Yea I think we should revise it for Mac.  The windows is not completely straightforward.  Even with MinGW the download script for the daspk31.tgz doesn't quite work due to missing programs like wget.  I guess I'll revise the instructions for windows to say manually download.

enochd:
I'm not sure if this is related, but for ubuntu, after pulling these and other changes, I am getting a make error:

`No PyDAS solvers found.  Please check if you have the latest version of PyDAS.
ImportError: No module named dassl
make: *** [solver] Error 1
`

I've made sure that i'm using the latest PyDAS version.

rwest:
Were you sure to `make` and `make install` the new pydas after you updated it?

enochd:
Yes, I was.

Issue #283, opened 2014-09-04
atomtypes module: keeping track of new atom types


nickvandewiele:
With the support of the new adjacency list format, new elements and a plethora of new atom types emerged as well.

the module`rmgpy.molecule.atomtype` does a number of things:
- it defines the supported atom types through the constructor of `AtomType(...)`, by providing 2 things:
  - the element symbol
  - relation to one another atom types: a list of **all** predecessors: `generic`, a list of **all** successors: `specific`
- it defines the conversion of one atom type to a sister atom type when an Action (`incrementBond, decrementBond, formBond, breakBond, incrementRadical, decrementRadical, incrementLonePair, decrementLonePair` ) is called upon the atom of this atom type.

E.g.: 

``` python
atomTypes['Cd'  ].setActions(
    incrementBond=['Cdd','Ct'],
    decrementBond=['Cs'],
    formBond=['Cd'],
    breakBond=['Cd'],
    incrementRadical=['Cd'],
    decrementRadical=['Cd'],
    incrementLonePair=[],
    decrementLonePair=[]
)
```
- it provides a method `getAtomType(atom, bonds)` that perceives the correct atom type for the given parameter atom surrounded by the parameter bonds.

``` python
...
    elif atom.symbol == 'C':
        if   double == 0 and doubleO == 0 and triple == 0 and benzene == 0: atomType = 'Cs'
        elif double == 1 and doubleO == 0 and triple == 0 and benzene == 0: atomType = 'Cd'
        elif double + doubleO == 2        and triple == 0 and benzene == 0: atomType = 'Cdd'
        elif double == 0 and doubleO == 0 and triple == 1 and benzene == 0: atomType = 'Ct'
        elif double == 0 and doubleO == 1 and triple == 0 and benzene == 0: atomType = 'CO'
        elif double == 0 and doubleO == 0 and triple == 0 and benzene == 2: atomType = 'Cb'
        elif double == 0 and doubleO == 0 and triple == 0 and benzene == 3: atomType = 'Cbf'
   elif atom.symbol == 'N':
...
```

Although this module seems to work fine and streamlined for fast comparison, the code itself seems:
- difficult to maintain: e.g. a variable `doubleO` counts the number of double bonds to oxygen, but what about CS (double bond to S), CN (triple bond to N), ... The above code will explode in terms of lines of code because of the combinatorial nature of the attributes to be checked.
[...]

PR #354, opened 2015-03-13
Aromatic Oxygen (and Nitrogen, for that matter)

rwest:
To try and address things like Furans, which should be aromatic, this topic branch (aromaticO) tries
to allow Ob atom types (and hopefully will enable the N3b and N5b atom types to function).
There is still some work to be done, but here is a start. Feel free to add commits to this branch and PR.
- [ ] check the LonePair settings are right on the Ob atom type
- [ ] add the Huckel rule check, in case someone crafts an adjacencylist with 7 Cb's in a cycle? Maybe redundant.
- [x] fix the GroupAtom bugs mentioned below
- [ ] add more, and more granular, unit tests

### DIFF
--- a/rmgpy/molecule/atomtype.py
+++ b/rmgpy/molecule/atomtype.py
@@ -228,7 +228,7 @@ atomTypes['N5dd'] = AtomType('N5dd', generic=['R','R!H','N','Val5'],  specific=[
 atomTypes['N5t' ] = AtomType('N5t',  generic=['R','R!H','N','Val5'],  specific=[])
 atomTypes['N5b' ] = AtomType('N5b',  generic=['R','R!H','N','Val5'],  specific=[])
 
-atomTypes['O'   ] = AtomType('O',    generic=['R','R!H','Val6'],      specific=['Os','Od','Oa','Ot'])
+atomTypes['O'   ] = AtomType('O',    generic=['R','R!H','Val6'],      specific=['Os','Od','Oa','Ot','Ob'])
 atomTypes['Os'  ] = AtomType('Os',   generic=['R','R!H','O','Val6'],  specific=[])
 atomTypes['Od'  ] = AtomType('Od',   generic=['R','R!H','O','Val6'],  specific=[])
 atomTypes['Oa'  ] = AtomType('Oa',   generic=['R','R!H','O','Val6'],  specific=[])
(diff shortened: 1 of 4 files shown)


rwest:
Currently failing two unit tests:
```python
FAIL: Test the GroupAtom.equivalent() method.
  File "/Users/rwest/Code/RMG-Py/rmgpy/molecule/groupTest.py", line 146, in testEquivalent
    self.assertTrue(atom1.equivalent(atom2), '{0!s} is not equivalent to {1!s}'.format(atom1, atom2))
AssertionError: ['Ob'] is not equivalent to ['O']

FAIL: Test the GroupAtom.isSpecificCaseOf() method.
  File "/Users/rwest/Code/RMG-Py/rmgpy/molecule/groupTest.py", line 174, in testIsSpecificCaseOf
    self.assertTrue(atom1.isSpecificCaseOf(atom2), '{0!s} is not a specific case of {1!s}'.format(atom1, atom2))
AssertionError: ['Ob'] is not a specific case of ['O']
```

connie:
Will it be ok to allow rings more than length 6 to become aromatic?  Do our group contributions account for that correctly?

Also, is this a good time to reorganize atomTypes to be more intuitive and more transparent?  We had talked about perhaps generating the list of atomTypes via a tree.  We could do this as part of the database loading step to initialize the atomTypes.  See previous issues https://github.com/GreenGroup/RMG-Py/issues/283 and https://github.com/GreenGroup/RMG-Py/issues/327

One thing I worry about is what the incrementBond, breakBond, loseLonePair, etc. attributes are used for.  Are they being used to generate groups for detecting the reverse reaction?  What kind of errors can writing these by hand lead to?

shamelmerchant:
Also the b in Cb stands for benzenic Ob as nomenclature might be confusing since it is furanic?

rwest:
I don't know what group contribution database will do. Maybe ring size should be capped at 6?

I agree that #283 is still a good idea.  Was @nickvandewiele working towards it in 75fb7697ebdf95539270b422bdc77b401262d367?
And yes, now may be a good time... But if we can make furan aromatic without having to refactor the entire AtomType mechanism, then I'd rather not wait.  If it turns out we can't, then someone will have to delve in and redo all the AtomTypes.

As for the naming convention...  I don't think there is one that won't cause some confusion. Should the carbon atoms in a furan ring still be called `Cb`? I agree that `Xa` for aromatic may be better than `Xb` for benzenic, but `Oa` is already taken for atomic...  I decided that for now, `b` just means aromatic! We can then check for aromatic rings by insisting every atom type in the ring ends in a 'b' (or 'bf'), for example.

mliu49:
Closing because this is extremely out of date. Aromatic atomtypes for oxygen and nitrogen have been implemented in #1206. However, aromatic resonance structure generation has not been extended to heteroatoms yet. There are other considerations which we will need to think about before taking that step.

Issue #342, opened 2015-02-19
Liquid Phase example does not work


connie:
Errored as follows:

```python
Generating thermodynamics for new species...
Found thermo for C in thermo library "corrected": primarythermolibrary
Error: Couldn't find in main abraham database:
Error: <Molecule "C">
Error: 1 C u0 p1 c0 {2,S} {3,S}
2 H u0 p0 c0 {1,S}
3 H u0 p0 c0 {1,S}

Traceback (most recent call last):
  File "../../../rmg.py", line 147, in <module>
    rmg.execute(args)
  File "/home/connie/code/RMG-Py/rmgpy/rmg/main.py", line 438, in execute
    self.initialize(args)
  File "/home/connie/code/RMG-Py/rmgpy/rmg/main.py", line 426, in initialize
    self.reactionModel.enlarge([spec for spec in self.initialSpecies if spec.reactive])
  File "/home/connie/code/RMG-Py/rmgpy/rmg/model.py", line 744, in enlarge
    spec.generateThermoData(database, quantumMechanics=self.quantumMechanics)
  File "/home/connie/code/RMG-Py/rmgpy/rmg/model.py", line 165, in generateThermoData
    return self.processThermoData(database, thermo0, thermoClass)
  File "/home/connie/code/RMG-Py/rmgpy/rmg/model.py", line 194, in processThermoData
    soluteData = database.solvation.getSoluteData(self)
  File "/home/connie/code/RMG-Py/rmgpy/data/solvation.py", line 602, in getSoluteData
    soluteData = self.getSoluteDataFromGroups(species)
  File "/home/connie/code/RMG-Py/rmgpy/data/solvation.py", line 662, in getSoluteDataFromGroups
    sdata = self.estimateSoluteViaGroupAdditivity(molecule)
  File "/home/connie/code/RMG-Py/rmgpy/data/solvation.py", line 757, in estimateSoluteViaGroupAdditivity
    self.__addGroupSoluteData(soluteData, self.groups['abraham'], molecule, {'*':atom})
  File "/home/connie/code/RMG-Py/rmgpy/data/solvation.py", line 791, in __addGroupSoluteData
    raise KeyError('Node has no parent with data in database.')
KeyError: 'Node has no parent with data in database.'
```

Update: duplicate of #325?

connie:
I think the problem is that the node the molecule hits is one of the top ones: 

    Css
    1 * Css u0

[...]

PR #364, opened 2015-03-30
Fix to solvation thermo for lone pairs

bslakman:
This addresses issue #342 .

When a lone pair is encountered, we turn it into two unpaired electrons for the purpose of adding hydrogens and looking up the solvation thermo of the saturated structure. But we only replace the lone pairs based on the usual valency of the atom. For example for CH2(S) we would replace that lone pair to make sum(unpaired electrons + bonds + charge) be equal to 4. But for H2O we wouldn't replace the lone pairs on Oxygen.

When we "unsaturate" the radicals, we should also correct the "A" H-bonding parameter that we previously calculated for the saturated structure. In the future we can have a radical solute library in the database, which "cancels" out the A value that we got from the molecule, as in RMG-Java. As a quick fix for right now, I am only correcting oxygen radicals, with a value of -0.345. 0.345 is the A value for -OH and -OOH. In the RMG-Java database, we have one additional entry for O in carboxylic acids; in RMG-Py, we will also need values for N and S radicals.

Also, an old fixup to the documentation that I never committed.

### DIFF
--- a/documentation/source/users/rmg/liquids.rst
+++ b/documentation/source/users/rmg/liquids.rst
@@ -108,7 +108,7 @@ dipole interactions related to the polarizability of the solvent) [Vitha2006]_, 
 `lL` term accounts for the contribution from cavity formation and dispersion (dispersion interactions are 
 known to scale with solute volume [Vitha2006]_, [Abraham1999]_. The `eE` term, like the `sS` term, 
 accounts for residual contributions from dipolarity/polarizability related interactions for solutes 
-whose blend of dipolarity/polarizability differs from that implicitly built into the `S` parameter [Vitha2006]_, [Abraham1990]_, [Jalan2010]_. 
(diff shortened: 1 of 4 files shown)


bslakman:
I just added a commit that actually looks up the radical correction to A. Goes along with my pull request on RMG-database https://github.com/GreenGroup/RMG-database/pull/62

connie:
Did you see my pull request on your master branch?

bslakman:
@connie I merged in your pull request and rebased onto official master. I added in data for N3sH2 (which was missing for some reason), but I didn't add data to the top nodes. I don't think it's a good idea; theoretically, there should be a Platts group for every heavy atom group we encounter. If we are falling up to the top node and we just put zeroes for everything, we won't get an error, and then we won't see that we are missing data.

The database branch is now ready to merge too (and I think the Travis test should pass when it is).

PR #365, opened 2015-03-30
Sort QM files by method subfolders

connie:
See https://github.com/GreenGroup/RMG-Py/pull/264 for detailed past commentary.  

* Fixes https://github.com/GreenGroup/RMG-Py/issues/360, now uses QMscratch as default folder or the user can specify the scratchDirectory in their input file (just like with QMfiles)
* Fixes https://github.com/GreenGroup/RMG-Py/issues/355, by using thermo only from a particular method because the method is now a subfolder: QMfiles/method_name/. Note that we will use existing .thermo files with the same method even if they come from different programs. i.e. if you run MOPAC you can use an existing .thermo file from GAUSSIAN, because they are supposed to give the same answer/same level of theory. 
* Fixes https://github.com/GreenGroup/RMG-Py/issues/358  for HBI corrections for both QM and normal group additivity based values for radical thermo.  It prioritizes library saturated values, then QM saturated values, then group additivity saturated values, then applies the radical correction.  

I've tested this using library values for saturated thermo for both a cyclic and noncyclic radical and it works.


connie:
Writing folders as SMILES strings can cause problems in Linux terminal:

    cd CCC=CC1CC=CC(CC)C1
    bash: syntax error near unexpected token `('

PR #376, opened 2015-04-21
Symmetry contribution for radicals

nickvandewiele:
- remove the symmetry contribution for saturated species if the thermo originates from thermo libraries, or QMTP
- check if pre-computed symmetry number exists before calculating it again


nickvandewiele:
This  PR results in the 'minimal' model containing 19/37 core species/reactions as opposed to the previous 17/35 that raised the initial doubts.

Issue #367, opened 2015-04-03
No solvation data for Ot / Negative lone pairs?


bslakman:
Get the following errors from one of my solvation jobs

```
ValueError: Sanitization error: Explicit valence for atom # 0 C, 5, is greater than permitted
Logged from file solvation.py, line 835
Error: 1 C u0 p-1 c+1 {2,T} {3,S} {4,S}
2 O u0 p1 c+1 {1,T}
3 H u0 p0 c0 {1,S}
4 H u0 p0 c0 {1,S}

Traceback (most recent call last):
  File "Code/RMG-Py/rmg.py", line 147, in <module>
    rmg.execute(args)
  File "/home/slakman.b/Code/RMG-Py/rmgpy/rmg/main.py", line 507, in execute
    self.reactionModel.enlarge(objectsToEnlarge)
  File "/home/slakman.b/Code/RMG-Py/rmgpy/rmg/model.py", line 739, in enlarge
    spec.generateThermoData(database, quantumMechanics=self.quantumMechanics)
  File "/home/slakman.b/Code/RMG-Py/rmgpy/rmg/model.py", line 160, in generateThermoData
    return self.processThermoData(database, thermo0, thermoClass)
  File "/home/slakman.b/Code/RMG-Py/rmgpy/rmg/model.py", line 189, in processThermoData
    soluteData = database.solvation.getSoluteData(self)
  File "/home/slakman.b/Code/RMG-Py/rmgpy/data/solvation.py", line 598, in getSoluteData
    soluteData = self.getSoluteDataFromGroups(species)
  File "/home/slakman.b/Code/RMG-Py/rmgpy/data/solvation.py", line 658, in getSoluteDataFromGroups
    sdata = self.estimateSoluteViaGroupAdditivity(molecule)
  File "/home/slakman.b/Code/RMG-Py/rmgpy/data/solvation.py", line 832, in estimateSoluteViaGroupAdditivity
    self.__addGroupSoluteData(soluteData, self.groups['abraham'], saturatedStruct, {'*':atom})
  File "/home/slakman.b/Code/RMG-Py/rmgpy/data/solvation.py", line 868, in __addGroupSoluteData
    raise KeyError('Node has no parent with data in database.')
KeyError: 'Node has no parent with data in database.'
```

[...]

PR #378, opened 2015-04-21
Fix issues with solvation thermo for C#O.

bslakman:
This addresses #367, and goes along with a corresponding database fix.

Added a new unit test for C#O and changed how we change lone pairs into bi-radicals to saturate.


connie:
Merged via https://github.com/GreenGroup/RMG-Py/commit/5a3ac97d5389c015462764a4b88da3817cffca1a

PR #383, opened 2015-05-12
Fix the way LogicOR groups are handled in descending kinetics tree

connie:
This pull request addresses https://github.com/GreenGroup/RMG-Py/issues/381

Change the way descendTree works, allowing an optional argument for 'strict' matching, in other words, requiring that groups and the molecules that match them completely match all atom labels.  Our previous check in generate reactions is not very robust.  I decided to alter the base function rather than implement lines in the generate reactions section.

We use strict atom label matching in finding the correct kinetic groups to associate
with a reaction and its reaction template.  However, the previous method used for LogicOr groups
was flawed.  It found the very first child node of an LogicOR group and used its atomLabels
to match the reactant.  However, sometimes the LogicOR group contained children that had
different atomLabels.  This resulted in some correct matches being thrown out.

By adding a ``strict`` argument to the descendTree function, the matchNodeToStructure, and
matchToStructure functions, we can be sure that full atom label matching takes place.  Note
that this is currently used primarily for kinetic groups and is not implemented for other
functions such as thermo estimation.

### DIFF
--- a/rmgpy/data/base.py
+++ b/rmgpy/data/base.py
@@ -871,13 +871,15 @@ class Database:
         atom in `structure`.
         
         Matching to structure is more strict than to node.  All labels in structure must 
-        be found in node.  However the reverse is not true.
+        be found in node.  However the reverse is not true, unless `strict` is set to True.
         
         Usage: node = either an Entry or a key in the self.entries dictionary which has
                       a Group or LogicNode as its Entry.item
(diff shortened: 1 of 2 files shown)


connie:
I also ran the minimal example before and after this change, and verified that there was no change in the core and edge model, which indicates to me that using strict matching for descending the tree for matching kinetics groups.

PR #384, opened 2015-05-13
New database unit tests

connie:
This addresses https://github.com/GreenGroup/RMG-database/issues/65

- Makes sure that rule entries are in the correct order like the family's top template.  
- Makes sure that the top templates are defined with products and reactants named differently if the family is not its own reverse.

### DIFF
--- a/databaseTest.py
+++ b/databaseTest.py
@@ -61,6 +61,12 @@ class TestDatabase():  # cannot inherit from unittest.TestCase if we want to use
             self.compat_func_name = test_name
             yield test, family_name
             
+            test = lambda x: self.kinetics_checkReactantAndProductTemplate(family_name)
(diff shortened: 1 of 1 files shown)


rwest:
How about a more general "no two nodes in the family (in any tree) can have the same label"?

connie:
That test already exists I think, and we have a check to make sure that the entries don't have identical names when we load it.  This one is only for the product template groups to ensure that they don't overwrite an existing group.

rwest:
How come the previous test didn't catch this then? I thought the problem was that the name was reused? Or am I confused?

connie:
We do the test for all the entries inside groups.py, but we don't check this part of it:

    template(reactants=["XSYJ"], products=["XYSJ"], ownReverse=False)

PR #391, opened 2015-06-06
Add two examples of pruning parameter set

KEHANG:
Two examples of pruning parameter set are provided. One is for pruning,
another is non-pruning. By comparing between two examples, users could get
ready to choose pruning parameter properly.


rwest:
Reads well to me. Thanks. Do the lists of parameter descriptions get rendered as a lost!

KEHANG:
The parameter descriptions are already in the paragraph I guess there's no need for hyperlink to external ref (and actually previously we don't have ref for the parameters anyway, the hyperlinks were invalid then) so here I changed them into normal code format rather than ref format.

PR #396, opened 2015-06-13
Documentation cleanup

rwest:
Somewhat urgent - we're aiming for release this week.
Please add commits (or pull requests) to this branch, to help improve and update the documentation.

The latex is formatted badly in many places. There are numerous warnings in the html, about incorrect cross-references etc, there are some formatting errors, there may be some class definitions etc missing from the API documentation, and it could do with a better introductory overview.

To build:
```
pip install --upgrade sphinx # if you don't already have it
cd documentation
make html
make latexpdf
```
then look at `documentation/build/html/index.html` and `documentation/build/latex/RMGPy.pdf`

Much of the documentation is in restructuredtext files in `documentation/source` and some parts are generated from the docstrings throughout the source code.

If you notice any weird things but aren't sure how to fix them, just mention them in a comment on this pull request.


enochd:
The RMG theory section has a subsection 'MEASURE theory guide'. There is also a corresponding user guide. Is MEASURE going to be part of the new release or is it replaced by Cantherm?

rwest:
Good question Enoch. Does measure still exist/work? is it entirely replaced by cantherm? @jwallen? I've never quite understood the measure/cantherm divide or transition.

enochd:
I've never tried to use it since it's my understanding that it does everything MEASURE does and more. Based on a cursory search through the code, I think it can be completely removed from RMGPy.

shamelmerchant:
The last time Connie and I checked there were still some function calls to
Measure in a pdep job in RMG-Py. For standalone pdep - Cantherm has all the
features in Measure. I don't know if this changed since then.

enochd:
The only place i see it's methods called is in rmg/pdep.py: 

`from rmgpy.measure.collision import SingleExponentialDown
from rmgpy.measure.reaction import fitInterpolationModel`

I think these can be replaced with other cantherm calls and we could entirely remove measure from RMG-Py, if desired.

jwallen:
I'd intended for MEASURE to be absorbed into CanTherm - the combination would then have a base feature set similar to other pressure dependence codes (i.e. the ability to compute k(T,P) from QC calculated parameters). This was pretty late in my tenure so I don't think I got around to deleting the MEASURE references. Based on the above it looks reasonably safe to remove it from the code and documentation.

bslakman:
I'm getting the following error when I do make html or make latexpdf... anyone else?

```
Belindas-MBP:documentation belinda$ make html
sphinx-build -b html -d build/doctrees   source build/html
  File "/Users/belinda/.virtualenvs/rmgpy4/bin/sphinx-build", line 9, in <module>
    load_entry_point('Sphinx==1.3.1', 'console_scripts', 'sphinx-build')()
  File "/Users/belinda/.virtualenvs/rmgpy4/lib/python2.7/site-packages/sphinx/__init__.py", line 51, in main
    sys.exit(build_main(argv))
  File "/Users/belinda/.virtualenvs/rmgpy4/lib/python2.7/site-packages/sphinx/__init__.py", line 61, in build_main
    from sphinx import cmdline
  File "/Users/belinda/.virtualenvs/rmgpy4/lib/python2.7/site-packages/sphinx/cmdline.py", line 23, in <module>
    from sphinx.application import Sphinx
  File "/Users/belinda/.virtualenvs/rmgpy4/lib/python2.7/site-packages/sphinx/application.py", line 30, in <module>
    from sphinx import package_dir, locale
  File "/Users/belinda/.virtualenvs/rmgpy4/lib/python2.7/site-packages/sphinx/locale/__init__.py", line 15, in <module>
    from six.moves import UserString
ImportError: cannot import name UserString
```

My sphinx and six are updated.

dgreen18:
Alright, here are a lot of problems with the PDF of the documentation (mostly formatting). All page #s refer to the numbers given by the PDF itself (out of 302 pgs), not the number in the footer of each page. Each footer number is 4 less than the numbers I give (i.e. pg 12  would be pg 8 in the printed PDF):

Code/words run into page margin or off of page: 12 13 14 33 41 62 68 70 205 207 208 209 226 261

pg 271 (Python Module Index) is very odd in context and perhaps should be deleted or moved?

page 226, 227 table runs off of page

page 266, 270, 272, and others are completely blank apart from header/footer- these blank pages between every section feel pretty odd when the section or final page of a section is sparse or tiny, but can be kept is that’s the aesthetic choice. (see pages 264-272 especially)

Entries in the index frequently bleed into the margin, but it’s merely an aesthetic distraction, no big deal

Lots of the spacing in the Index needs to be fixed- there are frequently tabs or huge gaps where there should just be a space (as early as the first addAction() on page 273)- I can do this?

In the Index: the first column runs into/overlaps with the second one: 273 275 280 281 284 285 286(not horrible) 289 294 295 296 298 299

Chapter 2, “Indices and tables,” claims on its cover page to have a genindex, mod index, and search, but contains absolutely nothing. I also don’t know why, if it’s legitimate, the “Python Module Index” would not be included in this indices chapter (though I’m in favor of both Ch. 2 and PMI being deleted altogether). Really it seems that the Contents, Bibliography, and Index are all that is necessary in the documentation, perhaps with a blank page between each. 

Why is page 5 (fig. 1) there? It stands as an isolated island with two blank pages surrounding it and no context.

Perhaps the entire documentation could be split into two overarching chapters, if we want two: the first as the first three sections of the contents and the second as the API reference. That would make a lot more sense thematically than the current system.

Every time verifyOutputFile() is repeated (like 9 times, do Ctrl+F), there’s a typo: there should be no apostrophe in “Check’s”!
There might be an apostrophe error every time inputFileKeywords(attempt) is used as well around “attempt”, but I’m not sure.
Every time writeInputFile(attempt) is repeated, there’s a typo: it says “attmept” instead of attempt.
Every time matchNodeToStructure(node, structure, atoms, strict=False) is repeated: under Usage, we should reformat list into bullet points or something on same indent. Also at bottom “strict = if set to True, ensures that all the node’s atomLabels are matched by in the structure” seems to be a typo.

pg 170 bottom table: first column text overlaps with second column
pg 169 table: 2nd column text overlaps with 3rd column
pg 164 2nd to bottom table: 1st column text overlaps 2nd column
pg 146 2nd from top table: 1st column text overlaps 2nd
pg 122 table: 1st column text overlaps 2nd
pg 108 table: 2nd column text overaps 3rd
pg 98 1st table, 3rd table, and 4th table: 1st column overlaps 2nd

bottom of pg 127: maybe move initialize() to next page?
bottom of pg 31: move “Libraries” to the next page?

top of pg 87: the headers “The Modified Strong Collision Method” etc. seem weird and out of place. Delete?
bottom of 75: “How Pruning Works” etc. same thing

pg 49 bottom image: is there a way to make the arrows not overlap with the other circles?

Okay, throughout this entire document there are sometimes the “genindex, modindex, search” bullet points and nothing else. What’s up with that? Can we delete them?

rwest:
Thanks. Also noticed we have no documentation on how to use CanTherm for pdep calculations. Could the deprecated measure documentation be moved across?

dgreen18:
I thought I had installed Latex, but I hadn't. When I made latexpdf, however, I got no error message and Terminal claimed that PDF(s) had been added to the new documentation/build/latex/ folder, even though there were none. This making processes should check whether the user has Latex installed and return an error if it is not.

goldmanm:
If we end up removing MEASURE, you may also consider removing/updating convertFAME (which creates a measure object)

rwest:
@pierrelb and/or @dgreen18 I believe you recently checked the MacOS Anaconda installation works? Please could you update the new documentation to be more platform-agnostic, ie. ensure that it works for Mac (without breaking it for Ubuntu?) 

The file to edit:
https://github.com/ReactionMechanismGenerator/RMG-Py/edit/docs/documentation/source/users/rmg/installation/anacondaDeveloper.rst

PR #400, opened 2015-06-21
Changes to Chemkin file parsing

rwest:
Various changes to the reading and interpretation of chemkin files.
These changes are all from our chemkin importer tool branch, but apply to code in the master RMG, so I've rebased them here and suggest we merge them. Mostly they make things more robust to a variety of weird chemkin files that are out there in the literature.


connie:
Closed via bd812db753823492a9ec36624131ce4af1c43698

PR #403, opened 2015-06-24
PAH bug checking valencies when reading adjacancy lists

rwest:
Trying to read the adjacency list for a polyaromatic hydrocarbon leads to this `InvalidAdjacencyListError: Invalid valency for atom C with 0 unpaired electrons, 0 pairs of electrons, 0 charge, and bonds [B,B,B].`

I guess the problem is those benzene bonds don't have an "order" of 1.5, because if they did then the atom would need a valence of 4.5.
e.g. http://chemistry.stackexchange.com/questions/1302/bond-order-for-carbonate-ion-for-resonance

Here's a failing unit test to get things started...
(do not merge this PR yet - at least until it's marked as "work in progress" (WIP))

Any ideas @nickvandewiele

### DIFF
--- a/rmgpy/molecule/adjlist.py
+++ b/rmgpy/molecule/adjlist.py
@@ -109,8 +109,15 @@ class ConsistencyChecker(object):
             theoretical = valence - order - atom.radicalElectrons - 2*atom.lonePairs
 
             if atom.charge != theoretical:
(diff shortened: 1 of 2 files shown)


rwest:
@nickvandewiele this now just skips the check_partial_charge validation on fused aromatic atoms (type `Cbf`). Can you think of a better fix?

mliu49:
Integrated into #769.

PR #407, opened 2015-07-03
Added Exception for user input of double bonded O2

nyee:
Many users erroneously use the double bonded singlet version for O2, which
does not react with most reaction families. Now RMG will throw an error
giving an explaination for why the user should use triplet O2.

If the user actually wants singlet O2, I have included a new option
"allowSingletO2" to the generatedSpeciesConstraints. Note that this
constraint is a little more restrictive by design. It will ignore
the allowed list, which usually lets users circumvent species
constraints for inputs such starting species and reaction libraries. The
only way to have singlet O2 is to set allowSingletO2=True. The point is
that we never want singlet O2 to fly under the radar.

### DIFF
--- a/examples/rmg/ch3no2/input.py
+++ b/examples/rmg/ch3no2/input.py
@@ -19,6 +19,7 @@ generatedSpeciesConstraints(
     #maximumSulfurAtoms = 0,
     #maximumHeavyAtoms = 3,
     maximumRadicalElectrons = 2,
+	allowSingletO2=False
 )
 
 # List of species
--- a/rmgpy/rmg/input.py
+++ b/rmgpy/rmg/input.py
@@ -289,6 +289,7 @@ def generatedSpeciesConstraints(**kwargs):
         'maximumSulfurAtoms',
         'maximumHeavyAtoms',
         'maximumRadicalElectrons',
+        'allowSingletO2',
     ]
     for key, value in kwargs.items():
         if key not in validConstraints:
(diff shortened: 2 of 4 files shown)


rwest:
Please add documentation of this new input file syntax.
I also suggest moving this check to only be when reading the input file. Currently it looks to me like every time any reaction is generated throughout the whole of RMG we run this check, which involves creating a brand new `O=O` molecule and doing isomorphism checks on it, which seems wasteful. To avoid user errors, we only need to check the input file species, not all species for ever.

connie:
I think we should additionally prevent it from existing in any form of
library in the database.

We can put it into globally forbidden species as well to prevent kinetic
groups from forming it.

On Fri, Jul 3, 2015 at 5:58 PM, Richard West <notifications@github.com>
wrote:

> Please add documentation of this new input file syntax.
> I also suggest moving this check to only be when reading the input file.
> Currently it looks to me like every time any reaction is generated
> throughout the whole of RMG we run this check, which involves creating a
> brand new O=O molecule and doing isomorphism checks on it, which seems
> wasteful. To avoid user errors, we only need to check the input file
> species, not all species for ever.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ReactionMechanismGenerator/RMG-Py/pull/407#issuecomment-118432352>
> .
>



-- 
Connie Gao
connieg@mit.edu || homepage <http://www.conniegao.com> || 978-500-5197
Chemical Engineering PhD Candidate
Massachusetts Institute of Technology

nyee:
Updated the check based on the suggestions. Now at initialization, it checks the species in input.py and all inputted seed mechanisms and reaction libraries. Documentation should be updated by goldmanm soon.

connie:
Closed via https://github.com/ReactionMechanismGenerator/RMG-Py/commit/d87b874d29a58aaf3ba2281d08f84e0deecd86e0.  Will add documentation through docs branch.

Issue #416, opened 2015-07-23
rdkit error in minimal example


bslakman:
I am getting the following error on updated master branch of RMG-Py and RMG-database, on both my Mac and a Linux cluster: 

```
Adding rate rules from training set in kinetics families...
Traceback (most recent call last):
  File "rmg.py", line 139, in <module>
    cProfile.runctx(command, global_vars, local_vars, stats_file)
  File "/usr/local/Cellar/python/2.7.6/Frameworks/Python.framework/Versions/2.7/lib/python2.7/cProfile.py", line 49, in runctx
    prof = prof.runctx(statement, globals, locals)
  File "/usr/local/Cellar/python/2.7.6/Frameworks/Python.framework/Versions/2.7/lib/python2.7/cProfile.py", line 140, in runctx
    exec cmd in globals, locals
  File "<string>", line 1, in <module>
  File "/Users/belinda/Code/RMG-Py/rmgpy/rmg/main.py", line 462, in execute
    self.initialize(args)
  File "/Users/belinda/Code/RMG-Py/rmgpy/rmg/main.py", line 354, in initialize
    self.loadDatabase()
  File "/Users/belinda/Code/RMG-Py/rmgpy/rmg/main.py", line 296, in loadDatabase
    family.addKineticsRulesFromTrainingSet(thermoDatabase=self.database.thermo)
  File "/Users/belinda/Code/RMG-Py/rmgpy/data/kinetics/family.py", line 893, in addKineticsRulesFromTrainingSet
    reactant.thermo = thermoDatabase.getThermoData(reactant, trainingSet=True)
  File "/Users/belinda/Code/RMG-Py/rmgpy/data/thermo.py", line 732, in getThermoData
    logging.info("Resonance isomer {0} {1} gives H298={2:.0f} J/mol".format(i+1, thermo[i][2].toSMILES(), thermo[i][1]))
  File "molecule.py", line 1424, in rmgpy.molecule.molecule.Molecule.toSMILES (build/pyrex/rmgpy/molecule/molecule.c:22385)
  File "molecule.py", line 1455, in rmgpy.molecule.molecule.Molecule.toSMILES (build/pyrex/rmgpy/molecule/molecule.c:22250)
  File "molecule.py", line 1530, in rmgpy.molecule.molecule.Molecule.toRDKitMol (build/pyrex/rmgpy/molecule/molecule.c:23447)
ArgumentError: Python argument types in
[...]

PR #423, opened 2015-07-30
Check RDKit version during 'make' so that people update to 2015+

connie:
This pull request tries to address https://github.com/ReactionMechanismGenerator/RMG-Py/issues/416

Not sure when the feature was actually updated in RDKit, but let's assume 2015 to be safe.

One thing to be cautious of is when building from Github source for RDKit.  Not sure if version is given from that. But for now assume that most people have a stable release.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,8 +6,6 @@
 
 DASPK=$(shell python -c 'import pydas.daspk; print pydas.daspk.__file__')
 DASSL=$(shell python -c 'import pydas.dassl; print pydas.dassl.__file__')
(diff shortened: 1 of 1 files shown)


rwest:
Good idea. I think it was this commit that added the needed implictH option to the API
https://github.com/rdkit/rdkit/commit/c41ffea23ba597e8e17dfa9db052aedeffe807d2
so 2015.03 is indeed the first official release with it in.

I simplified the test. If it looks ok to you, merge.

Issue #417, opened 2015-07-27
Solver does not use collision efficiencies??


rwest:
We can read the advanced collision efficiencies from chemkin files, store them in RMG P-Dep reactions, and output them again, but when integrating the reaction models internally, I think they are not used.

The `getRateCoefficient()` methods of all subclasses of `PDepKineticsModel` are given only `T` and `P` and not a vector of species. 

This would be taken care of by using the result of  `getEffectivePressure()` as `P` which is what happens if you use `Reaction.getRate()` but the actual ODE solver bypasses this and just calls `getRateCoefficient()`.  It looks like `Reaction.getRate()` is never actually used?

rwest:
I have a vague memory of people reporting that their RMG simulations don't quite match CHEMKIN simulations of the same thing. Could this be the cause? If you are such a person, please comment!

connie:
Confirmed that that `getEffectivePressure` is not being used in the solver. This indeed looks very troubling.  I'm guessing it never got caught because hardly anyone checked PDEP rates with collision efficiencies against CHEMKIN, especially because we often use single bath gases, the errors would've been hidden now.

PR #425, opened 2015-08-10
Fix handling of collision efficiencies in solver

connie:
This pull request addresses https://github.com/ReactionMechanismGenerator/RMG-Py/issues/417

Update k_f(T,P) and k_r(T,P) within the solver by updating values in the residual() calculation.  
Collision efficiencies and pdep kinetics are stored in the SimpleReactor object during initializeModel() so as to reduce computation time.  

Checked a few conditions against CHEMKIN, with unit test in place.


connie:
Comment directly to commits become lost when branches are rebased and force-pushed.  So I am recopying @rwest's comments from earlier.  Perhaps from now on comment directly within the pull request and make references to specific lines?

-----
Only mentioning because I expect this is in a tight loop that gets called a lot during the solve?

Evaluating y[:numCoreSpecies] twice within the loop like this may be more costly than evaluating it once and storing it in a new variable (if you have type declared said variable for Cython).

Also, we should check that for i in range(some_variable_not_cdef_declared) is C optimized. (i.e. check the html file made by Cython). And does it work with our cdef int i or does it need cdef size_t i?

----

Just checking: this isomorphism checking isn't done inside a tight loop in the solver? I don't think it is, but wanted to be sure...

----
No it won't be rechecked in the solver.  That was whole reason I created a `getEffectiveCollisionEfficiencies` function rather than using `getEffectivePressure` every time.  That way it is called in the ReactionSystem.initializeModel step, and then the coefficients are stored there until integration to t_end is completed.

rwest:
OK, commenting on the PR itself sounds like a good suggestion.
Is this still in progress? Let us know when to review and merge.

connie:
Still in the middle of making unit tests and checking robustness.

rwest:
I assume those results in the unit test came from a chemkin run?

connie:
Yes






On Wed, Aug 19, 2015 at 8:17 AM -0700, "Richard West" <notifications@github.com> wrote:












I assume those results in the unit test came from a chemkin run?

rmgpy/solver/simple.pyx
```diff
@@ -227,27 +259,48 @@ cdef class SimpleReactor(ReactionSystem):
...
+            colliderEfficiencies = self.colliderEfficiencies
+            for i in range(pdepColliderReactionIndices.shape[0]):
+                # Calculate effective pressure
+                Peff = P*numpy.sum(colliderEfficiencies[i]*y_coreSpecies / numpy.sum(y_coreSpecies)) 
+                kf[pdepColliderReactionIndices[i]] = pdepColliderKinetics[i].getRateCoefficient(T, Peff)
+                kr[pdepColliderReactionIndices[i]] = kf[pdepColliderReactionIndices[i]]/equilibriumConstants[pdepColliderReactionIndices[i]]
+                
+            # Update object's forward and reverse rate coefficients
+            self.forwardRateCoefficients = kr
+            self.reverseRateCoefficients = kf
```
rwest: 1) did you mean to switch forward and reverse here? (forward becomes kr, and reverse becomes kf)?
2) if not, I don't think these lines are needed at all - you update the arrays in place, the originals change too (`kr` is just the name of `self.reverseRateCoefficients` in the local scope, not a deep copy of it)

rwest:
@connie, can you check my last commit (5b9eb5ca673ee3354b17b4526fa0eb0930ef62ea).

It's strange that the unit tests passed both before and after my commit, so perhaps we can add one that would catch it? (I'm surprised your chemkin simulation one didn't but perhaps it's not sensitive enough?)

If that looks OK to you, I'm happy with this PR and you can merge it.

connie:
Thanks for catching the error, but we still need them to be saved since the rate coefficients are used in other functions in the solver, ie. the Jacobian calculation.  I'm guessing since the Jacobian is only really relevant for sensitivities, the solver still gets the correct answer for `y`, because the Jacobian doesn't actually have to be correct for the solver to proceed.

rwest:
But they already are saved, no?

```python
kf = self.some_array
#update some element of the array
kf[5] = 5
# but it is the same array
self.some_array[5] == 5
```
?

connie:
Oh I think I know why it worked, it's because all the reactions in there have colliders, so they were always  getting updated correctly even tho I saved them in reverse.  I have added in a non-pdep reaction and updated the unit test.  Now it fails on the previous commit when saved coefficients were reversed.  You are right about the array being saved automatically, so I cut that commit.

Will merge once travis passes.

nickvandewiele:
@connie I noticed that you initialized (11a00202adde2a0bf669488d65436d08ce558097) the `efficiencies` dict with `Molecule` objects rather than SMILES strings so you could easily call the `isomorphic` function to check equality.

In the near past, I altered this section (d91cbefa7234ceef518a55fe8900d0ff0d185c9d) to store efficiencies using the molecule label attribute and using string comparison for equality check (48a09b68f8e300b2e3971ea19323d4ead38fc20b). By doing so, I was able to read in chemkin/write files without dictionary files, which opened up a lot of new venues for using the RMG-Py functions. 

For example, one could read in a chemkin mechanism and use our solvers to run a reactor simulation.

Moreover, in the scope of RMG-Py and chemkin models, molecule labels are unique so that they can be used to determine equality. Any thoughts about this?

rwest:
@nickvandewiele, I see there are benefits to your string comparison, but there are drawbacks, eg. a name is not unique across all different models, i.e. you can read a self-consistent chemkin file by strings only, but can not pick-and-mix from a range of RMG libraries unless you knew the identities of the molecules. I've struggled with this issue in my "chemkin importer" project, and would love for us to devise a win-win solution. Perhaps discuss on a different issue, as this PR #425 does not change the status quo.

PR #426, opened 2015-08-12
Fixes for exporting Java-style libraries

rwest:
Neither the Kinetics nor the Thermo library export features actually output things in RMG-Java syntax.
Now we do so, although if you have species that can't be represented in and old adjacency list (eg. containing nitrogen) then it will obviously fail to output this. At least it won't crash, however, and in most cases will make a parseable and passable output.

Hopefully this fixes associated parts of the website too (like http://rmg.mit.edu/simulate/java_kinetics_library), though I've not tested it.

### DIFF
--- a/rmgpy/chemkin.py
+++ b/rmgpy/chemkin.py
@@ -1686,18 +1686,28 @@ def markDuplicateReactions(reactions):
         markDuplicateReaction(reaction1, remainingList)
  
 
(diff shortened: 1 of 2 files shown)


connie:
Merged via 946a305b590a8f3c8fd8d7a1763574a68de1f0be.  

I just tested it and it does fix the website Java kinetics library module.

Issue #432, opened 2015-08-18
Pressure Dependence instructions and input syntax a bit iffy.


rwest:
If you start with the minimal file, then follow the instructions for how to add pressure dependence, it doesn't help you to add inert species for colliders. Would be good to mention this is needed and link to a complete working example. 

Furthermore, the `interpolation=('PDepArrhenius')` instructions were broken (see #431).
That bit of code is a bit fragile - requiring a comma in the tuple is bound to catch people out, and the error isn't hit until well after the model generation has begun (rather than when reading the input file).

PR #433, opened 2015-08-19
Fix input reading of interpolation model for ('PDepArrhenius')

connie:
Closes https://github.com/ReactionMechanismGenerator/RMG-Py/pull/431, instead of using clunky input file tuple syntax.  See also: 
https://github.com/ReactionMechanismGenerator/RMG-Py/issues/432

This was already fixed in the Cantherm input reader, but
not for RMG's own input:
https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/cantherm/input.py#L305

### DIFF
--- a/rmgpy/rmg/input.py
+++ b/rmgpy/rmg/input.py
@@ -243,6 +243,8 @@ def pressureDependence(
     rmg.pressureDependence.method = method
     
     # Process interpolation model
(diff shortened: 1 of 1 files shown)


connie:
Closed via d0ddf2963db6628262e462344f5e2fddb33c6bd2

PR #435, opened 2015-08-19
Atom sorting, a new ID parser module and a new augmented inchi

nickvandewiele:
This PR bundles three major, related improvements to the `rmgpy.molecule` package:

1. A refactoring of the identifier generation and parsing functionality of the molecule module
2. A new way of sorting atoms
3. A first attempt in creating an canonical (augmented) inchi for species with complex electronic structures.

# Identifier generation and parsing

A subpackage of the molecule package is created: `rmgpy.molecule.parser`. This package bundles modules related to the generation and parsing of molecule identifiers such as SMILES and InChI. This stuff was originally to be found in the `rmgpy.molecule.molecule` module. With different backends (rdkit, openbabel) and different identifiers (SMILES, SMARTS, InChI, InChI key, augmented inchi) this became quite large, hence the idea of moving it to its own sub-package.

# Sorting atoms

Prior to this PR, vertices were sorted based on a vertex connectivity value that calculated the connectivity of a vertex until the 3rd degree. This is rather limited, oftentimes leading to multiple equivalent sorting possibilities of atoms.

The new sorting algorithm implements Morgan's idea, by recursively updating the connectivity value of the vertices, until the number of distinct connectivity values is maximized.

Next, `molecule.sortAtoms()` uses not only the connectivity value of the vertex (`sortVertices`), but also its atom attributes like charge, radicalElectrons, lonePairs.

I believe this leads to a unique sorting order of atoms in contrast to the previous implementation.

# Canonical (augmented) inchi

Prior to this PR, the augmented inchi was an inchi with the total multiplicity of the molecule appended at the end.

E.g.:
the triplet biradical .C=C. was represented by:
`InChI=1/C2H2/c1-2/h1-2H/mult3`

Because backends are only able to convert the InChI to a graph, this lead to wrong graphs.

E.g.:`InChI=1/C2H2/c1-2/h1-2H/mult3` would be converted into ethyne (C#C) rather than the
biradical with the identical atom connectivity.

This PR tries to heuristically correct for those mistakes, e.g. by converting the triple bond into
a double bond, with neighboring radical atoms.

In addition, in some cases the multiplicity info is not enough to unambiguously reconstruct the 
graph from the augmented InChI, leading to errors downstream. Therefore, a new (optional) layer is created that indicate the atoms that bear unpaired electrons.

.C=C. might be represented by

`InChI=1/C2H2/c1-2/h1-2H/mult3/u1,2`

with `/u1,2` referring to the location of the unpaired electrons (atoms 1, and 2).

Besides these 3 main features, I created some minor modifications:
- cache the InChI in the Molecule object
- a molecule hash and equality comparison method
- removing the identifier generation from the molecule constructor


rmgpy/molecule/parser/inchi.py
```diff
@@ -0,0 +1,109 @@
...
+    u_indices = [int(i) for i in ulayer.split(U_LAYER_SEPARATOR)]
+    return inchi, mult, u_indices
+
+def ignore_prefix(string):
+    """
+    Splits off the 'InChI=1S' or 'InChI=1' layer of an InChI
+    and returns the last part.
+    """
+
+    assert INCHI_PREFIX in string, 'Not a valid InChI: {}'.format(string)
```
rwest: Properly we should raise an exception like `InchiException` rather than `AssertionError` to enable smarter error handling downstream (eg. the website may want to implement a `Try ... Except` block for user mistakes.)   Ideally assertions should only check for bugs in the code, not mistakes in the input or chemistry.   (Though I often lazily use them for this too  :wink:)

rmgpy/molecule/parser/inchiTest.py
```diff
@@ -0,0 +1,92 @@
+import unittest
+
+from rmgpy.molecule.parser.inchi import AugmentedInChI, InChI, ignore_prefix, compose_aug_inchi, INCHI_PREFIX, U_LAYER_PREFIX
+
+
+class InChITest(unittest.TestCase):
+
+    def test_constructor(self):
+        inchi1 = InChI('InChI=1S/foo')
+        assert inchi1 is not None
```
rwest: I thought unit tests should use `self.assert` (or `self.assertTrue`, `self.assertEqual`, etc. etc.) rather than `assert` so that they're recognized as test failures rather than errors.

rmgpy/data/kinetics/family.py
```diff
@@ -1067,8 +1065,7 @@ def applyRecipe(self, reactantStructures, forward=True, unique=True):
         # If product structures are Molecule objects, update their atom types
         for struct in productStructures:
             if isinstance(struct, Molecule):
-                struct.updateAtomTypes()
-                struct.updateMultiplicity()
+                struct.update()
```
rwest: if it's *not* a `Molecule` then `updateConnectivityValues` is no longer called (removed from line 1046 above). What will this effect and does it matter? what could it be besides a `Molecule`? a `Group` maybe? (eg. in a reaction template?)

rmgpy/molecule/graph.pyx
```diff
@@ -427,7 +444,8 @@ cdef class Graph:
         # If we need to sort then let's also update the connecitivities so
         # we're sure they are right, since the sorting labels depend on them
         self.updateConnectivityValues()
-        self.vertices.sort(key=getVertexConnectivityValue)
+
+        self.vertices = sorted(self.vertices, key=getVertexConnectivityValue)
```
rwest: Is making a sorted copy of the list then storing it and discarding the old one better than sorting the original list in place?

nickvandewiele:
Additional commits:

- clean up imports
- I moved the 3 additional modules (parser, inchi, util) into the rmgpy.molecule package since these modules had dependencies from the molecule package.
- I added the 3 modules to setup.py so that they are cythonized, although they consist of pure python code for now.

nickvandewiele:
for some reason, RMG completely generates the wrong model for `make eg1` now. It starts off by adding [CH2] to the core, instead of [CH3]...

`make test` passes though.

connie:
See https://github.com/ReactionMechanismGenerator/RMG-Py/pull/443 for updated pull request.

PR #444, opened 2015-08-27
Species revamp

nickvandewiele:
migrating #439 here

### DIFF
--- a/rmgpy/cantherm/input.py
+++ b/rmgpy/cantherm/input.py
@@ -181,8 +181,8 @@ def reaction(label, reactants, products, transitionState, kinetics=None, tunneli
         if label in reactionDict:
             raise ValueError('Multiple occurrences of reaction with label {0!r}.'.format(label))
     logging.info('Loading reaction {0}...'.format(label))
-    reactants = sorted([speciesDict[spec] for spec in reactants])
-    products = sorted([speciesDict[spec] for spec in products])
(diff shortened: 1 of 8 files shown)


nickvandewiele:
make eg1 takes ~240s instead of 90s on travis. I need to find the bottleneck before this PR can be merged into master.

nickvandewiele:
I got the time to run make eg1 back to the original again by removing the `generateResonanceIsomers()` calls in `__compare_molecules` assuming that the resonance isomers are already created when two `Species` objects are compared against each other.

I made sure that resonance isomers are generated by including a call to `generateResonanceIsomers()` in the constructor, and in the 2 methods that enable the creation of Species objects (`fromAdjacencyList()`, and `fromSMILES`).

Thirdly, because of the latter, I removed redundant calls to `generateResonanceIsomers()`.

nickvandewiele:
Closing this PR because:

- some commits are part of another PR (e.g. getter to augmented inchi)
- the equality comparison is a bad idea: everytime you will do `spc in list_of_species`, it will call the equality methods, rather than simply checking the much faster references
- the generateResonanceIsomers() call in the constructor of Species slows the code down. We have a lot of empty constructors `Species().from...` for which the call is completely useless.

I've learned that messing around with the constructor of important classes can dramatically change the duration of simulations.

PR #453, opened 2015-09-16
Add more controls (minCoreSizeForPrune, minSpcExistIterForPrune) and garbage collection for prune

KEHANG:
This PR adds
- after every 2 times of pruning, `gc.collect()` will be executed, which further frees up memory.
- two prune related handles: `minCoreSizeForPrune` (avoid pruning when kinetic model is far away from completeness, default value is 50, but user can specify  their own value in the input file) and `minSpcExistIterForPrune` (make sure the pruned species exist at least for certain iterations. It was 2 previously, now 2 is the default value and user can specify their own value in the input file)


rwest:
Not a fan of the name `minSpcExistIterForPrune`. Hard to decipher meaning or remember how to spell. How about `minSpeciesAgeForPrune` or the full `minSpeciesExistIteraionsForPrune`?

While you're looking at this bit of code: does the iteration count get reset if an existing species is made again in a new generation? Eg if not then a radical formed by abstracting from a core species will be made *every* time a radical enters the core, so it might have Beene made each of the last 2 iterations (and will be in the next 5) but it has therefore existed for at least 2 generations and so will be pruned, even though it's about to be made again.

KEHANG:
@rwest Thanks for the suggestion! The current design does't reset the age of species; if a radical has been generated before, RMG will redirect this radical to the reference of the existing species object. Yes, for your case, the radical will first be pruned if the flux is really low and re-generated in later stage if necessary. So what is the ideal case for your example? this radical should only be pruned after 5 continuous generation by resetting age? But for cases where the radical will be generated 5 times but not continuously (some enlarge iterations with no generation of this radical are between the generations) resetting also doesn't help.

rwest:
Hmm. Perhaps the eligibility for pruning check could be  "was most recently made at least N cycles ago" rather than "was first made at least N cycles ago"? The choice of N would be up to the user. You're right that whatever you choose there may be some cases when a species is pruned only to be created again, but I think that's probably unavoidable with our current algorithm.

KEHANG:
@rwest but it's a good point you just brought. Perhaps I could make another pull request later for it since this pull request is only aiming to give users the options to choose these two parameters. I will rename the parameter you mentioned shortly.

PR #456, opened 2015-09-22
Fix bug91 website & ensure heavy atoms appear first in adj lists

nickvandewiele:
This PR addresses https://github.com/ReactionMechanismGenerator/RMG-website/issues/91 and ensures that heavy atoms appear first in the adjacency list.

bf72770 is a unit test contains a number of aromatic species that include the molecule mentioned in the issue.

794eb1a ensure that the risky combo of `.toRDKitMol(molecule, removeHs=True, returnMapping=True)` is not used anymore.

ab4d579 is an improvement of `toRDKitMol` to only add atoms to the index list that will actually exist in the generated rdkit molecule.

a625972 ensures that heavy atoms always preceed H-atoms, when `molecule.sortAtoms()` is called.

### DIFF
--- a/rmgpy/molecule/adjlistTest.py
+++ b/rmgpy/molecule/adjlistTest.py
@@ -670,12 +670,12 @@ class TestMoleculeAdjLists(unittest.TestCase):
         adjlist: Check that we can convert back to old style adjacency list
         """
         molecule2 = Molecule().fromSMILES('C=CC=C[CH]C')
-        string = """1 C 0 {3,D} {4,S}
-2 C 0 {4,S}
-3 C 0 {1,D} {5,S}
-4 C 1 {1,S} {2,S}
-5 C 0 {3,S} {6,D}
-6 C 0 {5,D}"""
+        string = """1 C 0 {2,D}
+2 C 0 {1,D} {4,S}
+3 C 1 {5,S} {6,S}
+4 C 0 {2,S} {6,D}
+5 C 0 {3,S}
+6 C 0 {3,S} {4,D}"""
         self.assertEqual(molecule2.toAdjacencyList(removeH=True,oldStyle=True).strip(),string.strip())
 ################################################################################
 class TestConsistencyChecker(unittest.TestCase):
--- a/rmgpy/molecule/atomtypeTest.py
+++ b/rmgpy/molecule/atomtypeTest.py
@@ -201,11 +201,11 @@ class TestGetAtomType(unittest.TestCase):
         """
         Test that getAtomType() returns appropriate carbon atom types.
         """
-        self.assertEqual(self.atomType(self.mol1, 0), 'Cs')
-        self.assertEqual(self.atomType(self.mol1, 6), 'Cd')
-        self.assertEqual(self.atomType(self.mol1, 4), 'Cdd')
-        self.assertEqual(self.atomType(self.mol1, 9), 'Ct')
-        self.assertEqual(self.atomType(self.mol1, 1), 'CO')
+        self.assertEqual(self.atomType(self.mol1, 5), 'Cs')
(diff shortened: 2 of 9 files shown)


connie:
I think the check_heavy_atoms function is unnecessary if the sorting is done correctly.  At most we should run it in an outside unit test.  Otherwise it will be costly to run it everytime during sorting.  

Is there any reason not to sort using atoms.sort(lambda a, a.getNumber)?  or a.isHydrogen?  This would do the same thing, rather than writing moveHs.  See 54e15d9

Also, I know these commits fail the unit tests.  When the inchi is malformed we run the risk of an infinite while loop and the molecule never being correct (https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/molecule/parser.py#L417)  I think this is something that must be fixed.

nickvandewiele:
The idea of the new atom sorting is that you come up with an order of atoms that is always the same, regardless of how you input the molecule.

If you sort atoms purely based on `atom.number`, then the order of the sorted atoms, is heavily dependent on how you input the atoms in the first place.

That's why `atom.connectivityvalue` exists. You generate an integer that tries to differentiate atoms of the same element based on their connectivity within the molecule.

Next, if you purely use atom connectivity values to sort atoms, then you run the risk that H-atoms might come before heavy atoms, purely based on their connectivity. That's why I added a secondary layer of sorting that compensates for that.

connie:
Is there a reason why we cannot sort based on both without having to write
our own sorter?  Prioritizing the atom.number and then connectivityvalue?

http://stackoverflow.com/questions/5212870/sorting-a-python-list-by-two-criteria



On Tue, Sep 22, 2015 at 6:19 PM, NICK VANDEWIELE <notifications@github.com>
wrote:

> The idea of the new atom sorting is that you come up with an order of
> atoms that is always the same, regardless of how you input the molecule.
>
> If you sort atoms purely based on atom.number, then the order of the
> sorted atoms, is heavily dependent on how you input the atoms in the first
> place.
>
> That's why atom.connectivityvalue exists. You generate an integer that
> tries to differentiate atoms of the same element based on their
> connectivity within the molecule.
>
> Next, if you purely use atom connectivity values to sort atoms, then you
> run the risk that H-atoms might come before heavy atoms, purely based on
> their connectivity. That's why I added a secondary layer of sorting that
> compensates for that.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ReactionMechanismGenerator/RMG-Py/pull/456#issuecomment-142438178>
> .
>



-- 
Connie Gao
connieg@mit.edu || homepage <http://www.conniegao.com> || 978-500-5197
Chemical Engineering PhD Candidate
Massachusetts Institute of Technology

nickvandewiele:
`self.atoms.sort(key=lambda a: a.getDescriptor(), ...)` does exactly that.

`getDescriptor()` refers to a tuple 
`(self.getAtomConnectivityValue(), self.symbol, self.radicalElectrons, self.lonePairs, self.charge)`.

The reason why I used the connectivity value as the first key, rather than the symbol is because it is the most logic choice. You generate connectivity values for each atom, and you sort the atoms based on that numerical order. The connectivity value has also much more discriminatory power than the atom symbol.

I wanted to keep that logic, numerical order, and only infrequently deviate from that order when ensuring that heavy atoms always come first. So `moveHs` only swaps a limited number of atoms and preserves the order of most other atoms.

nickvandewiele:
ea7e0a7 creates a number of malformed unittests, all caught by an Exception.

d4bb280 removes the check that heavy atoms precede H's, as requested by @connie 

8dce572 avoids an infinite loop when the indices of unpaired electrons are wrong. Great suggestion, @connie

PR #458, opened 2015-10-07
Changed 'drawMolecules' attribute to 'generateOutputHTML' for clarity.

bslakman:
Since the change in https://github.com/ReactionMechanismGenerator/RMG-Py/commit/a2acb569e85a8606ac7fe6fd507b6d0752048f66 , I thought the attribute name 'drawMolecules' was a little confusing since it generates the HTML file now (and it was set to False in all the examples). I changed the name to 'generateOutputHTML' and set it to 'True' for the minimal example only.


rwest:
Hmm. Build failed because conda installed an outdated RDKit?:
`rdkit-2014.09. 100% |################################| Time: 0:00:00  44.86 MB/s`

connie:
Can you modify the parts associated with saving and loading the input file in RMG-Py and RMG-website as well?

connie:
Yeah I noticed this about our travis builds also.  Somehow it was prioritizing 2014 version over 2015 version even though both are on the anaconda channel.  Also when I last checked, copying packages between channels was not working either.  This seems to be an issue on conda's side.

PR #462, opened 2015-10-15
make conda env use RDKit v 2015.03.1

rwest:
Trying to fix the travis build problems

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -9,7 +9,7 @@ dependencies:
   - numpy
   - cython ==0.21
   - matplotlib
-  - rdkit
+  - rdkit ==2015.03.1
   - pydas
   - pydqed
   - quantities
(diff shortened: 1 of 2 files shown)


rwest:
I think this is ready to merge. Once it is, and the master is passing travis tests again, we should rebase other pull requests and see if they start passing.

connie:
We may need to rebuild and reupload the binary versions of RMG-Py since they don't have the rdkit 2015 requirement in the recipes, which may lead to users getting the older dependency even when they install RMG-Py via binary.  Not sure why Anaconda has not fixed this dependency management issue.

rwest:
Yes, we will need to rebuild the binaries, as this PR changes the build recipe. Sorry - I forgot to increment the build number in meta.yaml. Please do so before merging (on my phone now)

PR #464, opened 2015-10-16
Newly added Retroen family into RMG

KEHANG:
This PR is to allow Retroen family to be in the scope of RMG-Py.


connie:
Related to https://github.com/ReactionMechanismGenerator/RMG-Py/issues/466
Am gonna merge this so it's working and implement having less hardcoding later.

PR #465, opened 2015-10-16
Consolidate version numbering

rwest:
Now anaconda and the documentation all get their version numbers from `rmgpy.__version__` which is set in the file `rmgpy/__init__.py`.

Furthermore, the anaconda build number is set to be the number of git commits since the last tag.
So the workflow should be something like:

 1. bump the build number in `rmgpy/__init__.py`
 2. get that committed to official master
 3. tag that commit, eg. `git tag -a 1.0.1  5733bf1e5` 
 4. push that tag to official, eg. `git push official 1.0.1`
 5. rebuild and publish the documentation (eg. `cd documentation; make html; make publish`)
 6. rebuild and upload conda binaries (eg. `conda build RMG-Py; binstar upload -u RMG path/to/binary`)


connie:
Good work. Does this mean that the tag function is well suited for build numbers and that we can essentially get rid of pushing to version branches? (Which mean we could delete 1.0.0)? Can tagging be done retroactively on older commits?


_____________________________

rwest:
Yes, I would recommend tagging releases. I have tagged the 1.0.0 release (or what I think was the release). We can delete the 1.0.0 branch, which we never intend to merge, though could make a similar branch to travis test the conda binaries (which we could rebase whenever we wish to run).
Once this PR is merged, we can tag it as 1.0.1.

rwest:
Can we merge this and push a working build out, then do #468 for version 1.0.2, or is #468 going to be a really quick fix?

connie:
I wouldn't consider the build to be working unless #468 was resolved.

PR #468, opened 2015-10-16
Fix group algorithm that was breaking database tests

connie:
This pull request tries to fix the breaking of the database tests and the subgraph isomorphism check between groups.  (DO NOT PULL UNTIL WORKING)

Summary of changes:
- Add 'make test-database' to RMG-Py travis build to ensure safety

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ install:
 
 script: 
   - make test
+  - make test-database
   - make eg1
   - make eg4
   - make eg5
--- a/rmgpy/data/base.py
+++ b/rmgpy/data/base.py
@@ -948,18 +948,17 @@ class Database:
                 # Passed semantic checks, so add to maps of already-matched atoms
                 initialMap[atom] = center
             # Labeled atoms in the structure that are not in the group should
-            # not be considered in the isomorphism check, so remove them temporarily
+            # not be considered in the isomorphism check, so flag them temporarily
             # Without this we would hit a lot of nodes that are ambiguous
-            removedAtoms = []
-            for label, atom in structure.getLabeledAtoms().iteritems():
-                if label not in centers:
-                    removedAtoms.append(atom)
-                    structure.atoms.remove(atom)
+            flaggedAtoms = [atom for label, atom in structure.getLabeledAtoms().iteritems() if label not in centers]
+            for atom in flaggedAtoms: atom.ignore = True
+            
             # use mapped (labeled) atoms to try to match subgraph
             result = structure.isSubgraphIsomorphic(group, initialMap)
-            # Restore atoms removed in previous step
-            for atom in removedAtoms:
-                structure.atoms.append(atom)
+            
+            # Restore atoms flagged in previous step
+            for atom in flaggedAtoms: atom.ignore = False
+                
             return result
 
     def descendTree(self, structure, atoms, root=None, strict=False):
--- a/rmgpy/molecule/graph.pxd
+++ b/rmgpy/molecule/graph.pxd
@@ -33,6 +33,7 @@ cdef class Vertex(object):
     cdef public short sortingLabel
     cdef public bint terminal
     cdef public Vertex mapping
+    cdef public bint ignore
 
     cpdef Vertex copy(self)
 
--- a/rmgpy/molecule/graph.pyx
+++ b/rmgpy/molecule/graph.pyx
@@ -57,6 +57,7 @@ cdef class Vertex(object):
     def __init__(self):
         self.edges = {}
         self.resetConnectivityValues()
+        self.ignore = False
 
     def __reduce__(self):
         """
--- a/rmgpy/molecule/vf2.pyx
+++ b/rmgpy/molecule/vf2.pyx
@@ -167,7 +167,7 @@ cdef class VF2:
         if callDepth == 0:
             if self.findAll:
                 mapping = {}
-                for vertex2 in self.graph2.vertices:
+                for vertex2 in [v for v in self.graph2.vertices if not v.ignore]:
                     assert vertex2.mapping is not None
                     assert vertex2.mapping.mapping is vertex2
                     mapping[vertex2.mapping] = vertex2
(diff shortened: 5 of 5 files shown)


nickvandewiele:
In https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/molecule/graph.pyx#L415 we iterate over the edges of `vertex1` and search for the index of the neighboring vertex in `self.vertices`:
```python
        for vertex1, old_value in zip(self.vertices, old_values):
            ...
            for vertex2 in vertex1.edges: count += old_values[self.vertices.index(vertex2)]
```

In https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/data/base.py#L953

vertices are removed from `self.vertices` while the references in `vertex.edges` are not removed at the same time.
```python
            removedAtoms = []
            for label, atom in structure.getLabeledAtoms().iteritems():
                if label not in centers:
                    removedAtoms.append(atom)
                    structure.atoms.remove(atom)
```


This leads to the following error:
```
  File "/vagrant_data/Code/RMG-Py/rmgpy/data/base.py", line 856, in matchNodeToNode
    return self.matchNodeToStructure(node,nodeOther.item, atoms=nodeOther.item.getLabeledAtoms()) and self.matchNodeToStructure(nodeOther,node.item,atoms=node.item.getLabeledAtoms())
  File "/vagrant_data/Code/RMG-Py/rmgpy/data/base.py", line 961, in matchNodeToStructure
    result = structure.isSubgraphIsomorphic(group, initialMap)
  File "rmgpy/molecule/group.py", line 782, in rmgpy.molecule.group.Group.isSubgraphIsomorphic (build/pyrex/rmgpy/molecule/group.c:15922)
    def isSubgraphIsomorphic(self, other, initialMap=None):
  File "rmgpy/molecule/group.py", line 809, in rmgpy.molecule.group.Group.isSubgraphIsomorphic (build/pyrex/rmgpy/molecule/group.c:15811)
    return Graph.isSubgraphIsomorphic(self, other, initialMap)
  File "rmgpy/molecule/graph.pyx", line 474, in rmgpy.molecule.graph.Graph.isSubgraphIsomorphic (build/pyrex/rmgpy/molecule/graph.c:8759)
    return vf2.isSubgraphIsomorphic(self, other, initialMap)
  File "rmgpy/molecule/vf2.pyx", line 77, in rmgpy.molecule.vf2.VF2.isSubgraphIsomorphic (build/pyrex/rmgpy/molecule/vf2.c:1595)
    self.isomorphism(graph1, graph2, initialMapping, True, False)
  File "rmgpy/molecule/vf2.pyx", line 102, in rmgpy.molecule.vf2.VF2.isomorphism (build/pyrex/rmgpy/molecule/vf2.c:2019)
    graph1.sortVertices()
  File "rmgpy/molecule/graph.pyx", line 448, in rmgpy.molecule.graph.Graph.sortVertices (build/pyrex/rmgpy/molecule/graph.c:8130)
  File "rmgpy/molecule/graph.pyx", line 292, in rmgpy.molecule.graph.Graph.updateConnectivityValues (build/pyrex/rmgpy/molecule/graph.c:5768)
    connectivityValues = self.__update([len(vertex.edges) for vertex in self.vertices], 0)
  File "rmgpy/molecule/graph.pyx", line 417, in rmgpy.molecule.graph.Graph.__update (build/pyrex/rmgpy/molecule/graph.c:7689)
    for vertex2 in vertex1.edges: count += old_values[self.vertices.index(vertex2)]
ValueError: <GroupAtom ['Ss']> is not in list
```

connie:
Interesting, I wonder why this function worked fine before.

nickvandewiele:
In the previous implementation of the sorting algorithm (cf. below), we did not search for the index of the neighbor vertex `self.vertices.index(vertex2)`:

```cython
    cpdef updateConnectivityValues(self):
        """
        Update the connectivity values for each vertex in the graph. These are
        used to accelerate the isomorphism checking.
        """
        cdef Vertex vertex1, vertex2
        cdef short count
        
        for vertex1 in self.vertices:
            count = len(vertex1.edges)
            vertex1.connectivity1 = count
        for vertex1 in self.vertices:
            count = 0
            for vertex2 in vertex1.edges: count += vertex2.connectivity1
            vertex1.connectivity2 = count
        for vertex1 in self.vertices:
            count = 0
            for vertex2 in vertex1.edges: count += vertex2.connectivity2
            vertex1.connectivity3 = count
```

nickvandewiele:
Possible solution strategies I see:

- Flag atoms that should not be considered in the isomorphism check rather than removing them.
The flag could be an attribute of Vertex that is set at runtime. 
The sorting algorithm will not crash anymore, but the isomorphism code of `vf2` will need to be updated.

- Fully remove/add atoms rather than only modifying `self.vertices`
There is `removeVertex` available which will update both vertices and edge. Adding the atoms and bond afterwards will require a bit more work.

- Updating the sorting algorithm such that we don't need to search the index of removed atoms

rwest:
Fully removing (including the bonds) sounds dangerous as part of an isomorphism `matchNodeToStructure` check. We'd be modifying the graph that we're checking, so if the group definition of the Node we're checking against specifies the presence of a bond, which we just removed from the structure, then we'll get a false negative?  Though I haven't looked fully at the context to see if this would be a problem here, it just feels risky.

connie:
I think I agree with Richard, go with option 3: Updating the sorting algorithm such that we don't need to search the index of removed atoms

nickvandewiele:
I agree with strategy _2_ being a risky road. I tried strategy _1_ for the following reason:

The removal of the atoms leads to a graph in which vertex/edge references are temporarily out of order. The new sorting implementation did not do implement any risky behavior, it simply searched for an index of vertex in the graph's vertex list. Hence, imho it revealed a piece of code that could lead to trouble in the future.

That being said, I do introduce a new boolean attribute in Vertex, which is in principle something inherent to the algorithm, and not to the data structure.

rwest:
Solution looks OK to me. Profiling data from Travis build suggests my last commit helped (though I've not repeated the test). All tests pass, so going to merge.

Issue #484, opened 2015-11-05
Parser does not fail politely if openbabel not installed


rwest:
There's a `try` / `except` block around the `import openbabel` in rmgpy/molecule/parser.py suggesting that it's an optional dependency, but if you try running without it you just crash with an unhelpful message:

```python
File "molecule.py", line 1173, in rmgpy.molecule.molecule.Molecule.fromSMILES (build/pyrex/rmgpy/molecule/molecule.c:19646)
File "molecule.py", line 1177, in rmgpy.molecule.molecule.Molecule.fromSMILES (build/pyrex/rmgpy/molecule/molecule.c:19525)
File "parser.py", line 425, in rmgpy.molecule.parser.fromSMILES (build/pyrex/rmgpy/molecule/parser.c:7627)
File "parser.py", line 431, in rmgpy.molecule.parser.fromSMILES (build/pyrex/rmgpy/molecule/parser.c:7509)
File "parser.py", line 193, in rmgpy.molecule.parser.__parse (build/pyrex/rmgpy/molecule/parser.c:4165)
File "parser.py", line 156, in rmgpy.molecule.parser.__fromSMILES (build/pyrex/rmgpy/molecule/parser.c:3789)
File "parser.py", line 209, in rmgpy.molecule.parser.parse_openbabel (build/pyrex/rmgpy/molecule/parser.c:4400)
NameError: name 'openbabel' is not defined
```

If we need it, we should just remove the try/catch and we'll fail with an ImportError. If we want it to remain optional, something else needs changing.

nickvandewiele:
I am working towards a solution that does not require OB necessarily.

PR #486, opened 2015-11-05
Anticipate that OB is not installed

nickvandewiele:
A patch that addresses #484

### DIFF
--- a/rmgpy/molecule/parser.pxd
+++ b/rmgpy/molecule/parser.pxd
@@ -9,6 +9,7 @@ cimport element as elements
 from .molecule cimport Atom, Bond, Molecule
 
 cpdef list BACKENDS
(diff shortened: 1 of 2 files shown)


nickvandewiele:
I successfully tested `make eg1` after removing OB from my PYTHONPATH.

nickvandewiele:
I noticed that when parsing a SMILES string, the order of the atoms returned by the various backends (OB, RDKit) actually differs for the same string.

Not that this is a big thing, but it failed some tests in which the atom indices were hardcoded, like 
https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/molecule/atomtypeTest.py#L103

Ideally, hard-coded atom numbers in unit tests should be avoided. If it is impossible not to hard-code atom numbers, they should preferably be invariant over code changes...

PR #492, opened 2015-11-13
RMG Output via observer pattern

nickvandewiele:
The idea is that we will eventually have an RMG simulation class (`rmgpy.rmg.RMG`) that is unaware (and doesn't care) of what is done with its state (reaction model, reaction system, ...).

For example: 
the chemkin file, and HTML output writers are moved to the corresponding modules that contain the code that writes the stuff. 

The idea is to decouple everything related to writing information to the user from the actual simulation. It's an easier design to extend in the future, and it could potentially be used as well to (asynchronously) offload these time consuming parts in the future.

### DIFF
--- a/rmgpy/chemkin.py
+++ b/rmgpy/chemkin.py
@@ -30,13 +30,14 @@
 """
 This module contains functions for writing of Chemkin input files.
 """
-
+import shutil
 import math
 import re
 import logging
 import textwrap
 import os.path
 import numpy
+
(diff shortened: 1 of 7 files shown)


rwest:
Looks like a reasonable idea, and includes some nice refactoring. A few more docstrings or comments may help people (our future selves) reading the code without reference to this pull request. eg. how should they set up a new "listener", when is it called, can you call one without the others, etc.

And how close is this to enabling multi-processing?

PR #493, opened 2015-11-16
RMG Listener: chapter 2

nickvandewiele:
This PR addresses some of the issues raised in https://github.com/ReactionMechanismGenerator/RMG-Py/pull/492#issuecomment-156718725:

- adding some documentation to the various implemented listeners
- remove PyDAS dependency, by duplicating the Subject class in a new `rmgpy.util` module.
- correct some bugs in the tools modules and tests
- ensure that people can choose their own listeners to be attached, by moving the `register_listeners` method from `initialize(..)` to `execute(...)`
- refactor the creation of output subdirectories: only create a subdir if the feature is actually used.
- replacing the `worksheet` functionality of `ReactionSystem` by attaching a listener (`SimulationProfileWriter`) instead.


rwest:
Looks good to me (skimming the commits online)

PR #496, opened 2015-11-18
Liquid reactor unit tests

nickvandewiele:
This PR adds a number of unit tests for the `LiquidReactor` model.

the unit tests are simply copied from simpleTest:test_Solve, refactored.

The test for the analytical jacobian currently fails, therefore added a WIP decorator.


nickvandewiele:
Probably the unit test is doing something wrong, but some quick questions:

- Was the analytical jacobian for the liquid reactor simply copied from the same method in SimpleReactor?
- Should both implementations be different?

c605ac0 seems to be the commit that added it.

rwest:
Looks like it's the same as the SimpleReactor, as the subsequent commit which converted things into LiquidReactor didn't change the Jacobian. Probably it should be different somehow.

nickvandewiele:
OK, I will remove the jacobian implementation in `LiquidReactor` to ensure that people don't mistakingly use it.

nickvandewiele:
I removed the analytical jacobian of LiquidReactor rather than commenting out the implementation and replacing it by an array of ones because it lead to solver divergence problems. ReactionSystem.advance(...) calls the analytical jacobian if it finds one.

The WIP unit test should remind future developers that this function needs attention. The previous implementation can easily be retrieved by reverting the removal commit.

Issue #495, opened 2015-11-17
Print input file near start of log file?


rwest:
This would help remove any confusion as to which version of an input file was used to create the log file!
(I think RMG-Java does this).
We could either just print it verbatim (with comments etc) or use the appropriate function to print it reformatted.

PR #500, opened 2015-11-24
print input file to log info at beginning of simulation

nickvandewiele:
cf. https://github.com/ReactionMechanismGenerator/RMG-Py/issues/495

Currently printed verbatim.


rwest:
Thanks!

PR #502, opened 2015-11-25
Extend generateReverseRateCoefficient for some PDep types.

rwest:
I think that the Troe broadening parameters etc. are 
the same in both directions, so as long as we correctly
reverse the high and low pressure limits, we should be 
able to reverse these reaction types too.

I think ThirdBody with +M and the low-pressure limits of
the PDep reaction types with (+M) should have one more
m3/mol in the units than they would otherwise.

I tried to reduce code duplication by making a special
method to reverse a given Arrhenius rate expression.

One thing this section of code would benefit from is some unit tests. Any volunteers??
(One way to test would be to reverse a reaction rate twice, and see if you end up close to where you started).

### DIFF
--- a/rmgpy/reaction.pxd
+++ b/rmgpy/reaction.pxd
@@ -84,6 +84,8 @@ cdef class Reaction:
 
     cpdef fixBarrierHeight(self, bint forcePositive=?)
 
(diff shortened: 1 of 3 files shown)


nickvandewiele:
I will add some tests.

nickvandewiele:
this PR can be merged as far as I am concerned.

PR #503, opened 2015-11-30
Database broadcast

nickvandewiele:
This PR takes the first step in supporting parallel computing in RMG-Py.

It broadcasts the thermo, kinetics, solvation, transport, and statmech databases to all workers through SCOOP.

If SCOOP is not imported or started properly, it should correctly fall back to the normal behavior as before.


nickvandewiele:
The build is currently failing because the SCOOP binary could not be found on anaconda yet. I will ensure that unit tests involving SCOOP don't fail when it could not be found, so that RMG-Py continues to work even when users are not using SCOOP.

rwest:
If we want to get towards scoop being the default, then we should also make a conda recipe for scoop and add it to our channel.

KEHANG:
I know this PR is only about broadcasting db, but I noticed the parallel module being created. I think the thing worth discussing is how future paralleled functions, say paralleled react, fit in this module?

nickvandewiele:
Scoop has already a binary in the RMG channel.

Issue #237, opened 2014-06-19
Move bimolecular/unimolecular lists to RMG-database project?


rwest:
In rmgpy.data.kinetics.common we have two lists: `BIMOLECULAR_KINETICS_FAMILIES`
and `UNIMOLECULAR_KINETICS_FAMILIES`.
I feel that these should be part of the RMG-database project, so that people can update them when they create new families.

PR #504, opened 2015-12-01
Reaction order

connie:
Addresses https://github.com/ReactionMechanismGenerator/RMG-Py/issues/237

Contrary to prior belief, we don't actually need to add the reaction family to the list of unimolecular and bimolecular families.  KineticsGroups.numReactants is capable of providing the reaction order.  However, I have migrated the hardcoded lists into the import old kinetics groups from RMG-Java function, since hardcoding in that function seems very reasonable.

### DIFF
--- a/rmgpy/cantherm/pdep.py
+++ b/rmgpy/cantherm/pdep.py
@@ -41,7 +41,7 @@ import logging
 
 import rmgpy.constants as constants
 import rmgpy.quantity as quantity
-from rmgpy.kinetics import Chebyshev, PDepArrhenius, getRateCoefficientUnitsFromReactionOrder
+from rmgpy.kinetics import Chebyshev, PDepArrhenius
 from rmgpy.reaction import Reaction
 from rmgpy.kinetics.tunneling import Wigner, Eckart
 
--- a/rmgpy/data/kinetics/common.py
+++ b/rmgpy/data/kinetics/common.py
@@ -40,58 +40,6 @@ from rmgpy.species import Species
 
 ################################################################################
 
-# The names of all of the RMG reaction families that are bimolecular
-BIMOLECULAR_KINETICS_FAMILIES = [
-    'H_Abstraction',
-    'R_Addition_MultipleBond',
-    'R_Recombination',
-    'Disproportionation',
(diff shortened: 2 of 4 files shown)


nickvandewiele:
In 9f78dda you stored the families as local variables within the scope of the `processOldLibraryEntry` method. Could we transform them into module constants, instead so that they remain accessible to other methods (in the future)?

The hard coded families still remain part of `RMG-Py` instead of migrating them to `RMG-database` like @rwest proposed. Do you think it's a bad idea to create a RMG-database dependency here? 

Instead of hard-coding, would it be possible to determine at _runtime_ whether a family is bimolecular? For example, based on the number of elements in the `reactants` variable of the `template` of a reaction family:

```python
template(reactants=["X_H_or_Xrad_H_Xbirad_H_Xtrirad_H", "Y_rad_birad_trirad_quadrad"], products=["X_H_or_Xrad_H_Xbirad_H_Xtrirad_H", "Y_rad_birad_trirad_quadrad"], ownReverse=True)

```

connie:
The reaction template length would probably be able give reaction order. However I'm not so sure where we load or parse this from a RMG-Java database file.  I'm hoping someday these Java functions could be deprecated entirely.

rwest:
Be careful: sometimes the template has additional trees that aren't reactants. Eg. a ring opening may have groups for "one end of the ring", "the other end of the ring", and "whats connecting the two ends". So a reaction with one reactant (when it comes to units of the rate constant) has three templates and would look trimolecular.

An easy solution could be to add a required "molecularity" attribute to the top of each template definition file in RMG-database. The only issue would be importing from RMG-Java, which at one point we wanted to be repeatable and automatic, but as Connie says, I actually hope never has to happen again :)
For the latter, you would need, somewhere, a hard-coded list of molecularities for each family, or to modify the RMG-Java database and code. Which might be worth it.

connie:
I checked and the way we extract numReactants does do it correctly. RMG-Py was never actively using the hard coded dictionaries. None of the functions were being actively used.

rwest:
*sigh*

connie:
So are we ok to merge this without undoing hardcoding for Java?

PR #509, opened 2015-12-08
added reaction filter by spec in output html

KEHANG:
This PR added a reaction filter in output html so that user can check if desired reactions with targeted species exist or not.


KEHANG:
don't merge, still some features to be added.

connie:
Fixed and placed in https://github.com/ReactionMechanismGenerator/RMG-Py/pull/527 instead

PR #511, opened 2016-01-03
Fix bugs in readReactionComments

KEHANG:
This PR mainly fixed two bugs

- parse comment like "! Template reaction: Disproportionation", where family info is in last token

- parse comment like "Disproportionation estimate: (Average:)". where family info is relied on checking whether the line has "estimate:" or "exact:"

### DIFF
--- a/rmgpy/chemkin.py
+++ b/rmgpy/chemkin.py
@@ -651,7 +651,7 @@ def readReactionComments(reaction, comments, read = True):
             )
             reaction.kinetics.comment = line
             
-        elif 'exact' in line or 'estimate' in line:
+        elif 'exact:' in line or 'estimate:' in line:
             index1 = line.find('[')
             index2 = line.find(']')
             template = [s.strip() for s in line[index1:index2].split(',')]
--- a/rmgpy/chemkin.py
+++ b/rmgpy/chemkin.py
@@ -560,7 +560,7 @@ def readReactionComments(reaction, comments, read = True):
             pass
         
         elif 'Template reaction:' in line:
-            label = str(tokens[-2])
+            label = str(tokens[-1])
             template = tokens[-1][1:-1].split(',')
             reaction = TemplateReaction(
                 index = reaction.index,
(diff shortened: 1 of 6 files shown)


nickvandewiele:
Is [this RMG-Website issue] (https://github.com/ReactionMechanismGenerator/RMG-website/issues/101) related?

KEHANG:
@nickvandewiele does that issue happen when you try to visualize a merged mechanism?

nickvandewiele:
Could you add a unit test?

For example, it reads in a chemkin file containing 1 reaction and its comments and check if the comments are correctly parsed?

KEHANG:
OK, I will add it.

KEHANG:
First of all, we have a `data/` folder already in `rmgpy/`, serving totally different purpose. It seems not appropriate to put the files into that folder. 
I had a quick chat with Connie about it as well, we agreed on that this `chemkin.py` and `chemkinTest.py` modules are very similar to the modules in `tools`, so I put the files in `tools/data` instead of creating a data folder directly under `rmgpy`.

KEHANG:
`chemkin_py` is for the files generated by RMG-Py, I expect there will be java files as well for unittest as well, so people later can add `chemkin_java`

nickvandewiele:
1) well, we don't have to stick to `data` as the name, something like  `test_data` could also work
2) if you think chemkin belongs to `tools`, why don't you move it there then? Now, the structure of the folders does not make sense to me.

KEHANG:
the two tests are very different in comments, the minimal model can work with the the first patch c105012, but pdd cannot work without second patch 067f422a2d63. So I have two tests here.

KEHANG:
probably you can quickly run the unittests, you'll see the difference.

KEHANG:
@nickvandewiele I've created the issue [https://github.com/ReactionMechanismGenerator/RMG-Py/issues/516] for your concern of file structure of RMG-Py. But for this PR, if no further comments, I'd like you to merge it so people can visualize their mechanisms correctly.

PR #518, opened 2016-01-07
generatePairs() unittest

KEHANG:
This PR added an unittest for problematic `generatePairs()` which currently determines pairs of the reaction non-deterministically.


connie:
Is this still necessary now that we have put in the native flux pairs?  Or is this to identify it for reactions that do not have the flux pairs in the comments?

KEHANG:
Yeah, I think most reactions should be fine since they will have flux pairs info. But non-TemplateReaction still need this method. So would be definitely better if generatePairs() can be more robust.

PR #524, opened 2016-01-13
Added flux pairs info to chemkin file

KEHANG:
This PR added

- flux pairs info to chemkin file when saving chemkin files. This info is particularly important for ROP analysis like aggregating flux by species

- when merging models, the pairs should be updated using the updated species


connie:
Placed in pull request https://github.com/ReactionMechanismGenerator/RMG-Py/pull/527 instead

PR #528, opened 2016-01-13
ignore RuntimeWarning emitted by warnScoopStartedProperly wrapper

### LINKED ISSUE #510: Runtime scoop error even when scoop flag is not used
Noticed that we now get default runtime error when running the minimal example: 

```
/Users/connie/code/RMG-Py/rmgpy/scoop_framework/util.py:77: RuntimeWarning: SCOOP was not started properly.
Be sure to start your program with the '-m scoop' parameter. You can find further information in the documentation.
```

When running without the `-m scoop` flag.  What kind of default behavior do we want?  If we want users to ALWAYS run with scoop, then why not get rid of this flag altogether and always run with scoop by default?  But if they do not use the flag, I feel that we should not by default display any sort of runtime warning.

nickvandewiele:
RuntimeWarnings will not be printed to output.

fixes #510

### DIFF
--- a/rmgpy/scoop_framework/util.py
+++ b/rmgpy/scoop_framework/util.py
@@ -52,6 +52,8 @@ def warnScoopStartedProperly(func):
         
         futures_not_loaded = 'scoop.futures' not in sys.modules
 
+        warnings.simplefilter('ignore', RuntimeWarning)
+
         try:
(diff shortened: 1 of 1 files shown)


connie:
Closed via 79749b7

Issue #521, opened 2016-01-11
SpeciesConstraints should not be an attribute of CoreEdgeReactionModel


nickvandewiele:
[A lot of water] (https://github.com/ReactionMechanismGenerator/RMG-Py/issues?utf8=%E2%9C%93&q=constraints) has flown under the bridge of molecular constraints and forbidden structures. This issue is meant to highlight some design issues from the coding point of view that will make the life hard for current and future coders of RMG-Py.

Currently, `speciesConstraints` are stored as an [instance attribute of `CoreEdgeReactionModel`] (https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/rmg/model.py#L370) AND as an [instance attribute of `rmgpy.rmg.main.RMG`] (https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/rmg/main.py#L180). In a rare use case of functional programming in RMG-Py, the method `CERM.failsSpeciesConstraints` is passed down to `KineticsFamily` through a series of helper methods like [here] (https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/rmg/model.py#L626) and alike. Hence, a lower level method is asking for information to a higher level object. This cycle of dependencies is a recipe for disaster for future coders. 

[With all the arguments considered] (https://github.com/ReactionMechanismGenerator/RMG-Py/pull/208) as to why species constraints need to be checked inside `KineticsFamily.__generateProductStructures` and not higher up, I would propose to move the species constraints away from CERM and store it in a different place. 

Like many other attributes `rmgpy.rmg.main.RMG`, the species constraints are simply read and do not change over time. One possibility is to contain this information in a global database like we do for the information loaded from `RMG-database`. This would lead to a design with less dependencies between the different modules, and happier coding in the future.

nickvandewiele:
closed by #529

PR #529, opened 2016-01-13
code redesign of species constraints

nickvandewiele:
This PR addresses some of the concerns vented in #521 regarding the manipulation of static user input information like species constraints.

In the current design:
- the `failsSpeciesConstraints` method is moved to the constraints module since it was called in various places; 
- passing the function name to various methods is deprecated: when a method needs the species constraints, it simply calls the global rmg variable that already stored that info
- the species constraints are broadcasted when the info is needed on a remote worker instead of the root worker, and a convenience function `getInput` is created that searches for that info in the global rmg variable or the broadcasted variable.

### DIFF
--- a/rmgpy/constraints.py
+++ b/rmgpy/constraints.py
@@ -1,0 +1,79 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+################################################################################
+#
+#   RMG - Reaction Mechanism Generator
+#
+#   Copyright (c) 2002-2010 Prof. William H. Green (whgreen@mit.edu) and the
+#   RMG Team (rmg_dev@mit.edu)
+#
+#   Permission is hereby granted, free of charge, to any person obtaining a
+#   copy of this software and associated documentation files (the 'Software'),
+#   to deal in the Software without restriction, including without limitation
+#   the rights to use, copy, modify, merge, publish, distribute, sublicense,
+#   and/or sell copies of the Software, and to permit persons to whom the
+#   Software is furnished to do so, subject to the following conditions:
+#
+#   The above copyright notice and this permission notice shall be included in
+#   all copies or substantial portions of the Software.
+#
+#   THE SOFTWARE IS PROVIDED 'AS IS', WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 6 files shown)


connie:
I like how `failsSpeciesConstraints` function is moved to an outside module under `rmgpy/constraints`. I know this was not in the original implementation, but is there a way to store the function attributes so it doesn't have to fill that in each time?  I guess it is probably negligible to call it from getInput, but we do use it every single time in `generateReactions()`.

The `getInput` function itself seems a bit unecessary.  What is the reasoning for not just pulling from `rmg.speciesConstraint`s directly?  ie. in the `failsSpeciesConstraints` just use
`rmg.speciesConstraints` by the global rmg broadcast, but not having to go through an additional function?  Or are you planning to do something like quickly read the input file rather than looking at the global RMG object in the future?  In that case, it should probably parse the input.py file rather than return rmg.speciesConstraints.

nickvandewiele:
the `getInput` function is essential to support both serial and parallel runs of RMG-Py.

In the parallel case, the global `rmg` variable populated with the `speciesConstraints` data will only be accessible in the process that started the whole RMG simulation (i.e. on the `root worker`).

On other processes that are spawned by the root process, none of the global data is available by simply importing it. First, that info has to be broadcasted by the root process, and then the slave process needs to explicitly look for a variable with the broadcasted name.

The `getInput` provides a uniform interface to access info that is stored in two different ways, regardless of the nature of the process it is called from.

I use a similar mechanism to retrieve data from the RMG-database.

connie:
I understand.  Is the reason why it is under `rmg/input.py` because we want it to be broadcasted immediately after initialization of the job?  Do you also forsee broadcasting other variables to that function from `input.py`?

nickvandewiele:
My intention is to use the same mechanism to _share_ user input information that is loaded through the `input` module. The `getInput` function would be expanded with an `elif` and the keyword that corresponds to the broadcasted info.

Like with the RMG-database data, I choose to broadcast the data, the moment it is loaded from the disk. The idea was that, the sooner you broadcast a variable to the workers, the sooner you can use that worker to do work with that data.

nickvandewiele:
Regarding, your comment on caching the constraints: 
I currently don't see a straightforward way of doing that.

But I also don't have any data that shows that this is expensive. Each line seems like a dictionary lookup.

PR #530, opened 2016-01-14
Parallel generation of reactions in `enlarge`

nickvandewiele:
This PR introduces a new design for the generation of reactions in parallel.

In the `rmgpy.rmg.model.CoreEdgeReactionModel.enlarge` method, reactions between a new core species and the existing list of core species, for the given list of available reaction families. This process is parallelized across the reaction families, across the list of core species and across the list of resonance isomers of a specific species both for uni- and bimolecular reaction families to minimize worker load imbalances due to the _heterogeneity_ of reaction families, species, and molecules. 

A lazy evaluation scheme of the tasks is chosen to maximize the number of tasks being run at the same time.

- A uniform `rmgpy.scoop_framework.util.map_` interface is introduced that either uses the parallel `map` function of SCOOP, or the serial map of python if scoop is not loaded.
- A new test target `make scoop` in Makefile is created to illustrate the launch of the parallel version using the minimal example.
- A uniform `rmgpy.data.rmg.getDB` interface allows retrieving RMG-database info, either from the global `rmgpy.data.rmg.database` variable, or from broadcasted variables when that global variable is not available (e.g. on remote workers).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,6 +136,14 @@ eg7: all
 	@ echo "Running gri_mech_rxn_lib example."
 	python rmg.py testing/gri_mech_rxn_lib/input.py
 	
+scoop: noQM
+	mkdir -p testing/scoop
+	rm -rf testing/scoop/*
+	cp examples/rmg/minimal/input.py testing/scoop/input.py
+	coverage erase
+	@ echo "Running minimal example with SCOOP"
+	python -m scoop -n 2 rmg.py -v testing/scoop/input.py
+
 ######### 
 # Section for setting up MOPAC calculations on the Travis-CI.org server
 ifeq ($(TRAVIS),true)
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1128,10 +1128,13 @@ class KineticsFamily(Database):
         Return ``True`` if the molecule is forbidden in this family, or
         ``False`` otherwise. 
         """
-        from rmgpy.data.rmg import database
+        from rmgpy.data.rmg import getDB
+        
+        forbidden_structures = getDB('forbidden')
+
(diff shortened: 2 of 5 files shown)


nickvandewiele:
`make eg1` with 4 workers. Note that the work among the 3 workers is equally distributed. The imbalance between the root worker and the 3 others is because the parallelized `react` method is only requiring only a small portion of work to be done. Broker queue length is not getting out of hand.

task_debug

time_debug

broker_debug

connie:
Nice work! Have you tested on other jobs for matching model results?

nickvandewiele:
butane oxidation (`make eg1` but with _butane_ instead of ethane), 8 workers, different machine.

```
Updating RMG execution statistics...
    Execution time (DD:HH:MM:SS): 00:12:42:01
    Memory used: 1669.14 MB


MODEL GENERATION COMPLETED

The final model core has 84 species and 1132 reactions
The final model edge has 31756 species and 69025 reactions
```

The worked time is well distributed among the 7 non-root workers:
time_debug

The root worker does more work in total, because this PR only parallelizes the `react` method. 

The number of tasks per worker seems constant across the non-root workers. The root worker has many more, not sure how to explain this. 

task_debug

This inequality between root/non-root workers is confirmed by the number of tasks completed per second (work density). The root worker has a different color than the others:

density_debug

Maybe it is related to the fact that the root process is still alive on the root worker when another task is assigned to the root worker during the time the root process has to wait for all the molecules to finish reacting.

**In conclusion:**
For a larger fuel, on a different machine with a different number of workers, the parallelization seems to be carried nicely: 
* The work is uniformly distributed across the workers despite the large differences in workload between tasks.
* The speed up is apparent when comparing the worked time by the root worker (equal to the wall clock time) to the worked time by each worker.

PR #532, opened 2016-01-15
sort reactant and product lists in checkForExistingReaction

nickvandewiele:
this ensures that :

if (rxn0.reactants == rxn.reactants and rxn0.products == rxn.products) or \
   (rxn0.reactants == rxn.products and rxn0.products == rxn.reactants):

does not produce false negatives because of order differences of the
list elements


nickvandewiele:
I accidentally left this out in #507, not knowing that it was important for reaction equality check in `checkForExistingReaction`.

However, it still does not solve duplicate reactions when p-dep is activated. I will report this in another issue.

PR #535, opened 2016-01-22
Filter Reactions during Enlarge Step in Flux Algorithm

connie:
These commits change our flux algorithm such that it prevents species from reacting when their concentrations are deemed too low at reaction conditions.  

screenshot 2016-01-22 11 13 43

screenshot 2016-01-22 11 13 59

**Key Variables**
`unimolecularThreshold` and `bimolecularThreshold`
- Binary arrays storing flag for whether a species or a pair of species are above a reaction threshold
- Threshold is set to True if rate = k_max*CA > toleranceMoveToCore*ratechar at any given time t in the reaction system
- Unimolecular k_max = kB*T/h
- Bimolecular k_max = k_diffusion = 1e13 cm^3/mol*s

`unimolecularReact` and `bimolecularReact`
- Binary arrays storing flags for when the unimolecularThreshold or bimolecularThreshold flag shifts from False to True 
- RMG should then react those species together

**Other considerations**
The `R_Recombination` reaction family is treated as an exception.  These reactions are now always added to core by default, because filtering method tends to omit them due to being the only case where 2 radicals combine.  They generally appear in a converged model, but may not enter the core until later, and may affect models with larger compounds or models that are not converged. This class of reactions are very important in models as a radical source and sink.  

- Bill also suggests that potentially other radical sources and sinks such as disproportionation may have to be treated differently as well.. which is for later consideration.

two radicals 
Note: Do not pull yet, still making some changes and need to perform tests.

### DIFF
--- a/rmgpy/chemkin.py
+++ b/rmgpy/chemkin.py
@@ -1959,7 +1959,7 @@ class ChemkinWriter(object):
     rmg.detach(listener)
     
     """
-    def __init__(self, outputDirectory):
+    def __init__(self, outputDirectory=''):
         super(ChemkinWriter, self).__init__()
         makeOutputSubdirectory(outputDirectory, 'chemkin')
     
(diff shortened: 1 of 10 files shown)


connie:
Oops not from official branch.

PR #537, opened 2016-01-22
Filter Reactions during Enlarge Step in Flux Algorithm

connie:
These commits change our flux algorithm such that it prevents species from reacting when their concentrations are deemed too low at reaction conditions.  

screenshot 2016-01-22 11 13 43

screenshot 2016-01-22 11 13 59

**Key Variables**
`unimolecularThreshold` and `bimolecularThreshold`
- Binary arrays storing flag for whether a species or a pair of species are above a reaction threshold
- Threshold is set to True if rate = k_max*CA > toleranceMoveToCore*ratechar at any given time t in the reaction system
- Unimolecular k_max = kB*T/h
- Bimolecular k_max = k_diffusion = 1e13 cm^3/mol*s

`unimolecularReact` and `bimolecularReact`
- Binary arrays storing flags for when the unimolecularThreshold or bimolecularThreshold flag shifts from False to True 
- RMG should then react those species together

**Other considerations**
The `R_Recombination` reaction family is treated as an exception.  These reactions are now always added to core by default, because filtering method tends to omit them due to being the only case where 2 radicals combine.  They generally appear in a converged model, but may not enter the core until later, and may affect models with larger compounds or models that are not converged. This class of reactions are very important in models as a radical source and sink.  

- Bill also suggests that potentially other radical sources and sinks such as disproportionation may have to be treated differently as well.. which is for later consideration.

Note: Do not pull yet, still making some changes and need to perform tests.

**Comparison test with filterReactions=False**

`filterReactions=False`

- eg1: core identical, edge has non-identical kinetics (probably non-deterministic database)

```
		Error: Non-identical kinetics for
		tested [CH2]C[CH]C(45) + [CH2][CH]C=C(39) <=> [CH2]CC=C(28) + C=C[CH]C(41)
		and original [CH2]C[CH]C(45) + [CH2][CH]C=C(43) <=> C=C[CH]C(39) + [CH2]CC=C(28) reaction.
		Error:         k(300K,1bar) k(400K,1bar) k(500K,1bar) k(600K,1bar) k(800K,1bar) k(1000K,1bar) k(1500K,1bar) k(2000K,1bar) 
		Error:            8.50    8.15    7.96    7.85    7.73    7.69    7.67    7.70
		Error:         k(300K,1bar) k(400K,1bar) k(500K,1bar) k(600K,1bar) k(800K,1bar) k(1000K,1bar) k(1500K,1bar) k(2000K,1bar) 
		Error:            7.37    7.30    7.25    7.21    7.16    7.12    7.05    7.00
		Error: Non-identical kinetics for
		tested [CH2]CC=C(28) + C=C[CH]C(41) <=> C=CCC(27) + C=CC=C(30)
		and original C=C[CH]C(39) + [CH2]CC=C(28) <=> C=CC=C(30) + C=CCC(27) reaction.
		Error:         k(300K,1bar) k(400K,1bar) k(500K,1bar) k(600K,1bar) k(800K,1bar) k(1000K,1bar) k(1500K,1bar) k(2000K,1bar) 
		Error:            7.19    7.15    7.12    7.09    7.04    7.01    6.95    6.91
		Error:         k(300K,1bar) k(400K,1bar) k(500K,1bar) k(600K,1bar) k(800K,1bar) k(1000K,1bar) k(1500K,1bar) k(2000K,1bar) 
		Error:            5.10    5.27    5.40    5.49    5.64    5.75    5.94    6.08
```

- eg2: micocanonical rate errors for both
- eg3: Identical
- eg5: Identical
- eg4: Not tested because it is thermoEstimator
- eg6: Identical
- eg7: Identical

**Model Comparison Test for a Heptane example with tolerance = .05**

Running time this branch with filterReactions = True: 11 min
Running time master branch = 72 min

Final core model size this branch: 73 species, 479 reactions
Final core model size master branch: 73 species, 493 reactions

This branch:
image
Master branch:
image

This branch:
image
Master branch:
image

For this particular model, the accuracy and size of the final model is nearly identical.  However with at least 7x speedup.


rwest:
Not sure how practical this is (haven't looked at commits in detail) but it would be nice if the general enabling changes to the master that don't actually change the algorithm, eg. (from the sound of it) "Allow OutputHTMLWriter and ChemkinWriter to be initialized with no directory", could be rebased to the beginning of the PR and merged first. Makes reviewing easier.
Not essential if interactive rebasing causes conflicts, but it's a nice thing to strive for.

connie:
Ok, will do.

nickvandewiele:
`rmgpy.rmg.model.CERM.enlargeEdge` and `rmgpy.rmg.model.CERM.enlargeCore` have large sections that are duplicates.

E.g.:
- Generate thermodynamics of new species
- Generate kinetics of new reactions
- For new reactions, convert ArrheniusEP to Arrhenius, and fix barrier heights
- Update unimolecular (pressure dependent) reaction networks
- Check new core and edge reactions for Chemkin duplicates

I would heavily advocate to refactor those parts into separate methods that are called in the respective methods.

nickvandewiele:
It's unclear to me, but it seems that the local variable `newReactions` variable in 

https://github.com/ReactionMechanismGenerator/RMG-Py/blob/enlargeReactionFilter/rmgpy/rmg/model.py#L814

will always be empty.

If so, then why is 
```
self.processNewReactions(newReactions, newSpecies, pdepNetwork)
```
called in that section. Since the main loop in `processNewReactions` iterates over that list, nothing will happen.

It's really unclear to what happens when `enlargeCore(self, newObject)` is called on a `Species` object.

Also, please remove those commented lines. Deprecated code will only confuse people more.

connie:
Added new commits to do as you suggested.  Simply consolidated enlargeEdge and enlargeCore into the same single function enlarge() as before.  I guess I coded it separately at first because it was easier.  Even though the function is still a bit too long for my liking, it no longer contains the code duplication.  

As you also mention, processNewReactions is uneeded in the enlargeCore step when a new species object is added.

Issue #533, opened 2016-01-19
Eckart tunneling factor failed unit test in 32-bit OS's


connie:
This particular unit test fails for 32-bit systems, both in Linux and Windows.  All other tests pass.  However, both 64-bit architectures pass.

```
======================================================================
FAIL: Test the Eckart.calculateTunnelingFactor() method.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Code\RMG-Py\rmgpy\kinetics\tunnelingTest.py", line 146, in test_calcu
lateTunnelingFactor
    self.assertAlmostEqual(kexp, kact, delta=1e-3*kexp)
AssertionError: 1623051.0 != 1344137.1281569458 within 1623.0509999999999 delta
```

PR #546, opened 2016-01-25
Fix 32-bit precision problem in eckhart tunneling value

connie:
This closes https://github.com/ReactionMechanismGenerator/RMG-Py/issues/533

By using numpy.cosh rather than libc's cosh, we can eliminate the 32-bit precision test which fails for eckhart tunneling.

It appears that we do not use the libc version of the cosh function anywhere else in the code, so hopefully this resolve these sorts of issues for good.

### DIFF
--- a/rmgpy/kinetics/tunneling.pyx
+++ b/rmgpy/kinetics/tunneling.pyx
@@ -244,7 +244,7 @@ cdef class Eckart(TunnelingModel):
             # If all of 2*pi*a, 2*pi*b, and 2*pi*d are sufficiently small,
             # compute as normal
             if twopia < 200. and twopib < 200. and twopid < 200.:
-                kappa[r] = 1 - (cosh(twopia-twopib)+cosh(twopid)) / (cosh(twopia+twopib)+cosh(twopid))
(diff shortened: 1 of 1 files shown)


nickvandewiele:
great catch! 

Is this a problem of `cosh` specifically, or should we replace libc.math calls more generally?

connie:
I don't quite know.... the other functions we use are things like exp and abs, which haven't given any problems so far, yet nobody has tested them.

nickvandewiele:
I haven't done any timings on these functions, but if numpy timings for these function are close the one of the libc cython library, maybe we could get rid of the entire library?

rwest:
It might not be worth sweating for the occasional cosh call, but a quick google suggests that for cythonized code, `libc.math` is faster than numpy.
http://stackoverflow.com/questions/18593308/tips-for-optimising-code-in-cython

connie:
I don't have confidence that I can fix the precision error with `libc.math.cosh` so I'd rather have it be a little slower than have a silent error for 32-bit systems

rwest:
Oh, absolutely. I like your fix (though I do wonder how numpy do it more precisely). My comment was addressed to Nick's suggestion of removing all calls to libc.math.

Issue #515, opened 2016-01-07
reaction library reactions are not added to core when all species are in core


nickvandewiele:
I ran `generateReactions` script of RMG-Py v1.0.1.

with the following input file:
```
# Data sources
database(
    thermoLibraries = ['primaryThermoLibrary'],
    reactionLibraries = [('Methylformate', False)],
    kineticsFamilies = ['R_Addition_MultipleBond'],
    kineticsEstimator = 'rate rules',
)

# List of species

species(
    label='HCjO',
    reactive=True,
    structure=adjacencyList(
        """
multiplicity 2
1 C u1 p0 c0 {2,D} {3,S}
2 O u0 p2 c0 {1,D}
3 H u0 p0 c0 {1,S}
        """),
)

species(
    label='CH2O',
    reactive=True,
    structure=adjacencyList(
        """
1 C u0 p0 c0 {2,D} {3,S} {4,S}
2 O u0 p2 c0 {1,D}
3 H u0 p0 c0 {1,S}
4 H u0 p0 c0 {1,S}
        """),
)

species(
    label='Fmoml',
    reactive=True,
    structure=adjacencyList(
        """
multiplicity 2
1 C u1 p0 c0 {3,S} {5,S} {6,S}
2 C u0 p0 c0 {3,S} {4,D} {7,S}
3 O u0 p2 c0 {1,S} {2,S}
4 O u0 p2 c0 {2,D}
5 H u0 p0 c0 {1,S}
6 H u0 p0 c0 {1,S}
7 H u0 p0 c0 {2,S}
        """),
)


simpleReactor(
    temperature=(1399,'K'),
    pressure=(1.93,'atm'),
    initialMoleFractions={
        "HCjO": 1,
    },
    terminationConversion={
        'HCjO': 0.999,
    },
    terminationTime=(1e-3,'s'),
)
```
The input file imports a reaction library (methylformate), and defines 3 input species that (HCjO, CH2O, Fmoml).

By default, the reactions from the reaction library will be added to the edge, while the 3 input species will be added to the core.

The reaction library contains 1 reaction involving those three species. Nevertheless, this library reaction does not make it to the core although all the species of the reaction belong to the core already. This is shown in the log below.

```
Summary of Model Enlargement
----------------------------
Added 1 new core species
    Fmoml(3)
Created 6 new edge species
    [CH2]OC([O])C=O(16)
    [CH2]O[CH]OC=O(17)
[...]

PR #547, opened 2016-01-26
Fix issue: reaction library rxns containing all input species not added to core

connie:
This closes https://github.com/ReactionMechanismGenerator/RMG-Py/issues/515

By adding the input species to the edge in the step before reaction libraries are added to edge,
we make sure that when the input species is added, a reaction on the edge containing all input species
can get added to the core.  

The source of the bug is due to the function:

 `spec, isNew = rmg.reactionModel.makeNewSpecies()`

giving `False` even though the species is in NEITHER the core nor edge.  (The input species are created but not yet added anywhere).

This was the simplest way I could think of for fixing the bug.  There should be no other instances in which a species is not in either the core or edge when it is created when this function returns `False`.

### DIFF
--- a/rmgpy/rmg/input.py
+++ b/rmgpy/rmg/input.py
@@ -96,6 +96,9 @@ def species(label, structure, reactive=True):
     spec, isNew = rmg.reactionModel.makeNewSpecies(structure, label=label, reactive=reactive)
     if not isNew:
         raise InputError("Species {0} is a duplicate of {1}. Species in input file must be unique".format(label,spec.label))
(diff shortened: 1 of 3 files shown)


nickvandewiele:
you could add a test inspired by this one:
https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/tools/testGenerateReactions.py#L26

and change the input file to add a reaction library, rather than a seed mech.
https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/tools/data/generate/duplicates/input.py

Issue #554, opened 2016-02-01
make eg3 crashes by memory error


nickvandewiele:
When running `make eg3` with v1.0.2 of RMG-Py, it crashes as follows:

```
After model enlargement:
    The model core has 12 species and 11 reactions
    The model edge has 95 species and 289 reactions

[*** glibc detected *** python: free(): invalid next size (fast): 0x000000000560b040 ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x7db26)[0x2b2218063b26]
/opt/anaconda/anaconda/envs/rmg_env/lib/python2.7/site-packages/numpy/core/multiarray.so(+0x1f5af)[0x2b221a6145af]
/home/nickv/ws/dev/RMG-Py/rmgpy/molecule/molecule.so(+0x9d3c)[0x2b2237c1fd3c]
/opt/anaconda/anaconda/envs/rmg_env/bin/../lib/libpython2.7.so.1.0(+0x134bce)[0x2b22173fabce]
/opt/anaconda/anaconda/envs/rmg_env/bin/../lib/libpython2.7.so.1.0(_PyObject_GC_Malloc+0xfe)[0x2b22173fb43e]
/opt/anaconda/anaconda/envs/rmg_env/bin/../lib/libpython2.7.so.1.0(PyType_GenericAlloc+0x49)[0x2b2217374599]
/home/nickv/ws/dev/RMG-Py/rmgpy/molecule/graph.so(+0x5988)[0x2b2237e7c988]
/home/nickv/ws/dev/RMG-Py/rmgpy/molecule/molecule.so(+0x16d08)[0x2b2237c2cd08]
/home/nickv/ws/dev/RMG-Py/rmgpy/molecule/graph.so(+0x1a98d)[0x2b2237e9198d]
/home/nickv/ws/dev/RMG-Py/rmgpy/molecule/molecule.so(+0x188de)[0x2b2237c2e8de]
/home/nickv/ws/dev/RMG-Py/rmgpy/molecule/molecule.so(+0xd7a5)[0x2b2237c237a5]
/opt/anaconda/anaconda/envs/rmg_env/bin/../lib/libpython2.7.so.1.0(PyEval_EvalFrameEx+0x8217)[0x2b22173c2747]
[...]

PR #556, opened 2016-02-01
Fix liquid reactor bug

nickvandewiele:
This addresses a small typo in base.pyx

it reused the `index` variable with the value it was last set to.

On the bright side, this serious bug did not lead to model differences for other simulations, which begs the question: "how is that even possible?".


nickvandewiele:
cf #554

nickvandewiele:
Another open question that remains: 
"how did this typo lead to a crash for make eg3"

Under the assumption that the value of `index` was set by this:
```
        speciesIndex = {}
        for index, spec in enumerate(coreSpecies):
            speciesIndex[spec] = index
```

and erroneously reused here:
```
                if maxCoreSpeciesRates[index] < coreSpeciesRates[index]:
                    maxCoreSpeciesRates[index] = coreSpeciesRates[index]
```

it should not lead to a crash: 
the size of the `maxCoreSpeciesRates` array is equal to the size of the `speciesIndex` dictionary because both data structures refer to the number of core species.

PR #560, opened 2016-02-02
Add git HEAD of RMG-database to output log.

connie:
This is very useful for tracking what version of RMG-database we are on.

Unfortunately we cannot track this is people are using a binary version of RMG-database.  Since RMG-database is not a package, we cannot check rmgdatabase.__version__  However, we can probably assume they are using the package corresponding to the version of rmgpy binary they are using.

### DIFF
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -781,14 +781,13 @@ class RMG(util.Subject):
         logging.info('')
         logging.info('RMG execution terminated at ' + time.asctime())
     
-    def getGitCommit(self):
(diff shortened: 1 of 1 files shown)


rwest:
Looks good. We could modify the RMG-database Conda recipe so the commit hash (or some other version identifier) gets hard coded into the database at compile time.

connie:
It's possible to actually using `conda list rmg-database` or `conda list rmg-py` to check the package as well. Perhaps we should call that if the git HEAD returns nothing in order to print the package name to check if they have them as binaries.

PR #561, opened 2016-02-02
avoid RMG-tests branch name collisions when deploying through travis

nickvandewiele:
before, when a RMG-Py travis build was triggered, it created a branch on the RMG-tests repo with the same name as the one that was being pushed by RMG-Py. That lead to name collisions, e.g. for master.

Now we prepend the branch name with `rmgpy-`.

The 2nd commit avoids loading matplotlib when we don't need it, which should speed up travis builds.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -64,12 +64,14 @@ cd $TARGET_DIR
 
 # create a new branch in RMG-tests with the name equal to
 # the branch name of the tested RMG-Py branch:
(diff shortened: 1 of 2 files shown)


nickvandewiele:
This is a summary of the latest build times of the RMG-Py / RMG-tests integration:

image

the compilation step is by far the largest chunk of the total time to run the build.

Issue #558, opened 2016-02-02
Pdep likely broken


connie:
Running a pdep job, it appears that it is stuck in a loop where nothing new gets added.  
```
Summary of Secondary Model Edge Enlargement
---------------------------------
Added 0 new core species
Created 0 new edge species
Added 0 new core reactions
Created 0 new edge reactions

After model enlargement:
    The model core has 17 species and 27 reactions
    The model edge has 112 species and 263 reactions


Saving current model core to Chemkin file...
Chemkin file contains 27 reactions.
Saving current model to verbose Chemkin file...
Chemkin file contains 27 reactions.
Updating RMG execution statistics...
    Execution time (DD:HH:MM:SS): 00:00:01:30
--
Updating 0 modified unimolecular reaction networks...

```

Towards the end we have this behavior, when grepping 'model core has' RMG.log:
```

--
    The model core has 29 species and 190 reactions
    The model edge has 864 species and 3573 reactions

--
    The model core has 29 species and 190 reactions
    The model edge has 864 species and 3573 reactions

--
    The model core has 29 species and 190 reactions
    The model edge has 864 species and 3573 reactions

--
    The model core has 29 species and 190 reactions
    The model edge has 864 species and 3573 reactions

--
    The model core has 29 species and 190 reactions
    The model edge has 864 species and 3573 reactions

--
    The model core has 29 species and 190 reactions
    The model edge has 864 species and 3573 reactions

```
Guess we didn't have a very good Pdep example to check this.... =/

connie:
Here is the `input.py` file

```
# Data sources
database(
    thermoLibraries = ['primaryThermoLibrary'],
    reactionLibraries = [],
    seedMechanisms = [],
    kineticsDepositories = ['training'], 
    kineticsFamilies = 'default',
    kineticsEstimator = 'rate rules',
)

# Constraints on generated species
generatedSpeciesConstraints(
    maximumCarbonAtoms = 7,
)

# List of species
species(
    label='n-heptane',
    structure=SMILES("CCCCCCC"),
)

species(
    label='Ar',
    reactive=False,
    structure=SMILES("[Ar]"),
)

simpleReactor(
    temperature=(1000,'K'),
    pressure=(400,'Pa'),
    initialMoleFractions={
        "n-heptane": 0.02,
        "Ar": 0.98,
    },
    terminationConversion={
        'n-heptane': 0.99,
    },
    terminationTime=(1e6,'s'),
)

[...]

PR #563, opened 2016-02-03
Fix pdep job convergence issue

connie:
Fixes https://github.com/ReactionMechanismGenerator/RMG-Py/issues/558

It turns out that pdep jobs do not produce the wrong models, but the convergence takes too long because `RMG.main` was calling `enlarge(reactEdge=True)` too often without updates to `unimolecularReact` and `bimolecularReact` arrays when filterReactions was off.

This fixes the issue and also restructures the `updateReactionThresholdAndReactFlags` function to have better readability.

Also the heptane-eg5 example was updated to reflect a pdep example that takes a little longer to converge without this fix.

### DIFF
--- a/examples/rmg/heptane-eg5/input.py
+++ b/examples/rmg/heptane-eg5/input.py
@@ -16,7 +16,6 @@ generatedSpeciesConstraints(
 # List of species
 species(
     label='n-heptane',
-    reactive=True,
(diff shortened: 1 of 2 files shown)


connie:
Checks on RMG-tests confirm that this does not change eg1-6.

Issue #564, opened 2016-02-03
Overflow error


abdulganij:
Hi

I tried to generate a mechansim for 30 carbon molecule (n-alkane) but RMG is unable to generate a mechasnim.  Screenshot attached.

Is there an upper C no limit above which RMG will not work.
Please help.

INPUT FILE

# Data sources
database(
    thermoLibraries = ['primaryThermoLibrary'],
    reactionLibraries = [],
    seedMechanisms = [],
    kineticsDepositories = ['training'],
    kineticsFamilies = 'default',
    kineticsEstimator = 'rate rules',
)

# List of species
species(
    label='triacontane',
    reactive=True,
    structure=SMILES("CCCCCCCCCCCCCCCCCCCCCCCCCCCCCC"),
)

# Reaction systems
simpleReactor(
    temperature=(1350,'K'),
    pressure=(1.0,'bar'),
    initialMoleFractions={
        "triacontane": 1.0,
    },
    terminationConversion={
        'triacontane': 0.9,
    },
    terminationTime=(1e7,'s'),
)

simulator(
    atol=1e-16,
    rtol=1e-8,
)

model(
    toleranceKeepInEdge=0.0,
    toleranceMoveToCore=0.1,
    toleranceInterruptSimulation=0.1,
    maximumEdgeSpecies=100000,
)

options(
    units='si',
    saveRestartPeriod=None,
    generateOutputHTML=True,
    generatePlots=True,
    saveEdgeSpecies=True,
    saveSimulationProfiles=True,
)
![capture](https://cloud.githubusercontent.com/assets/17053949/12797218/dfb41bd4-cad3-11e5-9d60-b0d2793c03f4.PNG)

connie:
So it seems that we have a unit test here: https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/molecule/moleculeTest.py#L1300

On my two 64 bit computers (mac and linux), I do not get any int overflow with 30 carbons, nor even with 60 carbon n-alkane.  but the error begins once I try n-C80. @nick do you see any difference in behavior with your 32-bit machine?

rwest:
Hi.
There will always be some finite upper limit.
It seems you have found the limit in the current implementation. 
We may be able to the particular crash you found, and then the limit will be higher. Eventually the limit will just be that it takes infinitely long time to find all the many many reactions feasible with such a large molecule (but at least it shouldn't crash).

[...]

PR #570, opened 2016-02-12
revert the vertex connectivity value algorithm

nickvandewiele:
revert it back to the triple connectivity values per Vertex,
which has less discriminatory power, but is more robust for larger
molecules.

should address #564


nickvandewiele:
do not merge yet, I will add a number of unit tests that create big molecules.

PR #575, opened 2016-02-12
Add a function for getting polycyclic cycles only

connie:
Was originally intending to return all distinct cycles, but this function now correctly returns a list of only the polycyclic cycles, which will be useful for fitting polycyclic thermo groups.


nickvandewiele:
I am unsure whether I understand the concept that is returned by the method. Could you define what you mean by a _continuousCycle_?

Could you add a unit tests that validates a simple example? 

For norbornane, consisting of 2 fused 5-membered rings, what is returned?

Issue #548, opened 2016-01-26
updateLonePairs() is not working as expected for SiH2(S)


bslakman:
I noticed this when trying to saturate a radical species in order to get thermo. In the Molecule.saturate() method, we call Molecule.updateLonePairs() after adding hydrogen(s). 

This unit test currently fails with AssertionError: 2 != 1 
```
    def testUpdateLonePairs(self):
        adjlist = """
1 Si u0 p1 c0 {2,S} {3,S}
2 H  u0 p0 c0 {1,S}
3 H  u0 p0 c0 {1,S}
"""

        mol = Molecule().fromAdjacencyList(adjlist)
        mol.updateLonePairs()
        lp = 0
        for atom in mol.atoms:
            lp += atom.lonePairs
        self.assertEqual(lp, 1)
```

This error is caused by this line in updateLonePairs(), which will never work if we have 0 radical electrons, two single bonds and 1 lone pair
```
atom1.lonePairs = 4 - atom1.radicalElectrons - int(order)
```

This works if atom1 is Nitrogen. A quick fix would be a series of 'if' statements for whether atom1 is N, Si, O, S, or C, but this will require updating anytime a new atom is added.

PR #582, opened 2016-02-22
Fix update lone pairs issue for Silicon.

bslakman:
Addresses #548.


nickvandewiele:
Thanks @bslakman for solving this issue!

1) Could you perhaps create a new branch, say `update_lone_pairs` with these changes and create a new PR?

2) Do you think the problem of wrong lone pair counts is more widespread than Silicon?

3) I see your formula uses integer arithmetic, such that 1/2 = 0. Is that done intentionally?

PR #583, opened 2016-02-22
Create rmgpy.tools.plot submodule for plotting simulations and sensitivities

connie:
Create `GenericData` and `GenericPlot` classes, which can be inherited to create more complex plotting classes.  These classes use matplotlib to generate plots intelligently.

From these classes, 3 new classes are built for easy plotting within RMG:
- `SimulationPlot`
- `ReactionSensitivityPlot`
- `ThermoSensitivityPlot`

Let simulation plots be automatically plotted if  `saveSimulationProfiles` is set to `True`.  This adds some CPU time but is worthwhile and uses a negligible fraction of time for larger jobs.  We can directly go from reaction objects to plotting... but for now use csv parser, since we can plot from any csv file in the correct format.

Sensitivity bar plots are now plotted automatically when sensitivity is turned on, or when the sensitivity script is used.

Test scripts can be found in https://github.com/connie/rmg_wiki/tree/master/tools in `plotSimulation` and `plotSensitivity`


nickvandewiele:
So, `RMG-tests` fails because a clean travis VM propably does not have the proper renderer
variables set.
```
RuntimeError: Invalid DISPLAY variable
make: *** [run] Error 1
```

connie:
Any way to make it set those variables properly?

PR #584, opened 2016-02-23
Update lone pairs

bslakman:
This is an update to pull request #582 which I will now close. 

@nickvandewiele to answer your questions, yes this will also be an issue for Carbon and possibly we never caught it earlier because CH2 singlet is less common. I updated the pull request to reflect this.
Also, I did not consider the integer arithmetic, but I think it's ok. We should not have fractional numbers of lone pairs, and I also cannot think of any situation (for carbon or silicon) where the number on the left side would be odd.

### DIFF
--- a/rmgpy/molecule/molecule.py
+++ b/rmgpy/molecule/molecule.py
@@ -286,6 +286,13 @@ class Atom(Vertex):
         """
         return self.element.number == 8
 
+    def isSilicon(self):
+        """
+        Return ``True`` if the atom represents an silicon atom or ``False`` if
+        not.
+        """
+        return self.element.number == 14
+
     def incrementRadical(self):
         """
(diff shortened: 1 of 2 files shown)


rwest:
Could you rebase it again, and hope that it magically passes Travis this time?

nickvandewiele:
I am working on this, because it's an authentication issue, and not related to code as far as I understand. Travis CI builds of PR's from users outside the RMG organization currently crash.

nickvandewiele:
cf #625

bslakman:
So is the solution to push changes to an official branch, and then pull request from there?

rwest:
I think it is now fixed so if you rebase once more it will (mostly) work.

rwest:
@nickvandewiele please could you take a quick look and merge this? Thanks

PR #586, opened 2016-02-25
Improve thermo estimation for cyclics

KEHANG:
Previously the ring correction is always same for all the rings in molecule. This PR fixed this issue by feeding information of each ring when adding ring correction thermo data. A demonstration of this PR can be found [here](https://github.com/KEHANG/rmg_issue/blob/master/thermo_issues/thermo_correction_for_multiple-ring_molecules.ipynb).


nickvandewiele:
can you add some unit tests?

PR #587, opened 2016-02-26
Add a profile decorator function to CPU profile any function of RMG

nickvandewiele:
Simply add the `@profilefn` decorator to a function `func`, and it will run that function through a cProfile run, and print out the statistics, and CPU graph to a pdf named `func.pdf`

```python
@profilefn
def func(...):
...
```


connie:
Good idea, but the profilefn needs to be imported from the module before use.  Can you write a short document with instructions on how to use this exactly in the Wiki?  Any other notes for best practices for profiling would also be nice.  

Also running into the following error when the `@profilefn` decorate is attached to the `parseCSVData()` function in `rmgpy/tools/plot.py`:

```
  File "rmg.py", line 156, in <module>
    cProfile.runctx(command, global_vars, local_vars, stats_file)
  File "/Users/connie/anaconda/envs/rmg_env/lib/python2.7/cProfile.py", line 49, in runctx
    prof = prof.runctx(statement, globals, locals)
  File "/Users/connie/anaconda/envs/rmg_env/lib/python2.7/cProfile.py", line 140, in runctx
    exec cmd in globals, locals
  File "/Users/connie/code/RMG-Py/rmgpy/rmg/main.py", line 567, in execute
    filterReactions=False,
  File "rmgpy/solver/base.pyx", line 346, in rmgpy.solver.base.ReactionSystem.simulate (build/pyrex/rmgpy/solver/base.c:11879)
    cpdef simulate(self, list coreSpecies, list coreReactions, list edgeSpecies, list edgeReactions,
  File "rmgpy/solver/base.pyx", line 565, in rmgpy.solver.base.ReactionSystem.simulate (build/pyrex/rmgpy/solver/base.c:10717)
    self.notify()
  File "pydas/observer.pyx", line 73, in pydas.observer.Subject.notify (pydas/observer.c:1246)
  File "/Users/connie/code/RMG-Py/rmgpy/rmg/listener.py", line 119, in update
    SimulationPlot(csvFile=csvFile, numSpecies=10, ylabel='Mole Fraction').plot(pngFile)
  File "/Users/connie/code/RMG-Py/rmgpy/tools/plot.py", line 314, in plot
    self.load()
  File "/Users/connie/code/RMG-Py/rmgpy/tools/plot.py", line 286, in load
    time, dataList = parseCSVData(self.csvFile)
TypeError: 'NoneType' object is not iterable
make: *** [eg1] Error 1
```

connie:
Update: It does appear that the profile pdf is generated properly even though this error message arises.  I find it in my RMG-Py main folder when running make eg1.  Maybe the error arises from running the profiler concurrently with -p on the `rmg.py` script?

nickvandewiele:
Thanks for the feedback.

I will add a unit test showing how to use it, and update the wiki.

The decorator was not designed to be used in combination with `-p`. Could you test what happens without the `-p`.

connie:
I did... same error appears, maybe you can attach it to that function and see if it's replicable?

nickvandewiele:
[wiki](https://github.com/ReactionMechanismGenerator/RMG-Py/wiki/CPU-Profiling-an-RMG-Py-function)

[unit test](https://github.com/ReactionMechanismGenerator/RMG-Py/commit/5116feb10b12ba3a96287e4abced30f60af04cff)

[fix to the bug that caused the crash](https://github.com/ReactionMechanismGenerator/RMG-Py/commit/774b27e20b92ccb814fc279b537776d98a631294)

rwest:
Haven't had a chance to play with this, but it sounds great! I notice the wiki warns:
> Warning If func multiple times, than the a profiling study will be done for each iteration separately. In the next iteration, the results of the previous iteration will be overridden. Hence, at the end of the simulation, only the files of the last profiling study will be available.

Is there any way to work around this? Have the profile data accumulate somehow? Perhaps too difficult, but would be nice to have.

nickvandewiele:
@rwest the aggregated profile data is indeed much more useful than snapshot of the last call to the function.

Afaik: one can read existing profiles, stitch them together and print the result:

```python
stats = pstats.Stats('profile_stats_0.stats')
for i in range(1, 5):
    stats.add('profile_stats_%d.stats' % i)

profile.dump_stats("aggregated.prof")
```

But in the framework of a decorator, I don't see right away how to do this:
* how do you determine the index of the iteration from within the decorator
* when/how can you create the aggregate data if everything happens within the decorator function.

I guess the constraint here is basically the use of the decorator.

rwest:
Could try something like https://wiki.python.org/moin/PythonDecoratorLibrary#Alternate_Counting_function_calls
Unsure of the overhead involved. Might be able to avoid reloading the stats files each time, and just hang onto them in memory. Else at least keep track of the index as you suggested. 

You could either print the cumulative data each time (never knowing which will be the last) or use a with statement to clean up the instance of the class that's hanging on to the profiling data http://stackoverflow.com/questions/865115/how-do-i-correctly-clean-up-a-python-object

Whether it's worth it is up to you :) 
What do you have planned for this profiling decorator?

nickvandewiele:
[profilehooks](https://pypi.python.org/pypi/profilehooks) does exactly what this PR is about. No need for me to reinvent the wheel. Closing this PR.

rwest:
`import antigravity`
[;-)](https://xkcd.com/353/)

PR #596, opened 2016-03-03
Add cantera and observable regression submodules

connie:
New submodules under `rmgpy.tools`: `canteraModel` and `observablesRegression`

`canteraModel`
Contains classes for creating cantera conditions and cantera jobs.  Is able to mostly automatically simulate reactor conditions and plot them
Cantera dependency as also been added to anaconda build and environment.  

`observablesRegression`
Contains a class for testing observables between two models.  Experimental data handling not yet implemented but will probably be implemented by @nyee 

Also added an `ipython` folder to the RMG-Py main folder, containing useful scripts or other educational ipython notebooks.  Currently only the cantera simulation ipynb is in there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,9 @@ rmgpy/solver/settings.pxi
 
 # bin/symmetry
 bin/symmetry
+
+# ipython checkpoints
+**/.ipynb_checkpoints/**
+
+# cantera files
+*.cti
--- a/environment.yml
+++ b/environment.yml
@@ -6,6 +6,7 @@ dependencies:
   - libgcc # [unix]
   - python
   - numpy >=1.10.0
+  - cantera
   - cython ==0.21
   - matplotlib
   - rdkit >=2015.09.2
--- a/environment_windows.yml
+++ b/environment_windows.yml
@@ -6,6 +6,7 @@ dependencies:
   - mingwpy
   - python
   - numpy >=1.10.0
+  - cantera
   - cython ==0.21
   - matplotlib
   - rdkit >=2015.09.2
--- a/ipython/README
+++ b/ipython/README
@@ -1,0 +1,12 @@
+This folder contains ipython notebooks containing scripts and other notes related to using RMG.
+
+If you do not have ipython installed, we suggest installing it through the jupyter package while your rmg environment is activated by typing the following commands:
+
+source activate rmg_env
+conda install jupyter
+
+
+For Windows users:
+
+activate rmg_env
+conda install jupyter
--- a/ipython/canteraSimulation.ipynb
+++ b/ipython/canteraSimulation.ipynb
@@ -17,7 +17,6 @@
    "source": [
     "from rmgpy.chemkin import *\n",
     "from rmgpy.tools.canteraModel import *\n",
-    "from rmgpy.quantity import *\n",
     "from IPython.display import display, Image"
    ]
   },
@@ -61,9 +60,9 @@
     "\n",
     "reactorType = 'IdealGasReactor'\n",
     "molFracList=[{ethane: 1}]\n",
-    "Tlist = Quantity([1300,1500,2000],'K')\n",
-    "Plist = Quantity([1],'atm')\n",
-    "reactionTime = Quantity(0.5, 'ms')"
+    "Tlist = ([1300,1500,2000],'K')\n",
+    "Plist = ([1],'atm')\n",
+    "reactionTime = (0.5, 'ms')"
    ]
   },
   {
--- a/ipython/canteraSimulation.ipynb
+++ b/ipython/canteraSimulation.ipynb
@@ -74,7 +74,7 @@
    "outputs": [],
    "source": [
     "# Create cantera object, loading in the species and reactions\n",
-    "job = Cantera(speciesList=speciesList, reactionList=reactionList)\n",
+    "job = Cantera(speciesList=speciesList, reactionList=reactionList, outputDirectory='temp')\n",
     "# The cantera file must be created from an associated chemkin file\n",
     "job.loadChemkinModel('data/minimal_model/chem_annotated.inp')\n",
     "# Generate the conditions based on the settings we declared earlier\n",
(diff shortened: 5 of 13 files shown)


nickvandewiele:
I will review this PR.

nickvandewiele:
Why do you choose for primitive types (float) for the `CanteraCondition` attributes, when the header of the `generateCanteraConditions` function mentions that similar attributes should be `Quantity` objects?

nickvandewiele:
Instead of hard-coding [`chem.cti`] (https://github.com/ReactionMechanismGenerator/RMG-Py/pull/596/files#diff-1d9f6852720bf4a1cc9e4d9988c8da8aR188), could you not adopt the filename of the passed-in chemkin file?

It may come in handy when having multiple chemkin files in the directory.

nickvandewiele:
Could you replace `print `statements by `logging.info`?

connie:
For `CanteraCondition`, I went with mostly what Nate wrote, and storing the values as floats in SI units allows them to represent the actual T, P, and V used within the Cantera program.  I think doing it either way works. If we make `CanteraCondition` use quantity objects, then the conversion would simply be switched to the `Cantera.simulate()` function.  If you have a strong case for not doing it this way then I can switch it.

For `chem.cti` naming, I suppose changing the name to be similar to the chemkin input would be ok.. ideally I would've wanted the .cti do be just passing a stream into the loading of the model and not saved at all..

Will logging.info work fine w RMG-tests without any initialization?

nickvandewiele:
I am looking at the dependencies of `canteraModel`:

- `from rmgpy.chemkin import loadSpeciesDictionary`  is not used
- could you explicitly mention the imports from plot? Ideally, the number of dependencies to another sister module like `plot `should be minimized if we want to avoid cyclic dependency errors in the future. E.g. the GenericData object should really be in a more low-level module than plot if it is being used by other modules.

nickvandewiele:
> Will logging.info work fine w RMG-tests without any initialization?

using logging without explicit initialization prints to the console, so that is not a problem.

nyee:
One of the major things that I did not address well was the fact that to specify the chemical system in Cantera is you specify exactly 2 of T,P and V. I suggest an assert statement that checks exactly one of T0, P0, V0 is None in the initalization of a CanteraCondition. This will also need to be the case in for the generateCanteraConditions function for the Tlist, Plist, and Vlist as well. 

The last part of this point is that we need to actually apply this correctly to the Simulation. In line 233 of canteraModel.py. Right now we always do `self.model.TPX = condition.T0, condition.P0, condition.molFrac`, but we need if statements and sometimes do `self.model.PVX` or `self.model.TPX`

nyee:
A few more minor comments:

In canteraModel.py:
-line 241 is not a very helpful exception statement. Maybe type out the supported types we have right now.
-line 284 we define 100 time steps for our output. Cantera does use an automatic time stepper and then interpolates for the time points we ask for. It might be better to just use the output from the automatic time stepper, since a human never reads this and we'll get better precision. 

@bryanwweber Could you advise us how to get this output, I know I saw you respond in an issue about the same request, but I can't seem to find it anymore...

In ObservablesRegression.py:
-We should remove the getNearestTime function as it is is almost the same as findNearest. The only place it is used is line 39-41 which we should just replace as:

`index=findNearest(t2, timepoint1)`
`t2sync[i]=t2[index]`
`y2sync[i]=y2[index]`

-line 104 there seems to be a typo in the comment with an extra or missing quotation mark
-The whole output logging for the ObservablesTestCase.compare() is a bit of a mess. (Completely my fault, I wrote it to just do something rather than as a polished output) Right now, it says "In condition x, "variable is broken", for all the broken variables in all the conditions, then after all that it tells you "condition x is T=blah, P=blah, X=blah". It should probably define the condition first, then all the variables broken for that specific condition, then define a new condition, give all broken variables for that condition, etc.

connie:
Is there a `self.model.VTX` as well?  I think raising exceptions will be better than assertions, just to let the user know the system is overspecified.  Can you write in the changes and make a new commit to this pull request?

connie:
>> Will logging.info work fine w RMG-tests without any initialization?

> using logging without explicit initialization prints to the console, so that is not a problem.

This doesn't seem to work.  I tried converting print statements to logging.info, but running on the console does not produce any stdout.

bryanwweber:
@nyee There are two time stepping methods in Cantera `ReactorNet`s. The first is `netw.step()` (with 2.2 and older you need to supply an end time, with 2.3 and newer, this is deprecated). `netw.step()` takes one internal time step, of size determined by the solver, and returns. The other method is `netw.advance()` which takes an end time as an argument and takes enough internal time steps until it gets to the time given as an argument. If the last time step took it past the argument, it interpolates backwards to set the state. In my experience, there can be accumulated interpolation errors if you use `advance` enough times in a row. Thus, I typically use `step` and write a `while` loop that checks the network time to break, something like

    while netw.time < 0.1:
        netw.step(1)
        get output here...

but you could just as easily write a loop over a number of time steps to take if you don't care about when it breaks. Again, the argument to `step` is deprecated in 2.3 (2.2.1 is the current stable release). You can see a use case here: https://github.com/bryanwweber/UConnRCMPy/blob/master/uconnrcmpy/dataprocessing.py#L548

connie:
Based on @nickvandewiele and @nyee's suggestions, the following has been changed:
- Selective importing in modules
- Migration of `GenericData` to `rmgpy.tools.data`
- Creation of matching cantera `.cti` files based on chemkin filenames
- Tuple objects that can reconstruct Quantity objects, rather than pure Quantity objects, very similar in usage to RMG's `input.py` file
- Deleted `getNearestTime` function

Critical fixes before merging this pull request:
- Adaptive timestepping

Good to have features to do after this pull request:
- Change from print to `logging.info`.  I saw that directly switching no longer prints to console
- Use of `Solution.PVX` or other thermophase settings in addition to `Solution.TPX`
- Check the T,P,V constraints during `generateCanteraConditions` and possibly the `CanteraCondition` class that the system is not under- or over-specified
- Better organization of `ObservablesTest.compare()` function print statements

I'd like to be able to merge this branch once we have most of the critical features in place, and for new pull requests to be made for good-to-have and extensions of features.

Issue #594, opened 2016-03-01
Solver .CSV files have alternating blank rows if RMG's run on a Win platform


alongd:
When opening .CSV files from the Solver folder after running RMG on Windows 10 I get alternating rows of data and blanks, i.e.:

```
Time (s)	Volume (m^3)	C2H6N(5)	HCO(6)
			
1.16E-16	0.059481028	9.53E-21	8.16E-46
			
3.25E-16	0.059481028	2.67E-20	9.69E-45
			
7.43E-16	0.059481028	6.10E-20	9.51E-44
```

Instead of
```
Time (s)	Volume (m^3)	C2H6N(5)	HCO(6)	
1.16E-16	0.059481028	9.53E-21	8.16E-46	
3.25E-16	0.059481028	2.67E-20	9.69E-45	
7.43E-16	0.059481028	6.10E-20	9.51E-44
```

This makes working with the data a bit frustrating.
(Until issue is resolved, users can refer to '[A quick way to delete blank rows in Excel](http://www.techrepublic.com/blog/microsoft-office/a-quick-way-to-delete-blank-rows-in-excel/)')

alongd:
Above error also holds for scripts output, such as sensitivity analysis

connie:
This is related to unix vs. windows style line endings it seems: http://stackoverflow.com/questions/8746908/why-does-csv-file-contain-a-blank-line-in-between-each-data-line-when-outputting

Can you test if http://stackoverflow.com/a/8747275 works by changing the opening of the csv file from 'w' to 'wb' and see if it fixes things?

alongd:
Yes, changing `w` to `wb` in rmgpy\rmg\listener.py line 59 (```with open(filename, 'wb') as csvfile:```) did the trick.

PR #597, opened 2016-03-04
Fix csv writing in Windows

connie:
Closes https://github.com/ReactionMechanismGenerator/RMG-Py/issues/594

Changes opening of csv files for writing from 'w' status to 'wb' status, preventing blank lines from being written in their .csv files when user runs RMG-Py in windows machines.


connie:
Confirmed and tested in windows 10 to be fixed.

PR #599, opened 2016-03-09
Parallel generation of reactions in `enlarge`

nickvandewiele:
This PR introduces a new design for the generation of reactions in parallel.

In the `rmgpy.rmg.model.CoreEdgeReactionModel.enlarge` method, reactions between a new core species and the existing list of core species, for the given list of available reaction families. This process is parallelized across the list of core species.

- A uniform `rmgpy.scoop_framework.util.map_` interface is introduced that either uses the parallel map function of SCOOP, or the serial map of python if scoop is not loaded.
- A new test target make scoop in Makefile is created to illustrate the launch of the parallel version using the minimal example.
- A uniform `rmgpy.data.rmg.getDB` interface allows retrieving RMG-database info, either from the global `rmgpy.data.rmg.database` variable, or from broadcasted variables when that global variable is not available (e.g. on remote workers).
- the `reverse` variable is now added as a real attribute of `TemplateReaction`. Before, it already existed, but was never added in the deserialization method of `TemplateReaction`.

To reduce communication costs of sending data to/from the workers:
- reactions that are send back to the root process will contain indices of core species for the reactants/products, rather than real Species objects. This is done in the _deflate_ function of the `rmgpy.rmg.react` module. Once the reaction arrives on the root process, the index will be _inflated_ again to the original Species object.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,6 +135,14 @@ eg7: all
 	@ echo "Running eg7: gri_mech_rxn_lib example"
 	python rmg.py testing/eg7/input.py
 	
+scoop: noQM
+	mkdir -p testing/scoop
+	rm -rf testing/scoop/*
+	cp examples/rmg/minimal/input.py testing/scoop/input.py
+	coverage erase
+	@ echo "Running minimal example with SCOOP"
+	python -m scoop -n 2 rmg.py -v testing/scoop/input.py
+
 ######### 
 # Section for setting up MOPAC calculations on the Travis-CI.org server
 ifeq ($(TRAVIS),true)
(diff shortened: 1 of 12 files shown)


connie:
In terms of releases, perhaps we should tag the current HEAD at release v1.0.4, and then once this pull request is merged, jump versions to v1.1 or even v2.0.

nickvandewiele:
sure, I'm ok with that.

nickvandewiele:
I think this PR is ready to be merged.
- I rebased it onto the master
- it passes all the unit and regression tests
- as indicated by RMG-tests, in serial mode it runs as fast as the current master.

PR #604, opened 2016-03-14
constant Concentration pull request

chatelak:
This pull request includes the constant concentration implementation in the new RMGPY.

I had to modify the implementation we did last year to make it fit to the new RMG-Py. At the end it does the same: apply a coreSpeciesRate=0 for the species (or list of species) that we want constant. But I changed the way constant species information is stored, like this it is way cleaner I think. I only had to create 2 new attributes  (very lightweight: Two list of x sting where x is the number of species constant) to the liquid reactor). Both are stored in reactor attributes, the first one is the constant species labels and the second store constant species positions into the coreSpeciesRate Array. As this position can differ from one liquidSystem to another, I think this implementation is safer.

Drawbacks of the previous developpments:
The old implementation was using props and a boolean for each species (even the non constant ones). And it was also inducing many changes in reactor functions (passing new arguments in both gas and liquid phase reactors). Finally, the old implementation did not support multi-reactor systems.

To conclude:
I think, This implementation is cleaner and lightweight than the previous one .
I wrote some unitests, the documentation and provided an example with constant species.

My validation cases are:
- [ ] gasPhaseMinimal & sensitivity is not broken
- [ ] liquidPhase with 0,1 and 2 spc constant (as shown in the documentation)
- [ ] Multireactor system are running correctly (even with one containing constant species and not the other one)
- [] Make test fails due to recent developments (for cantera).

Note: As mentioned in the documentation, it should be avoided to specify a product as a constant species (the mole fraction of the constant species rises over time due to a mass loss). Maybe this is obvious and can be removed from the documentation.


bslakman:
Hi @chatelak, thanks for this! I don't think your new example file is added to this pull request, `examples/rmg/liquid_phase_constSPC/input.py` so for me the test fails.

chatelak:
Thank you, I forgot I had to force this file addition due to the .gitignore configuration. It should work now.

chatelak:
Hi @bslakman , Did you try it again? I am abble to run it locally in several places (including on centos).

I actually have a question regarding one unitest I created. I one of them I am using a new RMG instance to use most of functions to read input file, store constant SPC names and then indices as it is done in regular simulation.

In the log (with make test command) it is printed out something like creating a new RMG instance should be avoided. Any idea why?

connie:
@chatelak, can you rebase this off of official/master and then force push back to this branch using the commands
```
git fetch official
git rebase official/master
git push origin your_branch -f
```

I'd like to see the travis unit testing pass.

chatelak:
Ok one of my unitest is not passing. I need to have a look on this.

chatelak:
Thanks @connie for the command. I have not seen the fix from @nickvandewiele.

I still get errors in the make test but all related to cantera.

connie:
You will need to install cantera on your environment. 

'''
source activate rmg_env
conda install -c rmg cantera 
'''

bslakman:
Looks good, I will merge this now.

Regarding your question about the new RMG instance warning during testing. I think it's fine since the purpose of that unit test is to see that an actual RMG simulation is working properly. That warning, I think, is so that we don't have another RMG instance floating around while running a main simulation. But for testing it should be okay.

connie:
Please do not merge branches that are not rebased if possible..

bslakman:
Oops sorry. I assumed it had been since Travis was passing and was not before.

PR #607, opened 2016-03-15
Read chemkin files literally: Don't load new bath gases when loading a chemkin file

connie:
There is a design change to the loading of chemkin files.  I think we should no longer
follow RMG-Java's way, and instead read in chemkin files more literally, especially
since all RMG-Py jobs do write almost all the bath gases.  This is also
so that the indices and readouts match cantera and chemkin more closely without unexpected
loading of species.


rwest:
Looks sensible to me. May surprise a few people but perhaps in fact remove surprises because if you have to declare ALL colliders you won't be caught out when you forget to declare an unusual one like CH4 for example.

connie:
The regression does seem to be done still on RMG-tests for pull requests, not sure why it says it doesn't deploy from pull requests...

Issue #590, opened 2016-02-27
Fatal error with explicit valence for nitrogen


ejtierne:
We're trying to build a mechanism for 2-nitrodiphenylamine: 

```
species(
    label='2NDPA',
    reactive=True,
    structure=adjacencyList(
    """
1  O u0 p3 c-1 {2,S}
2  N u0 p0 c+1 {1,S} {3,D} {4,S}
3  O u0 p2 c0 {2,D}
4  C u0 p0 c0 {2,S} {5,D} {9,S}
5  C u0 p0 c0 {4,D} {6,S} {17,S}
6  C u0 p0 c0 {5,S} {7,D} {18,S}
7  C u0 p0 c0 {6,D} {8,S} {19,S}
8  C u0 p0 c0 {7,S} {9,D} {20,S}
9  C u0 p0 c0 {4,S} {8,D} {10,S}
10 N u0 p1 c0 {9,S} {11,S} {21,S}
11 C u0 p0 c0 {10,S} {12,D} {16,S}
12 C u0 p0 c0 {11,D} {13,S} {22,S}
13 C u0 p0 c0 {12,S} {14,D} {23,S}
14 C u0 p0 c0 {13,D} {15,S} {24,S}
15 C u0 p0 c0 {14,S} {16,D} {25,S}
16 C u0 p0 c0 {11,S} {15,D} {26,S}
17 H u0 p0 c0 {5,S}
18 H u0 p0 c0 {6,S}
19 H u0 p0 c0 {7,S}
20 H u0 p0 c0 {8,S}
21 H u0 p0 c0 {10,S}
22 H u0 p0 c0 {12,S}
23 H u0 p0 c0 {13,S}
24 H u0 p0 c0 {14,S}
25 H u0 p0 c0 {15,S}
26 H u0 p0 c0 {16,S}
    """),
)
```

Once we comment out the forbidden structure "N_birad_singlet_2singleBonds", RMG-Py starts but then crashes with:

``` python
[15:01:43] Explicit valence for atom # 6 N, 4, is greater than permitted
[15:01:43] Explicit valence for atom # 6 N, 4, is greater than permitted
[15:01:43] Explicit valence for atom # 6 N, 4, is greater than permitted
[15:01:43] Explicit valence for atom # 6 N, 4, is greater than permitted
[15:01:43] Explicit valence for atom # 6 N, 4, is greater than permitted
[...]

Issue #405, opened 2015-07-02
Aromatic molecule generated from adjlist does not produced kekulized version in resonance isomers


connie:
I thought we do the kekulization at the start but this is problematic for importing from Java.

Loading a molecule from SMILES produces the two versions of isomers (both aromatic and kekulized).

    >>> m = Molecule().fromSMILES('Cc1ccccc1')
    >>> m.generateResonanceIsomers()
    [Molecule(SMILES="Cc1ccccc1"), Molecule(SMILES="Cc1ccccc1")]

Loading it from an aromatic adjlist retains only the aromatic version.  However, if you load from a kekulized adjlist you get both resonance isomers.

    >>> m2 = Molecule().fromAdjacencyList('''1  H 0 {2,S}
    ... 2  C 0 {3,S} {9,S} {10,S} {1,S}
    ... 3  C 0 {4,B} {8,B} {2,S}
    ... 4  C 0 {3,B} {5,B} {11,S}
    ... 5  C 0 {4,B} {6,B} {12,S}
    ... 6  C 0 {5,B} {7,B} {13,S}
    ... 7  C 0 {6,B} {8,B} {14,S}
    ... 8  C 0 {3,B} {7,B} {15,S}
    ... 9  H 0 {2,S}
    ... 10  H 0 {2,S}
    ... 11  H 0 {4,S}
    ... 12  H 0 {5,S}
    ... 13  H 0 {6,S}
    ... 14  H 0 {7,S}
    ... 15  H 0 {8,S}
    ... ''')

    >>> m2.generateResonanceIsomers()
    [Molecule(SMILES="Cc1ccccc1")]

rwest:
I think the call `fromSMILES`, via RDKit, kekulizes it in RDKit.
We could add an "if aromatic then kekulize using RDKit" step in the `generateResonanceIsomers()` method.

I also notice that `toSMILES` seems to "unkekulize" it - in that both resonance isomers in your example above are printed as `Molecule(SMILES="Cc1ccccc1")` even though their adjacency lists are different. This may be unavoidable and hopefully is not a problem if we're mindful that SMILES are a sometimes-conevinent shorthand, and the "true" representation is always the adjacency list.

connie:
Yeah it looks like SMILES always return the aromatic representation, so the shorthand is indeed deceiving.

Similarly on print issues, the other day I created a Species object with a single molecule:

    >>> from rmgpy.molecule import *
[...]

PR #609, opened 2016-03-16
Some changes to resonance isomer generation

rwest:
I think the net result of this is mostly that when it fails to generate kekule resonance form due to an RDKit error, it carries on silently without (just like was already done for failing to generate aromatic resonance forms). This fixes the probelm @ejtierne was having in #590.

I also created a "debugRDKitMol" function which is useful at times. We should put it in some error handlers in other places where we hit RDKit errors.

The next step would be to silence the very many `[15:01:43] Explicit valence for atom # 13 N, 4, is greater than permitted` messages that get printed to the standard error stream. This would require some fancy IO juggling like in my debugRDKitMol function. I was tempted to try this, but then came across [this blog post](http://rdkit.blogspot.com/2016/03/capturing-error-information.html) suggesting there'll be some built in support in the next RDKit (not sure if it captures and silences, or just captures).

### DIFF
--- a/rmgpy/molecule/generator.py
+++ b/rmgpy/molecule/generator.py
@@ -4,7 +4,6 @@ import cython
 import logging
 import itertools
 import sys
-from cStringIO import StringIO  # for python3: from io import StringIO
 
 # local imports
 try:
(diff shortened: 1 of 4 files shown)


connie:
Looks good to me.  Unfortunately still doesn't solve  https://github.com/ReactionMechanismGenerator/RMG-Py/issues/405 where only a single kekulized form can arise from the aromatic form, and that RDKit does it nondeterministically.  For instance, `Cc1ccccc1O` can generate either: 
```
[Molecule(SMILES="CC1=CC=CC=C1O"), Molecule(SMILES="Cc1ccccc1O")]
```
or
```
[Molecule(SMILES="CC1C=CC=CC=1O"), Molecule(SMILES="Cc1ccccc1O")]
```

So if you start with the kekulized form, you can either get lucky and end up with all 3 isomers, or you end up with 2. 

Anyway, the entire resonance isomer generation for aromatics is still a poor long term solution because it will go haywire with PAHs.  For now this fix is ok just so that RMG no longer crashes.

PR #613, opened 2016-03-17
Fixes to TransportData

connie:
What started as a small fix led to the discovery of multiple bugs.

Summary of fixes:
- Fixes loading of all the transport data from a chemkin `tran.dat` file into the `TransportData` object
- Estimate shape index for monatomic molecules correctly within their estimation methods, not when it reaches the `saveTransportFile` function in `rmgpy.chemkin`
- Deletes duplicate `dipoleMoment`, `polarzibility`, and `Zrot` attributes from the `Species` class and closes https://github.com/ReactionMechanismGenerator/RMG-Py/issues/615
- Correctly saves the rotational relaxation number from the `TransportData` object rather than from `Species.zRot`, which never had anything saved inside.

### DIFF
--- a/rmgpy/chemkin.py
+++ b/rmgpy/chemkin.py
@@ -1447,13 +1447,6 @@ def writeReactionString(reaction, javaLibrary = False):
     if len(reaction_string) > 52:
         logging.warning("Chemkin reaction string {0!r} is too long for Chemkin 2!".format(reaction_string))
     return reaction_string
-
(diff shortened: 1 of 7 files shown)


rwest:
Could you add any unit tests that would have caught these or similar bugs?

connie:
Adding one unit test for loading and saving transport..  don't have energy to write others.

PR #620, opened 2016-03-18
RMG direct species and reaction conversion to Cantera objects

connie:
Wasn't planning on working on this but needed to add the functionality now or later for uncertainty analysis capabilities.  This pull request generates cantera objects from RMG species and reaction objects.  A full suite of unit tests were written ensuring that conversion occurs smoothly (so far) for all `Reaction` objects with the accepted kinetic types: Arrhenius, MultiArrhenius, PDepArrhenius, MultiPDepArrhenius, Chebyshev, ThirdBody, Lindemann, and Troe.   For `Species` objects, thermo is only accepted in 7 coefficient NASA form, and transport is also directly converted.

Unit tests also cover the direct comparison of ck2cti parsed objects for a set of reactions and species with various kinetics and show direct match from RMG converted objects vs parsed.  These `checkEquivalency` functions can also be helpful in the case where we just have 2 cantera objects we want to compare in the general sense. This will make it easy for me to modify and change the cantera model kinetics and thermo as needed for uncertainty analysis.

Also added an ipython example for the `ObservablesTestCase` to do one final check on whether models loaded via `ck2cti` parser vs directly converted from RMG objects.  Visually they are identical, so I feel that these conversions are safe for use.  

Tagging @speth on this. =D


rmgpy/tools/canteraModel.py
```diff
@@ -405,3 +405,137 @@ def findIgnitionDelay(time, yVar=None, metric='maxDerivative'):
...
+            assert checkNearlyEqual(trans1.well_depth, trans2.well_depth, dE), "Identical well depth"
+
+        if ctSpec1.thermo or ctSpec2.thermo:
+            thermo1 = ctSpec1.thermo
+            thermo2 = ctSpec2.thermo
+
+            Tlist = [300,500,1000,1500,2000]
+            for T in Tlist:
+                assert checkNearlyEqual(thermo1.cp(T), thermo2.cp(T), dE),  "Similar heat capacity"
+                assert checkNearlyEqual(thermo1.h(T), thermo2.h(T), dE), "Similar enthalpy"
```
rwest: Doing a relative comparison could lead to false failures, or even division by zero errors, when H is close to zero. Maybe this one (and any others that can be arbitrarily close to 0) should be absolute comparisons instead of relative ones?

rmgpy/tools/canteraModel.py
```diff
@@ -405,3 +405,137 @@ def findIgnitionDelay(time, yVar=None, metric='maxDerivative'):
...
+
+    return True
+    
+def checkEquivalentCanteraReaction(ctRxn1, ctRxn2, checkID=False, dE=1e-5):
+    """
+    Checks that the two cantera species are nearly equivalent
+    if checkID is True, then ID's for the reactions will also be checked
+    """
+    def checkEquivalentArrhenius(arr1, arr2):
+        assert checkNearlyEqual(arr1.activation_energy, arr2.activation_energy, dE), "Similar Arrhenius Ea"
```
rwest: If `arr1.activation_energy` is 0 this might cause problems? (see comment above)

rmgpy/thermo/nasa.pyx
```diff
@@ -361,8 +362,8 @@ cdef class NASA(HeatCapacityModel):
         # high temperature
         coeffs = numpy.zeros(15)
         coeffs[0] = polys[0].Tmax.value_si # mid point temperature between two polynomials
-        coeffs[1:8] = polys[1].coeffs # 7 coefficients of the high temperature polynomial
-        coeffs[8:15] = polys[0].coeffs # 7 coefficients of the low temperature polynomial
+        coeffs[1:8] = copy.deepcopy(polys[1].coeffs) # 7 coefficients of the high temperature polynomial
+        coeffs[8:15] = copy.deepcopy(polys[0].coeffs) # 7 coefficients of the low temperature polynomial
```
rwest: Because you're putting into a slice of the coeffs array, rather than changing the object that the coeffs name points to, I'm pretty sure this is not necessary.
```python
In [1]: import numpy as np
In [2]: a=np.ones(5)
In [3]: b=np.zeros(3)
In [4]: a
Out[4]: array([ 1.,  1.,  1.,  1.,  1.])
In [5]: a[1:4]=b
In [6]: a
Out[6]: array([ 1.,  0.,  0.,  0.,  1.])
In [7]: b[1]=2
In [8]: a
Out[8]: array([ 1.,  0.,  0.,  0.,  1.])
In [9]: b
Out[9]: array([ 0.,  2.,  0.])
```
but it might not be expensive and I'm not sure how often this gets called.

rmgpy/tools/canteraModel.py
```diff
@@ -199,6 +199,32 @@ def generateConditions(self, reactorType, reactionTime, molFracList, Tlist=None,
         """
         self.conditions = generateCanteraConditions(reactorType, reactionTime, molFracList, Tlist, Plist)
 
+    def loadModel(self):
```
rwest: The name is a bit confusing, as I feel we're not loading anything.

rwest:
Apart from my 3 comments above
1. [] relative comparisons of values that may be close to 0 could lead to false failures
2. [] deepcopy not necessary when putting values into slices of an array
3. [] the name `loadModel` is confusing for making a cantera model from the current RMG model in memory (nothing is being loaded)
this looks great.  A model pull request! -- there are lots of unit tests, and every commit does one thing, and has a clear commit message.

connie:
Any suggestion on what `loadModel` function name should be changed to?

Issue #602, opened 2016-03-10
SpeciesConstraints applied to training reactions


nyee:
When setting species constraints in input file, it seems to affect the training reactions. When setting `maxCarbonAtoms=6`, we get the error:
rmgpy.data.kinetics.common.KineticsError: Unable to calculate degeneracy for reaction Molecule "C=C1C=CC[CH]C1" <=> Molecule "[CH2]C12C=CCC1C2" in reaction family Intra_R_Add_Exocyclic. Expected 1 reaction but generated 0.

If we set `maxCarbonAtoms=8`, we get:
rmgpy.data.kinetics.common.KineticsError: Unable to calculate degeneracy for reaction Molecule "[CH2]C=CCC=C1C=CC=C1" <=> Molecule "C1C=C[C](C=1)C1CC=CC1" in reaction family Intra_R_Add_Exocyclic. Expected 1 reaction but generated 0

The molecule RMG complains about is always one C higher than the max allowed. I haven't looked too thoroughly into this yet, but if I were to guess, I bet RMG is applying the speciesConstraints to these species in the training reaction, but still give back some Species object in the Reaction object. The Species object is probably flawed, maybe missing all its atoms. So when RMG tries to calculate the degeneracy it can't find any labelled atoms and gives back a degeneracy of 0.

connie:
Closed via aedfdeb46021dd29f4a9e68304a6b36784ebae70

PR #622, opened 2016-03-21
Add work around to turn off speciesConstraints for training reactions

nyee:
Work around for #602 

A cleaner way to do it is to load database before we load the species constraints. The problem is that setting species constraints is embedded in the loadInput function. loadinput is needed before loadDatabase because RMG needs to know what libraries, families, etc to load. Additionally, this would change the way we implement the "allowed" option in species constraints. Overall I think its better to just use this work around.

@nickvandewiele Can you test to see if this messes up a parallel job specifically in regards to broadcasting either the species constraints or database.

### DIFF
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -41,6 +41,7 @@ import shutil
 import numpy
 import csv
 import gc
+import copy
 
 from rmgpy.constraints import failsSpeciesConstraints
 from rmgpy.molecule import Molecule
@@ -64,7 +65,6 @@ from rmgpy.restart import RestartWriter
 from rmgpy.qm.main import QMDatabaseWriter
 from rmgpy.stats import ExecutionStatsWriter
 from rmgpy.tools.sensitivity import plotSensitivity
-
 ################################################################################
 
 solvent = None
@@ -311,8 +311,12 @@ class RMG(util.Subject):
         if self.kineticsEstimator == 'rate rules':
             if '!training' not in self.kineticsDepositories:
                 logging.info('Adding rate rules from training set in kinetics families...')
+                # Temporarily remove species constraints for the training reactions
+                copySpeciesConstraints=copy.copy(self.speciesConstraints)
+                self.speciesConstraints={}
                 for family in self.database.kinetics.families.values():
                     family.addKineticsRulesFromTrainingSet(thermoDatabase=self.database.thermo)
+                self.speciesConstraints=copySpeciesConstraints
             else:
                 logging.info('Training set explicitly not added to rate rules in kinetics families...')
             logging.info('Filling in rate rules in kinetics families by averaging...')


nyee:
I think the failed Travis check is because github went down in the middle. Is there a way to re-run the check?

nickvandewiele:
Although this workaround solves the issue, I assure you it will haunt future developers.

> A cleaner way to do it is to load database before we load the species constraints.

Species constraints are loaded when the RMG input file is loaded. The RMG input file allows the user to control which parts of the database are loaded (libraries, families) . If you want to load the database _before_ the RMG input file, you will have to decouple the species constraints from the RMG input file.

nickvandewiele:
Thinking out loud:

species constraints are only checked in the method `rmgpy.data.kinetics.family.KineticsFamily.__generateProductStructures`.

Why is this method called in the method `rmgpy.data.kinetics.family.KineticsFamily.addKineticsRulesFromTrainingSet`?
- Because it needs to compute the reaction path degeneracy (rpd) for those training reactions that were defined in the reverse direction of the reaction family.

What about the strategy to insert a boolean flag (default=True) in the signature `rmgpy.data.kinetics.family.KineticsFamily.__generateProductStructures` that tells the method if it needs to check for constraints?

- Well, the distance between both methods is large, so the signatures of the intermediate methods will have to reflect the change as well.

nickvandewiele:
Can we deduce whether we should check constraints from the object type passed to `rmgpy.data.kinetics.family.KineticsFamily.calculateDegeneracy` ?

Well, training reactions that are converted into kinetics rules use the `rmgpy.reaction.Reaction` object, cf. 

```python
item = Reaction(reactants=[m.molecule[0].copy(deep=True) for m in entry.item.products], products=[m.molecule[0].copy(deep=True) for m in entry.item.reactants])
template = self.getReactionTemplate(item)
item.degeneracy = self.calculateDegeneracy(item)
```

whereas reactions in the core/edge of RMG are `rmgpy.data.kinetics.family.TemplateReaction` objects. 

Since the signature of `rmgpy.data.kinetics.family.__generateProductStructures` does not take reaction objects, but molecule objects, you would still need to pass in a flag, as the previous line of thinking was explaining. So, the same amount of effort, and not bullet proof neither when future developers will accidentally create Reaction objects.

nickvandewiele:
Why are molecule constraints checked in `__generateProductStructures` anyways? You might as well check those constraints elsewhere where you have better information?

Molecules that do not satisfy one of the constraints will raise a `rmgpy.data.base.ForbiddenStructureException`. As far as I can see, none of the methods that calls `__generateProductStructures` catches the exception. Hence, it will always lead to a crash of the RMG simulation.

One argument in favor of checking it in `__generateProductStructures` is to avoid continuing the simulation even when something already went wrong. But if we move the constraints checking to a couple of methods upstream, I don't see how it would make much of a difference.

connie:
`rmgpy.data.family.KineticsFamily.__generateReactions` will sometimes check for `ForbiddenStructureException` and pass when it finds it, thereby skipping the reaction when a forbidden structure is generated. This same exception handling is not done in `rmgpy.data.kinetics.family.KineticsFamily.calculateDegeneracy`, which means the RMG will crash when converting training reactions to rules.  

I am against testing for `TemplateReaction` vs `Reaction` object for avoiding species constraints.  I believe it is quite dangerous to do so without an explicit flag.  

I believe the reason this test is in `__generateProductStructures` is to avoid further kinetics or reaction object creation in the process as soon as possible, for time savings.  Not sure if there is a better place to put it.

nickvandewiele:
Ok, by the lack of better alternatives than the one proposed in the current PR, I will proceed and merge this.

PR #626, opened 2016-03-23
ArrayQuantity: better uncertainty handling

connie:
Make the `ArrayQuantity` class more like the `ScalarQuantity` class in the way they store and use uncertainties as a property, with an `uncertainty_si` attribute. This eliminates confusion due to the differences in the two classes and makes `ArrayQuantity` more robust.

Also added a suite of unit tests.


nickvandewiele:
So, now both the `ScalarQuantity` and `ArrayQuantity` classes have the same attributes and methods?

connie:
Yes.

Issue #572, opened 2016-02-12
if rmg says my species is forbidden, tell me why.


cfgoldsmith:
We have a reactant that is forbidden. It would be good to know which part of the forbidden structure list it hits, so that I may comment it out (without having to guess repeatedly). Can we just append that to the error message?
Thanks!
Franklin

PR #628, opened 2016-03-23
Verbose forbidden structure debugging info

connie:
Closes https://github.com/ReactionMechanismGenerator/RMG-Py/issues/572 as suggested by @cfgoldsmith 

Adds the feature of printing more verbose debugging info when raising a `ForbiddenStructureException`.  This helps the user quickly identify why the exception was raised and allows them to address it as either a human error or database problem.

Also fixes the bug introduced in  2e2c2bedbd where forbiddenStructures are set to None permanently.

Don't merge yet, somebody should review/test.

### DIFF
--- a/rmgpy/constraints.py
+++ b/rmgpy/constraints.py
@@ -32,10 +32,14 @@ import logging
 
 from rmgpy.species import Species
 
-def failsSpeciesConstraints(species):
+def failsSpeciesConstraints(species, verbose=False):
     """
     Pass in either a `Species` or `Molecule` object and checks whether it passes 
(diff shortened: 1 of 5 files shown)


nickvandewiele:
My intuition said: the logging module already has a verbosity filtering mechanisms, why create a new one.

Now, I understand the tactics of the verbose flag, but it's a lot of code to avoid printing the forbidden structures for molecules outside of the database loading. 

I wonder if there are other, more elegant solutions.

connie:
I agree... if there's a more elegant solution it would be preferred, since these functions are called very frequently.

Issue #271, opened 2014-08-04
Error restarting job with solvation


bslakman:
After my computer restarted, I tried to restart the lipid peroxidation example I'd be running with solvation on and got the following error: 

```python
Loading previous restart file...
Traceback (most recent call last):
  File "rmg.py", line 147, in <module>
    rmg.execute(args)
  File "/Users/belinda/Code/RMG-Py/rmgpy/rmg/main.py", line 385, in execute
    self.initialize(args)
  File "/Users/belinda/Code/RMG-Py/rmgpy/rmg/main.py", line 338, in initialize
    self.loadRestartFile(os.path.join(self.outputDirectory,'restart.pkl'))
  File "/Users/belinda/Code/RMG-Py/rmgpy/rmg/main.py", line 642, in loadRestartFile
    self.reactionModel = cPickle.load(f)
  File "reaction.py", line 117, in rmgpy.reaction.Reaction.__init__ (build/pyrex/rmgpy/reaction.c:3973)
AttributeError: ("'rmgpy.reaction.Reaction' object has no attribute '__k_effective_cache'", <type 'rmgpy.reaction.Reaction'>, (-1, '', [<Entry index=-1 label="RJ_R_Y">], [<Entry index=1 label="R_R">, <Entry index=2 label="YJ">], None, True, None, False, 1, None))
```

I'm looking into it but let me know if anyone has had similar errors

bslakman:
Closing this now, as Reaction().k_effective_cache is no longer private, and we can pickle/unpickle a liquid phase reaction. (#633)
There is still an issue with restart, however; will open an issue or add to existing ones.

PR #633, opened 2016-04-11
un/pickle reaction when diffusionLimiter.enabled is set to True

nickvandewiele:
A few small changes to the way Reaction.__k_effective_cache is stored.

https://github.com/ReactionMechanismGenerator/RMG-Py/issues/271

- renaming it to k_effective_cache
- a unit test that shows the un/pickling is correctly done

The whole design of this attribute can really be improved a lot. This is only some patchwork.


nickvandewiele:
@bslakman could you review this PR (and merge if you see fit) as it pertains to some features you must be familiar with?

PR #634, opened 2016-04-11
Regression part1

nyee:
First part of a series of commits to implement comparisons with Experimental Data for regression testing.

Most significant contribution in this part is a  new class ComparisonBundle for storing data to be compared. It is rigidly checks that species, units, etc between all data is the same as well as has a few other utility features I thought might be necessary in the future.


JacksonBurns:
Wildly stale. Closing.

PR #637, opened 2016-04-20
Add a method "toGroup" for the conversion of Molecule to Group objects

nickvandewiele:
This PR adds a method that converts a Molecule to Group object.

a unittest checks instance type, atom type, atom count, bond count.

### DIFF
--- a/rmgpy/molecule/group.py
+++ b/rmgpy/molecule/group.py
@@ -534,9 +534,9 @@ class Group(Graph):
     Corresponding alias methods have also been provided.
     """
 
-    def __init__(self, atoms=None):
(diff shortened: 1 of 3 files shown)


connie:
I think these groups will end up being pretty restrictive and identical to their original molecules, since their charges and lone pairs are completely specified.  If this is the intent, then I think it is fine. but it doesn't serve much other purpose than to select for that exact molecule.  

Also, I'm guessing this is intentional, but the `Group.multiplicity` is not set, which is the case for most groups in the database.  

If you wanted to make them more loosely defined, maybe adding a `removeHydrogen` flag might be useful.

Issue #643, opened 2016-05-05
checkForExistingSpecies not working for biaromatics


KEHANG:
During RMG model generation, if some species has more than one aromatic rings and only
some of them are kekulized, current `checkForExistingSpecies` will fail. 

For example, species (`SMILES="C1=CC=C(C=C1)CCCC1C=CC=CC=1"`) already existed, which has two resonance isomers:
![image](https://cloud.githubusercontent.com/assets/2739496/15056234/8e5c2878-12de-11e6-864d-a22d3d487693.png) and 
![image](https://cloud.githubusercontent.com/assets/2739496/15056242/97bc1f0e-12de-11e6-98bb-be7e3cb1853b.png), 
then if new molecule is generated in a partially kekulized form (product from the reaction below): 
![image](https://cloud.githubusercontent.com/assets/2739496/15056314/edbbb770-12de-11e6-855a-2c69378bbf3d.png), 
`checkForExistingSpecies` will fail and treat it as a new species.

KEHANG:
It's fixed by this [PR#644](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/644).

nickvandewiele:
I guess RMG-Py will have more than two resonance isomers for Species C1=CC=C(C=C1)CCCC1C=CC=CC=1, not just the kekulized and aromatic one. 

If you call `spc.generateResonanceIsomers()` on a given isomer of that species how many isomers will be generated?

KEHANG:
@nickvandewiele only two isomers are generated. This is actually an [unittest](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/644/commits/93873bdff90435ae8d112ec3eb31ce4765e96949) I made here.

rwest:
I would only expect 2 resonance isomers for that because it is so symmetrical.
Break the symmetry a bit and you are more likely to run into trouble with making all the resonance forms every time (see #405).
`C1=C(C)C=C(C=C1)CCC=CC1C=CC=C(O)C=1`
![biaromatic](https://cloud.githubusercontent.com/assets/93807/15076656/b7290182-1378-11e6-812c-1ed9804d386e.png)

KEHANG:
@rwest the current issue is the asymmetrical biaromatics is generated by RMG as edge species, which actually should be recognized as same as the symmetrical one that already exists in speciesDict.

rwest:
[...]

PR #644, opened 2016-05-05
Fix of checkForExistingSpecies for biaromatics according to issue#643

KEHANG:
This PR includes
- fix of checkForExistingSpecies for biaromatics according to [issue#643](https://github.com/ReactionMechanismGenerator/RMG-Py/issues/643) where if the molecule has more than one aromatic rings and only some of them are kekulized, the checkForExistingSpecies will fail.By running generateResonanceIsomers in checkForExistingSpecies, it will create isomers with fully
kekulized version, which makes checkForExistingSpecies work again.

- a unittest of checkForExistingSpecies for biaromatics.

### DIFF
--- a/rmgpy/rmg/model.py
+++ b/rmgpy/rmg/model.py
@@ -399,14 +399,29 @@ class CoreEdgeReactionModel:
         and the matched species (if found) or
         ``False`` and ``None`` (if not found).
         """
-        new_spec = Species(molecule=[molecule])
-        new_spec.generateResonanceIsomers()
+        # Create obj to check against existing species
+        # obj can be `Molecule` object or `Species` object
+
+        # For non-cyclic molecules, obj is `Molecule` object
+        # We expect it to be part of the list of isomers in a species
+        # object if it has a match
+        obj = molecule
+
+        # For cyclic molecules, obj is `Species` object and aromatic resonance
+        # isomers are generated.  This is due to the hysteresis of isomer generation
+        # for aromatic/polyaromatic compounds: not all kekulized forms can be found
+        # within the list of isomers for a species object describing a unique aromatic compound
+        if molecule.isCyclic():
+            obj = Species(molecule=[molecule])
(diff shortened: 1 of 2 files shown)


nickvandewiele:
can you replace the under_score style variable name by camelCase?

nickvandewiele:
Seeing this fix, makes me question why the method signature of `CERM.checkForExistingSpecies` contains a `Molecule` parameter instead of a `Species`, especially given the method name...

I understand why this fix works, but calling that very expensive method without reusing any of the results (the newly created species object is never returned/stored), is really a pity.

I wonder if the method could be called elsewhere such that its generated resonance isomers will reused?

connie:
The reason we had to do this was because of faulty resonance isomer generation for aromatics (unidirectional). Which means for some cases you cannot find a particular isomer. But you can generate resonance isomers for it and get the more generalized form. This function using just molecules would work fine if the resonance isomer generation worked perfectly. I also wonder if we can have speed up and do this for only aromatics because as far as I know only the aromatics are affected.

connie:
Also @kehang you should rebate this branch.

rwest:
Yes, seems costly and I think this gets called a lot. At very least only do to cyclic species and only generate the aromatic isomer not all resonance isomers. 
As Connie says, better would be to have resonance forms reliably generated in the first place, but there can be very many and perhaps in the long run the whole aromatic/resonance thing needs rethinking. 

Another thought: Is a type of graph isomorphism possible that treats bond orders flexibly in such a way as to allow different resonance forms to appear isomorphic?

connie:
I believe the graph isomorphism check between two graphs is very rigid/checks for exact bond/multiplicity/electronic structure.  I think it probably needs to stay that way for the purpose of generating reactions..  The flexible bond orders could potentially be applied to list checking molecule to species.molecule, or species to species?

KEHANG:
@rwest I added a new commit which only generate aromatic isomers for cyclic species, so for normal molecules, the method `checkForExistingSpecies` is still the same. Can you review this PR?

PR #649, opened 2016-05-13
do not recompute thermo when it is already available

nickvandewiele:
When an existing RMG job is loaded and its species are used in the RMG enlargement algorithm, then one can re-use the thermochemistry of the loaded species. 

Currently, the thermo is not retained, and recomputed again, with the risk of generating thermo that is disparate from what the loaded RMG model has.

This PR checks if a species that is used as a parameter in `CERM.makeNewSpecies` has a `thermo` attribute. If it is present, it is re-used, and not generated again.

### DIFF
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -460,7 +460,7 @@ class RMG(util.Subject):
                         """.format(spec.label))
                         
             for spec in self.initialSpecies:
-                spec.generateThermoData(self.database, quantumMechanics=self.quantumMechanics)
+                spec.getThermoData(self.database, quantumMechanics=self.quantumMechanics)
                 spec.generateTransportData(self.database)
                 
             # Add nonreactive species (e.g. bath gases) to core first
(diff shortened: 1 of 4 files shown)


rwest:
I like what you're trying to do, but I don't like that `generateThermoData` no longer generates thermo data.
Could we move the check to wherever it is that this method is called? (I haven't looked into this idea in detail - it may be worse; just a thought).

nickvandewiele:
Currently, it will not call the `database.thermo.getThermoData` when a thermo object was found in the Species object. Are you advocating of making that call regardless of the fact of the thermo attribute already exists?

Moving the existence of a thermo attribute in Species could indeed be moved upstream, but I currently don't quite understand the potential risk of the current commit. What situations are you thinking of?

nickvandewiele:
Oh, one minute of thinking got me this scenario: 

1) one generates the thermo for an existing species object.
2) one modifies the species object, e.g. by replacing an unpaired electron by a hydrogen atom
3) one wants to generate the thermo of that modified species

rwest:
If `s` is some instance of `Species` then I would like `s.generateThermoData` to generate thermo data, or else be renamed if that is no longer what it does. (eg. `getThermoData`). The method name and docstrings should reflect what it does.  I think your scenario is a good example.

nickvandewiele:
Understood. I reverted previous changes to the `generateThermoData` method of Species, and instead created a new method `getThermoData`.

Issue #648, opened 2016-05-12
regex used in rmgpy.data.thermo.getRingGroupsFromComments not strict enough


nickvandewiele:
In [rmgpy.data.thermo.getRingGroupsFromComments](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/data/thermo.py#L1312):

```python
        for token in tokens:
            if 'ring' in token:
                splitTokens = re.split("\(|\)",token)
                assert len(splitTokens) == 3
                groupLabel = splitTokens[1]
                ringGroups.append(self.groups['ring'].entries[groupLabel])
            if 'polycyclic' in token:
                splitTokens = re.split("\(|\)",token)
                assert len(splitTokens) == 3
                groupLabel = splitTokens[1]
                polycyclicGroups.append(self.groups['polycyclic'].entries[groupLabel])
        return ringGroups, polycyclicGroups
```

the regex's that search for parentheses, used to identify `polycyclic(...)` are also hitting parentheses that are part of the name of the node, like [here](https://github.com/ReactionMechanismGenerator/RMG-database/blob/master/input/thermo/groups/polycyclic.py#L70), and will lead to an assertion error since the node name is split in more than 3 pieces.

The regex should be constrained to ignore node name parentheses.

nickvandewiele:
closed by #652

PR #652, opened 2016-05-13
Improve the (poly)cyclic node name parser of rmgpy.data.ThermoDatabase

nickvandewiele:
This PR addresses the issue with the (poly)cyclic node name parser, cf. https://github.com/ReactionMechanismGenerator/RMG-Py/issues/648 of `rmgpy.data.ThermoDatabase.getRingGroupsFromComments`

The PR constraints the regex by only matching outermost parentheses of a string, instead of every pair of parentheses. The latter would also hit parentheses contained in the names of polycyclic nodes.


nickvandewiele:
@connie : could you perhaps review this PR? Thanks!

PR #653, opened 2016-05-17
compare reactant/product references in checkForExistingReaction

nickvandewiele:
The current state only of `CERM.checkForExistingReaction` compares reaction labels, which are a
concatenation of species labels. Species labels are often non-canonical strings, like SMILES or anything that is compatible with the chemkin format. As a result, there's a risk that two distinct
species may lead to the same label, leading to a false
positive match in `CERM.checkForExistingReaction`. 

This PR creates a new function, `areIdenticalSpeciesReferences` that check references
of reactants/products in both directions. It is called within `CERM.checkForExistingReaction`. It uses the lazy `any` operator that returns if one element of the parameter array is True.

### DIFF
--- a/rmgpy/rmg/model.py
+++ b/rmgpy/rmg/model.py
@@ -530,14 +530,17 @@ class CoreEdgeReactionModel:
             if (rxn_id == rxn_id0):
                 if isinstance(familyObj, KineticsLibrary):
                     # If the reaction comes from a kinetics library, then we can retain duplicates if they are marked
-                    if not rxn.duplicate:
-                        return True, rxn0
+                    if areIdenticalSpeciesReferences(rxn, rxn0):
+                        if not rxn.duplicate:
+                            return True, rxn0
                 else:
-                    return True, rxn0
+                    if areIdenticalSpeciesReferences(rxn, rxn0):
+                        return True, rxn0
             
             if isinstance(familyObj, KineticsFamily) and familyObj.ownReverse:
                 if (rxn_id == rxn_id0[::-1]):
-                    return True, rxn0
+                    if areIdenticalSpeciesReferences(rxn, rxn0):
+                        return True, rxn0
 
         # Now check seed mechanisms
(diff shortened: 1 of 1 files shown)


nickvandewiele:
@KEHANG : could you review this PR?

KEHANG:
I thought the species label are actually SMILES (see [here](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/67ed48863fa4eae12819dc3e7627b040a99d0309/rmgpy/rmg/model.py#L456)), so two different species shouldn't be having same SMILES/labels, right? Also in method of `areIdenticalSpeciesReferences`, what if we have two reactants for `rxn1` and `rxn2`, if the order is different in the list `rxn.reactants`, `areIdenticalSpeciesReferences` should also return `True`, right?

nickvandewiele:
Thanks for your comments.

- Yes, currently SMILES are used to represent generate species labels, but that can change in the future.
- If the SMILES string is too long to fit into the chemkin format, then it will be replaced by another chemkin-compatible string. If the species originates from a seed mechanism or as a reactant it is using that name as a label. In other words, it's not guaranteed that the species label is a SMILES. 
- the generated SMILES is also a function of what the SMILES generator has implemented. For example, for the application I am intersted in, species containing non-normal isotopes (e.g. C-13) will have the same SMILES as the equivalent with C-12. 

The current PR makes sure that, if the species label is changed to another representation, it won't lead to false positives, because there is the reference comparison as a safety net.

nickvandewiele:
reaction reactants/products are sorted [here:](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/rmg/model.py#L510):

```python
def checkForExistingReaction(self, rxn):
...
    # Make sure the reactant and product lists are sorted before performing the check
    rxn.reactants.sort()
    rxn.products.sort()
```

before the reaction is sent to `areIdenticalSpeciesReferences `.

PR #657, opened 2016-05-18
optimize the failsSpeciesConstraints function

nickvandewiele:
This PR optimizes the frequently called function `rmgpy.constraints.failsSpeciesConstraints`. 

Previously, the atom count for C, H, N, ... was calculated for every molecule, regardless of whether there was a constraint in place. Now, the default value for a constraint is set to -1, so that it can be used to determine whether an atom count should be calculated or not.

### DIFF
--- a/examples/generateReactions/input.py
+++ b/examples/generateReactions/input.py
@@ -59,7 +59,6 @@ simpleReactor(
 #generatedSpeciesConstraints(
 #    allowed=['input species','seed mechanisms','reaction libraries'],
 #    maximumCarbonAtoms=4,
-#    maximumHydrogenAtoms=10,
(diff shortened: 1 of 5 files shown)


connie:
There's no way to get `hydrogenCount` unless both `maxHydrogenAtoms` and `maxHeavyAtoms` are set.  Probably should just reevaluate it for maxHeavyAtoms.

nickvandewiele:
you are right. I will correct that. Thanks!

nickvandewiele:
now that I look at it: who will ever need the `maximumHydrogenAtoms` constraint?

connie:
Good question-- seems like a bad constraint to set...

PR #659, opened 2016-05-19
Isotope support: reading/writing isotope-aware adjacency lists

nickvandewiele:
[edit]: I changed the adjacency list representation to `i13`, with mass in g/mol rather than kg/mol

This PR is a first step in supporting molecules with non-normal isotopes, like C-13.

Isotopic information for atoms with non-normal isotopes is represented by a `i` layer, e.g.: ethane with one carbon-13 is represented as follows:
```python
        1 C u0 p0 c0 i13 {2,S} {3,S} {4,S} {5,S}
        2 C u0 p0 c0 {1,S} {6,S} {7,S} {8,S}
        3 H u0 p0 c0 {1,S}
        4 H u0 p0 c0 {1,S}
        5 H u0 p0 c0 {1,S}
        6 H u0 p0 c0 {2,S}
        7 H u0 p0 c0 {2,S}
        8 H u0 p0 c0 {2,S}
```

This PR implements the following: 
- It writes out adjacency lists with `i`-layers, if present.
- It reads adjacency lists with `i`-layers and converts it into the correct Molecule object.
- It updates the `Atom.equivalent()` method to allow for distinguishing between atoms with distinct masses.
- It adds a new species constraint `maximumIsotopeCount`, that can be added in the RMG-Py input file.

### DIFF
--- a/examples/rmg/commented/input.py
+++ b/examples/rmg/commented/input.py
@@ -207,6 +207,8 @@ generatedSpeciesConstraints(
     #If this is false or missing, RMG will throw an error if the more less-stable form of O2 is entered 
     #which doesn't react in the RMG system. normally input O2 as triplet with SMILES [O][O]
     #allowSingletO2=False,
+    # maximum allowed number of non-normal isotope atoms:
+    #maximumIsotopicAtoms=2,
 )
 
 #optional block allows thermo to be estimated through quantum calculations
--- a/rmgpy/constraints.py
+++ b/rmgpy/constraints.py
@@ -29,7 +29,8 @@
 ################################################################################
 
 import logging
-
+from numpy import isclose
+from rmgpy.molecule.element import getElement
 from rmgpy.species import Species
 
 def failsSpeciesConstraints(species):
@@ -91,5 +92,13 @@ def failsSpeciesConstraints(species):
     if maxRadicals != -1:
         if (struct.getNumberOfRadicalElectrons() > maxRadicals):
             return True
-            
+
+    maxIsotopes = speciesConstraints.get('maximumIsotopicAtoms', -1)
+    if maxIsotopes != -1:
+        counter = 0
+        for atom in struct.atoms:
+            if not isclose(atom.mass, getElement(atom.symbol).mass, atol=1e-04):
+                counter += 1
+            if counter > maxIsotopes: return True
+
     return False
--- a/rmgpy/molecule/adjlist.py
+++ b/rmgpy/molecule/adjlist.py
@@ -638,11 +638,23 @@ def fromAdjacencyList(adjlist, group=False, saturateH=False):
             if not group:
                 partialCharges.append(0)
         
+
+        # Next the isotope (if provided)
+        isotope = -1
+        if len(data) > index:
+            iState = data[index]
+            if iState[0] == 'i':
+                isotope = int(iState[1:])
+                index += 1
+
+
         # Create a new atom based on the above information
         if group:
             atom = GroupAtom(atomType, unpairedElectrons, partialCharges, label, lonePairs)
         else:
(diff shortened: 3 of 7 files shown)


connie:
This looks ok for the most part.  However, I think it might be better to use integers instead of floats to represent the isotopes, particularly in the adjacency lists, I think we would prefer to see `i13` always rather than `i13.0`.  Or for Germanium, we'd see differences like `i71.9` vs `i72` that describe the same isotope, which is a possibility if people write in their own asses.  Each integer is unique in representing an isotopic atom: http://physics.nist.gov/cgi-bin/Compositions/stand_alone.pl, so I think it would be better to use integer values.

This does lead to the issue of decoupling the actual mass of the isotope which differs just slightly from its isotopic value, but it would lead to better uniformity overall and eliminate the need for the `numpy.isclose()` function and going towards a normal equivalency check.  I feel that it would be better if users did NOT have the flexibility of declaring their own masses.  Declaring just the isotopic integer value such as H(2) or C(13) is better.  We might want to create the common isotopic elements, actually create all of them by importing from the list on the NIST website into `element.py` mapping the isomer number to the masses that there is no ambiguity regarding their masses.  

Furthermore, I think the `maximumIsotopeCount` would be better if it was named `maximumIsotopicAtoms`.  I think the former descriptor  could be a bit ambiguous and  indicate the total number of isotopic molecules, not just the max number of isotopic atoms in a single molecule.

nickvandewiele:
@connie Thanks for your remarks, they are very pertinent.

I will change the constraints name to `maximumIsotopicAtoms`.

I like your idea of using a unique integer value that represents the isotope, it would definitely remove the float comparison. 

I am OK with a fixed list of possible isotopes based, as your NIST suggested.

I am still unsure how to implement this in practice: would you create an additional attribute `Element.isotope = Integer` that is decoupled from the `Element.mass = Float`, and a global dictionary that maps the isotope value to the corresponding mass?

When you mention `C(13)`, are you referring to the representation in the adjacency list? Are you saying the isotope information should not be captured in a `i` layer, but appended to the element symbol?

connie:
Regarding the implementation, I think the dictionary mapping would probably work best, as you suggested. So I see that in `element.py` we have a list of created elements, i.e.:

```
C  = Element(6,   'C' , 'carbon'        , 0.0120107)
N  = Element(7,   'N' , 'nitrogen'      , 0.01400674)
O  = Element(8,   'O' , 'oxygen'        , 0.0159994)
```

Potentially with the `Element.isotope` attribute we could then declare a type of isotopic carbon in this list like so, as well as the other ones from NIST (which would be time consuming, but probably better for extensibility. Or shortcut it by just putting in the most common isotopes is fine too):

```
C13  = Element(6,   'C' , 'carbon'  , 0.013003354835, isotope=13)
```

And possibly also append it to the `elementList` at the bottom.  And the `equivalent` function could then just check equivalency of all the attributes exactly (besides mass, which we would fix internally).

The mapping could be through a dictionary just in the adjacency reading code, probably similar in implementation to the `PeriodicSystem` class you have in `rmgpy.molecule.adjlist`.  I think one way to do it is to just add another dictionary within that class called `isotopic_atoms`.

I didn't mean anything by the `C(13)`.  I think the `i` layer is fine, because chemically speaking the molecules are still equivalent.  I think the adjlists should still retain the `C` symbol.  So I think I would stick with the `i` layer.

nickvandewiele:
@connie :

I reshuffled this PR, and introduced the `isotope` attribute in the `element.Element` class, by default set to the integer -1.

I hard-coded a number of new element instances, for D, T, C-13 and O-18, as a starter, and added them the `elementList`. In addition, the `getElement` function is now also comparing the isotope attribute to retrieve the correct element.

As a result, float comparison is removed, and the original atom equivalence method is restored. I updated the affected unit tests with these changes.

PR #660, opened 2016-05-19
Set reactive species attribute to False for inerts in loadSpeciesDictionary

nickvandewiele:
When creating `Species` objects by loading a species dictionary, the `reactive` attribute of the created Species object is not explicitly stored, and results in setting the `reactive` attribute of the created Species object is set to `True` by default.

For inerts like Helium, this is incorrect.

This PR checks a hard-coded tuple with 4 elements (the usual inerts) and compares the species label to the elements in that tuple. If a match is found, the reactive attribute is set to `False` for the created species.

### DIFF
--- a/rmgpy/chemkin.py
+++ b/rmgpy/chemkin.py
@@ -691,6 +691,7 @@ def loadSpeciesDictionary(path):
     """
     speciesDict = {}
     
+    inerts = [Species().fromSMILES(inert) for inert in ('[He]', '[Ne]', 'N#N', '[Ar]')]
     with open(path, 'r') as f:
         adjlist = ''
         for line in f:
@@ -699,6 +700,10 @@ def loadSpeciesDictionary(path):
                 species = Species().fromAdjacencyList(adjlist)
                 species.generateResonanceIsomers()
                 label = species.label
+                for inert in inerts:
+                    if inert.isIsomorphic(species):
+                        species.reactive = False
+                        break
                 speciesDict[label] = species
                 adjlist = ''
             else:


nyee:
@alongd You should probably be made aware that N2 is being made inert by default on master branch any time it is loaded from an RMG-dictionary

Do you think it is safer to do an isomorphism check against a species. Something like:

`Ar=Species().fromSMILES(['Ar']) #Add other inerts same way`
`inertList=[Ar, He, N2, Ne]`
.
.
.
`for inert in inertList:`
    `if inert.isIsomorphic(species): species.reactive=False`

This adds 4 isomorphism checks per species, so it will slow it down a bit. Maybe its not worth it...If you prefer string labels, it is probably okay also. Let me know and I'll merge it.

nickvandewiele:
yes, isomorphisms is definitely the solution here. I will implement this.

alongd:
Alright.
If N2 is predefined as reactive, this attribute isn't overridden, right?

nyee:
That's a good question Alon, I'm not actually sure. Just note that this is for when we load in a species dictionary, and not  for RMG input files. I am not sure how it will interact between an RMG input file with N2.reactive=True and then loading a species dictionary.

connie:
I think this is going to be a problem: the `loadSpeciesDictionary` function gets used through the `loadChemkinFile` function, which is used in `loadRMGPyJob` here: https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/tools/loader.py#L83 

So function specifically replaces the input file species with the chemkin species objects, which would have `reactive=False` even if it was set to `True` by the user.  The `loadRMGPyJob` function so far is not used for much other than rerunning a simulation or sensitivity analysis, but this could be unexpected functionality if there is disagreement with what's in the input.py file.

The CHEMKIN reaction list could also have species like N2 reacting, but the `loadSpeciesDictionary` function is indiscriminatory in setting `reactive=False` flags: https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/chemkin.py#L705

Issue #661, opened 2016-05-19
Reactions created that violate mass balance


nyee:
On an RMG job for phenol pyrolysis, I am getting reactions like this:

![image](https://cloud.githubusercontent.com/assets/1444635/15409828/95dbd59c-1de4-11e6-9066-9f13fd95b5f2.png)

<img width="840" alt="screen shot 2016-05-19 at 5 28 34 pm" src="https://cloud.githubusercontent.com/assets/1444635/15410341/281c5862-1de7-11e6-9402-01f5525b3377.png">

Both Connie and I have seen this occur on the master branch. Here is what we know so far:
1. So far it only happens with reactions with acetylene
2. It occurs for multiple reaction families, not from any library reactions
3. RMG does not throw any exceptions or crash. I have gotten to a fully converged RMG mechanism while making many imbalanced reactions. 
4. I get the error when phenol and Ar start in the input file. I do not get the error if I start with acetylene, phenol, and Ar to try to get the reaction to appear earlier.
5. I have only been able to get the error when I use a specific reaction library. This reaction library was converted from an RMG-generated mechanism for n-butane oxidation. I don't know if the library is causing the error or just influencing RMG to generate specific reactions where the error occurs. 

From above, you can see it we have had difficulties trouble-shooting the error. At best, it takes us about ten minutes to verify if the occurs or not. Anybody have any idea what is going on?

nickvandewiele:
call [reaction.isBalanced()](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/reaction.py#L879) after a reaction is created and raise an Exception when the method returns False.

rwest:
I notice that the nodes it hit when finding kinetics for the reactions are unambiguously an H atom radical, `1 * H u1`, which means that at the time of searching for kinetics the Molecules were correct and the reaction balanced. At some point that Molecule must be being swapped for the pointer to the C2H2 Species. Possibly some "check for existing" is going wrong. Or maybe it's to do with the scoop worker/master infrastructure eg `inflate` is getting the wrong species index?  I'd love to understand that new code better than I currently do.

Nick's is a good suggestion to help debug. Use a debugger that pauses when you hit that exception.

nickvandewiele:
Can someone add an input file so that we can all test and enjoy this?

nyee:
Here is a reaction library and input file that I see the error on. I don't seem to get the error if I don't use the reaction library. 

[...]

PR #662, opened 2016-05-20
Fix species index bug in react module

nickvandewiele:
This PR fixes a bug mentioned [here](https://github.com/ReactionMechanismGenerator/RMG-Py/commit/c147f77a056531d7c135b66b267ae1e29d0b0919#commitcomment-17551953), which might be possibly related to [this issue](https://github.com/ReactionMechanismGenerator/RMG-Py/issues/661) as well.

The bug stored the wrong species index as a reactant in the reaction. When the reaction is _inflated_ again, it should result in replacing the index with the wrong species. This may result in unbalanced reactions.

A number of unit tests were added, and a more comprehensive inline documentation was added to explain the functionality of the functions.

How this bug did not end up generating _more_ recent RMG crashes is really still a mystery to me, and  is being investigated.

### DIFF
--- a/rmgpy/rmg/react.py
+++ b/rmgpy/rmg/react.py
@@ -43,14 +43,35 @@ def react(spcA, speciesList=[]):
     """
     Generate reactions between spcA and the list of 
     species for all the reaction families available.
+
+    Returns an empty list if the spcA is non-reactive.
+
+    For the spcA, a list of tuples is created for each
+    resonance isomer of the species. Each tuple consists of (Molecule, index)
+    with the index the species index of the Species object.
+
+    For each Species in the the speciesList, its corresponding
+    resonance isomers are stored in a tuple ([Molecule], index) with the index 
+    the species index of the Species object.
+
+    Each tuple ([Molecule], index) is expanded into a list of tuples (Molecule, index)
+    resulting in one large list [(Molecule, index)].
+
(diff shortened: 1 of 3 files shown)


nickvandewiele:
@nyee Could you review this PR and perhaps use this branch to test issue #661 ?

rmgpy/rmg/react.py
```diff
@@ -43,14 +43,35 @@ def react(spcA, speciesList=[]):
     """
     Generate reactions between spcA and the list of 
     species for all the reaction families available.
+
+    Returns an empty list of the parameter species is non-reactive.
```
rwest: "if" not "of"

rwest: Plus I think it would be good to mention the name `spcA` (also on line 49, if that's what the 'method parameter species' means.

rwest:
Thanks Nick - the docstrings are a great help. 
Are any higher level unit tests to catch downstream consequences possible? Tests that would have failed before this fix. This increase in coverage would prevent other bugs.

nickvandewiele:
- improved the docstrings following Richard's advice.
- added some unit tests for CoreEdgeReactionModel that should run a lot of code within `rmgpy.rmg.model`. 

Although it did not solve #661, could someone review this PR and merge it?

PR #664, opened 2016-06-01
Fix trees

nyee:
Some changes related to my attempts to fix the trees as related to RMG-database issue

### DIFF
--- a/rmgpy/data/base.py
+++ b/rmgpy/data/base.py
@@ -875,10 +875,10 @@ class Database:
             return False
         
         #If the parentNode is a Group and the childNode is a LogicOr there is nothing to check,
-        #so it gets an automatic pass. However, we do need to check that everything down this
+        #except that the parent is listed in the attributes. However, we do need to check that everything down this
         #family line is consistent, which is done in the databaseTest unitTest
         elif isinstance(parentNode.item, Group) and isinstance(childNode.item, LogicOr):
-            return True
+            return childNode.parent is parentNode
         
         elif isinstance(parentNode.item,LogicOr):
             return childNode.label in parentNode.item.components
--- a/rmgpy/data/base.py
+++ b/rmgpy/data/base.py
@@ -1003,7 +1003,48 @@ class Database:
             #logging.warning('For {0}, a node {1} with overlapping children {2} was encountered in tree with top level nodes {3}. Assuming the first match is the better one.'.format(structure, root, next, self.top))
             return self.descendTree(structure, atoms, next[0], strict)
 
-################################################################################
+    def areSiblings(self, node, nodeOther):
+        """
+        Return `True` if `node` and `nodeOther` have the same parent node.  Otherwise, return `False`.
+        Both `node` and `nodeOther` must be Entry types with items containing Group or LogicNode types.
+        """
+        if node.parent is nodeOther.parent: return True
+        else: return False
+
+    def removeGroup(self, groupToRemove):
+        """
+        Removes a group that is in a tree from the database. In addition to deleting from self.entries,
+        it must also update the parent/child relationships
+
+        Returns the removed group
+        """
+        #Don't remove top nodes or LogicOrs as this will cause lots of problems
+        if groupToRemove in self.top:
+            raise Exception("Cannot remove top node: {0} from {1} because it is a top node".format(groupToRemove, self))
+        elif isinstance(groupToRemove.item, LogicOr):
+            raise Exception ("Cannot remove top node: {0} from {1} because it is a LogicOr".format(groupToRemove, self))
+        #Remove from entryToRemove from entries
+        self.entries.pop(groupToRemove.label)
+
+        #If there is a parent, then the group exists in a tree and we should edit relatives
+        parentR=groupToRemove.parent
+        if not parentR is None:
+            #Remove from parent's children attribute
+            parentR.children.remove(groupToRemove)
+
+            #change children's parent attribute to former grandparent
+            for child in groupToRemove.children:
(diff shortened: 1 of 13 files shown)


rwest:
I made some suggestions. Once you've checked them, @nyee, you should be able to
`git rebase -i --autosquash master fixTrees`  and it will put them into place as fixup commits and merge them with yours.

nyee:
Thanks for the suggestions, Richard. I have merged them in and forced pushed the updated version to this branch.

nyee:
This latest test passes for the RMG-database branch [fixTrees](https://github.com/nyee/RMG-database/tree/fixTrees).  The same test should be implemented on kinetics, solvation, transport, and statmech also, but I didn't want to put it in yet because no RMG-database commit will pass those tests.

In terms of debugging those other databases: It is easier to use this script to automatically rearrange the trees:

[fixSiblings.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/298335/fixSiblings.txt)

The above script should **not** be trusted used blindly. For example, it might be that groups are not [explicit enough](https://github.com/nyee/RMG-database/commit/71ea6851397cfb808a41ea7e6812a1e50ca1b7b4), rather than the tree is set up wrong.

rmgpy/data/base.py
```diff
@@ -875,10 +875,10 @@ def matchNodeToChild(self, parentNode, childNode):
             return False
         
         #If the parentNode is a Group and the childNode is a LogicOr there is nothing to check,
-        #so it gets an automatic pass. However, we do need to check that everything down this
+        #except that the parent is listed in the attributes. However, we do need to check that everything down this
         #family line is consistent, which is done in the databaseTest unitTest
         elif isinstance(parentNode.item, Group) and isinstance(childNode.item, LogicOr):
-            return True
+            return childNode.parent is parentNode
```
KEHANG: in order to spot possible suspicious parent/child, for this if situation `isinstance(parentNode.item, Group) and isinstance(childNode.item, LogicOr)`, we  might also want to check the items in the `LogicOr` are children of the `parentNode`

rmgpy/data/thermo.py
```diff
@@ -423,6 +423,53 @@ def processOldLibraryEntry(self, data):
...
+        #Remove from parent's children attribute
+        parentR.children.remove(entryToRemove)
+
+        #change children's parent attribute to former grandparent
+        for child in entryToRemove.children:
+            child.parent = parentR
+
+        #extend parent's children attribute with new children
+        parentR.children.extend(entryToRemove.children)
+
```
KEHANG: please add an unittest for this helpful method

testing/databaseTest.py
```diff
@@ -259,6 +277,26 @@ def kinetics_checkChildParentRelationships(self, family_name):
                     nose.tools.assert_true(family.matchNodeToChild(ancestorNode, childNode),
                                     "In {family} family, group {ancestor} is not a proper ancestor of its child {child}.".format(family=family_name, ancestor=ancestorNode, child=nodeName))
 
+    def kinetics__checkSiblingsForParents(self, family_name):
```
KEHANG: remove one underscore in the method name?

rmgpy/data/kinetics/common.py
```diff
@@ -83,40 +83,44 @@ def sortEfficiencies(efficiencies0):
     if entry.label != '':
         f.write('    label = "{0}",\n'.format(entry.label))
 
+
+
     if isinstance(entry.item, Reaction):
-        if entry.label != str(entry.item):
```
KEHANG: seems this `saveEntry` method is used for many purposes, can we also add some comments in the method to clarify which case (e.g., save to `groups.py`, `rules.py`, `reactions.py`) is for which if statement?

testing/databaseTest.py
nyee: Did some rebasing and this is now fixed in commit 1bba6c

rmgpy/data/kinetics/common.py
nyee: Fixed up in commit fdb29

nyee:
Added unittest for removeGroup with latest commit. Discovered some bugs along the way and fixed them in the fixup! commit for the removeGroup method.

Unfortunately, I can't run make test on my computer, and I'm not sure I used the correct path to the test_data, so I need travis to tell me whether the unittest passes.

rmgpy/data/base.py
nyee: As said in the comment above, we actually do this in [databaseTest.py](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/059b895b67370b2daa5699c3b5c84ffc87a36add/testing/databaseTest.py#L245-L253
l). 

In my opinion, the current implementation, unless we want to start using the matchNodeToChild for things other than database unittests.

rmgpy/data/thermo.py
```diff
@@ -421,6 +421,66 @@ def processOldLibraryEntry(self, data):
...
+                parentR.reference = groupToRemove.reference
+                parentR._longDesc = groupToRemove._longDesc
+                parentR._shortDesc = groupToRemove._shortDesc
+                parentR._longDesc = groupToRemove.longDesc
+                parentR._shortDesc = groupToRemove.shortDesc
+                parentR.rank = groupToRemove.rank
+                parentR.referenceType = groupToRemove.referenceType
+
+        #look for other pointers that point toward entry
+        for entryName, entry in self.entries.iteritems():
```
KEHANG: I think this `for` should be combined with the `if` statement starting at line 457 in order to make it more concise.

PR #666, opened 2016-06-03
Fixed unit error in the solute radius

yunsiechung:
There was a unit error in the conversion of the solute's McGowan volume to the radius when calculating the solute's diffusivity. The unit of the McGowan volume is cm^3/mol/100, and the unit of the radius should be meter. Therefore, the McGowan volume needs to be divided by Avogadro's number to cancel out mol^-1. There was the same error in the calculation of diffusion rates. 

The two errors canceled out completely so the final values of diffusion rates were not affected at all. But the corrections were made for clarification

### DIFF
--- a/rmgpy/data/solvation.py
+++ b/rmgpy/data/solvation.py
@@ -211,8 +211,9 @@ class SoluteData():
         Get diffusivity of solute using the Stokes-Einstein sphere relation. Radius is 
         found from the McGowan volume.
         """
+        N_a=6.0221409e+23 #Avogadro's number, mol^-1
         k_b = 1.3806488e-23 # m2*kg/s2/K
-        radius = math.pow((75*self.V/3.14159),(1.0/3.0))/100 # in meters
+        radius = math.pow((75*self.V/(3.14159*N_a)),(1.0/3.0))/100 # in meters
         D = k_b*T/6/3.14159/solventViscosity/radius # m2/s
         return D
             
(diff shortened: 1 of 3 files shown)


nickvandewiele:
great catch, @yunsiechung !

I noticed that you use hard-coded constants, eg for avogadro. Could you perhaps import them from the constants module:

```python
# at the import section:
...

foo = bar * constants.Na

```

The less hard-coded constants we have, the better!

yunsiechung:
That's a better idea. I will do that now. Since I don't know whether there are any ways to modify the current, I will just close this and add another pull request using the constants module.
Thanks Nick!

bslakman:
Also check the liquid branch in ReactionMechanismGenerator? I think @goldmanm  and @rwest made some changes regarding this

yunsiechung:
@bslakman Thank you for letting me know. The same corrections were made in the liquid branch. 

Why are these changes not included in master branch? Is there reason why these are kept separately in the liquid branch?

rwest:
See https://github.com/ReactionMechanismGenerator/RMG-Py/compare/liquid
I thought @goldmann was going to make a pull request from this branch, but maybe @yunsiechung would like to instead.  Or I could if you prefer.

yunsiechung:
@rwest  I could make a pull request from this branch after I make some modifications that Nick suggested. I am also trying to learn how to make a pull request, so it would be a good practice for me.
Thank you for help!

PR #669, opened 2016-06-06
Fixed unit error in liquid phase reactions and added diffusionLimiter function

yunsiechung:
The unit error regarding the diffusivity is corrected. 
Also, all the constants, such as pi, Na and kB are all replaced by the values from rmgpy.constants module. 

fixDiffusionLimitedA(T) function is added for liquid phase reactions. When it is enabled, the Pre-exponential factor A is decreased by the diffusion factor evaluated at a reaction temperature. This corrected pre-exponential factor can be used only for isothermal liquid reactions.

### DIFF
--- a/rmgpy/data/solvation.py
+++ b/rmgpy/data/solvation.py
@@ -208,13 +208,16 @@ class SoluteData():
     
     def getStokesDiffusivity(self, T, solventViscosity):
         """
-        Get diffusivity of solute using the Stokes-Einstein sphere relation. Radius is 
-        found from the McGowan volume.
+        Get diffusivity of solute using the Stokes-Einstein sphere relation. 
+        Radius is found from the McGowan volume.
+        solventViscosity should be given in  kg/s/m which equals Pa.s
+        (water is about 9e-4 Pa.s at 25C, propanol is 2e-3 Pa.s)
+        Returns D in m2/s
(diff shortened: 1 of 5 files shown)


rmgpy/data/solvation.py
```diff
@@ -214,9 +215,8 @@ def getStokesDiffusivity(self, T, solventViscosity):
         (water is about 9e-4 Pa.s at 25C, propanol is 2e-3 Pa.s)
         Returns D in m2/s
         """
-        k_b = 1.3806488e-23 # m2*kg/s2/K
-        radius = math.pow((75*self.V/3.14159/6.0221409e23),(1.0/3.0))/100 # in meters, V is in MgGowan volume in cm3/mol/100
-        D = k_b*T/6/3.14159/solventViscosity/radius # m2/s
+        radius = math.pow((75*self.V/constants.pi/constants.Na),(1.0/3.0))/100 # in meters, V is in MgGowan volume in cm3/mol/100
+        D = constants.kB*T/6/3.14159/solventViscosity/radius # m2/s
```
rwest: You still have a 3.14159 here

bslakman:
I changed the variable name to the new name introduced by Nick in db16db82. Will merge once Travis passes.

PR #679, opened 2016-06-13
update the docstring of KineticsFamily.generateReactions

nickvandewiele:
This PR updates the docstring of one method, to reflect that the reactants/products of some reactions may store indices referring to Species objects rather than the Species objects themselves.


rwest:
I think this function returns reactions with `Molecule` objects as reactants and products. The deflating to `Species` and `int` objects happens elsewhere.

PR #680, opened 2016-06-14
Add CoolProp dependency to anaconda build and environment

connie:
Coolprop is a fluids property estimator (http://coolprop.org) and its binary now found on https://anaconda.org/RMG/coolprop

Adding it as a dependency in anticipation for future solvation module reliance on this program.


yunsiechung:
Thanks Connie!

PR #686, opened 2016-06-22
Fix parallel generation of reactions

nickvandewiele:
Before the reaction filtering modifications, the `enlarge` method would react one new core species against N existing species. After that change, the `enlarge` method would react all N core species against all N species, but filter out those combinations that would lead to negligible reactions.

The original design of the parallel generation of reactions would be used in the new O(N²) design of `enlarge`, resulting in zero speedup.

This PR fixes that by modifying the parallel design to the new structure of `enlarge`. A new function `rmgpy.rmg.react.reactAll` encapsulates the reaction generation.

A critical section in CERM.enlarge is shown here:
```python
rxns = reactAll(self.core.species, numOldCoreSpecies, unimolecularReact, bimolecularReact)
spcs = [self.retrieveNewSpecies(rxn) for rxn in rxns]
            
for rxn, spc in zip(rxns, spcs):
     rxn = self.inflate(rxn)    
     self.processNewReactions([rxn], spc)
```
`CERM.processNewReactions` requires passing the Species `spc` that was used to originally generate the reactions. In the parallel design the reference to that species is lost due to de/serialization to/from workers. Hence we need to find it back somehow and this is done in `CERM.retrieveNewSpecies`. The method searches for any reactant/product of the given reaction that is an integer instead of a Molecule object. An integer is an indication that the species was already part of the core, and hence refers to the original Species object.

### DIFF
--- a/rmgpy/rmg/model.py
+++ b/rmgpy/rmg/model.py
@@ -1563,7 +1563,7 @@ class CoreEdgeReactionModel:
                 self.networkList.append(network)
 
         # Add the path reaction to that network
-        network.addPathReaction(newReaction, newSpecies)
+        network.addPathReaction(newReaction)
         
         # Return the network that the reaction was added to
         return network
--- a/rmgpy/rmg/model.py
+++ b/rmgpy/rmg/model.py
@@ -54,7 +54,7 @@ from rmgpy.data.kinetics.library import KineticsLibrary, LibraryReaction
 
 from rmgpy.kinetics import KineticsData
 import rmgpy.data.rmg
-from .react import react
+from .react import reactAll
 
 from pdep import PDepReaction, PDepNetwork
 # generateThermoDataFromQM under the Species class imports the qm package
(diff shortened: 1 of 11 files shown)


rmgpy/rmg/model.py
```diff
@@ -768,23 +768,24 @@ def enlarge(self, newObject=None, reactEdge=False, unimolecularReact=None, bimol
...
-                        reactions = [self.inflate(reaction) for reaction in reactions]
-                        self.processNewReactions(reactions, self.core.species[j], None)
+                        if self.core.species[i].reactive and self.core.species[j].reactive:
+                            spcTuples.append((self.core.species[i].copy(deep=True), self.core.species[j].copy(deep=True)))
+
+            rxns = list(react(*spcTuples))
+            spcs = [self.retrieveNewSpecies(rxn) for rxn in rxns]
+            
+            for rxn, spc in zip(rxns, spcs):
+                rxn = self.inflate(rxn)
```
KEHANG: from reading the steps in method `inflate`, it doesn't seem to take care of the reactions which have `reverse` attribute which is also a `Reaction` type containing reactants and products, they may still be integer?

KEHANG:
the unittest `testDeflate` in `reactTest.py` should be changed accordingly

nickvandewiele:
2b2fdcf inflates the reverse attribute of reaction

PR #696, opened 2016-06-30
Avoid error in database test for nodes with no parents.

rwest:
Nothing in the official master database currently fails this test, but it can catch some errors when adding new reaction families or editing the database.

### DIFF
--- a/testing/databaseTest.py
+++ b/testing/databaseTest.py
@@ -298,6 +298,11 @@ class TestDatabase():  # cannot inherit from unittest.TestCase if we want to use
             #top nodes and product nodes don't have parents by definition, so they get an automatic pass:
             if childNode in originalFamily.groups.top or childNode in originalFamily.forwardTemplate.products: continue
             parentNode = childNode.parent
+            
+            if parentNode is None:
+                # This is a mistake in the database, but it should be caught by kinetics_checkGroupsFoundInTree
+                # so rather than report it twice or crash, we'll just silently carry on to the next node.
+                continue
             # Check whether the node has proper parents unless it is the top reactant or product node
             # The parent should be more general than the child
             nose.tools.assert_true(family.matchNodeToChild(parentNode, childNode),


connie:
I think this is redundent.  We already have it in `general_checkNodesFoundInTree` in the `databaseTest.py`, right? https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/testing/databaseTest.py#L437

nyee:
Is this catching something that is not being caught by the test Connie pointed out? 

Also when I was looking at the code, it occurred to me that the product nodes created by training reactions should be caught. It doesn't look like we have any statements to catch that. Are we loading the the training reactions during the databasetest? https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/testing/databaseTest.py#L29


I thought we do by default.

rwest:
Without this patch one gets 
```
ERROR: Kinetics family Cl_Abstraction: parent-child relationships are correct?
  File "/Users/rwest/anaconda/envs/rmg5/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/rwest/Code/RMG-Py/testing/databaseTest.py", line 59, in <lambda>
    test = lambda x: self.kinetics_checkChildParentRelationships(family_name)
  File "/Users/rwest/Code/RMG-Py/testing/databaseTest.py", line 303, in kinetics_checkChildParentRelationships
    nose.tools.assert_true(family.matchNodeToChild(parentNode, childNode),
  File "/Users/rwest/Code/RMG-Py/rmgpy/data/base.py", line 871, in matchNodeToChild
    if isinstance(parentNode.item, Group) and isinstance(childNode.item, Group):
AttributeError: 'NoneType' object has no attribute 'item'
```
and 
```
FAIL: Kinetics family Cl_Abstraction: groups are in the tree with proper parents?
  File "/Users/rwest/anaconda/envs/rmg5/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/rwest/Code/RMG-Py/testing/databaseTest.py", line 47, in <lambda>
    test = lambda x: self.kinetics_checkGroupsFoundInTree(family_name)
  File "/Users/rwest/Code/RMG-Py/testing/databaseTest.py", line 272, in kinetics_checkGroupsFoundInTree
    nose.tools.assert_true(ascendParent is not None, "Group {group} in {family} family was found in the tree without a proper parent.".format(group=child, family=family_name))
AssertionError: Group Cd_rad in Cl_Abstraction family was found in the tree without a proper parent.
    'Group Cd_rad in Cl_Abstraction family was found in the tree without a proper parent.' = self._formatMessage('Group Cd_rad in Cl_Abstraction family was found in the tree without a proper parent.', "%s is not true" % safe_repr(False))
>>  raise self.failureException('Group Cd_rad in Cl_Abstraction family was found in the tree without a proper parent.')
```

The second FAIL does report the problem, but the first ERROR is confusing. `AttributeError: 'NoneType' object has no attribute 'item'` does not help you debug the database.

With this patch one still gets the kinetics_checkGroupsFoundInTree FAIL
```
FAIL: Kinetics family Cl_Abstraction: groups are in the tree with proper parents?
  File "/Users/rwest/anaconda/envs/rmg5/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/rwest/Code/RMG-Py/testing/databaseTest.py", line 47, in <lambda>
    test = lambda x: self.kinetics_checkGroupsFoundInTree(family_name)
  File "/Users/rwest/Code/RMG-Py/testing/databaseTest.py", line 272, in kinetics_checkGroupsFoundInTree
    nose.tools.assert_true(ascendParent is not None, "Group {group} in {family} family was found in the tree without a proper parent.".format(group=child, family=family_name))
AssertionError: Group Cd_rad in Cl_Abstraction family was found in the tree without a proper parent.
    'Group Cd_rad in Cl_Abstraction family was found in the tree without a proper parent.' = self._formatMessage('Group Cd_rad in Cl_Abstraction family was found in the tree without a proper parent.', "%s is not true" % safe_repr(False))
>>  raise self.failureException('Group Cd_rad in Cl_Abstraction family was found in the tree without a proper parent.')
```
but now the confusing ERROR message is replaced with a more helpful FAIL
```
FAIL: Kinetics family Cl_Abstraction: parent-child relationships are correct?
  File "/Users/rwest/anaconda/envs/rmg5/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/rwest/Code/RMG-Py/testing/databaseTest.py", line 59, in <lambda>
    test = lambda x: self.kinetics_checkChildParentRelationships(family_name)
  File "/Users/rwest/Code/RMG-Py/testing/databaseTest.py", line 304, in kinetics_checkChildParentRelationships
    originalFamily
AssertionError: Parent of Cd_rad is None but it's not in <ReactionFamily "Cl_Abstraction">.groups.top
    'Parent of Cd_rad is None but it\'s not in <ReactionFamily "Cl_Abstraction">.groups.top' = self._formatMessage('Parent of Cd_rad is None but it\'s not in <ReactionFamily "Cl_Abstraction">.groups.top', "%s is not false" % safe_repr(True))
>>  raise self.failureException('Parent of Cd_rad is None but it\'s not in <ReactionFamily "Cl_Abstraction">.groups.top')
```


So it doesn't really catch anything new, but it does turn an unhelpful ERROR into a helpful FAIL.

Not sure I understand @nyee's second point. Is that something I changed or did you just notice a test that you could add?

nyee:
My second point has to do with an exception that we don't directly address. Sometimes we get group entries in the family that are actually products with some generic name 'productx', where 'product' is a label from forwardTemplate.products, and 'x' an integer. These entries will all be direct children of the entries from fowardTemplate.products. You can see we have some lines of code that work around this issue: https://github.com/rwest/RMG-Py/blob/3385acdde8189ccc88d8396132860988577481f0/testing/databaseTest.py#L269

I was thinking that not all our tests are set up to handle this situation. Even backing up, I'm not even sure when or what causes there products nodes to be created, although I think its related to training reactions. I tried loading the database right now and didn't see any product nodes, but I know I've had to deal with the issue before. Anyway it doesn't cause any problems right now, so maybe it doesn't warrant further investigation. 

@rwest If nobody wants to look into it, I'll merge your PR. Its definitely an improvement over the current code.

rwest:
Do you prefer the solution in the most recent commit? Silence this error, because it's caught in another test.

(As for the product nodes, they're probably auto-generated for reverse template matching purposes. If the top node of reactants is a LogicOR then there will need to be several product node templates created. I don't think it's causing problems here.)

rwest:
I've squashed the commits and rebased. Any objections to merging if Travis passes?

nyee:
Looks fine to me.

PR #698, opened 2016-06-30
Stops the markDuplicateReaction function from marking edge species duplicates

rgillis8:
…if the edge species are not going to be saved.


connie:
@nickvandewiele Is adding this `saveEdgeSpecies` flag the best way to do this, and will it affect the parallel enlarge?

nickvandewiele:
I would advocate against this implementation as it adds an additional parameter to the method.
Why not simply store that flag as an object attribute in a CERM instance, and access it from the instance method when needed?

connie:
Makes sense, seeing as we don't need to continually pass the flag in because it is the same throughout a job.

rgillis8:
I'll try and re-implement it as per Nick's suggestion.

rgillis8:
Ok. I think that this is a better implementation. i have retested it as well and it has identical behavior to the previous implementation.

PR #701, opened 2016-07-01
Strict parents

nyee:
This PR fixes a bug where matchNodeToChild (a function entirely used to check the database) was not checking atomLabels. Right now this will fail the database test on travis until the PR with the same name on RMG-database is merged.


nyee:
@connie reviewed this and gave me permission to merge

PR #705, opened 2016-07-06
Atomtypes

nyee:
This PR moves the hard-coded portion of atomTypes from the function getAtomType to the initialization of each atomType. I think this is better because each Atomtype instance previously was not tied in anyway with its definition and could not actually be defined without the getAtomType function.

The new implementation of getAtomType might be slower than previously as it matches more objects, but @KEHANG was telling me that atomtyping is not really limiting. I'm not very familiar with speed checks, so it might be worthwhile for somebody else to try it.

I seemed to have made a mess the branch naming for #700, so I've created a new pull request. Much more updated

### DIFF
--- a/rmgpy/molecule/atomtype.py
+++ b/rmgpy/molecule/atomtype.py
@@ -75,9 +75,21 @@ class AtomType:
     `decrementRadical`  ``list``            The atom type(s) that result when the number of radical electrons is decremented
     `incrementLonePair` ``list``            The atom type(s) that result when the number of lone electron pairs is incremented
     `decrementLonePair` ``list``            The atom type(s) that result when the number of lone electron pairs is decremented
+
+    The following features are what are required in a given atomtype. Any int in the list is acceptable.
+    An empty list is a wildcard
+    'single'            ''list''            The total number of single bonds on the atom
+    'allDouble'         ''list''            The total number of double bonds on the atom
+    'rDouble'           ''list''            The number of double bonds to any non-oxygen, nonsulfur
+    'oDouble'           ''list''            The number of double bonds to oxygen
+    'sDouble'           ''list''            The number of double bonds to sulfur
+    'triple'            ''list''            The total number of triple bonds on the atom
+    'benzene'           ''list''            The total number of benzene bonds on the atom
+    'lonePairs'         ''list''            The number of lone pairs on the atom
     =================== =================== ====================================
 
     """
+    
 
     def __init__(self, label='', generic=None, specific=None,
                  single=None,
@@ -99,15 +111,14 @@ class AtomType:
         self.decrementRadical = []
         self.incrementLonePair = []
         self.decrementLonePair = []
-        #Number of a bonds allowed in the atomtype, (empty list is a wildcard)
-        self.single = [] if single is None else single #total number of single bonds
-        self.allDouble = [] if allDouble is None else allDouble #total number of double bonds
-        self.rDouble = [] if rDouble is None else rDouble #double bonds to anything except an O or S
+        self.single = [] if single is None else single
+        self.allDouble = [] if allDouble is None else allDouble
+        self.rDouble = [] if rDouble is None else rDouble
         self.oDouble = [] if oDouble is None else oDouble
         self.sDouble = [] if sDouble is None else sDouble
-        self.triple = []  if triple is None else triple #total number of triple bonds
-        self.benzene = [] if benzene is None else benzene #total number of benzene bonds
-        self.lonePairs = [] if lonePairs is None else lonePairs #total number of lonePairs
+        self.triple = [] if triple is None else triple
+        self.benzene = [] if benzene is None else benzene
+        self.lonePairs = [] if lonePairs is None else lonePairs
 
     def __repr__(self):
         return '<AtomType "%s">' % self.label
@@ -187,6 +198,20 @@ class AtomType:
         atom type `atomType2` or ``False``  otherwise.
         """
         return self is other or self in other.specific
+    
+    def getFeatures(self):
+        """
+        Returns a list of the features that are checked to determine atomtype
+        """
+        features=[self.single,
+                  self.allDouble,
+                  self.rDouble,
+                  self.oDouble,
+                  self.sDouble,
+                  self.triple,
+                  self.benzene,
+                  self.lonePairs,]
+        return features
 
 ################################################################################
 
@@ -245,7 +270,7 @@ atomTypes['Cs'  ] = AtomType('Cs',   generic=['R','R!H','C','Val4'],  specific=[
 atomTypes['Cd'  ] = AtomType('Cd',   generic=['R','R!H','C','Val4'],  specific=[],
                              single=[], allDouble=[1], rDouble=[1], oDouble=[0], sDouble=[0], triple=[0], benzene=[0])
 atomTypes['Cdd' ] = AtomType('Cdd',  generic=['R','R!H','C','Val4'],  specific=[],
-                             single=[0], allDouble=[2], rDouble=[], oDouble=[], sDouble=[], triple=[0], benzene=[0])
+                             single=[0], allDouble=[2], rDouble=[0,1,2], oDouble=[0,1,2], sDouble=[0,1,2], triple=[0], benzene=[0])
 atomTypes['Ct'  ] = AtomType('Ct',   generic=['R','R!H','C','Val4'],  specific=[],
                              single=[], allDouble=[0], rDouble=[], oDouble=[], sDouble=[], triple=[1], benzene=[0])
 atomTypes['CO'  ] = AtomType('CO',   generic=['R','R!H','C','Val4'],  specific=[],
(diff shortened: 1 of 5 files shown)


nickvandewiele:
RMG-tests failed: https://travis-ci.org/ReactionMechanismGenerator/RMG-tests/builds/142657448

```
$ ./run.sh NC
Running NC example
Version of RMG: /home/travis/miniconda/envs/benchmark/bin
/home/travis/miniconda/envs/benchmark/lib/python2.7/site-packages/matplotlib/font_manager.py:273: UserWarning: Matplotlib is building the font cache using fc-list. This may take a moment.
  warnings.warn('Matplotlib is building the font cache using fc-list. This may take a moment.')
Version of RMG: /home/travis/build/ReactionMechanismGenerator/code/tested/RMG-Py
/home/travis/miniconda/envs/tested/lib/python2.7/site-packages/matplotlib/font_manager.py:273: UserWarning: Matplotlib is building the font cache using fc-list. This may take a moment.
  warnings.warn('Matplotlib is building the font cache using fc-list. This may take a moment.')
  File "/home/travis/build/ReactionMechanismGenerator/code/tested/RMG-Py/rmg.py", line 164, in <module>
    rmg.execute(**kwargs)
  File "/home/travis/build/ReactionMechanismGenerator/code/tested/RMG-Py/rmgpy/rmg/main.py", line 653, in execute
    bimolecularReact=self.bimolecularReact)
  File "/home/travis/build/ReactionMechanismGenerator/code/tested/RMG-Py/rmgpy/rmg/model.py", line 785, in enlarge
    rxns = reactAll(self.core.species, numOldCoreSpecies, unimolecularReact, bimolecularReact)
  File "/home/travis/build/ReactionMechanismGenerator/code/tested/RMG-Py/rmgpy/rmg/react.py", line 155, in reactAll
    rxns = list(react(*spcTuples))
  File "/home/travis/build/ReactionMechanismGenerator/code/tested/RMG-Py/rmgpy/rmg/react.py", line 75, in react
    combos
  File "/home/travis/build/ReactionMechanismGenerator/code/tested/RMG-Py/rmgpy/scoop_framework/util.py", line 154, in map_
    return map(WorkerWrapper(args[0]), *args[1:], **kwargs)
  File "build/bdist.linux-x86_64/egg/scoop/fallbacks.py", line 49, in wrapper
  File "/home/travis/build/ReactionMechanismGenerator/code/tested/RMG-Py/rmgpy/scoop_framework/util.py", line 112, in __call__
    return self.myfn(*args, **kwargs)
  File "/home/travis/build/ReactionMechanismGenerator/code/tested/RMG-Py/rmgpy/rmg/react.py", line 96, in reactMolecules
    rxns = family.generateReactions(molecules)
  File "/home/travis/build/ReactionMechanismGenerator/code/tested/RMG-Py/rmgpy/data/kinetics/family.py", line 1308, in generateReactions
    reactionList.extend(self.__generateReactions(reactants, forward=True))
  File "/home/travis/build/ReactionMechanismGenerator/code/tested/RMG-Py/rmgpy/data/kinetics/family.py", line 1439, in __generateReactions
    productStructures = self.__generateProductStructures(reactantStructures, [mapA, mapB], forward)
  File "/home/travis/build/ReactionMechanismGenerator/code/tested/RMG-Py/rmgpy/data/kinetics/family.py", line 1200, in __generateProductStructures
    productStructures = self.applyRecipe(reactantStructures, forward=forward)
  File "/home/travis/build/ReactionMechanismGenerator/code/tested/RMG-Py/rmgpy/data/kinetics/family.py", line 1166, in applyRecipe
    struct.update()
  File "rmgpy/molecule/molecule.py", line 721, in rmgpy.molecule.molecule.Molecule.update (build/pyrex/rmgpy/molecule/molecule.c:14763)
  File "rmgpy/molecule/molecule.py", line 917, in rmgpy.molecule.molecule.Molecule.updateAtomTypes (build/pyrex/rmgpy/molecule/molecule.c:17910)
  File "rmgpy/molecule/atomtype.py", line 452, in rmgpy.molecule.atomtype.getAtomType (build/pyrex/rmgpy/molecule/atomtype.c:6373)
  File "rmgpy/molecule/atomtype.py", line 483, in rmgpy.molecule.atomtype.getAtomType (build/pyrex/rmgpy/molecule/atomtype.c:6249)
rmgpy.molecule.atomtype.AtomTypeError: Unable to determine atom type for atom O-, which has 0 double bonds to C, 0 double bonds to O, 0 double bonds to S, 0 triple bonds, and 0 benzene bonds.
```

rmgpy/molecule/group.py
```diff
@@ -870,3 +918,160 @@ def isIdentical(self, other):
...
+
+            groupFeatures = getFeatures(atom, atom.bonds)
+            # print groupFeatures
+
+            single = groupFeatures[0]
+            allDouble = groupFeatures[1]
+            triple = groupFeatures[5]
+            benzene = groupFeatures[6]
+
+            bondValency = single + 2 * allDouble + 3 * triple + 4.0/3.0 * benzene
```
mliu49: I'm unsure about the multiplier for benzene bonds. Should it be 3/2 or 4/3, or does it matter in this case?

rmgpy/molecule/group.py
```diff
@@ -433,6 +453,34 @@ def copy(self):
         """
         return GroupBond(self.vertex1, self.vertex2, self.order[:])
 
+    def isSingle(self):
+        """
+        Return ``True`` if the bond represents a single bond or ``False`` if
+        not.
+        """
+        return self.order[0] == 'S' and len(self.order) == 1
```
mliu49: So these methods only check if the first bond option is of the specified type? What about cases with a list of possible types?

rmgpy/molecule/atomtype.py
```diff
@@ -91,6 +111,14 @@ def __init__(self, label='', generic=None, specific=None):
...
         self.incrementLonePair = []
         self.decrementLonePair = []
+        self.single = [] if single is None else single
+        self.allDouble = [] if allDouble is None else allDouble
+        self.rDouble = [] if rDouble is None else rDouble
+        self.oDouble = [] if oDouble is None else oDouble
+        self.sDouble = [] if sDouble is None else sDouble
+        self.triple = [] if triple is None else triple
+        self.benzene = [] if benzene is None else benzene
+        self.lonePairs = [] if lonePairs is None else lonePairs
```
mliu49: I think these can all be simplified to `single or []`, etc. similar to lines 104 and 105.

mliu49:
I have a few additional questions and comments.

Questions:

- Why are there fewer single bond options for Ss compared to Os?
- It seems that you added [0,1,2] specifications for oDouble and sDouble in Cdd and Sidd later, was there an issue with leaving them as []?
- If there was an issue, does N5dd need to have the same added specifications?

Some minor comments:

- Use ` something not in list` instead of `not something in list` for readability, although they're identical, same thing with using `a is not b` instead of `not a is b`:
  - atomtype.py: Line 478
  - group.py: Lines 398, 408 - could potentially be replaced by `return all(checklist)`
  - group.py: Lines 950, 1000
- Remove commented `print` statements:
  - group.py: Lines 969, 978
- Should atom types be compared using `==` or `is`? Both are used (group.py, line 966 vs. 1019).
- group.py, line 1077 - could potentially be replaced by `return any(checklist)`
- group.py, line 945 - spacing (sorry, it was bothering me)

rmgpy/molecule/group.py
nyee: I'll add a comment a clarifying comment, but I think that is the behavior we want.

rmgpy/molecule/group.py
nyee: I'm unsure as well, but I have rewritten if benzene==3, use 4/3 otherwise use 3/2. I think this is the best compromise. 

I thought about making a new bond order 'FB' for fused benzene bond, but even that is not satisfactory because its valency is still unclear. In naphthalene, you might say the two B bonds are 3/2 and FB is 1. However in the case of the central atom in phenalene would you say it has 3 FB bonds with 4/3 each or 2 B bonds and 1 FB bond?

nyee:
@mliu49 I implemented all the minor comments you mentioned with the last few commits. Regarding the questions about the required bonding for different atomtypes, I am running into problems on another feature from those poor decisions. I'm going to debug before coming up with a final answer. 

Regarding the '==' vs 'is', it should be 'is' as AtomType's '== 'is defined like:

    def equivalent(self, other):
        """
        Returns ``True`` if two atom types `atomType1` and `atomType2` are
        equivalent or ``False``  otherwise. This function respects wildcards,
        e.g. ``R!H`` is equivalent to ``C``.
        """
        return self is other or self in other.specific or other in self.specific

Each AtomType is only instatiated once, (only one atomTypes['C'] in the entire RMG run). I could run into trouble if I use deep copys though because that 'is' will fail

mliu49:
I think it looks good, pending what you find out about the required bond specifications.

On the `==` topic, I don't think that the `equivalent` method you referenced is a special method. The method for `==` should be `__eq__`. Since you say that each AtomType object is only created once, then `is` should be fine.

nyee:
@mliu49 I just reverted the atomtypes to what they were before this branch, which seems to work well enough. I have also rebased most of the recent commits to clean up. If this passes the travis, I will merge in unless you have any additional comments.

mliu49:
Ok, I think you can merge it.

Issue #703, opened 2016-07-04
Non-identical mechanisms for eg5 between v1.0.4 and current master


nickvandewiele:
RMG-tests has recently been generating mechanisms for eg5 that have diverged from the previous benchmark (v.1.0.4 of RMG).

see this [build](https://travis-ci.org/ReactionMechanismGenerator/RMG-tests/builds/142251189)

```
core for eg5:
Test model has 14 species.
Original model has 13 species.
Test model has 14 reactions.
Original model has 8 reactions.
The tested model has 1 species that the original model does not have.
spc: [CH2]CCCCC(9)
The tested model has 6 reactions that the original model does not have.
rxn: [CH2]CC(3) + [CH2]CCC(2) <=> [CH2]CCCC(4) + C[CH2](5)      origin: PDepNetwork #11
rxn: [CH2]CCCCC(9) + [CH3](10) <=> n-heptane(1)     origin: PDepNetwork #18
rxn: [CH2]CCCCC(9) + [CH3](10) <=> [CH2]CC(3) + [CH2]CCC(2)     origin: PDepNetwork #18
rxn: [CH2]CCCCC(9) + [CH3](10) <=> [CH2]CCCC(4) + C[CH2](5)     origin: PDepNetwork #18
rxn: C=C(16) + [CH2]CCC(2) <=> [CH2]CCCCC(9)        origin: PDepNetwork #23
rxn: [CH3](10) + [CH3](10) <=> C[CH2](5) + [H](7)       origin: PDepNetwork #16
edge for eg5:
Test model has 55 species.
Original model has 47 species.
Test model has 299 reactions.
Original model has 168 reactions.
The tested model has 8 species that the original model does not have.
spc: C=CCCCC(39)
spc: [CH2]CC[CH]CC(40)
spc: [CH2]C[CH]CCC(41)
spc: [CH2]CCC[CH]C(42)
spc: [CH2][CH]CCCC(43)
spc: [CH2]CCCC[CH2](44)
spc: C[CH]CCCC(45)
spc: CC[CH]CCC(46)
The tested model has 131 reactions that the original model does not have.
...
```

Issue #677, opened 2016-06-13
Non-identical thermo for [C-]#[N+] between v1.0.4 and current master


nickvandewiele:
RMG-tests has recently been generating mechanisms for methylamine that have diverged from the previous benchmark (v.1.0.4 of RMG). 

I believe that a different thermo calculation 
See this [build](https://travis-ci.org/ReactionMechanismGenerator/RMG-tests/builds/136575046)

```
Non-identical thermo!
tested: [C-]#[N+]
original:   [C-]#[N+]
Hf(300K)  |S(300K)   |Cp(300K)  |Cp(400K)  |Cp(500K)  |Cp(600K)  |Cp(800K)  |Cp(1000K) |Cp(1500K) 
     95.89|     51.61|      8.62|      8.68|      8.61|      8.50|      8.34|      8.33|      8.35
     76.55|      8.98|      2.40|      2.02|      1.88|      1.79|      1.43|      1.12|      1.74
thermo: Thermo library: thermo_DFT_CCSDTF12_BAC + radical(Cs_P)
thermo: Thermo group additivity estimation: group(N) + other(R) + group(Ct-CtCs) + other(R) + radical(Cs_P)

```

is at the basis of the much larger discrepancies in terms of kinetics:

```
Non-identical kinetics!
tested:
rxn: [C-]#[N+](926) + CH3CHO(33) <=> CH2CHO(32) + HNC(76)       family: H_Abstraction
benchm:
rxn: CH2CHO(32) + HNC(76) <=> CH3CHO(33) + [C-]#[N+](926)       family: H_Abstraction
k(1bar)|300K   |400K   |500K   |600K   |800K   |1000K  |1500K  |2000K  
k(T):  |  -2.06|   0.03|   1.34|   2.25|   3.46|   4.24|   5.40|   6.07
k(T):  |   1.01|   2.13|   2.84|   3.34|   4.02|   4.46|   5.14|   5.55
Kinetics: Arrhenius(A=(48170,'cm^3/(mol*s)'), n=2.56, Ea=(9.731,'kcal/mol'), T0=(1,'K'), comment="""Estimated using template (X_H;Y_rad) for rate rule (C/H3/CO;Y_rad)
[...]

PR #707, opened 2016-07-07
Broadcast quantumMechanics 

nickvandewiele:
This PR changes the way the quantumMechanics object is passed around throughout RMG.

In the past, the object was created upon reading the RMG input file, and passed down as a parameter throughout multiple functions.

This PR removes the QM parameter from these functions, and instead fetches it from the global RMG object. In addition, the QM object is broadcast to all available workers: when thermochemistry is calculated on a remote worker (not yet implemented, WIP), this PR allows the retrieval of the QM broadcasted object when it is needed.

_Implementation details:_

**Broadcast:**
the quantumMechanics attribute of the global RMG object is a `QMCalculator` instance. Upon reading the RMG input file, the `rmg.quantumMechanics` attribute is broadcast under the key "quantumMechanics" when it is available.

**Retrieval:**
in the method `rmgpy.data.thermo.ThermoDatabase.getThermoData`, it is verified whether then quantumMechanics object exists. If it exists, it is used in the same hierarchical way as before.

### DIFF
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -47,6 +47,7 @@ from rmgpy.molecule import Molecule, Atom, Bond, Group
 import rmgpy.molecule
 from rmgpy.species import Species
 
+from rmgpy.scoop_framework.util import get
 
 #: This dictionary is used to add multiplicity to species label
 _multiplicity_labels = {1:'S',2:'D',3:'T',4:'Q',5:'V',}
@@ -763,7 +764,7 @@ class ThermoDatabase(object):
         )
 
 
-    def getThermoData(self, species, trainingSet=None, quantumMechanics=None):
+    def getThermoData(self, species, trainingSet=None):
         """
         Return the thermodynamic parameters for a given :class:`Species`
         object `species`. This function first searches the loaded libraries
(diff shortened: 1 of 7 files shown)


rwest:
@pierrelb and @nathandavid13, please have a look at this - I imagine it'll affect the QMTS things not yet on master.

connie:
Some comments:

- Duplicated calls of `quantumMechanics.initialize()`.  It is called in `rmg.input.readInputFile` and `rmg.input.readThermoInputFile` but also called in `rmg.main.initialize`
- Loading of the database has been changed in `thermoEstimator.py` (I guess you are trying to save time by not loading any of the kinetics or transport) but is not going to inherit the user-specified `thermoLibraries`.  Previously, `RMG.loadDatabase()` would load the user preferences from the RMG object (i.e. `RMG.thermoLibraries`), but loading using the path alone won't.
- 1,3-hexadiene example now fails due to `QMSettings.checkAllSet` checking if `quantumMechanics.fileStore` is specified.  In the thermoEstimator, not having this attribute in the input file is fine because of line 409: 
        `rmg.quantumMechanics.setDefaultOutputDirectory(rmg.outputDirectory)`

nickvandewiele:
Thanks for the review.

I will fix these bugs.

The problem with the previous `RMG.loadDatabase()` in `thermoEstimator.py` is that it wanted to calculate the QMTP thermo for all of the species in the kinetics libraries, wasting a lot of time.

nickvandewiele:
@connie : the reported bugs have been fixed, and the commit is rebased. 

RMG-tests shows no _new_ divergence issues (besides the ones reported here: #703, and #677), while `make eg2` generates a different bug ( #709 ) which is unrelated to this PR.

connie:
Ok I've checked that the bugs are now resolved.  Thanks.

PR #501, opened 2015-11-24
Fix bug in flux diagram script

nickvandewiele:
addresses https://github.com/ReactionMechanismGenerator/RMG-Py/issues/479

Issue #479, opened 2015-10-30
generateFluxDiagram fails with minimal example (on local machine)


KEHANG:
I used `minimal` example to generateFluxDiagram by running `python generateFluxDiagram.py testing/minimal/input.py`. It gave the error:
```
Conducting simulation of reaction system 1...
Generating flux diagram for reaction system 1...
Traceback (most recent call last):
  File "scripts/generateFluxDiagram.py", line 42, in <module>
    main()
  File "scripts/generateFluxDiagram.py", line 39, in main
    fluxdiagram.run(inputFile, useJava)
  File "/home/kehang/Code/edge_inchi_proj/RMG-Py/rmgpy/tools/fluxdiagram.py", line 517, in run
    os.path.join(rmg.outputDirectory, 'flux', '{0:d}'.format(index+1)), centralSpecies, speciesPath
  File "/home/kehang/Code/edge_inchi_proj/RMG-Py/rmgpy/tools/fluxdiagram.py", line 265, in generateFluxDiagram
    graph.write_dot(os.path.join(outputDirectory, 'flux_diagram_{0:04d}.dot'.format(frameNumber)))
  File "build/bdist.linux-x86_64/egg/pydot.py", line 1809, in <lambda>
  File "build/bdist.linux-x86_64/egg/pydot.py", line 1911, in write
  File "build/bdist.linux-x86_64/egg/pydot.py", line 2023, in create
pydot.InvocationException: Program terminated with status: 1. stderr follows: Warning: syntax ambiguity - badly delimited number '3.75015410405e' in line 8 of /tmp/tmpnMujyK splits into two tokens
Error: /tmp/tmpnMujyK: syntax error in line 8 near '-05'
```
The temporary file looks like this:
```
  1 digraph flux_diagram {
  2 label="t = 10^-6.5 s";
  3 graph [rankdir=LR, fontname=sans, fontsize=10, overlap=false, bb="0,0,788.26,243.41"];
  4 node [label="\N"];
  5 "ethane(1)" [penwidth=9.99989414208, width="0.9027", image="/home/kehang/Code/edge_inchi_proj/RMG-Py/testing/minimal/species/ethane(1).png", label=" ", pos="32.497,14    9.29", height="0.5"];
[...]

PR #711, opened 2016-07-12
Fix issues with flux diagram generation

### LINKED ISSUE #636: Flux diagram generation error
Hi, 

I have been trying to generate the flux diagram for a completed RMG run, both via online interface and the local script. However, it always ends up in an error typically as the one shown below (from online interface):

********************************************************************************************************************
IOError at /simulate/flux
[Errno 2] No such file or directory: '/var/www/RMG-website/rmgweb/media/rmg/tools/flux/1/flux_diagram_0049.png'

Request Method:
POST 

Request URL:
http://rmg.mit.edu/simulate/flux 

Django Version:
1.8.5 

Exception Type:
IOError 

Exception Value:
[Errno 2] No such file or directory: '/var/www/RMG-website/rmgweb/media/rmg/tools/flux/1/flux_diagram_0049.png'

Exception Location:
build/bdist.linux-x86_64/egg/pydot.py in write, line 1895 

Python Executable:
/usr/bin/python 

Python Version:
2.7.6 

Python Path:
['/var/www/RMG-website/rmgweb',
 '/var/www/RMG-Py',
 '/opt/anaconda/envs/rmg/lib/python2.7/site-packages/gprof2dot-2015.2.3-py2.7.egg',
 '/opt/anaconda/envs/rmg/lib/python2.7/site-packages/pydot2-1.0.32-py2.7.egg',
 '/opt/anaconda/envs/rmg/lib/python2.7/site-packages/setuptools-19.1.1-py2.7.egg',
 '/opt/anaconda/envs/rmg/lib/python2.7/site-packages',
 '/usr/local/lib/python2.7/dist-packages/PyDAS-1.0.0-py2.7-linux-x86_64.egg',
 '/usr/lib/python2.7',
 '/usr/lib/python2.7/plat-x86_64-linux-gnu',
 '/usr/lib/python2.7/lib-tk',
 '/usr/lib/python2.7/lib-old',
 '/usr/lib/python2.7/lib-dynload',
 '/usr/local/lib/python2.7/dist-packages',
 '/usr/lib/python2.7/dist-packages',
 '/usr/lib/python2.7/dist-packages/PILcompat',
 '/usr/lib/python2.7/dist-packages/gtk-2.0',
 '/usr/lib/python2.7/dist-packages/ubuntu-sso-client',
 '/var/www/RMG-website',
 '/var/www/RMG-website/rmgweb',
 '/var/www/RMG-database',
 '/var/www/RMG-Py']
 

Server time:
Tue, 19 Apr 2016 00:01:20 -0400 

********************************************************************************************************************

mliu49:
Addresses many bugs with flux diagram generation from both command line and website, and incorporates all changes from #501.

Additionally, pydot will need to be updated to 1.2.2, so merging this PR should wait until that is finalized.

This PR should fix ReactionMechanismGenerator/RMG-website#82. Mark's original issue in ReactionMechanismGenerator/RMG-website#82, ReactionMechanismGenerator/RMG-website#40, and #636 seem to all be related, but I don't know if this will fix those, as I haven't been able to reproduce that error. It might also have something to do with the website server itself, if folders are getting deleted somehow.

Also, I think #479 was fixed by 12c0db0d6d7f477326266f08436eb7a1b21ba780.


mliu49:
Travis might fail for the same reason it did previously in #501.

connie:
Can you add the pydot 1.2.2 dependencies to our anaconda channel and also change the `travis.yml` and `environment.yml` files as needed so that travis passes?  Will the dependency version change affect anything else that we draw?

mliu49:
I think that the issue might be that the subprocess call cannot find mencoder. [This post](http://happyproblems.blogspot.com/2013/02/python-subprocess-throws-errno-2-no.html) suggests that subprocess calls do not have access to the PATH variable, so the full path to mencoder should be specified. The alternative is to use shell=True, which directly uses the command in the shell, so it has access to PATH.

However, neither of those methods worked for Travis, so I'm pretty sure the issue in this case is that the Travis build does not have mencoder installed.

On the other hand, the dev website is currently experiencing the same error is being as Travis, but mencoder is installed. In this case, specifying the full path to mencoder (`/usr/bin/mencoder`) fixes the error.

Therefore, I think that `.travis.yml` will need to be changed to install mencoder, if we want to test flux diagram generation.

I'm not sure is specifying the full path to mencoder is a good idea though, since different users may have it installed in a different location. Using shell=True is also not generally recommended due to security issues. Is it possible to determine the location of mencoder within the code?

mliu49:
Yay! Installing mencoder worked!

The remaining issue is to decide whether to use the full path, shell=True, or another alternative. Any thoughts would be appreciated!

connie:
I would recommend against the shell=True if we think it's a security concern.  We should probably be able to figure out the relative path based on where we think mencoder should be installed.  Sort of like how we search for the MOPAC executable in the same way in `rmgpy.qm.mopac`

Is there any way to convert mencoder to a conda package instead?  Not every OS will have apt-get.

mliu49:
I switched from using mencoder to ffmpeg, which is a more actively developed open-source video encoder. It also has existing anaconda packages, which makes things much easier.

I think this PR is ready for merging if someone wants to review it.

rmgpy/tools/fluxtest.py
```diff
@@ -0,0 +1,24 @@
...
+import os
+import os.path
+import shutil
+
+from rmgpy.tools.fluxdiagram import *
+
+class FluxDiagramTest(unittest.TestCase):
+
+    def test_avi(self):
+        folder = os.path.join(os.getcwd(),'rmgpy/tools/data/flux/')
```
nickvandewiele: could you perhaps use 

```python
folder = os.path.join(os.path.dirname(rmgpy.__file__),'tools/data/flux')
```

rather than `os.getcwd()`

This way, you can run the test from other locations besides the RMG-Py root directory.

rwest: Good suggestion. I don't actually use Windows so can't test this, and it may make no difference, but I am in the habit of writing it like this (because Windows uses backslashes in paths instead of slashes?):
```python
folder = os.path.join(os.path.dirname(rmgpy.__file__),'tools','data','flux')
```

(if it does make a difference, may want to change lines 16 and 17 below)

mliu49: Good points. I made both changes.

Issue #712, opened 2016-07-13
Tree averaging does not work as expected


connie:
@KEHANG and I discovered this behavior yesterday.  We found that RMG does not create the proper averaged nodes with closest distance to the exact match, and can instead move further up the tree to an even more general set of nodes.  We found this in the R_Addition_MultipleBond family after adding some a new groups `Cds_Cds_benzene` and `Cds-CdH_Cds-CdR_benzene`: 

```
L1: R_R                                                           L1: Y_J
  L2: Cd_R                                                         L2: CJ
    L3: Cds_Cds                                                     L3: CsJ 
       L4: Cds_Cds_benzene                                           L4: CsJ-HHH
         L5: Cds-CdH_Cds-CdR_benzene                                 L4: CsJ-CsHH
```

He then added a training reaction which matches with `Cds-CdH_Cds-CdR_benzene; CsJ-HHH`, the reaction between benzene and methyl radical. Then we tried to see whether a reaction between benzene and ethyl radical, which is a sibling of methyl, will use the expected rate rule which adopts from the training reaction added.  We would expect RMG to find `Cds-CdH_Cds-CdR_benzene; CsJ` which uses the value from the training rule.  But what we actually get is:

![image](https://cloud.githubusercontent.com/assets/583115/16805855/13691c92-48e0-11e6-99ee-18f2643917d9.png)

```
! Average of (Cds-CdH_Cds-CdR_benzene;CsJ-HHH from training reaction 73 + 
! Average of (Cds-HH_Cds-HH;CsJ-HHH + Cds-HH_Cds-CsH;CsJ-HHH + Cds-HH_Cds-CsCs
! CsJ-HHH + Cds-HH_Cds-OsH;CsJ-HHH +
.
.
.
!  + Cds-CdCd_Cds-CdCd;CsJ-CdCdCs)))
! Estimated using template (Cds_Cds;CsJ) for rate rule (Cds-CdH_Cds-CdR_benzene;CsJ-CsHH)
! Multiplied by reaction path degeneracy 6
ethyl(2)+benzene(1)=C8H11(5)                        1.306e+04 2.433     7.412 
```

[...]

PR #713, opened 2016-07-14
Fix tree averaging issue #712

KEHANG:
This PR tries to fix the issue #712 by

- generating all combinations of cross-level rules
- when calculating the estimated rules, the averaging scheme is only choose distance-1 rules

### DIFF
--- a/rmgpy/data/base.py
+++ b/rmgpy/data/base.py
@@ -451,7 +451,7 @@ class Database:
 
     def __loadTree(self, tree):
         """
-        Parse an old-style RMG tree located at `tree`. An RMG tree is an n-ary
+        Parse an group tree located at `tree`. An RMG tree is an n-ary
         tree representing the hierarchy of items in the dictionary.
         """
 
@@ -498,7 +498,10 @@ class Database:
                 else:
                     entry.parent = None
                     self.top.append(entry)
-
+                    
+                # Save the level of the tree into the entry
+                entry.level = level
+                
                 # Add node to list of parents for subsequent iteration
                 parents.append(label)
 
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1039,15 +1039,15 @@ class KineticsFamily(Database):
         else:
             return self.groups.top
     
-    def fillKineticsRulesByAveragingUp(self, rootTemplate=None, alreadyDone=None):
+    def fillKineticsRulesByAveragingUp(self):
         """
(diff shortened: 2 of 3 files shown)


rwest:
No time to look into it further myself now, but I think this is being killed because it's run out of memory on travis, not because it's run out of time.

connie:
In addition to kehang's notes on this pull request, I have also updated:


- The storage of kinetics comments for averaged rules is streamlined.  Previous storage of fully recursive kinetics comments into the rule entries that are averaged during `fillRulesByAveragingUp` spent more than 5 GB memory during RMG runs.  Therefore, I have simplified it to listing the distance 1 children that contribute to kinetics only.  In order to find the actual sources of the original kinetics, one would need to look into the code further.  This means that verbose comments are no longer as verbose as before.  To turn back on the truly recursive comments, you can go into the code to uncomment a few lines.

- Euclidean distance algorithm implemented for the selection of the best kinetics to use in `estimateKinetics`.  Previously, we went with the Manhattan distance, which meant an entry that was a distance of (3,0) away was equally favorable as an entry of distance (2,1) away.  Now, switching to a euclidean distance norm we will choose the (2,1) entry over the (3,0) entry.  This is because we saw in the models that higher top level nodes were becoming increasingly incorporated.  I think we want the algorithm to prevent top level nodes from appearing as infrequently as possible.


image + image = image + image

C2H3(13) + C2H3(13)	⇔	C2H4(8) + C#C(25)	

Before:
```
! Template reaction: Disproportionation
! Flux pairs: C2H3(13), C2H4(8); C2H3(13), C#C(25); 
! Estimated using average of templates (Y_rad_birad_trirad_quadrad;Cd_Cdrad) + (Y_rad;CH_d_Rrad) + (Cd_pri_rad;XH_Rrad) for rate rule
! (Cd_pri_rad;Cd_Cdrad)
! Multiplied by reaction path degeneracy 2
C2H3(13)+C2H3(13)=C2H4(8)+C#C(25)                   6.005e+11 0.167     1.127    
```

After:

```
! Template reaction: Disproportionation
! Flux pairs: C2H3(13), C2H4(8); C2H3(13), C#C(25); 
! Estimated using template (Y_rad;CH_d_Rrad) for rate rule (Cd_pri_rad;Cd_Cdrad)
! Multiplied by reaction path degeneracy 2
C2H3(13)+C2H3(13)=C#C(25)+C2H4(8)                   6.447e+06 1.902     -1.131   
```

PR #714, opened 2016-07-14
Fix some issues with rmgpy.qm.qmdata

nickvandewiele:
This PR fixes issues with some of the qm classes:

- limit the no. of threads used by MOPAC calculations to 1 (to avoid CPU contention when concurrent thermo will be implemented)
- remove mutable argumets in QMData constructor
- remove CCLibData class and replace it by a module function. It's easier, and more pythonic. The class was not adding anything anyway.

### DIFF
--- a/rmgpy/qm/qmdata.py
+++ b/rmgpy/qm/qmdata.py
@@ -23,13 +23,9 @@ class QMData(object):
         self.groundStateDegeneracy = groundStateDegeneracy
         self.numberOfAtoms = numberOfAtoms #: Number of atoms.
         self.stericEnergy = Energy(stericEnergy)
-        """
-        Steric energy.
-        """
         self.molecularMass = Mass(molecularMass)
         self.energy = Energy(energy)
         self.atomicNumbers = atomicNumbers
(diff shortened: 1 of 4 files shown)


rwest:
@nateharms please could you look at this? a good way to get familiar with this bit of code...

rmgpy/qm/qmdata.py
```diff
@@ -58,44 +54,36 @@ def __repr__(self):
...
         return 'QMData({0!s})'.format(string)
         
-class CCLibData(QMData):
+def parseCCLibData(cclibData, groundStateDegeneracy):
     """
-    QM Data extracted from a cclib data object
-    
-    This data objects collects information that CCLib was able to
-    retrieve from a quantum chemistry output file.
+    Parses a CCLib data object and returns CCLibData object.
```
rwest: It returns a QMData object

rwest:
Looks fine to me (apart from the typo in the docstrings commented above).

nickvandewiele:
thanks! Docstring is fixed.

PR #716, opened 2016-07-18
Asynchronous computing of thermo data

nickvandewiele:
This PR adds asynchronous (concurrent) computing of thermo data, and primarily aims at reducing the wall clock time of QMTP calculations in RMG.

The biggest change of this PR is the way thermo is computed and stored. Conceptually, there is now a difference between _submitting_ a request to compute thermo, and _retrieving_ the thermo when needed.

### Thermo calculation submission
Thermo calculation submission for a `Species` object is done as follows:
`rmgpy.thermo.thermoengine.submit(species)`

This function call will create a task to compute the thermo and send it to a worker, which computes the thermo while the main code is continuing to execute.

### Thermo calculation retrieval
Thermo retrieval for a `Species` object is done as follows:
`species.getThermoData()`

This method call encapsulates a number of different scenarios: if a previous call was made to submit a thermo calculation task, the `species.getThermoData()` will check if the thermo is already available. If it is not yet available, it will wait for it. If it _is_ available, it will return the thermo right away. If `species.getThermoData()` finds out there was not yet a thermo calculation task created, it will create one and wait for its completion.

### Code changes

In the context of RMG runs, a thermo calculation task is submitted whenever a new species is created, i.e. in the `CERM.makeNewSpecies()` method.

In terms of code, a new module `rmgpy.thermo.thermoengine` moves the thermo generation methods previously stored in `rmgpy.model.Species` to module functions. Parameters such as db, and qm are no longer passed down from one method to another, but fetched from the globally accessible or broadcasted variables.

### Missing pieces

**1**
The previous design that allowed to retrieve the thermo by accessing the instance attribute `spc.thermo` is bound to fail in many scenarios.

**2**
The last two commits are temporary in nature as I currently do not see yet how the `ThermoLibrary.loadEntry` functionality can function correctly. The bug in this part of RMG has nothing to do with this PR, though.

**3**
RMG-website calls


nickvandewiele:
@pierrelb : care to review as this must be up your alley?

nickvandewiele:
@mliu49 
In addition, I have completely ignored any calls made by RMG-website, so that will pose some problems as well. 

But it may be an opportunity to revisit some code in RMG-website and migrate it back to RMG-Py , or replace it by RMG-Py code altogether.

mliu49:
As far as I can tell, these changes don't affect RMG-website. Thermo generation is only done through calling `ThermoDatabase.getThermoData()`, which was was not changed much.

rmgpy/thermo/thermoengine.py
```diff
@@ -0,0 +1,145 @@
...
+            if "Thermo library" in thermo0.comment and isinstance(thermo0,NASA):
+                thermo=thermo0
+                if thermo.E0 is None:
+                    thermo.E0 = wilhoit.E0
+            else:
+                thermo = wilhoit.toNASA(Tmin=100.0, Tmax=5000.0, Tint=1000.0)
+    else:
+        raise Exception('thermoClass neither NASA nor Wilhoit.  Cannot process thermo data.')
+    
+    if thermo.__class__ != thermo0.__class__:
```
connie: Currently this is just an extra computation since we now longer print RMS error

PR #720, opened 2016-07-19
Allow the profiling of RMG Jobs that do not converge

rgillis8:
Allows a value wallTime to be read in on the input file. After wallTime lapses the RMG job terminates and profiling data is created.

### DIFF
--- a/examples/rmg/commented/input.py
+++ b/examples/rmg/commented/input.py
@@ -165,6 +165,9 @@ options(
 	#gets RMG to generate edge species chemkin files. Uses lots of memory in output.
 	#Helpful for seeing why some reaction are not appearing in core model.  
     saveEdgeSpecies=False,
+    #Sets a time limit in the form DD:HH:MM:SS after which the RMG job will stop. Useful for profiling on jobs that
+    #do not converge.
+    #wallTime = '00:00:00',
 )
 
 # optional module allows for correction to unimolecular reaction rates at low pressures and/or temperatures.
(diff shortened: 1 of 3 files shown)


nickvandewiele:
so when you specify 00:00:00 should it be interpreted as DD:HH:MM or HH:MM:SS ?

rgillis8:
HH:MM:SS although if the user wrote DD:HH:MM:SS that would also work

nickvandewiele:
Perhaps clarify in the documentation or input file what 00:00:00 means as it could be misinterpreted.

rgillis8:
Sure. It is described in the "commented" input file, but I can also clarify in the attribute description.

rgillis8:
What documentation would I clarify this in?

nickvandewiele:
ok, I missed that. Ignore my comment then.

rgillis8:
Well now that you mentioned this I am seeing that in the RMG object description (in main.py) the wallTime description that was already there doesn't quite match what I've done.

rgillis8:
it says it is a time in seconds when really what was already partially implemented and what I fixed was this HH:MM:SS format which I think is better anyways. I should probably change it? What do you think.

nickvandewiele:
1) I don't see many people wanting to specify a wall clock with a precision up until seconds.
2) I would go for a single supported format DD:HH:MM and not support DD:HH:MM:SS for now.

rgillis8:
Hmm, I think that having seconds precision is important when you are trying to stop at certain steps, especially in short < 5 min jobs. If you were just caring about stopping on long jobs I agree that seconds is pretty silly for a 12 hr job.

rgillis8:
This is ready I think.

Issue #718, opened 2016-07-19
Travis CI builds fail with cantera.ck2cti.InputParseError


nickvandewiele:
Currently, Travis CI builds fail with the following error.

```python
Traceback (most recent call last):
  File "/home/travis/build/ReactionMechanismGenerator/code/tested/RMG-Py/rmgpy/tools/regression.py", line 192, in <module>
    main()
  File "/home/travis/build/ReactionMechanismGenerator/code/tested/RMG-Py/rmgpy/tools/regression.py", line 189, in main
    run(benchmark, tested, *args)
  File "/home/travis/build/ReactionMechanismGenerator/code/tested/RMG-Py/rmgpy/tools/regression.py", line 151, in run
    observables = {'species': observables}
  File "/home/travis/miniconda/envs/benchmark/lib/python2.7/site-packages/rmgpy/tools/observablesRegression.py", line 105, in __init__
    self.newSim.loadChemkinModel(os.path.join(newDir,'chem_annotated.inp'), transportFile=newTransportPath, quiet=True)
  File "/home/travis/miniconda/envs/benchmark/lib/python2.7/site-packages/rmgpy/tools/canteraModel.py", line 262, in loadChemkinModel
    parser.convertMech(chemkinFile, transportFile=transportFile, outName=outName, **kwargs)
  File "/home/travis/miniconda/envs/benchmark/lib/python2.7/site-packages/cantera/ck2cti.py", line 1964, in convertMech
    raise InputParseError("No transport data for species '{0}'.".format(s))
cantera.ck2cti.InputParseError: No transport data for species 'Ar'.
```

connie:
On master right now, after running minimal example the `tran.dat` is completely incorrect:

```
! Species         Shape    LJ-depth  LJ-diam   DiplMom   Polzblty  RotRelaxNum Data     
! Name            Index    epsilon/k_B sigma     mu        alpha     Zrot      Source   
! Ar                  None
! He                  None
! Ne                  None
! N2                  None
! ethane(1)           None
! CH3(2)              None
! C2H5(3)             None
! C(6)                None
! H(4)                None
! C2H4(8)             None
! H2(12)              None
! C3H7(14)            None
! C2H3(13)            None
! C#C(25)             None
! C4H7(28)            None
! C4H7(38)            None
! C3H5(32)            None
! C4H6(30)            None
! C4H7(42)            None
```

Works fine on the code before your PR thermo_engine.

[...]

PR #721, opened 2016-07-19
Fix bug that forgets generating transport data when needed.

nickvandewiele:
This PR fixes a bug (cf. #718 ) that results in empty transport data files (`tran.dat`) after the recent merge of the `thermo_engine` branch. Transport data is also needed for p-dep simulations, so they would have crashed as well.

This PR replaces access of `spc.transportData` by method calls `spc.getTransportData()`. The latter checks if the transport data is available, and if not calculates the transport data on the spot.

In terms of code, I moved `generateTransportData` from `rmgpy.rmg.model.Species` to `rmgpy.species.Species`, where it really belongs. This way, `rmgpy.rmg.model.Species` becomes even more empty; only methods for stat. mech and solvent are still part of that class.

I added a unit test for the `getTransportData()` method.

### DIFF
--- a/rmgpy/chemkin.py
+++ b/rmgpy/chemkin.py
@@ -1744,7 +1744,7 @@ def saveTransportFile(path, species):
         f.write("! {0:15} {1:8} {2:9} {3:9} {4:9} {5:9} {6:9} {7:9}\n".format('Species','Shape', 'LJ-depth', 'LJ-diam', 'DiplMom', 'Polzblty', 'RotRelaxNum','Data'))
         f.write("! {0:15} {1:8} {2:9} {3:9} {4:9} {5:9} {6:9} {7:9}\n".format('Name','Index', 'epsilon/k_B', 'sigma', 'mu', 'alpha', 'Zrot','Source'))
         for spec in species:
-            transportData = spec.transportData
+            transportData = spec.getTransportData()
             if (not transportData):
                 missingData = True
             else:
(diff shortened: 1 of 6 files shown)


nickvandewiele:
[this Travis CI build](https://travis-ci.org/ReactionMechanismGenerator/RMG-tests/builds/145917133) shows that this version does not lead to #718 anymore.

connie:
The code looks fine to me.  But, I think we should change the names a bit for the functions in the `species.Species` class.   `getData` is a bit too generate, since we now have `generateTransportData` and `getTransportData` in the same class.

I think if we are going ahead w the transport data changes, we might as well move  `generateStatMech` and `generateEnergyTransferModel` to the regular `rmg.species.Species` class for consistency, and call the statmech function the same way as we do for transport and thermo.

This is unrelated, but the solvent properties really need some cleanup...  Should we consider moving all that data to the `species.Species` class and getting rid of the `rmg.model.Species` class?  @yunsiechung Or perhaps consolidating that stuff into a new consolidated module like `rmgpy.solvent`?

nickvandewiele:
Thanks for the review, @connie 

- I agree, `getData` could be renamed in a better way. I will change that. Its name was chosen, because the original design for concurrent thermo, also included the calculation of transport.

- I will move generateStatMech also to the `species.Species` class.

- I agree we could/should revisit the architecture of the `solvent` functionality. Preferably, I'd like to get rid of these class attribute of `Species` altogether, rather than moving them as you propose.

connie:
I think it would be nice if the `getData` function could be simply merged into `getThermoData`, that way only `getThermoData` and `generateThermoData` functions exists, paralleling the other species properties.  would this work?  

```
if not self.thermo:
    submit(self)
else:
    if isinstance(self.thermo, (NASA, Wilhoit, ThermoData)):
        return self.thermo
    else: 
        return self.thermo.result()
```

nickvandewiele:
I can consolidate both methods into one single, but there will be some code duplication: after you call `submit`, you will still need to go through the other steps of checking the type of thermo since a _serial_ run will have a different behavior than a _parallel_ run.

yunsiechung:
@connie @nickvandewiele 
Yeah, I agree with you guys. I think it would be nice if we could get rid of `rmg.model.Species` class. 

I am considering moving `rmg.model.Species` class attributes such as `solventName` and `solventData` to `species.Species` class attributes. Also, I want to make a module such as `rmgpy.thermo.thermosolvation` that contains all the solvation thermo related parts. I have some thought and maybe we can discuss about it.

PR #663, opened 2016-05-23
Can't run scoop unit tests on my Mac

rwest:
This is more an issue than a pull request, but at least there's now a place for someone to fix it!

When I run `make test` on my mac with an anaconda python stack, it gets as far as the (skipped) test

```
Test that we can broadcast a simple string. ... SKIP: test currently only runs on linux
```

which is one of the [scoop](http://pyscoop.org/) tests, and then it freezes, presumably on the next scoop test.
When I press `control-C` to abort, it then prints:

```
^C
======================================================================
ERROR: test suite for <class 'rmgpy.rmg.parreactTest.ParallelReactTest'>
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/rwest/anaconda/envs/rmg5/lib/python2.7/site-packages/nose/suite.py", line 209, in run
    self.setUp()
  File "/Users/rwest/anaconda/envs/rmg5/lib/python2.7/site-packages/nose/suite.py", line 292, in setUp
    self.setupContext(ancestor)
  File "/Users/rwest/anaconda/envs/rmg5/lib/python2.7/site-packages/nose/suite.py", line 315, in setupContext
    try_run(context, names)
  File "/Users/rwest/anaconda/envs/rmg5/lib/python2.7/site-packages/nose/util.py", line 471, in try_run
    return func()
  File "/Users/rwest/Code/RMG-Py/rmgpy/scoop_framework/framework.py", line 99, in setUpClass
    raise Exception('Could not start server!')
Exception: Could not start server!
```

Followed by all the coverage reporting, which has this stack trace buried in the middle of it (presumably printed from a different thread):

```
Traceback (most recent call last):
  File "/Users/rwest/anaconda/envs/rmg5/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/Users/rwest/anaconda/envs/rmg5/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/Users/rwest/anaconda/envs/rmg5/lib/python2.7/site-packages/scoop-0.7.2.0-py2.7.egg/scoop/broker/__main__.py", line 114, in <module>
  File "/Users/rwest/anaconda/envs/rmg5/lib/python2.7/site-packages/scoop-0.7.2.0-py2.7.egg/scoop/broker/brokerzmq.py", line 184, in run
[...]

PR #722, opened 2016-07-20
fix getThermoData method of species.Species

nickvandewiele:
this PR fixes a bug introduced in d227a597e9424fd96797d337cef3b2077784f156 after merging two methods into one.

The bug only occurs in parallel runs, and was caught in the RMG-tests Travis CI builds, but not in the unit tests of this repo. It only emphasizes the importance of having unit tests with scoop enabled also working for this repo, cf. #663 (WIP). If not, the parallel functionality will be broken in no time...

### DIFF
--- a/rmgpy/species.py
+++ b/rmgpy/species.py
@@ -461,16 +461,12 @@ class Species(object):
         from rmgpy.thermo.thermoengine import submit
         
         if self.thermo:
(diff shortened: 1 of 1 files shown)


nickvandewiele:
RMG-tests [passes](https://travis-ci.org/ReactionMechanismGenerator/RMG-tests/builds/146234840).

PR #727, opened 2016-07-22
Thermo group validation

nyee:
Adds script for comparing group additivity estimate with libraries and between commits.


rmgpy/tools/thermo_group_validation.py
```diff
@@ -0,0 +1,237 @@
...
+    #Save other variables into inputDict
+    if isinstance(commitStrs, str): commitStrs = [commitStrs]
+    if isinstance(commitNames, str): commitNames = [commitNames]
+    inputDict['commitStrs'] = commitStrs
+    inputDict['commitNames'] = commitNames
+    inputDict['variable'] = variable
+    inputDict['temperature'] = float(temperature)
+
+
+def loadInput(path, speciesThermo0):
```
KEHANG: speciesThermo0 is never used in the method, a redundant argument?

rmgpy/tools/thermo_group_validation.py
```diff
@@ -0,0 +1,237 @@
...
+    if commitStr == 'HEAD': yield
+    else:
+        try:
+            #Check that the commit str is not a file, so that you don't accidently check out unsaved changes
+            if os.path.exists(commitStr):
+                raise Exception("Argument commitStr must be a sha1, not a file")
+            else:
+                subprocess.call(['git', 'checkout', commitStr])
+                actualCommit = subprocess.check_output(['git', 'log','--format=%H%n%cd', '-1'], cwd=gitDirectory).splitlines()[0]
+                assert commitStr in actualCommit, "Checkout to another commit failed, please make sure your changes are stashed or committed."
```
KEHANG: if you have some uncommitted changes on top of HEAD, maybe you should also give assertion warning?

rmgpy/tools/thermo_group_validation.py
```diff
@@ -0,0 +1,237 @@
...
+    csvList = [[name]+data for name, data in parityData.iteritems()]
+
+    #Put into kcal/cal
+    conversionCsvList=[]
+    for point in csvList:
+        newPoint=[point[0]]
+        for value in point[1:]:
+            if variable == 'H' or variable == 'G':
+                newPoint.append(value/1000.0/4.182)
+            else:
```
KEHANG: also explicitly list out 'Cp' and 'S' instead of `else:`?

rmgpy/tools/thermo_group_validation.py
```diff
@@ -0,0 +1,237 @@
...
+    prunedCsvList=[]
+    #don't include header in pruned csv
+    for item in csvList:
+        if not std(item[2:]) == 0:
+            prunedCsvList.append(item)
+
+    xvar = [x+1 for x in range(len(prunedCsvList))]
+    for index, x in enumerate(range(len(prunedCsvList[0])-2)):
+        yvar = [abs(x[1]-x[index+2]) for x in prunedCsvList]
+        plt.semilogy(xvar, yvar, 's', label = inputDict['commitNames'][index])
```
KEHANG: so any reason for only plotting the prunedCsvList?

rmgpy/tools/thermo_group_validation.py
```diff
@@ -0,0 +1,237 @@
...
+    with open(os.path.join(os.getcwd(), heading+'.csv'), 'wb') as output:
+        spamwriter=csv.writer(output)
+        #add Header
+        spamwriter.writerow(['SMILES', 'Library Value at '+str(T0)+'K']+ inputDict['commitNames'])
+        for line in csvList:
+            spamwriter.writerow(line)
+
+    #Remove points where group estimates are the same for plotting
+    prunedCsvList=[]
+    #don't include header in pruned csv
```
KEHANG: would be nicer if can add more detailed description for data structure of `prunedCsvList ` and `csvList` since they're non-homogeneous list (containing string, float, etc). E.G., comments like each item in those lists is a list, in that list first item is name, second item is library value, third item is the estimated value at commit1, ...

rmgpy/tools/thermo_group_validation.py
nyee: fixed with 145df84

rmgpy/tools/thermo_group_validation.py
nyee: fixed with 9f9386c

rmgpy/tools/thermo_group_validation.py
nyee: Fixed with 58c8011

rmgpy/tools/thermo_group_validation.py
nyee: Yeah it probably depends on what the user wants. I was thinking the main purpose is to compare between commits, so the prunedCsvList ignores entries where the group additivity has not changed. 

It could be argued that we just want to generally visualize how good our group additivity works. In that case, we would not want to use the prunedCsvList. Earlier I found that graphs get pretty busy when putting in multiple libraries though.

rmgpy/tools/thermo_group_validation.py
nyee: Fixed with 07f409ab

rmgpy/tools/thermo_group_validation.py
```diff
@@ -0,0 +1,248 @@
...
+
+    #change to new directory for RMG-database
+    with cd(settings['database.directory']):
+        for commitStr in inputDict['commitStrs']:
+            with previousCommit(commitStr, settings['database.directory']):
+                newGroups= ThermoDatabase()
+                newGroups.load(os.path.join(settings['database.directory'], 'thermo'))
+                for smiles in speciesThermo:
+                    newSpecies = Species().fromSMILES(smiles)
+                    speciesThermo[smiles].append(newGroups.getThermoDataFromGroups(newSpecies))
```
mliu49: Should resonance isomers be generated before getting thermo data?

nyee:
Will be moved to RMG-tests

PR #739, opened 2016-08-04
Support for isotope information in chemkin files

nickvandewiele:
This PR adds support for isotope information in chemkin files.

Specifically:

- a new attribute `chemkinName` in the Element class ensures that chemkin-compatible element symbols are written. Chemkin only allows up names of to two characters long, so alternative had to be provided for longer, more straightforward names like "C-13".
- Writing the elements section of the chemkin file is refactored in its own function. However, rather than searching for all unique elements (and isotopes) among the species of the model, the elements are still hard-coded, for efficiency reasons. 
- the utility function `retrieveElementCount` of the `rmgpy.molecule.util` module now supports isotopes as well. That is, non-normal isotopes like C-13 will not be counted as a normal carbon atom, but as a separate key in the returned dictionary.
- the `writeThermoEntry ` function of the `rmgpy.chemkin` module now correctly writes the elements of the compound. It does this by removing the hard-coded heuristic, by a more thorough search via `retrieveElementCount`.

### DIFF
--- a/rmgpy/chemkin.py
+++ b/rmgpy/chemkin.py
@@ -52,6 +52,7 @@ from rmgpy.data.kinetics.library import LibraryReaction
 from rmgpy.data.kinetics.family import TemplateReaction
 from rmgpy.rmg.pdep import PDepNetwork
 from rmgpy.molecule import Molecule
+from rmgpy.molecule.util import retrieveElementCount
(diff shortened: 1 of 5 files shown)


nickvandewiele:
This PR is ready to be merged. Could someone review it?

Issue #736, opened 2016-08-03
Some PDep reactions are generated as irreversible


alongd:
Some PDep reactions in many RMG models are irreversible (See e.g., issue #731). From an email exchange with @jwallen, this is indeed unintentional. As a first stage or short-term solution, @whgreen suggested to add a module that runs subsequent to model completion that by default converts all `=>` signs to `=` (unless the user explicitly chooses not to in the input file). This will also affect many ERC-FoundationFuelv0.9 library reactions which are defined as irreversible, e.g.:
```
CH3(10)+O(26)=>H(5)+H2(7)+CO(29)
O(26)+CH2(34)=>H(5)+H(5)+CO(29)
O2(4)+CH2(34)=>O(26)+CH2O(35)
```
Since we don't want to allow the reactions with three products to become reversible, perhaps the suggested module should prohibit this?

rwest:
A post-processing fix won't help flux analysis (and thus model expansion decisions) while building the model, and if it's going to involve special cases for seed mechanisms and reaction libraries, it may end up being as difficult to write as fixing the problem at source. So I'd recommend spending a few more minutes looking into fixing the problem where it occurs before starting to implement a post-processing fix.

whgreen:
Alon,
  Aye reversible reactions with three products really a problem? Can you test in CHEMKIN and in the RMG ode solver?
    Bill

Sent from my iPhone

On Aug 2, 2016, at 8:37 PM, Alon Grinberg Dana <notifications@github.com<mailto:notifications@github.com>> wrote:


Some PDep reactions in many RMG model are irreversible (See e.g., issue #731<https://github.com/ReactionMechanismGenerator/RMG-Py/issues/731>). From an email exchange with @jwallen<https://github.com/jwallen>, this is indeed unintentional. As a first stage or short-term solution, @whgreen<https://github.com/whgreen> suggested to add a module that runs subsequent to model completion that by default converts all => signs to = (unless the user explicitly chooses not to in the input file). This will also affect many ERC-FoundationFuelv0.9 library reactions which are defined as irreversible, e.g.:

CH3(10)+O(26)=>H(5)+H2(7)+CO(29)
O(26)+CH2(34)=>H(5)+H(5)+CO(29)
O2(4)+CH2(34)=>O(26)+CH2O(35)


Since we don't want to allow the reactions with three products to become reversible, perhaps the suggested module should prohibit this?

alongd:
[...]

PR #746, opened 2016-08-17
Debugging (and fixing?) issue #736

rwest:
This should probably be tidied up and perhaps squashed before being merged (or discarded) but I thought I'd share it via a Pull Request (and let Travis do it's checking). See further discussion on issue #736


rmgpy/rmg/model.py
```diff
@@ -1404,6 +1405,7 @@ def addReactionToUnimolecularNetworks(self, newReaction, newSpecies, network=Non
 
         # Add the path reaction to that network
         network.addPathReaction(newReaction)
+        network.invalidate()
```
mliu49: This one might be unnecessary since `addPathReaction` will invalidate the network.

PR #750, opened 2016-08-22
Possible fix for the duplicate training reactions problem

connie:
Bug was found where reaction 2 in  R_Addition_MultipleBond/training
C2H3O3 <=> C2H2O + HO2
was getting duplicated in the final model.  It appears
that we fail the checkForExistingReaction check because we typically
only check in reverse direction when the family is its ownReverse.
However, this forgets that there are some training reactions that
are in fact in the reverse direction!

Therefore we should more expand the searchRetrieveReactions()
function to generate reverse direction reactions for all families.
This may slow things down, leave optimization till later.

### DIFF
--- a/rmgpy/rmg/model.py
+++ b/rmgpy/rmg/model.py
@@ -395,8 +395,8 @@ class CoreEdgeReactionModel:
                 else:
                     if areIdenticalSpeciesReferences(rxn, rxn0):
                         return True, rxn0
-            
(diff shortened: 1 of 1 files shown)


connie:
Manual testing of `official/master` versus this branch shows that runtimes do not increase with these changes, and duplicate training reactions have been successfully removed from the MCH, NC, and methane examples in `RMG-tests`.

PR #751, opened 2016-08-22
Kinetics comments: use square brackets instead of parentheses

connie:
Because kinetics groups often contain parentheses, use square brackets instead inside kinetics comments for better differentiation from the actual names of the groups.  This is so that we can also parse the comments more easily later.


rwest:
This does limit the use of SMILES strings in node names, but perhaps we never did that anyway?

connie:
We did do it, but there were only 3 nodes I had to change out of using square brackets.  I'd say the restriction is fine.  People can use SMILES in their training reactions.

rwest:
Sounds good - Training Reactions is probably where things with SMILES belong anyway.

PR #752, opened 2016-08-22
Mark core pdep reactions as reversible

mliu49:
Instead of looping over net reactions of all pdep networks and checking if they're in the core, loop over all core reactions and check if they're pdep.

Alternative solution for #736.

### DIFF
--- a/rmgpy/rmg/model.py
+++ b/rmgpy/rmg/model.py
@@ -1460,13 +1460,12 @@ class CoreEdgeReactionModel:
         # direction from the list of core reactions
         # Note that well-skipping reactions may not have a reverse if the well
         # that they skip over is not itself in the core
-        for network in updatedNetworks:
-            for reaction in network.netReactions:
-                try:
-                    index = self.core.reactions.index(reaction)
-                except ValueError:
-                    continue
-                for index2, reaction2 in enumerate(self.core.reactions):
+        index = 0
+        coreReactionCount = len(self.core.reactions)
+        while index < coreReactionCount:
+            reaction = self.core.reactions[index]
+            if isinstance(reaction, PDepReaction):
+                for reaction2 in self.core.reactions[index+1:]:
                     if isinstance(reaction2, PDepReaction) and reaction.reactants == reaction2.products and reaction.products == reaction2.reactants:
                         # We've found the PDepReaction for the reverse direction
                         dGrxn = reaction.getFreeEnergyOfReaction(300.)
@@ -1493,10 +1492,13 @@ class CoreEdgeReactionModel:
                             self.core.reactions.remove(reaction2)
                             self.core.reactions.insert(index, reaction2)
                             reaction2.reversible = True
+                        coreReactionCount -= 1
                         # There should be only one reverse, so we can stop searching once we've found it
                         break
                 else:
                     reaction.reversible = True
+            # Move to the next core reaction
+            index += 1
 
 
     def markChemkinDuplicates(self):
--- a/rmgpy/rmg/model.py
+++ b/rmgpy/rmg/model.py
@@ -1444,7 +1444,7 @@ class CoreEdgeReactionModel:
                         index += 1
 
         count = sum([1 for network in self.networkList if not network.valid and not (len(network.explored) == 0 and len(network.source) > 1)])
-        logging.info('Updating {0:d} modified unimolecular reaction networks...'.format(count))
+        logging.info('Updating {0:d} modified unimolecular reaction networks (out of {1:d})...'.format(count, len(self.networkList)))
         
         # Iterate over all the networks, updating the invalid ones as necessary
         # self = reactionModel object
--- a/rmgpy/rmg/pdep.py
+++ b/rmgpy/rmg/pdep.py
@@ -491,6 +491,8 @@ class PDepNetwork(rmgpy.pdep.network.Network):
         if self.valid: return
         # Do nothing if there are no explored wells
         if len(self.explored) == 0 and len(self.source) > 1: return
+        # Log the network being updated
+        logging.info("Updating {0:s}".format(self))
 
         # Generate states data for unimolecular isomers and reactants if necessary
         for isomer in self.isomers:


rwest:
Modifying a list while iterating over it is usually discouraged and called a "code smell". In this case, it seems to be bug free, and possibly quite efficient, but the fact that I just wasted <redacted> time to convince myself why that's the case, illustrates why it's discouraged!  

An example of behavior that might surprise people looking at code that modifies a list while iterating:
```python
example = range(20)
for i in example:
    if i<10:
        example.remove(i)
print example
```
gives `[1, 3, 5, 7, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19]`.
Even if that's what you intended, there's a high chance someone editing your code later will mess it up.

The Python 2 documentation 7.3. "The for statement" gives the same advice:

>Note: There is a subtlety when the sequence is being modified by the loop (this can only occur for mutable sequences, i.e. lists). An internal counter is used to keep track of which item is used next, and this is incremented on each iteration. When this counter has reached the length of the sequence the loop terminates. This means that if the suite deletes the current (or a previous) item from the sequence, the next item will be skipped (since it gets the index of the current item which has already been treated). Likewise, if the suite inserts an item in the sequence before the current item, the current item will be treated again the next time through the loop. This can lead to nasty bugs that can be avoided by making a temporary copy using a slice of the whole sequence, e.g.,

mliu49:
Ah, that make sense. My inexperience with python clearly shows...

I'll refactor it.

mliu49:
@rwest, do you think this can be merged? It would be nice if we could get it into the 2.0 release (@KEHANG).

rwest:
I think so.

But when you run it do you find the line `logging.info("Updating {0:s}".format(network))` prints networks that aren't actually changing, and does more than the `Updating {0:d} modified unimolecular reaction networks` line would suggest?

mliu49:
It seems that networks without any explored isomers will not be updated even if they are invalid. I moved the line into `rmgpy.rmg.pdep.update` so that it only prints if the network is actually updated.

rwest:
When I run the 1,3-hexadiene example (eg2) on this branch, with QMThermo turned off for speed, I get a lot of warnings like this:
```
Warning: For path reaction [CH2]C(C=CCC)C([CH]CC)C=C(33) <=> [CH2]C(C=CCC)[C](C=C)CCC(252):
Warning:     Expected kf(1896.74 K) = 4.83322e+08
Warning:       Actual kf(1896.74 K) = 3.67284e+08
Warning:     Expected Keq(1896.74 K) = 1.10939
Warning:       Actual Keq(1896.74 K) = 0
Warning: Significant corrections to k(E) to be consistent with high-pressure limit for path reaction "[CH2]C(C=CCC)C([CH]CC)C=C(33) <=> [CH2]C(C=CCC)[C](C=C)CCC(252)".
Warning: For path reaction [CH2]C(C=CCC)C([CH]CC)C=C(33) <=> [CH2]C(C=CCC)C(C=C)C[CH]C(254):
Warning:     Expected kf(1896.74 K) = 2.16857e+08
Warning:       Actual kf(1896.74 K) = 1.54733e+08
Warning:     Expected Keq(1896.74 K) = 1.22516
Warning:       Actual Keq(1896.74 K) = 0
Warning: Significant corrections to k(E) to be consistent with high-pressure limit for path reaction "[CH2]C(C=CCC)C([CH]CC)C=C(33) <=> [CH2]C(C=CCC)C(C=C)C[CH]C(254)".
Warning: For path reaction [CH2]C(C=CCC)C([CH]CC)C=C(33) <=> [CH2][C](C=CCC)C(C=C)CCC(255):
Warning:     Expected kf(1896.74 K) = 8.55574e+08
Warning:       Actual kf(1896.74 K) = 6.62849e+08
Warning:     Expected Keq(1896.74 K) = 2.79274
Warning:       Actual Keq(1896.74 K) = 0
Warning: Significant corrections to k(E) to be consistent with high-pressure limit for path reaction "[CH2]C(C=CCC)C([CH]CC)C=C(33) <=> [CH2][C](C=CCC)C(C=C)CCC(255)".
```
always it says the Actual Keq = 0, which suggests something funny with our reversibility assumptions/checks?

Eventually it crashed on me with this:
```python
Warning: For path reaction C=C[CH]C(CC)C[CH]C=CCC(35) <=> [CH2]C=CC([CH]C=CCCC)CC(341):
Warning:     Expected kf(1896.74 K) = 0.0600994
Warning:       Actual kf(1896.74 K) = 0.00876467
Warning:     Expected Keq(1896.74 K) = 1
Warning:       Actual Keq(1896.74 K) = 0
Error: Invalid k(E) values computed for path reaction "C=C[CH]C(CC)C[CH]C=CCC(35) <=> [CH2]C=CC([CH]C=CCCC)CC(341)".
Error: Increasing number of grains did not decrease error enough (Current badness: 0.8, previous 0.8). Something must be wrong with network 9
  File "../../rmg.py", line 164, in <module>
    rmg.execute(**kwargs)
  File "/Users/rwest/Code/RMG-Py/rmgpy/rmg/main.py", line 600, in execute
    self.reactionModel.enlarge(objectToEnlarge)
  File "/Users/rwest/Code/RMG-Py/rmgpy/rmg/model.py", line 675, in enlarge
    self.updateUnimolecularReactionNetworks()
  File "/Users/rwest/Code/RMG-Py/rmgpy/rmg/model.py", line 1454, in updateUnimolecularReactionNetworks
    network.update(self, self.pressureDependence)
  File "/Users/rwest/Code/RMG-Py/rmgpy/rmg/pdep.py", line 561, in update
    K = self.calculateRateCoefficients(Tlist, Plist, method)
  File "/Users/rwest/Code/RMG-Py/rmgpy/pdep/network.py", line 210, in calculateRateCoefficients
    self.setConditions(T, P)
  File "/Users/rwest/Code/RMG-Py/rmgpy/pdep/network.py", line 325, in setConditions
    raise error
rmgpy.pdep.network.InvalidMicrocanonicalRateError
```

Here's the [input.py](https://github.com/ReactionMechanismGenerator/RMG-Py/files/438036/input.py.txt) file


I would prefer to see these two things explained and understood, if not fixed, before merging.

mliu49:
Those warnings and the error are unrelated to the changes introduced in this branch. Running the example on master gives the same result.

I can try to look into it though.

rwest:
Ah, good point @mliu49. They also persist with QMTP on. I'll merge this PR now, but please do look into the problems in eg2 and let us know what you find!

Issue #611, opened 2016-03-17
Requested thermo libraries not checked for existence.


rwest:
You can request in your input file a thermo library that doesn't exist, and it doesn't cause a problem until trying to look thermo up in the library when it gives a KeyError. It would be nice if it checked when loading the database (or the input file), and warned the user with a helpful message.

mliu49:
Addressed by #766.

PR #766, opened 2016-09-07
Thermo library existence checking when loading database

KEHANG:
This PR is mainly to address the issue #611 

- checks thermo library existence when loading database instead of crashing after job starts

- adds an unit test for the change method `loadLibraries` in `thermo.py`

### DIFF
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -751,18 +751,30 @@ class ThermoDatabase(object):
         points to the top-level folder of the thermo database.
         """
         self.libraries = {}; self.libraryOrder = []
-        for (root, dirs, files) in os.walk(os.path.join(path)):
-            for f in files:
-                name, ext = os.path.splitext(f)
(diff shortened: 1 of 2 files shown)


nyee:
I don't understand the purpose of this pull request. Don't we want RMG to crash if the user puts in existent library?

I think it is rare that a user will decide to run the job without the library anyway. Even worse, the user might not notice their library didn't load and let the job keep running.

mliu49:
It seems that the current way will not load the library if it doesn't exist, but still add the name to `self.libraryOrder`, which won't cause errors until later with a non-helpful KeyError.

I do think that making this an error instead of a warning might be better so the user knows right away that something was wrong with the input file. It could be something minor like a spelling mistake, and it would bad for user to be unaware and have the job keep running.

KEHANG:
@mliu49 I agree we should let users know the issues as early as possible. Thanks for the comment.

KEHANG:
@nyee @mliu49 you guys have more suggestions?

Issue #640, opened 2016-04-25
Mutable constructors in classes


connie:
Nathan brought this to my attention: http://stackoverflow.com/questions/4841782/python-constructor-and-default-value

Initializing a class with a mutable object such as an empty default list `[]` can sometimes like to problems when recreating another instance of that class.  I wasn't super aware of this, and so far it seems that RMG doesn't immediately have this problem but we could maybe scour the classes with a script just to be sure.

rwest:
Yes, I think the standard procedure is to initialize it with `None` and have the `__init__` method create the empty list. One of the python "gotchas" like [integer division](https://github.com/ReactionMechanismGenerator/RMG-Py/commit/63298c2a45c40b0ecef28bfca1a9429bbaafd4c5).  If someone is into static analysis scripting, you could check for instances of that too!

nickvandewiele:
https://github.com/ReactionMechanismGenerator/RMG-Py/issues/292

rwest:
Hah - thanks @nickvandewiele, I'd forgotten that one!

connie:
Mutable constructors found in the following code through regex search:

https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/molecule/group.py#L537

https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/tools/canteraModel.py#L166

Use of `[]` and `{}` in:
https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/tools/observablesRegression.py#L67

Mutiple uses of `{}` in:
https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/tools/plot.py

PR #768, opened 2016-09-09
Long distance interaction

nyee:
This pull request changes the way we define benson group corrections to include :class:Groups with multiple labelled atoms.


rmgpy/molecule/molecule.py
```diff
@@ -1595,4 +1595,66 @@ def toGroup(self):
         
         return group
 
-
+    def getFirstNeighbor(self, atom):
```
nyee: All these getNeighbor functions seem like they should be done recursively. Just as quick pseudo-code suggestion.

def recursivelyGetNthNeighbors( startingAtoms, n, blacklist):
   if startingAtoms is not a list: [startingAtoms] 
   neighbors = []
   for atom in startAtoms:
       neighbors.extend([neighbor if neighbor.isNonHydrogen() for neighbor in self.getBonds(atom)])
   neighbors = list(set(neighbors)-set(blackList)) #remove duplicates and atoms already covered in blacklist
   if n >0:
     neighbors = recursivelyGetNthNeighbors(neighbors, n-1, startingAtoms)
   return neighbors

Above pseudo-code might need polish/debugging, but you get the idea.

rmgpy/data/thermo.py
```diff
@@ -1403,13 +1405,15 @@ def computeGroupAdditivityThermo(self, molecule):
                     raise
                 # Correct for gauche and 1,5- interactions
                 if not cyclic:
```
nyee: Aren't your ortho/meta/para corrections going to never be used because these are always on cyclic?

rmgpy/data/thermo.py
```diff
@@ -1404,7 +1405,7 @@ def computeGroupAdditivityThermo(self, molecule):
                     logging.error(molecule.toAdjacencyList())
                     raise
                 # Correct for gauche and 1,5- interactions
-                if not cyclic:
+                if molecule.isAtomInCycle(atom) and (atom.atomType.label!='Cb'):
```
nyee: This work around seems too non-general to me. I think we should separate interactions between cyclic ones and non-cyclic ones. Maybe the gauche and ortho should be in separate files, and analogously for meta and int15

rmgpy/data/thermo.py
```diff
@@ -193,16 +193,19 @@ def addThermoData(thermoData1, thermoData2, groupAdditivity=False):
         if len(thermoData1.Tdata.value_si) != len(thermoData2.Tdata.value_si) or any([T1 != T2 for T1, T2 in zip(thermoData1.Tdata.value_si, thermoData2.Tdata.value_si)]):
             raise Exception('Cannot add these ThermoData objects due to their having different temperature points.')
         
+        testZero = abs(thermoData2.H298.value_si)+abs(thermoData2.S298.value_si) # Used to check if all of the entries in thermoData1 are zero
```
nyee: A more concise implementation would be:

testZero=sum([abs(value) for value in [H298, S298] + Cpdata]]

Then just check if this is zero.

rmgpy/data/thermo.py
```diff
@@ -193,16 +193,19 @@ def addThermoData(thermoData1, thermoData2, groupAdditivity=False):
...
         for i in range(thermoData1.Tdata.value_si.shape[0]):
             thermoData1.Cpdata.value_si[i] += thermoData2.Cpdata.value_si[i]
+            testZero += abs(thermoData2.Cpdata.value_si[i]) # Used to check if all of the entries in thermoData1 are zero
```
nyee: Isn't this testing thermoData2? maybe irrelelvant if you use my above suggestion.

rmgpy/data/thermo.py
```diff
@@ -193,16 +193,19 @@ def addThermoData(thermoData1, thermoData2, groupAdditivity=False):
...
         thermoData1.H298.value_si += thermoData2.H298.value_si
         thermoData1.S298.value_si += thermoData2.S298.value_si
 
         if groupAdditivity:
-            if thermoData1.comment:
-                thermoData1.comment += ' + {0}'.format(thermoData2.comment)
-            else:
-                thermoData1.comment = 'Thermo group additivity estimation: ' + thermoData2.comment
+            if testZero !=0: # Used to check if all of the entries in thermoData1 are zero
```
nyee: comment should say if thermoData2 is zero

rmgpy/data/thermo.py
```diff
@@ -214,16 +217,19 @@ def removeThermoData(thermoData1, thermoData2, groupAdditivity=False):
...
     thermoData1.S298.value_si -= thermoData2.S298.value_si
 
     if groupAdditivity:
-        if thermoData1.comment:
-            thermoData1.comment += ' - {0}'.format(thermoData2.comment)
-        else:
-            thermoData1.comment = 'Thermo group additivity estimation: ' + ' - {0}'.format(thermoData2.comment)
+        if testZero !=0: # Used to check if all of the entries in thermoData1 are zero
+            if thermoData1.comment:
+                thermoData1.comment += ' - {0}'.format(thermoData2.comment)
```
nyee: This gives you an unsimplified comment at the end. A better way to do it is to regular expression substitute it out. We probably need to think about the regex a bit more, but off-the-cuff suggestion:

`thermoData1.comment = re.sub(re.escape(thermoData2.comment)+'\b', '', thermoData1.comment)`

rmgpy/molecule/molecule.py
```diff
@@ -1595,4 +1595,38 @@ def toGroup(self):
         
         return group
 
-
+    def getNthNeighbor(self, startingAtoms, n, blacklist):
```
nyee: Try changing the variable name from 'blacklist' to 'ignore', as that is more clear. My bad for the stupid terminology.  I suggest making 'ignore' an empty list by default:

`def getNthNeighbor(self, startingAtoms, n, ignore = []):`

Then, be a little bit more specific on defining each variable: E.g. startingAtoms is a list of :class:Atom for which we will get the nth partner, n is an interger, ignore is a list of :class:Atoms that will not be returned in the neighboring atoms. etc.

rmgpy/molecule/molecule.py
```diff
@@ -1595,4 +1595,38 @@ def toGroup(self):
...
+                if neighbor.isNonHydrogen():
+                    neighbors.append(neighbor)
+        neighbors = list(set(neighbors)-set(blacklist))
+        for atom in startingAtoms:
+            blacklist.append(atom)
+        if n>1:
+            neighbors = self.getNthNeighbor(neighbors, n-1, blacklist)
+        return neighbors
+
+    def getBaseCb(self, atom):
```
nyee: I don't understand what this is doing. It returns [atom] if its a Cb atom or a list of neighboring Cb atoms? What does the labeling of atoms do?

rmgpy/data/thermo.py
```diff
@@ -1327,6 +1335,23 @@ def estimateRadicalThermoViaHBI(self, molecule, stableThermoEstimator ):
                 logging.error(molecule)
                 logging.error(molecule.toAdjacencyList())
                 raise
+            
+            # Add the radical interactions which are ignored in the thermo calculation of molecule.
+            if saturatedStruct.isAromatic() and (not saturatedStruct.isAtomInCycle(atom)):
+                baseAtom = saturatedStruct.getBaseCb(atom)[0]
```
nyee: Probably related to the fact that I don't understand the getBaseCb method, but I also don't understand what is going on here. You get the closest Cb to H? Do you need to check all Cbs?

rmgpy/data/thermo.py
```diff
@@ -1336,6 +1361,28 @@ def estimateRadicalThermoViaHBI(self, molecule, stableThermoEstimator ):
             for H, bond in added[atom]:
                 thermoData.H298.value_si -= 52.103 * 4184
 
+            # Remove the redundant interactions introduced by saturating the radical.
+            if saturatedStruct.isAromatic() and (not saturatedStruct.isAtomInCycle(atom)):
+                baseAtom = saturatedStruct.getBaseCb(atom)[0]
+                for atom_1 in saturatedStruct.getNthNeighbor([baseAtom],1,[]):
+                    try:
+                        # Try to match twice. As 'baseAtom' could be labeled as '*1' or '*2'.
+                        self.__removeGroupThermoData(thermoData,self.groups['interactionDistance1'], saturatedStruct, {'*1':baseAtom, '*2':atom_1})
```
nyee: If you remove twice here, do you need to add twice in 1344 also?

rmgpy/data/thermo.py
```diff
@@ -1456,7 +1517,7 @@ def __addPolycyclicCorrectionThermoData(self, thermoData, molecule, polyring):
         # polycylic ring in molecule and match group)
         # otherwise, apply heuristic algorithm
         if not isPolyringPartialMatched(polyring, matched_group):
-            thermoData = addThermoData(thermoData, matched_group_thermodata, groupAdditivity=True)
+            thermoData = addThermoData(thermoData, matched_group_thermodata, groupAdditivity=True, verbose=True)
```
nyee: Do we want verbose to be True or False here? I was under the impression that verbose is more for debugging purposes.

rmgpy/molecule/molecule.py
```diff
@@ -1595,4 +1595,38 @@ def toGroup(self):
...
-
+    def getNthNeighbor(self, startingAtoms, n, blacklist):
+        '''
+        Recursively get the Nth nonHydrogen neighbors of the startingAtoms.
+        StartingAtoms should be a list.
+        Return a list.
+        '''
+        neighbors = []
+        for atom in startingAtoms:
+            for neighbor in self.getBonds(atom):
```
nyee: Nitpick: but you could remove the second layer of the for loop by putting it in the list (replacing lines 1606-1608):

```
newNeighbors = [neighbor for neighbor in self.getBonds(atom) if neighbor.isNonHydrogen()]
neighbors.extend(newNeighbors)
```

Guess it only saves one line of code...

rmgpy/data/thermo.py
```diff
@@ -183,7 +183,7 @@ def processOldLibraryEntry(data):
     )
 
 
-def addThermoData(thermoData1, thermoData2, groupAdditivity=False):
+def addThermoData(thermoData1, thermoData2, groupAdditivity=False, verbose=False):
```
PengZhang13: That's right. Fixed with c484dcc92da067d7a5beacbbfb11543d82295b29

rmgpy/data/thermo.py
PengZhang13: Here we want to turn on the comments of polycyclic correction to pass the unittest. 
Typically this comment is very short and also very helpful to check if the ring correction is calculated correctly. 
So I think probably here we could leave it on.

rmgpy/molecule/molecule.py
```diff
@@ -1618,4 +1618,21 @@ def toGroup(self):
...
+        for atom in startingAtoms:
+            newNeighbors = [neighbor for neighbor in self.getBonds(atom) if neighbor.isNonHydrogen()]
+            neighbors.extend(newNeighbors)
+
+        neighbors = list(set(neighbors)-set(ignoreList))
+        for atom in startingAtoms:
+            ignoreList.append(atom)
+        if n>1:
+            neighbors = self.getNthNeighbor(neighbors, n-1, ignoreList)
+        return neighbors
```
PengZhang13: I removed the confusing getBaseCb method in 32705582bb4d1adcaab97fd61cb10f79c360ff7a since we don't need it anymore.

rmgpy/data/thermo.py
```diff
@@ -219,11 +223,12 @@ def removeThermoData(thermoData1, thermoData2, groupAdditivity=False):
...
 
+    testZero = sum(abs(value) for value in [thermoData2.H298.value_si, thermoData2.S298.value_si]+thermoData2.Cpdata.value_si.tolist())
+    # Used to check if all of the entries in thermoData2 are zero
+
     if groupAdditivity:
-        if thermoData1.comment:
-            thermoData1.comment += ' - {0}'.format(thermoData2.comment)
-        else:
-            thermoData1.comment = 'Thermo group additivity estimation: ' + ' - {0}'.format(thermoData2.comment)
+        if verbose or testZero !=0: # Used to check if all of the entries in thermoData2 are zero
```
PengZhang13: You're right. Fixed with b943cd0bb33e795869f8ec1051fddfc021435ac8.

rmgpy/data/thermo.py
```diff
@@ -1327,6 +1335,23 @@ def estimateRadicalThermoViaHBI(self, molecule, stableThermoEstimator ):
                 logging.error(molecule)
                 logging.error(molecule.toAdjacencyList())
                 raise
+            
+            # Add the radical interactions which are ignored in the thermo calculation of molecule.
+            if saturatedStruct.isAromatic() and (not saturatedStruct.isAtomInCycle(atom)):
```
PengZhang13: Thanks for the clarification that parenthesis is unnecessary.  
The 'atom' here is the atom that has the radical site, not the H to be added. So it could be in a cycle.

rmgpy/molecule/molecule.py
PengZhang13: Fixed with 32705582bb4d1adcaab97fd61cb10f79c360ff7a

rmgpy/molecule/molecule.py
PengZhang13: Thanks for the great suggestion! Fixed with 32705582bb4d1adcaab97fd61cb10f79c360ff7a

mliu49:
I've been thinking about how to organize these corrections, and I agree that they shouldn't be grouped by distance only (which you've already reverted).

It seems that the cleanest way would be to have two sets of corrections: cyclic and non-cyclic, then match by distance and groups within each type. However, that wouldn't allow multiple corrections to be applied from a single file. Given that, the way it is now is probably the best option. I think it makes sense logically.

I do think that the interactions should be renamed to `cyclic (or aromatic) _ortho/_meta/_para` for clarity.

nyee:
I agree with Max about the cyclics vs noncyclics.  I was also thinking maybe we should just some variables at the top of each correction.py file.

```
multipleCorrectionsPerAtom = False/True
centerDistance = 3
```

the tells the code whether to look for more than one correction and how far to distance between labelled atoms are.

mliu49:
I really like these new changes! I'll do a more detailed code review sometime this weekend.

PengZhang13:
Thanks. Me too! Your SSSR suggestion works fantastically!

One thing to be discussed: 
For now, we pair the atoms using their permutations, then half some of the values in the database since they'll be hit twice. Should we migrate to combinations and use full values for all the entries as we discussed before? In that case, the database will be more clear but the code might be tricky. Because actually we could always hit something (at least some general entries) for both {'*1', atom1, '*2',atom2} and {'*1', atom2, '*2',atom1}. So it might be useless to try to match the second one if the first one fails (as it will never fail).

rmgpy/data/thermo.py
```diff
@@ -1414,16 +1441,42 @@ def computeGroupAdditivityThermo(self, molecule):
...
-                try:
-                    self.__addGroupThermoData(thermoData, self.groups['int15'], molecule, {'*':atom})
-                except KeyError: pass
+                # Pair atom with its 1st and 2nd nonHydrogen neighbors, 
+                # Then match the pair with the entries in the database longDistanceInteraction_noncyclic.py
+                # Currently we only have gauche(1,4) and 1,5 interactions in that file. 
+                # If you want to add more corrections for longer distance, please call getNthNeighbor() method accordingly.
+                # Potentially we could include other.py in this database, but it's a little confusing how to label atoms for the entries in other.py
+                if not molecule.isAtomInCycle(atom):
+                    for atom_2 in molecule.getNthNeighbor([atom],1,[])+molecule.getNthNeighbor([atom],2,[]):
```
mliu49: I think there's potential here to have `getNthNeighbor` accept a list of distances and return all neighbors at those distances. For example, since getting distance 2 neighbors requires finding distance 1 neighbors, the current way actually requires 3 function calls, one of which is repeated. I'll comment on the actual function with a code suggestion.

While it's not that big of a deal for distance 1 and 2 neighbors, it would help for future extensibility if people add corrections for longer distances.

rmgpy/molecule/molecule.py
```diff
@@ -1618,4 +1618,21 @@ def toGroup(self):
         
         return group
 
-
+    def getNthNeighbor(self, startingAtoms, n, ignoreList = []):
```
mliu49: One possible way to allow a list of distances:
```
def getNthNeighbor(self, startingAtoms, distances, ignoreList=[], n=1):
    `distances` is a list of integers, corresponding to desired neighbor distances
```
Then most of the rest stays the same, until the last part:
```
    if n < max(distances):
        if n in distances:
            neighbors += self.getNthNeighbor(neighbors, distances, ignoreList, n+1)
        else:
            neighbors = self.getNthNeighbor(neighbors, distances, ignoreList, n+1)

    return neighbors
```
I haven't tested this, but I think it should work. Feel free to ignore this if you prefer the current method (or if it doesn't work like I think it does).

rmgpy/data/thermoTest.py
```diff
@@ -171,13 +171,11 @@ def testParseThermoComments(self):
         # Pure group additivity thermo
         propane = Species(index=3, label="Propane", thermo=NASA(polynomials=[NASAPolynomial(coeffs=[3.05257,0.0125099,3.79386e-05,-5.12022e-08,1.87065e-11,-14454.2,10.0672], Tmin=(100,'K'), Tmax=(986.57,'K')),
          NASAPolynomial(coeffs=[5.91316,0.0218763,-8.17661e-06,1.49855e-09,-1.05991e-13,-16038.9,-8.86555], Tmin=(986.57,'K'), Tmax=(5000,'K'))],
-         Tmin=(100,'K'), Tmax=(5000,'K'), comment="""Thermo group additivity estimation: group(Cs-CsCsHH) + gauche(Cs(CsCsRR)) + other(R) + group(Cs-CsHHH) + gauche(Cs(Cs(CsRR)RRR)) + other(R) + group(Cs-CsHHH) + gauche(Cs(Cs(CsRR)RRR)) + other(R)"""), molecule=[Molecule(SMILES="CCC")])
+         Tmin=(100,'K'), Tmax=(5000,'K'), comment="""Thermo group additivity estimation: group(Cs-CsCsHH) + group(Cs-CsHHH) + group(Cs-CsHHH)"""), molecule=[Molecule(SMILES="CCC")])
```
mliu49: Instead of only removing the gauche and other corrections, should this be updated to match what RMG would return for propane with this update? I would guess that the gauche corrections would be replaced by longDistanceInteraction_noncyclic.

mliu49:
> Because actually we could always hit something (at least some general entries) for both {'1':atom1, '2':atom2} and {'1':atom2, '2':atom1}. So it might be useless to try to match the second one if the first one fails (as it will never fail).

Do you mean that it would most likely hit a more general node? If so, is that also a problem with using permutations? For example, could {'1':atom1, '2':atom2} hit a specific node and {'1':atom2, '2':atom1} hit a general node? I'm not sure what the best way to do it would be.

I don't think that halving the values for entries with idential substituents would be an issue though, if we decide to do it that way, as long as it's documented.

PengZhang13:
Yes, it would hit a more general node. But the corrections of general nodes are zero. So it will not be a problem for permutations. For combinations, I think what we are trying to do is avoiding halving the values for entries with identical substituents by doing this:
`if {'1':atom1, '2':atom2} hit a node: pass`
`else: try {'1':atom2, '2':atom1}`
In this way we match the entries with identical substituents only once. That's exactly what we want. However, the problem is, even for the entries with nonidentical substituents,  the `if` statement is always true and `else:` statement will never be excuted. That's not what we want.
Maybe we could find a solution to work around this or just use the 'permutations' version.

rmgpy/molecule/molecule.py
PengZhang13: That's smart! I'll try. Thanks for the code suggestion.

rwest: Remember to be very careful using an empty list as the default parameter to a function. I haven't checked this particular instance, but it is very rarely what you mean to do!  See http://docs.python-guide.org/en/latest/writing/gotchas/ and https://github.com/ReactionMechanismGenerator/RMG-Py/issues/640

rmgpy/molecule/molecule.py
rwest: Nit-nit-pick: the for loop is still there, just implicit, and thus (arguably) harder to understand 😉. Not sure if speed is an issue here, nor how these alternatives would compare, with or without cython. (Probably not worth worrying about)

PengZhang13: Thanks for the comment 😃. Take a shot with 957480c384dce6ab14cf51b7512c7db226618966 by separating the double layer 'for' loop into two individual loops. I think speed is not an issue for the getNthNeighbor method though. But it might be an issue for this branch. Because the descendTree method will be called much more frequently than that in the master branch.

rmgpy/data/thermo.py
PengZhang13: Implemented in ca714963918e081a2c55f658dfec1dd344f49bbe according to your code suggestion. It works! Thanks!

rmgpy/molecule/molecule.py
PengZhang13: Fixed with 243a93bc970925a9b4315b2befe37d97b2a69104.

rmgpy/data/thermoTest.py
PengZhang13: You're right. Fixed with ca714963918e081a2c55f658dfec1dd344f49bbe. Replaced the 'propane' example with a more complicated species to test longDistanceInteraction_cyclic and longDistanceInteraction_noncyclic simultaneously.

rmgpy/molecule/molecule.py
mliu49: I ran some tests out of interest, which you can see [here](https://github.com/mliu49/RMG-stuff/blob/master/Sandbox.ipynb). All three versions of this loop were timed, taking the cumulative time for 100,000 calls, repeated 5 times, then averaged.

I agree that speed is probably not an issue for this method, so it doesn't really matter too much which way you choose. However, the current method (adding all atoms, then removing non-hydrogens) is marginally slower than only adding non-hydrogens in the first place. Whether you use two for loops or list comprehension doesn't make that much of a difference.

PengZhang13: Nice tests. The list comprehension version and the current version are much cleaner for me than the two for loops version. Based on the tests, which show the current version is slower, I tend to choose the list comprehension version.

Issue #767, opened 2016-09-08
valency of elements and benzene bonds


nyee:
@KEHANG, @mliu49, and I were discussing refactoring and fixes for valency. First, we seem to have places where we have duplicated hard-coding for calculating valency, such as [here](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/molecule/adjlist.py#L43) and [here](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/molecule/molecule.py#L359). We suggest moving all of these to the rmgpy.molecule.element.py so that at least it is all hard-coded in the same place.

Second, benzene bond valency is unclear because of Cbf atomtype, which can have three 'B' bonds. After looking at a few example PAH's, we think the best fix is to use a valency of 4/3 if there are exactly three benzene bonds and 3/2 in all other cases. We considered creating a new bond order 'FB' for fused benzene bonds, but rejected it because it still has unclear valency when PAHs get big enough. For example, in pyrene the apparent bond order for 'FB' could be 1 or 2 depending on which bond you are looking at.

nyee:
#403 is related

mliu49:
Addressed by #769.

PR #769, opened 2016-09-09
Dynamic benzene bond order to fix issue #767

KEHANG:
This PR starts with commits made by @rwest aiming to address the bond order issue for `Cbf`. The detailed issue and solution were recorded [in issue 767] (https://github.com/ReactionMechanismGenerator/RMG-Py/issues/767). 

In summary, we try to modify a little to the current benzene bond order assignment: dynamic bond order depending on how many benzene bonds an atom has; if atom has three benzene bonds, the bond order for each benzene bond is 4/3, while other cases, the benzene bond order stays unchanged. This fix will have fused aromatics parsed correctly from adjacencylist.

In addition, 3 fused aromatics unittests and 1 adjacencylist unittest are added to make sure new changes are valid. so this PR includes

- new way of assigning benzene bond order

- additional unittests for fused aromatics parsing

- code clean-up: moving `bond_orders` to `molecule.py` and `PeriodicSystem` to `element.py`

### DIFF
--- a/rmgpy/molecule/adjlist.py
+++ b/rmgpy/molecule/adjlist.py
@@ -35,23 +35,9 @@ import logging
 import re
 from .molecule import Atom, Bond, getAtomType
 from .group import GroupAtom, GroupBond
-from .element import getElement
+from .element import getElement, PeriodicSystem
 #import chempy.molecule.atomtype as atomtypes
 
-class PeriodicSystem(object):
-    valence_electrons_first_period_elements  = {'H':1, 'He':2}
-        
-    valence_electrons_second_period_elements = {'C':4, 'N':5, 'O':6, 'Ne':8}
-        
-    valence_electrons_third_period_elements  = {'Si':4, 'S':6, 'Cl':7, 'Ar':8}
-        
-    valence_electrons = {}
-    valence_electrons.update(valence_electrons_first_period_elements)
-    valence_electrons.update(valence_electrons_second_period_elements)
-    valence_electrons.update(valence_electrons_third_period_elements)
-    
-    lone_pairs         = {'H': 0, 'He':1, 'C': 0, 'N': 1, 'Ne': 4, 'O': 2, 'Si':0, 'S': 2, 'Cl':3, 'Ar':4 }
-    
 class Saturator(object):
     @staticmethod
     def saturate(atoms):
@@ -310,13 +296,12 @@ def fromOldAdjacencyList(adjlist, group=False, saturateH=False):
                                  )
                 
             else:
-                standardLonePairs = {'H': 0, 'C': 0, 'O': 2, 'S': 2, 'Si': 0, 'Cl': 3, 'He': 1, 'Ne': 4, 'Ar': 4}
                 if lonePairsOfElectrons is not None:
                     # Intermediate adjlist representation
                     lonePairsOfElectrons = lonePairsOfElectrons + additionalLonePairs[0]
                 else:
                     # Add the standard number of lone pairs with the additional lone pairs
-                    lonePairsOfElectrons = standardLonePairs[atomType[0]] + additionalLonePairs[0]
+                    lonePairsOfElectrons = PeriodicSystem.lone_pairs[atomType[0]] + additionalLonePairs[0]
                     
                 atom = Atom(element=atomType[0],
                         radicalElectrons=radicalElectrons[0],
@@ -381,17 +366,15 @@ def fromOldAdjacencyList(adjlist, group=False, saturateH=False):
         if not group:
             if saturateH:
                 # Add explicit hydrogen atoms to complete structure if desired
-                standardLonePairs = {'H': 0, 'C': 0, 'O': 2, 'S': 2, 'Si': 0, 'Cl': 3, 'He': 1, 'Ne': 4, 'Ar': 4}
-                valences = {'H': 1, 'C': 4, 'O': 2, 'N': 3, 'S': 2, 'Si': 4, 'Cl': 1, 'He': 0, 'Ne': 0, 'Ar': 0}
                 newAtoms = []
                 for atom in atoms:
                     try:
-                        valence = valences[atom.symbol]
+                        valence = PeriodicSystem.valences[atom.symbol]
                     except KeyError:
                         raise InvalidAdjacencyListError('Error in adjacency list:\n{1}\nCannot add hydrogens: Unknown valence for atom "{0}".'.format(atom.symbol, adjlist))
                     radical = atom.radicalElectrons
                     order = atom.getBondOrdersForAtom()
-                    count = valence - radical - int(order) - 2*(atom.lonePairs-standardLonePairs[atom.symbol])
+                    count = valence - radical - int(order) - 2*(atom.lonePairs-PeriodicSystem.lone_pairs[atom.symbol])
                     for i in range(count):
                         a = Atom(element='H', radicalElectrons=0, charge=0, label='', lonePairs=0)
                         b = Bond(atom, a, 'S')
@@ -882,8 +865,7 @@ def getOldElectronState(atom):
     """
     Get the old adjacency list format electronic state
     """
-    standardLonePairs = {'H': 0, 'C': 0, 'O': 2, 'S': 2, 'Si': 0, 'Cl': 3, 'He': 1, 'Ne': 4, 'Ar': 4}
-    additionalLonePairs = atom.lonePairs - standardLonePairs[atom.element.symbol]
+    additionalLonePairs = atom.lonePairs - PeriodicSystem.lone_pairs[atom.element.symbol]
     electrons = atom.radicalElectrons + additionalLonePairs * 2
     if electrons == 0:
         electronState = '0'
--- a/rmgpy/molecule/adjlist.py
+++ b/rmgpy/molecule/adjlist.py
@@ -93,7 +93,6 @@ class Saturator(object):
                         - 2* atom.lonePairs - order - atom.charge
             
             '''
-            global bond_orders
             newAtoms = []
             for atom in atoms:
                 try:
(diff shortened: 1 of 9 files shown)


rmgpy/molecule/molecule.py
```diff
@@ -356,16 +362,9 @@ def updateCharge(self):
...
-        order = 0
-        for atom2, bond in self.bonds.items():
-            order += orders[bond.order]
-        if self.symbol == 'H' or self.symbol == 'He':
-            self.charge = 2 - valence - order - self.radicalElectrons - 2*self.lonePairs
-        else:
-            self.charge = 8 - valence - order - self.radicalElectrons - 2*self.lonePairs
+        valence = elements.PeriodicSystem.valence_electrons[self.symbol]
+        order = self.getBondOrdersForAtom()
+        self.charge = valence - order - self.radicalElectrons - 2*self.lonePairs
```
mliu49: I think it might be more clear if valence is renamed to valence_electrons here. Previously, the valence was hard coded, then the charge was calculated by subtracting the valence from the size of the valence shell, but now it directly uses the number of valence electrons, but the variable is still called valence.

Again, I don't know how we'll deal with multivalent elements...

rmgpy/molecule/parser.py
```diff
@@ -571,6 +571,7 @@ def convert_3_atom_2_bond_path(start, mol):
     electrons that the molecule should undergo.
     """
     from rmgpy.data.kinetics.family import ReactionRecipe
+    from .molecule import bond_orders
 
     def is_valid(mol):
         """Check if total bond order of oxygen atoms is smaller than 4."""
```
mliu49: For some reason, I can't comment directly on line 581, but could that line be replaced with `getBondOrdersForAtom`?

rmgpy/molecule/parser.py
```diff
@@ -768,15 +769,15 @@ def reset_lone_pairs(mol, p_indices):
...
     """
-
+    from .molecule import bond_orders
     for at in mol.atoms:
         index = mol.atoms.index(at) + 1 #1-based index
         count = p_indices.count(index)
         if count != 0:
             at.lonePairs = count
         else:    
             order = sum([bond_orders[b.order] for _,b in mol.getBonds(at).iteritems()])
```
mliu49: Could this also be replaced by `getBondOrdersForAtom`?

rmgpy/molecule/molecule.py
```diff
@@ -402,6 +401,30 @@ def setSpinMultiplicity(self,spinMultiplicity):
...
+        This helper function is to help calculate total bond orders for an
+        input atom.
+
+        Some special consideration for the order `B` bond. For atoms having 
+        three `B` bonds, the order for each is 4/3.0, while for atoms having other
+        than three `B` bonds, the order for  each is 3/2.0
+        """
+        num_B_bond = 0
+        order = 0
+        for _, bond in self.bonds.items():
```
mliu49: Could this be reduced to `for bond in self.bonds.values()`?

rmgpy/molecule/molecule.py
```diff
@@ -60,6 +60,12 @@
 
 ################################################################################
 
+bond_orders = {'S': 1, 'D': 2, 'T': 3}
+
+globals().update({
+    'bond_orders': bond_orders,
```
mliu49: Just for the sake of me learning, what's the purpose of updating the `globals()` dictionary?

rmgpy/molecule/adjlistTest.py
```diff
@@ -349,6 +349,55 @@ def testFromAdjacencyList4(self):
...
+            """
+        molecule = Molecule().fromAdjacencyList(adjlist, saturateH=True)
+        
+        self.assertTrue(molecule.multiplicity == 1)
+        
+        atom1 = molecule.atoms[0]
+        atom2 = molecule.atoms[1]
+        atom3 = molecule.atoms[2]
+        atom7 = molecule.atoms[6]
+        atom11 = molecule.atoms[10]
```
mliu49: Does the order of atoms always remain constant in RMG? If so, then this should be fine.

rmgpy/molecule/adjlist.py
```diff
@@ -381,7 +367,6 @@ def fromOldAdjacencyList(adjlist, group=False, saturateH=False):
         if not group:
             if saturateH:
                 # Add explicit hydrogen atoms to complete structure if desired
-                orders = {'S': 1, 'D': 2, 'T': 3, 'B': 1.5}
                 standardLonePairs = {'H': 0, 'C': 0, 'O': 2, 'S': 2, 'Si': 0, 'Cl': 3, 'He': 1, 'Ne': 4, 'Ar': 4}
```
mliu49: After looking at this a bit more, I think the two dictionaries should be the same, and nitrogen should be included. Otherwise, there will be a KeyError in line 380 if this function is ever called with a nitrogen atom. I suppose this has never come up because `fromOldAdjacencyList` doesn't get used very much?

KEHANG: Good catch and thanks for the commets! I'm modifying accordingly.

rmgpy/molecule/molecule.py
KEHANG: this is cython related consideration. This step is necessary when you have cythonized constant need to be imported by other python modules. So here `bond_orders` is defined in `molecule.py` which is going to be cythonized, in order to let say `parser.py`, a python module, to import successfully, you have to declare it in a globally accessible dict `globals`.

rmgpy/molecule/molecule.py
KEHANG: thanks for pointing it out, I was trying to use `self.bonds.iteritems()`instead of `self.bonds.items()`, since former is a type of `iterator/generator` which is more memory efficient. For more detail, you can take a look [here](http://stackoverflow.com/questions/24189379/what-is-the-advantage-of-iteritems).

rmgpy/molecule/adjlistTest.py
KEHANG: it's fine here since we didn't do `mol.sortVertices()`

rmgpy/molecule/element.py
```diff
@@ -106,19 +106,21 @@ def __reduce__(self):
...
-    valence_electrons_third_period_elements  = {'Si':4, 'S':6, 'Cl':7, 'Ar':8}
+    """
+    Collects hard-coded information of elements in periodic table. 
 
-    valence_electrons = {}
-    valence_electrons.update(valence_electrons_first_period_elements)
-    valence_electrons.update(valence_electrons_second_period_elements)
-    valence_electrons.update(valence_electrons_third_period_elements)
+    Currently it has static attributes:
+    `valenes`: the number of bonds an element is able to form around it.
```
mliu49: fixup to correct 'valences'?

rmgpy/molecule/parser.py
```diff
@@ -571,13 +571,14 @@ def convert_3_atom_2_bond_path(start, mol):
     electrons that the molecule should undergo.
     """
     from rmgpy.data.kinetics.family import ReactionRecipe
+    from .molecule import bond_orders
```
mliu49: This import and the one on line 772 are no longer needed.

rmgpy/molecule/moleculeTest.py
```diff
@@ -1276,6 +1278,123 @@ def testSaturateAromaticRadical(self):
...
+16 C u0 p0 c0 {6,D} {15,D}
+17 H u0 p0 c0 {7,S}
+18 H u0 p0 c0 {13,S}
+19 H u0 p0 c0 {8,S}
+20 H u0 p0 c0 {9,S}
+21 H u0 p0 c0 {10,S}
+22 H u0 p0 c0 {11,S}
+23 H u0 p0 c0 {14,S}
+24 H u0 p0 c0 {12,S}
+25 H u0 p0 c0 {15,S}
```
mliu49: I didn't notice this earlier, but this adjlist seems a bit weird. Was this the intended molecule?

image

KEHANG: yes, that's intended

rmgpy/molecule/parser.py
KEHANG: good point, thanks!

rmgpy/molecule/moleculeTest.py
rwest: Looks weird, but I think that's one valid way to represent a resonance form of a pyrene radical:
screen shot 2016-09-14 at 9 39 37 pm

rmgpy/molecule/moleculeTest.py
```diff
@@ -1276,6 +1278,123 @@ def testSaturateAromaticRadical(self):
...
+""")
+        pyrene_rad2 = Molecule().fromSMILES('[C]1C=C2C=CC=C3C=CC4=CC=CC=1C4=C23')
+        for isomer in pyrene_rad2.generateResonanceIsomers():
```
mliu49: Should this be `generateAromaticResonanceIsomers(pyrene_rad2)`?

On a related note, RMG goes crazy trying to generate resonance isomers for the kekulized form of this pyrene radical, moving the radical all over the molecule. However, this radical can't shift because it's in an sp2 orbital that's not in the same plane as the delocalized pi system. Another issue to add to the aromaticity list...

KEHANG: Probably I will add another item to your aromaticity list, `generateAromaticResonanceIsomers(pyrene_rad2)` currently cannot generate the intended isomer. Only using `pyrene_rad2.generateResonanceIsomers()` includes that.

rmgpy/molecule/moleculeTest.py
mliu49: Sorry, could you explain a bit more? I don't quite understand why it's a valid resonance form. Also, @rwest, is the pyrene in your image missing two hydrogens?

rwest: Haha oops. I wasted time photoshopping out one of the hydrogens, without realizing I'd successfully removed the one opposite it before generating the image (I thought my attempt had failed). Try this one:
screen shot 2016-09-14 at 11 01 48 pm

It may not be a sensible resonance form; I meant "valid" for the way RMG currently handles delocalization and Kekulization. 

This may be naïve, but my quick thinking (I think based on what RMG attempts)  is any bonding arrangement which allows you to put all the atoms (including H's) in the right place, and has the same overall number of radicals (hence multiplicity), is a resonance structure.

mliu49:
Everything looks good! I'll go ahead and merge this.

The resonance isomer question is something that I will work on later, and we can change that unit test if needed in the future.

PR #774, opened 2016-09-19
Entry accessibility

nyee:
Making this pull request as a place to write down work that still needs to be done. This is not ready to be merged.

This pull request addresses the issue of entry accessibility unit tests for the RMG-database. The main testDatabases other than kinetics are almost complete, and should probably go into their own pull request. Work that still needs to be done on these:
1. Functions and unit tests dealing with addImplicitBenzene. Right now it does too much and doesn't do those things right. I shouldn't try to handle Cbf3 as its more complicated than anything in the database and phenalene (my simple test case for Cbf3) is not actually fully aromatic on every ring. 
2. Refactor valency in standardizeAtomTypes method to use methods from #769  
3. Finish making a method to test that either a group has either <3 Cbf atoms or has all Cb/Cbf atoms in explicit rings. Otherwise we shouldn't allow these groups in the database because it is too hard to make sample molecules (by hand or by code)

I also want to use some of these commits to fix entry accessibility in kinetics, which will probably make a second pull request. There is still a lot of work to do for that:
1. Problem of conflicting ends and backbones. Some unimolecular kinetics families follow the paradigm of having a 'backbone' subtree describing the overall shape of the molecule and 'end' groups describing the reactive sites. We need to refactor the backbone entries to be orthogonal to the end entries. That is, any labelled atom overlapping between an end Group and backbone Group should be as general as the top of the end subtree. In the backbone, we only describe atoms not accounted for in the end groups (non-overlapping), or the bonding between non-overlapping atoms and overlapping atoms. This will make organization of the groups better as well as fix difficulties in making sample atoms.
2. To make sample atoms to descend the tree, backbone groups are the same as thermo groups, but end groups need to be merged with a general backbone before creating the molecule. This is partially implemented, but does not work well for backbones that are cyclic (e.g. entry R3H_SS_12cy3 i intra_H_migration)
3. Diels_alder_addition is a special case of having a pseudo-backbone/end paradigm. It is not actually unimolecule, but one of the reactants has backbone/end subtrees. It is a little more complicated, because usually the ends are terminal from non-reacting backbone labelled atoms: i.e. end1-nonOverlappingBackboneAtoms-end2. However diels alder actually has the "end" sandwiched in the middle of the backbone. This special case needs to be handled differently
4. I briefly discussed a problem of subtree leakage with Kehang and Richard, but further testing makes it look like it is actually already handled correctly. RMG labels all atoms on the backbone before descending the trees, matching with strict = True. Might want to confirm this after fixing above problems.
5. Once test is working and backbone/end is cleaned up. We still need to use the database test to find real non-accessible entries like we did for thermo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,9 @@ before_install:
   - conda update --yes conda
   - cd ..
   - git clone https://github.com/ReactionMechanismGenerator/RMG-database.git
-  - cd RMG-Py
+  - cd RMG-database
+  - git checkout unimol_accessibility
+  - cd ../RMG-Py
 
 install:
   - conda env create -f environment_linux.yml
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -2182,3 +2182,57 @@ class KineticsFamily(Database):
 
         return groupList
 
+    def mergeGroups(self, backbone, end):
+        """
+        This function takes two :class:Group objects and returns a merged based on overlapping atoms
+
+        backbone is the group that has greater labelled atoms usually showing the length of a chain in a family
+        end is the group that has fewer labelled atoms usually defining a reacting site at the end of a chain
+
+        Currently assumes the end can be merged at the closest labelled atom
+        """
+        labeled1 = backbone.getLabeledAtoms()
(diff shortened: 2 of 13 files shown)


nyee:
Moved pull request to official/entryAccessibility

PR #781, opened 2016-10-06
Modified to prevent user-specified PList from being overwritten

zjburas:
Changed if statement in PressureDependenceJob.generatePressureList to elif, so that the user-specified PList in Cantherm doesn't always get overwritten. This way, if the user wants cantherm to return k's at specific P's they can do so. I'm not sure if the original if statement was intentional or not, specifically, if it was written that way because of how cantherm is run in RMG mode. The analogous generateTemperatureList function uses an elif statement.


mliu49:
This looks like it was a bug, since `generateTemperatureList` has the same structure with `elif` there. I think it's safe to merge.

mliu49:
Sorry Zach, I didn't realize what "Update branch" did. Could you reset your master branch to 90b463f4de3a29a018b939847e799180c7cbfab6 and then rebase on top of official master?

zjburas:
@mliu49 , I'll just close this PR and reopen another one the right way.

Issue #783, opened 2016-10-11
RMG should allow users to specify transport libraries


alongd:
RMG's database currently have two transport libraries, GRIMech and PrimaryTransportLibrary (might be expanded soon to include JetSurF2 as well), but the user cannot specify transport libraries in the input file. As discussed with @KEHANG , it would be great to add a feature to allow users to specify this (and use the GRIMech and Primary as defaults).

PR #793, opened 2016-10-24
enable users to select customized transport liraries

KEHANG:
This PR tries to address issue #783 that currently users are not allowed to use their own transport libraries. With this patch, everything stays same if user doesn't specify the argument `transportLibraries` in input file. But if one specifies, RMG will use the specified ones.


KEHANG:
@alongd could you test this functionality since you have this need.

alongd:
@KEHANG , seems to work! :)

PR #805, opened 2016-10-28
Resonance Part 1: Clar Structure Generation

mliu49:
This PR is not ready for merging yet, but I'm opening the request so it can start being reviewed.

This is the first part of an overhaul of resonance structure generation in RMG. This PR does not actually affect any functionality. It only adds methods for Clar structure generation that will be actually integrated into the resonance algorithm in the second part.
### Clar Structures

The motivation behind using Clar structures is the lack of a consistent and deterministic resonance structures for polycyclic aromatic compounds. We currently have a fully delocalized structure, where all bonds in rings are benzene bonds. This works well for thermo estimation because it's consistent with Benson group additivity, but it is not great for kinetics since bonds in PAHs can have different reactivity depending on location in the molecule.

The basis of Clar structures is a more rigid electron counting system, wherein sets of 6 pi-electrons are assigned to specific rings. As a result, not all rings in a PAH will have a sextet assigned to them, and any leftover pi-electrons are kept as double bonds. The Clar structure(s) are defined as the structures with the greatest number of assigned sextets. A basic example is phenanthrene, shown here:
image
Phenanthrene has a total of 14 pi-electrons, which allows the assignment of two sextets and one double bond. Experiments show that the double bond in the Clar structure is the most reactive to addition reactions, and is also the shortest bond in the molecule.

The intent is to use Clar structures to fully replace Kekule structures because Clar structures are easier to enumerate, there are fewer Clar structures than Kekule structures for any PAH, and they provide a more chemically accurate representation of reactivity.
### Implementation

The Clar structure generation algorithm implemented here is based on a optimization algorithm proposed by Hansen and Zheng (J. Math. Chem. 1994). The task is formulated as an integer linear programming problem, where the number of sextets is maximized, subject to the constraint that each atom can only be part of one sextet or one double bond (ie. an atom valency constraint). There are two sets of binary variables, one set corresponding to whether or not each ring contains a sextet, and the other corresponding to whether or not each bond is a double bond. The objective function is simply the sum of the sextet variables.

All Clar structures are enumerated by repeating the optimization after adding a constraint that eliminates the already found solution as a feasible solution. The algorithm terminates when there are no more feasible solutions or when the objective value decreases, indicating that fewer sextets are being assigned.
### To Do

The open source GNU Linear Programming Kit was used to perform the optimization step, in conjunction with the PyGLPK wrapper. Anaconda packages exist for Linux and Mac, but not Windows. This is also complicated by the fact that PyGLPK is not under active development anymore, so it is only compatible with GLPK <v4.52.

I'm planning on making a Windows package for both GLPK and PyGLPK, and have it ready before this is merged into master. However, if anyone has suggestions for a better (faster?) solver/python binding combination, we can also switch to something else. I've looked at PuLP, but it creates input files for various solvers and runs them via command line, which seems sub-optimal. I also looked at PySCIPOpt, which is a wrapper for SCIP (supposedly the fastest non-commercial solver currently), but had difficulty getting it working. It also does not seem to support Windows at this time, and it's distributed under a unique ZIB Academic License.

### DIFF
--- a/rmgpy/molecule/molecule.py
+++ b/rmgpy/molecule/molecule.py
@@ -1621,4 +1621,48 @@ class Molecule(Graph):
         
         return group
 
+    def getAromaticSSSR(self):
+        """
+        Returns the smallest set of smallest aromatic rings
 
+        Identifies rings using `Graph.getSmallestSetOfSmallestRings()`, then uses RDKit to perceive aromaticity.
+        RDKit uses an atom-based pi-electron counting algorithm to check aromaticity based on Huckel's Rule.
+        Therefore, this method identifies "true" aromaticity, rather than simply the RMG bond type.
+
+        The method currently restricts aromaticity to six-membered carbon-only rings. This is a limitation imposed
+        by RMG, and not by RDKit.
+        """
+        cython.declare(rdAtomIndices=dict, aromaticRings=list, aromaticBonds=list)
+        cython.declare(rings=list, ring0=list, i=cython.int, atom1=Atom, atom2=Atom)
+
+        from rdkit.Chem.rdchem import BondType
+
+        AROMATIC = BondType.AROMATIC
+
+        rings = [ring0 for ring0 in self.getSmallestSetOfSmallestRings() if len(ring0) == 6]
+        if not rings:
+            return []
+
+        try:
+            rdkitmol, rdAtomIndices = generator.toRDKitMol(self, removeHs=False, returnMapping=True)
+        except ValueError:
+            return []
+
+        aromaticRings = []
+        for ring0 in rings:
+            aromaticBonds = []
+            # Figure out which atoms and bonds are aromatic and reassign appropriately:
+            for i, atom1 in enumerate(ring0):
+                if not atom1.isCarbon():
+                    # all atoms in the ring must be carbon in RMG for our definition of aromatic
+                    break
+                for atom2 in ring0[i + 1:]:
+                    if self.hasBond(atom1, atom2):
+                        if rdkitmol.GetBondBetweenAtoms(rdAtomIndices[atom1],
+                                                        rdAtomIndices[atom2]).GetBondType() is AROMATIC:
+                            aromaticBonds.append(self.getBond(atom1, atom2))
+            else:  # didn't break so all atoms are carbon
+                if len(aromaticBonds) == 6:
+                    aromaticRings.append(ring0)
+
+        return aromaticRings
--- a/rmgpy/molecule/resonance.py
+++ b/rmgpy/molecule/resonance.py
@@ -38,6 +38,7 @@ import rmgpy.molecule.parser as parser
 
 from .graph import Vertex, Edge, Graph, getVertexConnectivityValue
 from .molecule import Atom, Bond, Molecule
+from .atomtype import AtomTypeError
 import rmgpy.molecule.pathfinder as pathfinder
 
 def populate_resonance_generation_algorithm():
@@ -387,6 +388,54 @@ def generate_isomorphic_isomers(mol):
         index += 1
 
     return isomorphic_isomers
+
+
+def generateClarStructures(mol):
+    """
+    Generate Clar structures for a given molecule.
+
+    Returns a list of :class:`Molecule` objects corresponding to the Clar structures.
+    """
+    cython.declare(output=list, molList=list, newmol=Molecule, asssr=list, bonds=list, solution=list,
+                   y=list, x=list, index=cython.int, bond=Bond, ring=list)
+
+    if not mol.isCyclic():
+        return []
+
+    output = clarOptimization(mol)
+
+    molList = []
+
+    for newmol, asssr, bonds, solution in output:
+
+        # The solution includes a part corresponding to rings, y, and a part corresponding to bonds, x, using
+        # nomenclature from the paper. In y, 1 means the ring as a sextet, 0 means it does not.
+        # In x, 1 corresponds to a double bond, 0 either means a single bond or the bond is part of a sextet.
+        y = solution[0:len(asssr)]
+        x = solution[len(asssr):]
+
+        # Apply results to molecule - double bond locations first
+        for index, bond in enumerate(bonds):
+            if x[index] == 0:
+                bond.order = 'S'
+            elif x[index] == 1:
+                bond.order = 'D'
+            else:
+                raise ValueError('Unaccepted bond value {0} obtained from optimization.'.format(x[index]))
+
+        # Then apply locations of aromatic sextets by converting to benzene bonds
+        for index, ring in enumerate(asssr):
+            if y[index] == 1:
+                clarTransformation(newmol, ring)
+
+        try:
+            newmol.updateAtomTypes()
+        except AtomTypeError:
+            pass
+        else:
+            molList.append(newmol)
+
+    return molList
 
 
 def clarOptimization(mol, constraints=None, maxNum=None):
--- a/rmgpy/molecule/resonance.py
+++ b/rmgpy/molecule/resonance.py
@@ -519,6 +519,29 @@ def clarOptimization(mol, constraints=None, maxNum=None):
     return innerSolutions + [(molecule, asssr, bonds, solution)]
 
 
+def clarTransformation(mol, aromaticRing):
+    """
+    Performs Clar transformation for given ring in a molecule, ie. conversion to aromatic sextet.
+
+    Args:
+        mol             a :class:`Molecule` object
+        aromaticRing    a list of :class:`Atom` objects corresponding to an aromatic ring in mol
+
+    This function directly modifies the input molecule and does not return anything.
+    """
+    cython.declare(bondList=list, i=cython.int, atom1=Atom, atom2=Atom, bond=Bond)
+
+    bondList = []
+
+    for i, atom1 in enumerate(aromaticRing):
+        for atom2 in aromaticRing[i + 1:]:
+            if mol.hasBond(atom1, atom2):
+                bondList.append(mol.getBond(atom1, atom2))
+
+    for bond in bondList:
+        bond.order = 'B'
+
+
 class ILPSolutionError(Exception):
     """
     An exception to be raised when solving an integer linear programming problem if a solution
--- a/rmgpy/molecule/resonance.py
+++ b/rmgpy/molecule/resonance.py
@@ -388,3 +388,141 @@ def generate_isomorphic_isomers(mol):
 
     return isomorphic_isomers
 
+
+def clarOptimization(mol, constraints=None, maxNum=None):
+    """
+    Implements linear programming algorithm for finding Clar structures. This algorithm maximizes the number
+    of Clar sextets within the constraints of molecular geometry and atom valency.
+
+    Returns a list of valid Clar solutions in the form of a tuple, with the following entries:
+        [0] Molecule object
+        [1] List of aromatic rings
+        [2] List of bonds
+        [3] Optimization solution
+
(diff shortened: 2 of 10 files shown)


rwest:
This looks very promising 😃 

As for the GLPK interface,
I'm not sure if any of these python / GLPK interfaces are better or more maintained https://en.wikibooks.org/wiki/GLPK/Python
or https://github.com/cvxopt/cvxopt

Is it too much for [scipy.optimize.linprog](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.linprog.html) to handle?

mliu49:
I started out with `scipy.optimize.linprog`, but it's not able to solve integer LP's. I made a basic linear relaxation and branching method to obtain integer solutions, but ultimately decided to switch to an existing ILP solver because it would be much more robust, and likely faster than my version.

Travis is failing because I haven't added pyglpk to the dependencies list. I'll look into other options a bit more, then add anaconda packages to the RMG channel.

mliu49:
I have a licensing question related to the use of an external optimization solver. It seems that open source licenses are much more complicated than I thought.

GLPK and PyGLPK are released under GPLv3, which is a copyleft license that requires any derivative works to also be licensed under GPLv3. However, RMG is released under the MIT license, which is a permissive license that pretty much allows users to do whatever they want.

From what I could gather from online sources, if RMG depends on GLPK/PyGLPK, ie. it won't work without those libraries (which would happen with this PR), then RMG must also be released under GPLv3. I assume we purposely chose MIT, and don't want to change to GPLv3.

Is this correct, and does that mean I should use a different solver with a more permissive license?

rwest:
I think we chose the MIT license to avoid causing to others the kind of problem that this GPL license is causing us now!
I think it depends how the wrapper works. If we're essentially writing an input file to GLPK and running it as a separate program, then I think we're OK. If we compile and link to libraries, we're maybe not.  
https://www.gnu.org/licenses/gpl-faq.en.html#GPLInProprietarySystem
https://lists.gnu.org/archive/html/help-glpk/2011-01/msg00045.html

mliu49:
I found a different solver that seems very promising: [lpsolve](http://lpsolve.sourceforge.net/5.5/). It's licensed under LGPL, which allows use in other codes without license restrictions. It is ~20% faster than GLPK in preliminary tests. It has a python interface, although it is separate from the core lpsolve code. Anaconda packages exist for lpsolve itself, for all OS's, but not for the python interface, though I've already made a package for linux ([here](https://anaconda.org/mjliu/lpsolve55/files)). I think the next steps will be to test and clean up the new code some more, and make macOS/windows packages.

rmgpy/molecule/molecule.py
```diff
@@ -1621,4 +1621,41 @@ def toGroup(self):
         
         return group
 
+    def getAromaticSSSR(self):
+        """
+        Returns the smallest set of smallest aromatic rings
```
nyee: You should probably mention that this is what rdkit thinks are aromatic rings and maybe make the distinction between with the two types of rings you could have in clar structures?

rmgpy/molecule/resonance.py
```diff
@@ -341,17 +389,187 @@ def generate_isomorphic_isomers(mol):
...
+
+    # Check that we the result contains at least one aromatic sextet
+    if objVal == 0:
+        return []
+
+    # Check that the solution contains the maximum number of sextets possible
+    if maxNum is None:
+        maxNum = objVal  # This is the first solution, so the result should be an upper limit
+    elif objVal < maxNum:
+        raise ValueError('Sub-optimal solution obtained.')
```
nyee: IMO (which is not necessarily that of a good programmer), it is okay to raise built-in error types if you don't ever expect to catch the error because the person running the code can always read the message and tell what happened. 

In this case, I think you should create a new error type like "InvalidILPSolutionError", so that you don't confuse it with actual Value Errors that could occur in other places of your code.

rmgpy/molecule/resonance.py
```diff
@@ -341,17 +389,187 @@ def generate_isomorphic_isomers(mol):
...
+
+    if any([x != 1 and x != 0 for x in solution]):
+        raise ValueError('Non-integer solution obtained from optimization.')
```
nyee: Same as above, you could probably use the same error class.

rmgpy/molecule/resonance.py
```diff
@@ -341,17 +389,187 @@ def generate_isomorphic_isomers(mol):
...
+    new_b = sum(y) - 1
+    if constraints is not None:
+        constraints.append((new_a, new_b))
+    else:
+        constraints = [(new_a, new_b)]
+
+    # Run optimization with additional constraints
+    try:
+        innerSolutions = clarOptimization(mol, constraints=constraints, maxNum=maxNum)
+    except (ValueError, RuntimeError):
```
nyee: Runtime Error is for if the optimization fails?

rmgpy/molecule/resonance.py
```diff
@@ -341,17 +389,187 @@ def generate_isomorphic_isomers(mol):
...
+
+
+def clarOptimization(mol, constraints=None, maxNum=None):
+    """
+    Implements linear programming algorithm for finding Clar structures. This algorithm maximizes the number
+    of Clar sextets within the constraints of molecular geometry and atom valency.
+
+    Method from:
+        Hansen, P.; Zheng, M. The Clar Number of a Benzenoid Hydrocarbon and Linear Programming.
+            J. Math. Chem. 1994, 15 (1), 93–107.
```
nyee: Always good practice to explain what is returned from the function

rmgpy/molecule/molecule.py
KEHANG: more doctring? (e.g., based on aromatic criterion from RDKit) to show this method is done in a more fundamental way of checking aromaticity instead of just by checking `bond.isBenzene()`

KEHANG: also I realize you might not want other people to use this method once the `generateResonanceIsomers()` is finally done. Because then a more elegant way of getting smallest set of smallest aromatic ring is 
1. get a Species object from the target molecule
2. generateResonanceIsomers()
3. loop over all the molecules to get the aromatic isomers
4. examine these aromatic isomers to figure which rings are aromatic

rmgpy/molecule/resonance.py
```diff
@@ -341,17 +389,187 @@ def generate_isomorphic_isomers(mol):
...
+                   y=list, x=list, index=cython.int, bond=Bond, ring=list)
+
+    if not mol.isCyclic():
+        return []
+
+    output = clarOptimization(mol)
+
+    molList = []
+
+    for newmol, SSSR, bonds, solution in output:
```
mliu49: Ok, good to know. I used caps because most usages of getSmallestSetOfSmallestRings() assigns the result to SSSR. I'm fine with changing to lower case sssr though.

rmgpy/molecule/resonanceTest.py
```diff
@@ -314,3 +314,92 @@ def testKekulizedResonanceIsomersFused(self):
...
+
+    def testPhenanthrene(self):
+        """
+        Test phenanthrene, which is a basic case that should work.
+        """
+        mol = Molecule().fromSMILES('C1=CC=C2C(C=CC3=CC=CC=C32)=C1')
+        newmol = generateClarStructures(mol)
+
+        self.assertEqual(len(newmol), 1)
+        self.assertTrue(newmol[0].isAromatic())
```
KEHANG: the test here should be more detailed: e.g, which ring is aromatic, which is not, (only testing overall aromaticity is not enough)

rmgpy/molecule/resonanceTest.py
```diff
@@ -314,3 +314,92 @@ def testKekulizedResonanceIsomersFused(self):
...
+            for bond in bonds:
+                if bond.order == 'D':
+                    pi += 2
+
+            # Count pi electrons in solution
+            y = solution[0:len(SSSR)]
+            x = solution[len(SSSR):]
+            pi_solution = 6 * sum(y) + 2 * sum(x)
+
+            self.assertEqual(pi, pi_solution)
```
KEHANG: would be better to test directly: 
```
self.assertEqual(pi, 10)
self.assertEqual(pi_solution, 10)
```

KEHANG: you may also want to test there's only 1 aromatic ring in nathalene clar structure

rmgpy/molecule/resonance.py
```diff
@@ -341,17 +389,187 @@ def generate_isomorphic_isomers(mol):
...
+    for atom in atoms:
+        bonds.update([atom.bonds[key] for key in atom.bonds.keys() if key.isNonHydrogen()])
+    bonds = list(bonds)
+
+    # Identify exocyclic bonds, and save their bond orders
+    exo = []
+    for bond in bonds:
+        if bond.atom1 not in atoms or bond.atom2 not in atoms:
+            if bond.order == 'D':
+                exo.append(1)
```
mliu49: At this point, exocyclic bonds should only be either single or double bonds. If it were a benzene bond then the bond should be included in the list of aromatic rings.

rmgpy/molecule/resonance.py
mliu49: I will create a new error class.

rmgpy/molecule/resonance.py
```diff
@@ -341,17 +389,187 @@ def generate_isomorphic_isomers(mol):
...
+        innerSolutions = clarOptimization(mol, constraints=constraints, maxNum=maxNum)
+    except (ValueError, RuntimeError):
+        innerSolutions = []
+
+    return innerSolutions + [(molecule, ASSSR, bonds, solution)]
+
+
+def clarTransformation(mol, ring):
+    """
+    Performs Clar transformation for given ring in a molecule, ie. conversion to aromatic sextet.
```
mliu49: Yeah, I'll add more details to the docstring.

rmgpy/molecule/molecule.py
mliu49: Yes, that's a good point. I'll clarify that.

rmgpy/molecule/resonance.py
KEHANG: made it as `def clarTransformation(mol, aromaticRing)` to imply it's only for aromatic ring

rmgpy/molecule/resonance.py
mliu49: Sort of. I don't think the solver itself will raise a RuntimeError, but I was raising it if the optimization did not succeed. I'm also going to replace this with a new error class.

KEHANG:
@mliu49 final comment: could you squash some commits to make clean history and rebase, then I'll merge it.

mliu49:
How much squashing should I do?

nyee:
Yes, I go as far as doing fix ups instead of squashes to the final version. No reason to keep things that are not final drafts IMO.

rmgpy/molecule/resonance.py
```diff
@@ -341,17 +389,211 @@ def generate_isomorphic_isomers(mol):
 
     return isomorphic_isomers
 
-def populate_resonance_generation_algorithm():
+def generateClarStructures(mol):
     """
-    A list with the current set of resonance generation algorithms.
+    Generate Clar structures for a given molecule.
+
+    Returns all Clar structures as a list.
```
nyee: Its easier for other people if you give the object type for each input/output. For example you would say the output is a list of of :class:Molecule objects representing the Clar structures.

mliu49:
Yeah, they licensing issue was solved by switching to lpsolve. It's licensed under the Lesser GPL, which doesn't place any restrictions on use.

PR #807, opened 2016-10-31
Improve error handling for wrong number of reverse rxns

mliu49:
Give a different message when more than one reverse reactions are found, to differentiate from cases where no reverse reactions are found. Also log the reverse reactions for easier debugging.


mliu49:
Incorporated into #813.

PR #808, opened 2016-11-01
Fix lone pair thermo

### LINKED ISSUE #792: Discontinuities in heat capacity curves?
Search thermo for this C3H2 species  `InChI=1S/C3H2/c1-2-3-1/h1H2`

```
1 C u0 p0 c0 {2,S} {3,S} {4,S} {5,S}
2 C u0 p1 c0 {1,S} {3,S}
3 C u0 p1 c0 {1,S} {2,S}
4 H  u0 p0 c0 {1,S}
5 H  u0 p0 c0 {1,S}
```

eg. by clicking [here](http://rmg.mit.edu/database/thermo/molecule/1%20C%20u0%20p0%20c0%20%7B2,S%7D%20%7B3,S%7D%20%7B4,S%7D%20%7B5,S%7D%0A2%20C%20u0%20p1%20c0%20%7B1,S%7D%20%7B3,S%7D%0A3%20C%20u0%20p1%20c0%20%7B1,S%7D%20%7B2,S%7D%0A4%20H%20u0%20p0%20c0%20%7B1,S%7D%0A5%20H%20u0%20p0%20c0%20%7B1,S%7D%0A) 
and you will notice a few things.
1.  the picture looks like cyclopropane, C3H6, because it does not draw the lone pairs,
2. the heat capacity, entropy, and gibbs free energy, curves all have big discontinuities at 1500K.

screen shot 2016-10-21 at 2 11 06 pm

bslakman:
This pull request fixes #792. Now RMG will convert lone pairs to biradicals for the purpose of finding thermo via group additivity. 

There are no Benson groups for the lone pair atoms, so we end up falling up the thermo group tree to inappropriate groups since they are neither radicals nor are they attached to hydrogens. This is the same thing we decided to do for solvation in #364, since there are no Platts groups for the lone pair species. The better thing would be to add groups for these or put some of these molecules in a library, but for Carbon, we may not encounter many of these molecules anyway. 

Now there is no Cp discontinuity when searching for the problematic molecule - this is what I get when I run the website with this version of RMG-Py for cyclopropane with a lone pair on two of the carbons:
cp

### DIFF
--- a/rmgpy/data/solvation.py
+++ b/rmgpy/data/solvation.py
@@ -701,55 +701,7 @@ class SolvationDatabase(object):
         soluteData.comment = "Average of {0}".format(" and ".join(comments))
 
         return soluteData
-   
-    def transformLonePairs(self, molecule):
-        """
-        Changes lone pairs in a molecule to two radicals for purposes of finding
-        solute data via group additivity. Transformed for each atom based on valency.
-        """
-        saturatedStruct = molecule.copy(deep=True)
-        addedToPairs = {}
 
-        for atom in saturatedStruct.atoms:
-            addedToPairs[atom] = 0
-            if atom.lonePairs > 0:
-                charge = atom.charge # Record this so we can conserve it when checking
-                bonds = saturatedStruct.getBonds(atom)
-                sumBondOrders = 0
-                for key, bond in bonds.iteritems():
-                    if bond.order == 'S': sumBondOrders += 1
-                    if bond.order == 'D': sumBondOrders += 2
-                    if bond.order == 'T': sumBondOrders += 3
-                    if bond.order == 'B': sumBondOrders += 1.5 # We should always have 2 'B' bonds (but what about Cbf?)
-                if atomTypes['Val4'] in atom.atomType.generic: # Carbon, Silicon
-                    while(atom.radicalElectrons + charge + sumBondOrders < 4):
-                        atom.decrementLonePairs()
-                        atom.incrementRadical()
-                        atom.incrementRadical()
-                        addedToPairs[atom] += 1
-                if atomTypes['Val5'] in atom.atomType.generic: # Nitrogen
-                    while(atom.radicalElectrons + charge + sumBondOrders < 3):
-                        atom.decrementLonePairs()
-                        atom.incrementRadical()
-                        atom.incrementRadical()
-                        addedToPairs[atom] += 1
-                if atomTypes['Val6'] in atom.atomType.generic: # Oxygen, sulfur
-                    while(atom.radicalElectrons + charge + sumBondOrders < 2):
-                        atom.decrementLonePairs()
-                        atom.incrementRadical()
-                        atom.incrementRadical()
-                        addedToPairs[atom] += 1
-                if atomTypes['Val7'] in atom.atomType.generic: # Chlorine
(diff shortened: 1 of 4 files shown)


nyee:
I have an alternate approach, even though you already coded all this up... You could make thermo entries in /thermo/group.py for the atoms with extra lone pairs and then just have them point towards the bi-radicals. I think this would be easier to "reverse" if we ever get better data. If you consider the database easier to read than the code (questionable), then it is probably more transparent than pre-processing under the hood.

bslakman:
Hmm.. so actually I'd have to point them towards the species without any radicals? Because there is no data for biradicals in group.py (I think) because we do the stable species and then lookup the HBI correction in radical.py . Or we make a group in radical.py that's a lone pair. But we'd still have to know to do the HBI for lone pairs in the first place, which requires changing the actual code. What do you think?

nyee:
Okay yeah, what I said earlier doesn't make sense. I have one minor suggestion and one other possibly conflicting feature. 

1. You could change the way you generate bond order to use the new methods recently added in these #767. 

2. The way you are handling lone pairs will probably conflict with what @alongd and I have been compiling for new atom types on the [NatomType branch](https://github.com/ReactionMechanismGenerator/RMG-Py/tree/NAtomType). Specifically we have created new atomtypes defined by their number of lone pairs, and associated thermo group (not HBI) entries. Off the top of my head, I know the new "N1s" atomtype will be handled in both our method and this change. We might need a bit of time to sort out how best to just handle it once.

zjburas:
There is at least one system where carbons with lone pairs (singlet carbenes) are important:

In propargyl radical recombination and rearrangement to make benzene (thought to be the major route from small molecules to aromatics) there are two singlet carbene intermediates on the surface that are critical  to the overall pathway:

image

and

image

I would like RMG to find this pathway on its own, but it can't do that if the GAV thermo estimates for the above species are bad. Currently, as discussed by @bslakman and @nyee above, GAV thermo for singlet carbenes is way off because although HBI corrections exist in the database for singlet carbenes, they are never found by the current algorithm. So, for example, the thermo of a simple singlet carbene such as the following:

image

is estimated by the following groups:

image
+
image+some Gauche Corrections

Note two things:

1. The carbon with the lone pair is saturated with Cs's instead of H's
2. No HBI correction is applied for the singlet carbene

Effectively, RMG is estimating the thermo for the singlet carbene above to be the same as for the following stable molecule (neopentane):

image

Clearly this GAV calculation overestimates how stable the singlet carbene actually is. 

In contrast, the GAV calculation for triplet carbenes, such as:

image

is much more accurate (compared to existing library calculations) because HBI corrections are accessed that have data.

@bslakman 's suggestion of simply treating singlet carbenes as triplet carbenes (converting a lone pair into two unpaired electrons) may be a reasonable approximation. Certainly it would be better than how singlet carbenes are currently treated. However, @nyee thinks it might be possible to keep singlet and triplet carbenes distinct, and have both access their respective HBI corrections. I think this would be the preferred approach.

Once the HBI corrections for singlet carbenes can be accessed, I can add in some new corrections based on the calculated thermo for the two cyclic species above.

mliu49:
Closing because this is out of date and most likely incompatible with ReactionMechanismGenerator/RMG-database#206 and #1206.

Issue #809, opened 2016-11-05
Missing the species name when RMG's unable to determine atom type


alongd:
Recent changes to nitrogen atom type representations made in RMG (nitrogen branch) predicted a new N atom type with 2 lone pairs and 2 single bonds (e.g., `[NH-][S+]=C`, `[NH-][N+]#C`) with the following error in `rmgpy.molecule.atomtype.getAtomType`:
```
Traceback (most recent call last):
  File "D:\Code\RMG-Py\rmgpy\scoop_framework\util.py", line 112, in __call__
    return self.myfn(*args, **kwargs)
  File "D:\Code\RMG-Py\rmgpy\rmg\react.py", line 96, in reactMolecules
    rxns = family.generateReactions(molecules)
  File "D:\Code\RMG-Py\rmgpy\data\kinetics\family.py", line 1308, in generateReactions
    reactionList.extend(self.__generateReactions(reactants, forward=True))
  File "D:\Code\RMG-Py\rmgpy\data\kinetics\family.py", line 1460, in __generateReactions
    productStructures = self.__generateProductStructures(reactantStructures, [mapA, mapB], forward)
  File "D:\Code\RMG-Py\rmgpy\data\kinetics\family.py", line 1200, in __generateProductStructures
    productStructures = self.applyRecipe(reactantStructures, forward=forward)
  File "D:\Code\RMG-Py\rmgpy\data\kinetics\family.py", line 1166, in applyRecipe
    struct.update()
  File "rmgpy\molecule\molecule.py", line 742, in rmgpy.molecule.molecule.Molecule.update (build\Release\pyrex\rmgpy\molecule\molecule.c:14864)
    self.updateAtomTypes()
  File "rmgpy\molecule\molecule.py", line 938, in rmgpy.molecule.molecule.Molecule.updateAtomTypes (build\Release\pyrex\rmgpy\molecule\molecule.c:18011)
    atom.atomType = getAtomType(atom, atom.edges)
  File "rmgpy\molecule\atomtype.py", line 479, in rmgpy.molecule.atomtype.getAtomType (build\Release\pyrex\rmgpy\molecule\atomtype.c:6464)
  File "rmgpy\molecule\atomtype.py", line 523, in rmgpy.molecule.atomtype.getAtomType (build\Release\pyrex\rmgpy\molecule\atomtype.c:6338)
[...]

PR #810, opened 2016-11-06
Log the species RMG crashes on if atomType doesn't exist

alongd:
Logging the species RMG crashes on when an unknown atomType is discovered. Important for debugging when modifying atom types, see issue #809.
For now I only saw a need to add this fix to molecule.py, and not to adjlist.py (where `getAtomType` is also being called).

### DIFF
--- a/rmgpy/molecule/molecule.pxd
+++ b/rmgpy/molecule/molecule.pxd
@@ -198,7 +198,7 @@ cdef class Molecule(Graph):
 
     cpdef double calculateCpInf(self) except -1
     
-    cpdef updateAtomTypes(self)
+    cpdef updateAtomTypes(self, logSpecies=?)
     
     cpdef bint isRadical(self) except -2
 
--- a/rmgpy/molecule/molecule.py
+++ b/rmgpy/molecule/molecule.py
@@ -939,7 +939,7 @@ class Molecule(Graph):
                 atom.atomType = getAtomType(atom, atom.edges)
             except:
                 logging.error("Problematic species: {}".format(self))
-                atom.atomType = getAtomType(atom, atom.edges)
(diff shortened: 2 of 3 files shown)


mliu49:
You should be able to do `from .atomtype import AtomTypeError`, then have `except AtomTypeError:` and `raise`.

Also, the resonance module frequently calls `updateAtomTypes()`, with the expectation that encountering an AtomTypeError indicates an infeasible resonance structure has been generated. It might be undesirable to have all of those species printed as well. Maybe you could add a boolean argument to indicate whether or not to print the species? Then we could update the calls from resonance.py later to turn off the logging.

alongd:
Thanks @mliu49 , let me know if you have more comments

rwest:
Based on max's comment it sounds like this is currently in a place where we don't always want logging. Could we move it further up the stack trace to somewhere where the error is always an error that's worth logging? (Rather than having to pass in extra arguments about whether we want logging or not). 
(This comment made without looking at the code, so treat accordingly)

mliu49:
I think you need to modify molecule.pxd as well, since the module is cythonized. I can work with you on this tomorrow if you want.

alongd:
I added a flag set with a default value for presenting the log info. If we want to call the func without logging, just send 0 in addition to self. Will that be OK, or do you think we still need to change?

mliu49:
I pushed a commit so that we only catch an AtomTypeError, and we also print the adjacency list of the problematic molecule.

I think it is ok to default to printing the extra information, because usually it will be helpful for debugging. I think resonance structure generation is the only place where an AtomTypeError is acceptable.

alongd:
Will add these commits into a more comprehensive PR.

PR #814, opened 2016-11-09
Fix bug in addImplicitBenzene regarding step 3 sorting by connectivity

nyee:


rmgpy/molecule/group.py
```diff
@@ -1368,20 +1368,22 @@ def sortByConnectivity(atomList):
             Returns: a sorted list of atoms where each atom is connected to a previous
```
KEHANG: Can you add a few unittests for just this method `sortByConnectivity` rather than other higher level methods in `groupTest.py`; also choose at least examples that fall into different cases, e.g. one can be whole molecule is connected by `B` bond, the other can be divided into several sub-molecules.

nyee: Looked into it a little bit, but I don't think I can without changing the whole code structure. Because `sortByConnectivity()` is a function underneath `addImplicitBenzene() ` , I cannot seem to access sortByConnectivity except in the scope of addImplicitBenzene.

nyee:
@KEHANG, after thinking about it for a bit, I moved some of the helper functions to be methods for :class:Group instead, so that I could add unit tests. A few of the helper functions I decided should stay private, and are hopefully simple enough to not need unit tests.

PR #820, opened 2016-11-15
Reduction Documentation

goldmanm:
RMG has a functioning (though I don't think optimal) method to reduce final mechanisms. 

I modified an error message to help me use the algorithm, and I wrote a 'how to use' in the documentation.

### DIFF
--- a/rmgpy/reduction/reduction.py
+++ b/rmgpy/reduction/reduction.py
@@ -268,8 +268,11 @@ def searchTargetIndex(targetLabel, reactionModel):
     """
     Searches for the Species object in the core species
     of the reaction that has the same label as the parameter string.
-    """
+    reactionModel must be of class CoreEdgeReactionModel
 
+    Has known issues dealing with duplicate labels. See reductionTest.py
+    for a unittest of this issue.
+    """
     for i, spc in enumerate(reactionModel.core.species):
         if spc.label == targetLabel:
             return i
--- a/documentation/source/users/rmg/modules/reduction.rst
+++ b/documentation/source/users/rmg/modules/reduction.rst
@@ -1,0 +1,42 @@
+.. _reduction:
+
+***********************************
+Reducing an RMG output
+***********************************
+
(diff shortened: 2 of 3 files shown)


rmgpy/reduction/reduction.py
```diff
@@ -268,13 +268,20 @@ def searchTargetIndex(targetLabel, reactionModel):
     """
     Searches for the Species object in the core species
     of the reaction that has the same label as the parameter string.
-    """
 
+    This method does not differentiate between multiple species
```
KEHANG: Good catch! It would be better to use unittest to show the issue, doc-string might confuse users.

goldmanm: Great idea. I wrote a unittest for the issue. Should I leave the failing test or comment with @work_in_progress?

KEHANG:
can you use `git squash` to clean up a bit for the history and rebase? I can help you do that if you need

rmgpy/reduction/reduction.py
```diff
@@ -268,13 +268,18 @@ def searchTargetIndex(targetLabel, reactionModel):
...
+    for a unittest of this issue.
+    """
+    spc_labels = []
     for i, spc in enumerate(reactionModel.core.species):
         if spc.label == targetLabel:
             return i
+        spc_labels.append(spc.label)
 
-    raise Exception('{} could not be found...'.format(targetLabel))
+    raise Exception('{} could not be found in species labels {}.'.format(targetLabel,spc_labels))
```
KEHANG: I don't think it's a good way to print out all the species labels in the core; some mechanism can be pretty big and printing out is not so helpful.

If we are concerned about the situation where two species have same labels, maybe we should use isomorphism check instead. An idea would be to require user to feed SMILES instead of labels in the reduction_input.py

KEHANG: sure, let's revert to the previous exception, and focus on the documentation part which is actually well written.

PR #823, opened 2016-11-16
Resonance Part 2: Resonance Structure Generation for Aromatic Species

mliu49:
This is the second part of the resonance structure generation overhaul, and it builds on top of #805. This PR significantly changes how aromatic species are handled during resonance structure generation. Non-aromatic species should not be affected and are treated similarly to how they were before.

### Molecule Analysis
The first major change is the addition of a new method `analyzeMolecule()`, which identifies important features with the goal of using those features to determine how resonance structure generation is carried out. It identifies features such as important elements, radical character, and aromaticity. The analysis significantly affects treatment of aromatics. The affects non-aromatics to a small extent, by reducing the list of algorithms that are applied, through `populateResonanceAlgorithms()`.

### Aromaticity
Aromaticity detection is really difficult, which is why we rely on RDKit to do most of the heavy lifting. However, RDKit is not always correct. There are two main situations where RDKit is wrong:
1) False positive: If the species has exocyclic double bonds, the pi-electron counting algorithm used by RDKit fails, eg. dimethylene cyclohexadiene
image
2) False negative: If the species has a delocalized radical, it may have an aromatic resonance structure which RDKit won't recognize, eg. benzyl radical
image

False positives were relatively straightforward to handle, since RMG does not allow conversion of the ring bonds to benzene bonds. For false negatives, I ultimately decided to improve `generateAromaticResonanceIsomers()` rather than the main `generateResonanceIsomers()` method. The approach I took was to apply `generateAdjacentResonanceIsomers()`, then iterate through all of the results and find the structures with the most aromatic rings. Unfortunately, this is a bit tedious because of the repeated calls to `getAromaticSSSR()`.

However, it has the benefit of consistently generating the same aromatic structure regardless of which resonance form is inputted. There does remain a possibility of generating more than one aromatic structure, in cases where multiple positions of the radical give the same number of aromatic rings, eg.
image image
Both of these are generated given any of its resonance structures, but it's done consistently!

### Other Resonance Structures
For aromatic species, the process is now to first generate the aromatic resonance structure(s), which gives a consistent starting point regardless of the input. Then, the process diverges based on the features of the molecule.
- For radical polycyclic aromatic species, Kekule structures are generated in order to generate adjacent resonance structures. The resulting structures are then used for Clar structure generation. After all three steps, any non-aromatic structures are removed, under the assumption that they are not important resonance contributors.
- For radical monocyclic aromatic species, Kekule structures are generated along with adjacent resonance structures. All are kept regardless of aromaticity because the radical is more likely to delocalize into the ring. Ideally, the aromatic structures would be weighted more than the non-aromatic ones, but that is for future development.
- For stable polycyclic aromatic species, Clar structures are generated
- For stable monocyclic aromatic species, Kekule structures are generated

The last point is mainly to maintain compatibility with existing code. Once reactivity of benzene bonds is developed (the next stage of this project), then there will no longer be any point in keeping kekulized structures for stable monocyclic aromatic species.

### To Do
- More testing would be best, given the role of resonance structures. Checking the results of RMG-tests will be very important for this PR, to make sure it doesn't break anything.
- Suggestions for additional unit tests are welcome
- Need to decide what to do with two tests that were marked as @work_in_progress in thermoTest (9dbfe1a5582f6b23c495cf13a2111c24f34720d2). They are related to a pyrene radical resonance structure that has been determined to be invalid. @KEHANG


rwest:
@cfgoldsmith this and #805 are the pull requests I told you about...

mliu49:
Replaced by #835, which compares with an official branch instead of branch in my fork to enable triggering of RMG-tests.

PR #835, opened 2016-12-07
Resonance Part 2: Resonance Structure Generation for Aromatic Species

mliu49:
This is the second part of the resonance structure generation overhaul, and it builds on top of #805. This PR significantly changes how aromatic species are handled during resonance structure generation. Non-aromatic species should not be affected and are treated similarly to how they were before.

### Molecule Analysis
The first major change is the addition of a new method `analyzeMolecule()`, which identifies important features with the goal of using those features to determine how resonance structure generation is carried out. It identifies features such as important elements, radical character, and aromaticity. The analysis significantly affects treatment of aromatics. The affects non-aromatics to a small extent, by reducing the list of algorithms that are applied, through `populateResonanceAlgorithms()`.

### Aromaticity
Aromaticity detection is really difficult, which is why we rely on RDKit to do most of the heavy lifting. However, RDKit is not always correct. There are two main situations where RDKit is wrong:
1) False positive: If the species has exocyclic double bonds, the pi-electron counting algorithm used by RDKit fails, eg. dimethylene cyclohexadiene
image
2) False negative: If the species has a delocalized radical, it may have an aromatic resonance structure which RDKit won't recognize, eg. benzyl radical
image

False positives were relatively straightforward to handle, since RMG does not allow conversion of the ring bonds to benzene bonds. For false negatives, I ultimately decided to improve `generateAromaticResonanceIsomers()` rather than the main `generateResonanceIsomers()` method. The approach I took was to apply `generateAdjacentResonanceIsomers()`, then iterate through all of the results and find the structures with the most aromatic rings. Unfortunately, this is a bit tedious because of the repeated calls to `getAromaticSSSR()`.

However, it has the benefit of consistently generating the same aromatic structure regardless of which resonance form is inputted. There does remain a possibility of generating more than one aromatic structure, in cases where multiple positions of the radical give the same number of aromatic rings, eg.
image image
Both of these are generated given any of its resonance structures, but it's done consistently!

### Other Resonance Structures
For aromatic species, the process is now to first generate the aromatic resonance structure(s), which gives a consistent starting point regardless of the input. Then, the process diverges based on the features of the molecule.
- For radical polycyclic aromatic species, Kekule structures are generated in order to generate adjacent resonance structures. The resulting structures are then used for Clar structure generation. After all three steps, any non-aromatic structures are removed, under the assumption that they are not important resonance contributors.
- For radical monocyclic aromatic species, Kekule structures are generated along with adjacent resonance structures. All are kept regardless of aromaticity because the radical is more likely to delocalize into the ring. Ideally, the aromatic structures would be weighted more than the non-aromatic ones, but that is for future development.
- For stable polycyclic aromatic species, Clar structures are generated
- For stable monocyclic aromatic species, Kekule structures are generated

The last point is mainly to maintain compatibility with existing code. Once reactivity of benzene bonds is developed (the next stage of this project), then there will no longer be any point in keeping kekulized structures for stable monocyclic aromatic species.

### To Do
- More testing would be best, given the role of resonance structures. Checking the results of RMG-tests will be very important for this PR, to make sure it doesn't break anything.
- Suggestions for additional unit tests are welcome
- Need to decide what to do with two tests that were marked as @work_in_progress in thermoTest (9dbfe1a5582f6b23c495cf13a2111c24f34720d2). They are related to a pyrene radical resonance structure that has been determined to be invalid. @KEHANG


mliu49:
I'm going to go ahead and squash some commits here as well, to make reviewing easier.

KEHANG:
Just to give myself a reminder to run PDD case using this branch to see any changes.

nyee:
The other day we were talking about how this will disable some bonds from reacting in larger polycyclics because all clar structures may be represent that particular bond as benzene.

I spoke with @zjburas today and confirmed that this pull request will not interfere with the mechanism he is trying to generate as he only needs to get the naphthalene, which will have all its bonds reactive with its two Clar structures.

rmgpy/molecule/moleculeTest.py
```diff
@@ -1278,124 +1278,6 @@ def testSaturateAromaticRadical(self):
...
-        pyrene_rad2 = Molecule().fromSMILES('[C]1C=C2C=CC=C3C=CC4=CC=CC=1C4=C23')
-        for isomer in pyrene_rad2.generateResonanceIsomers():
-            if pyrene_rad.isIsomorphic(isomer):
-                break
-        else:  # didn't break
-            self.fail("{} isn't isomorphic with any aromatic forms of {}".format(
-                            pyrene_rad.toSMILES(),
-                            pyrene_rad2.toSMILES()
-                        ))
-
```
goldmanm: There is likely a reason for this change. for my knowledge though, what motivated the removal of these? Took a long time to run, likely redundant, were failing, etc?

goldmanm: I see they were moved to the resonance test section

mliu49:
I've added unit tests for isArylRadical(), and made the appropriate changes following #864.

rmgpy/molecule/resonance.py
```diff
@@ -130,34 +130,107 @@ def analyzeMolecule(mol):
 
 def generateResonanceIsomers(mol):
     """
-    Generate and return all of the resonance isomers of this molecule.
+    Generate and return all of the resonance structures for the input molecule.
```
KEHANG: I find this summary in a previous PR is helpful, please somehow add it to the docstring?:

### Other Resonance Structures
For aromatic species, the process is now to first generate the aromatic resonance structure(s), which gives a consistent starting point regardless of the input. Then, the process diverges based on the features of the molecule.
- For radical polycyclic aromatic species, Kekule structures are generated in order to generate adjacent resonance structures. The resulting structures are then used for Clar structure generation. After all three steps, any non-aromatic structures are removed, under the assumption that they are not important resonance contributors.
- For radical monocyclic aromatic species, Kekule structures are generated along with adjacent resonance structures. All are kept regardless of aromaticity because the radical is more likely to delocalize into the ring. Ideally, the aromatic structures would be weighted more than the non-aromatic ones, but that is for future development.
- For stable polycyclic aromatic species, only Clar structures are generated
- For stable monocyclic aromatic species, Kekule structures are generated

rmgpy/molecule/resonanceTest.py
```diff
@@ -272,6 +393,36 @@ def testMultipleKekulizedResonanceIsomers(self):
...
+        molList = generateResonanceIsomers(molecule)
+        self.assertEqual(len(molList), 6, "Expected 6 resonance structures, but generated {0}.".format(len(molList)))
+        aromatic = 0
+        for mol in molList:
+            if mol.isAromatic():
+                aromatic += 1
+        self.assertEqual(aromatic, 1, "Should only have 1 aromatic resonance structure")
+
     @work_in_progress
     def testKekulizedResonanceIsomersFused(self):
```
KEHANG: I find `generateResonanceStructures` is not fully tested. Since you have four categories (whether polyaromatic/monoaromatic * whether radical/stable), seems most cases are monocyclics except this `work_in_progress` one (even this one only covers stable polyaromatic category)

mliu49:
I think I've addressed your comments, @KEHANG. Let me know what you think, and then I can squash the fixups.

Also, while changing isArylRadical, I discovered a duplicated method in Molecule, and added a commit to remove it.

KEHANG:
@mliu49 cool, it looks good. Please go ahead and finalize.

KEHANG:
@mliu49 merged, congrats!

PR #838, opened 2016-12-09
Update charge

### LINKED ISSUE #837: R_Addition_COm family not functional
As title says, this family doesn't seem to work at all. You can see yourself using the following 
[input.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/643010/input.txt) for generateReactions.py

It turns out the problem is that we [do not update the charge on all the atoms after applying the recipe](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/7b4de65bb814440b9c9d848f0c983ca41a3ee066/rmgpy/molecule/molecule.py#L736). In fact, we [selectively only update charge when we modify the number of lone pairs ](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/7b4de65bb814440b9c9d848f0c983ca41a3ee066/rmgpy/molecule/molecule.py#L347). Depending on what order the recipe actions are written in, we could end up with an incorrect charge on an atom. 

In the case of R_Addition_COm, because [we lose a lone pair on the carbon atom and update the charge before forming a bond with *2](https://github.com/ReactionMechanismGenerator/RMG-database/blob/491502314284e8dde746d90c9c7647f504bbf736/input/kinetics/families/R_Addition_COm/groups.py#L15), we always end up with a negative charge on the carbon. In 1,2_Insertion_CO, [we lose the lone pair as the last step of the recipe](https://github.com/ReactionMechanismGenerator/RMG-database/blob/491502314284e8dde746d90c9c7647f504bbf736/input/kinetics/families/1%2C2_Insertion_CO/groups.py#L20), so we get the correct charge and the family works as intended.

nyee:
Added a step for updating the charge of all atoms after we have applied reaction recipes. This pull request fixes #837, and possibly other families where charge can change.


mliu49:
Removing the `updateCharge()` calls from increment and decrement lone pair might not be safe to do. Those functions are also used in `rmgpy.molecule.resonance`, where there is no subsequent call to `Molecule.update()`. (The resonance algorithms currently only call `Molecule.updateAtomTypes()`) There is another usage in SolvationDatabase, I think related to calculating the solvation correction.

Perhaps we should just leave them there?

nyee:
I'm fine with that too. We did run the RMG-tests, so the original hope was that it would be caught, but let's just leave it in there.

rmgpy/molecule/atomtype.py
```diff
@@ -370,7 +370,7 @@ def getFeatures(self):
 atomTypes['Cs'  ].setActions(incrementBond=['Cd','CO','Cs'], decrementBond=[],               formBond=['Cs'],        breakBond=['Cs'],        incrementRadical=['Cs'],   decrementRadical=['Cs'],   incrementLonePair=['Cs'],      decrementLonePair=['Cs'])
```
mliu49: This is somewhat unrelated, but should Cs be able to increment to CS?

nyee: Yeah it looks like just a typo to me. It certainly shouldn't be incrementing to Cs.

Issue #737, opened 2016-08-03
RMG should also output Cantera input files


alongd:
A suggestion to generate a Cantera input file (.cti) based on the annotated and non-annotated RMG-generated Chemkin files. The ck2cti script already exists, it doesn't treat our annotations so great, but overall does a good job.
Should this be a choice in the RMG input file (and if yes, then be also set to default?), or simply always enforced?

rwest:
👍  I think we should always generate a .cti file.

alongd:
How about creating a "model" directory, instead of "chemkin" with files:
chem.inp
chem.cti
chem_annotated.inp
chem_annotated.cti
chem_edge.inp (optional)
chem_edge.cti (optional)
species_dictionary.txt
species_edge_dictionary.txt (optional)
tran.dat
[cti files will include transport data I guess, if not than a separate Cantera transport file]

and all the chemxxxx.inp and chem_edgexxxx.inp will be relocated to a subfolder named "model_generation" inside "model", or something of that sort?

This should all be done after RMG has finished generating the model, I guess.

PR #839, opened 2016-12-14
Generate Cantera files using ck2cti

alongd:
At the end of an RMG run, save both chem.inp and chem_annotated.inp
as respective .cti files in a new `cantera` folder. See issue #737

### DIFF
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -66,6 +66,7 @@ from rmgpy.qm.main import QMDatabaseWriter
 from rmgpy.stats import ExecutionStatsWriter
 from rmgpy.thermo.thermoengine import submit
 from rmgpy.tools.sensitivity import plotSensitivity
+from cantera import ck2cti
 
 ################################################################################
 
@@ -681,6 +682,13 @@ class RMG(util.Subject):
                 )
                 
                 plotSensitivity(self.outputDirectory, index, reactionSystem.sensitiveSpecies)
+
+        # generate Cantera files chem.cti & chem_annotated.cti in a designated `cantera` output folder
+        try:
+            self.generateCanteraFiles(os.path.join(self.outputDirectory, 'chemkin', 'chem.inp'))
+            self.generateCanteraFiles(os.path.join(self.outputDirectory, 'chemkin', 'chem_annotated.inp'))
+        except EnvironmentError:
+            logging.error('Could not generate Cantera files due to EnvironmentError. Check read\write privileges in output directory.')
                 
         # Write output file
         logging.info('')
@@ -691,6 +699,25 @@ class RMG(util.Subject):
         logging.info('The final model edge has %s species and %s reactions' % (edgeSpec, edgeReac))
         
         self.finish()
+
+    def generateCanteraFiles(self, chemkinFile, **kwargs):
+        """
+        Convert a chemkin mechanism chem.inp file to a cantera mechanism file chem.cti
+        and save it in the cantera directory
+        """
+        transportFile = os.path.join(os.path.dirname(chemkinFile), 'tran.dat')
+        fileName = os.path.splitext(os.path.basename(chemkinFile))[0] + '.cti'
+        outName = os.path.join(self.outputDirectory, 'cantera', fileName)
+        canteraDir = os.path.dirname(outName)
+        try:
+            os.makedirs(canteraDir)
+        except OSError:
+            if not os.path.isdir(canteraDir):
+                raise
+        if os.path.exists(outName):
+            os.remove(outName)
+        parser = ck2cti.Parser()
+        parser.convertMech(chemkinFile, transportFile=transportFile, outName=outName, quiet=True, permissive=True, **kwargs)
 
     def initializeReactionThresholdAndReactFlags(self):
         numCoreSpecies = len(self.reactionModel.core.species)


rmgpy/rmg/main.py
```diff
@@ -709,14 +734,14 @@ def updateReactionThresholdAndReactFlags(self, rxnSysUnimolecularThreshold=None,
         numCoreSpecies = len(self.reactionModel.core.species)
         prevNumCoreSpecies = len(self.unimolecularReact)
         stale = True if numCoreSpecies > prevNumCoreSpecies else False
-        
-        
+
+
```
KEHANG: can you remove those empty line changes from the commit? there are several occurrance down the file as well

rmgpy/rmg/main.py
```diff
@@ -689,9 +697,26 @@ def execute(self, **kwargs):
         coreSpec, coreReac, edgeSpec, edgeReac = self.reactionModel.getModelSize()
         logging.info('The final model core has %s species and %s reactions' % (coreSpec, coreReac))
         logging.info('The final model edge has %s species and %s reactions' % (edgeSpec, edgeReac))
-        
         self.finish()
 
+    def generateCanteraFiles(self, chemkinFile, **kwargs):
+        """
```
KEHANG: I know it's probably not straightforward to write a unit test for this, but have you tested locally?

rmgpy/rmg/main.py
```diff
@@ -689,9 +697,26 @@ def execute(self, **kwargs):
         coreSpec, coreReac, edgeSpec, edgeReac = self.reactionModel.getModelSize()
         logging.info('The final model core has %s species and %s reactions' % (coreSpec, coreReac))
         logging.info('The final model edge has %s species and %s reactions' % (edgeSpec, edgeReac))
-
```
KEHANG: remove the line change here from this commit?

rmgpy/rmg/main.py
alongd: Yes, I tested for several systems (CH$-air, ethylamine oxidation), and it works fine

rmgpy/rmg/main.py
alongd: Sure, the deleted tabs were removed from the commit

rmgpy/rmg/main.py
```diff
@@ -692,6 +700,24 @@ def execute(self, **kwargs):
...
+        and save it in the cantera directory
+        """
+        transportFile = os.path.join(os.path.dirname(chemkinFile), 'tran.dat')
+        fileName = os.path.splitext(os.path.basename(chemkinFile))[0] + '.cti'
+        outName = os.path.join(self.outputDirectory, 'cantera', fileName)
+        canteraDir = os.path.dirname(outName)
+        try:
+            os.stat(canteraDir)
+        except:
+            os.mkdir(canteraDir)
```
mliu49: An alternative way to only create a directory if it doesn't exist would be to try to create it first, then catch if it fails:
```
try: 
    os.makedirs(path)
except OSError:
    if not os.path.isdir(path):
        raise
```
This will directly create the directory when it doesn't exist (most cases), do nothing if the directory exists, and raise an error if os.mkdir fails for other reasons (eg. if a file exists with the same name).

This method will also prevent a "race condition" where an error can occur if the directory is somehow created in between the call to os.stat and the call to os.mkdir.

rmgpy/rmg/main.py
```diff
@@ -681,6 +682,13 @@ def execute(self, **kwargs):
                 )
                 
                 plotSensitivity(self.outputDirectory, index, reactionSystem.sensitiveSpecies)
+
+        # generate Cantera files chem.cti & chem_annotated.cti in a designated `cantera` output folder
+        try:
+            self.generateCanteraFiles(os.path.join(self.outputDirectory, 'chemkin', 'chem.inp'))
+            self.generateCanteraFiles(os.path.join(self.outputDirectory, 'chemkin', 'chem_annotated.inp'))
+        except:
```
mliu49: It's generally not a good idea to catch all exceptions in a try, except block because you may catch errors that you weren't expecting or are actual bad errors that should be fixed.

I think in this case, catching EnvironmentError might be ok, which includes any issues with reading/writing to the filesystem. Any potential issues with RMG or Cantera would still be raised, which we would then need to fix.

alongd: Agree, updated

rmgpy/rmg/main.py
```diff
@@ -681,6 +682,15 @@ def execute(self, **kwargs):
...
                 plotSensitivity(self.outputDirectory, index, reactionSystem.sensitiveSpecies)
+
+        # generate Cantera files chem.cti & chem_annotated.cti in a designated `cantera` output folder
+        try:
+            self.generateCanteraFiles(os.path.join(self.outputDirectory, 'chemkin', 'chem.inp'))
+            self.generateCanteraFiles(os.path.join(self.outputDirectory, 'chemkin', 'chem_annotated.inp'))
+        except EnvironmentError:
+            logging.info('Error: Could not generate Cantera files due to EnvironmentError. Check read\write privileges in output directory.')
+        except:
+            raise
```
mliu49: Sorry, more minor points. You could use `logging.error` instead of writing `logging.info('Error...')`.

Also, there's no need to do `except: raise`, because Python will raise the error regardless. You can just remove those two lines. You generally use try...except blocks if you're going to do something about the error besides just raising it (eg. printing a more helpful error message, or somehow fixing the error).

KEHANG:
Is this PR ready to merge?

alongd:
I believe it's ready. I made minor fixes as asked by @mliu49 , if you think it's good now then let's merge.

PR #842, opened 2016-12-15
Fix #841

### LINKED ISSUE #841: mergeModels tool gives wrong info saying "kinetics from model 1 did not match that of model 2" for :class:ThirdBody and :class:Lindemann
I'm trying to merge two identical toy models together to test mergeModels tool. For third-body reactions and Lindemann reactions, it says that the "kinetics from model 1 did not match that of model 2". I think the reason is that in [this line](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/kinetics/falloff.pyx#L107) and [this line](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/kinetics/falloff.pyx#L201), the second parameter of the method `isinstance()` should be `ThirdBody` and `Lindemann` respectively, instead of `Troe`.

PengZhang13:
Fix #841: mergeModels tool gives wrong info about third-body reaction and lindemann reaction.

### DIFF
--- a/rmgpy/kinetics/falloff.pyx
+++ b/rmgpy/kinetics/falloff.pyx
@@ -104,7 +104,7 @@ cdef class ThirdBody(PDepKineticsModel):
         Checks to see if kinetics matches that of other kinetics and returns ``True``
         if coeffs, kunits, Tmin,
         """
-        if not isinstance(otherKinetics, Troe):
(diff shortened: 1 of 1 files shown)


KEHANG:
good catch!

PR #843, opened 2016-12-15
Correct degeneracy for R + R recombination reactions

goldmanm:
Our R+R recombination reactions have incorrect reaction values when R1 = R2. When the two reactants are the same, the entropy should be greater, which should decrease the reaction rate. 

The previous method of RMG always gave R_Recombination a degeneracy of 1 (despite RMG finding 2 paths). This used family-specific, which isn't ideal in the general Family.py method. (even though this should only apply to R1=R2). 

This pull request removes the family-specific language in family.py, and ensures the reaction values are thermodynamically consistent. This may affect estimation of some reactions as well. 

reaction | current degeneracy | degeneracy (from this pull request)
----------|--------------------|---------------
[CH3]+[CH3] | 1 | 1
[CH3]+[CH2]C | 1 | 2
[CH2]CCCC + [CH2]CCCC | 1 | 1
[CH2]CCCC + [CH2]CCCCC | 1 | 2

This pull request reduces the degeneracy of R+R recombination reactions when R1 = R2 and increases them when R1 /= R2, and it adds test methods that test each change.

Another alternative would be to divide degeneracy the by 2 when there are same reactants and divide by 2 again if it is an r_recombination, which would result in a degeneracy of 0.5. To get identical results to the previous RMG jobs, the current item in the pull request may benefit from changing rate rules originating with different reactants and the alternative may benefit from changing rate rules originating from identical reactants. I don't think training reactions would be affected. 

Maybe the two alternatives ask the question: Should "R1 + R2" and "R2 + R1" be counted as degenerate if R2 is different from R1?


rmgpy/data/kinetics/kineticsTest.py
```diff
@@ -1,32 +1,240 @@
...
+        self.assertEqual(len(pb_kinetics_list), 1, pb_kinetics_list)
+        
+        # the same reaction group must be found or this test will not work
+        self.assertIn(pp_kinetics_list[0][0].comment,pb_kinetics_list[0][0].comment,
+                         'this test found different kinetics for the two groups, so it will not function as expected\n' + 
+                         str(pp_kinetics_list)+str(pb_kinetics_list))
+        
+        # test that the kinetics are correct
+        self.assertNotEqual(pp_kinetics_list[0][0].getRateCoefficient(300), pb_kinetics_list[0][0].getRateCoefficient(300))
+        self.assertEqual(pp_kinetics_list[0][0].getRateCoefficient(300) * 2, pb_kinetics_list[0][0].getRateCoefficient(300))
```
goldmanm: this is actually asserting that propyl-butyl reaction is faster, which is not the correct test. This test should be failing.

goldmanm:
don't merge this yet. I am fixing up some functionality in this first, and then cleaning up commits. I'll submit another pull request when this is ready.

Issue #840, opened 2016-12-14
merging documentation


goldmanm:
It seems like RMG has multiple places where different pieces of documentation are kept: 

1. The [wiki page](https://github.com/ReactionMechanismGenerator/RMG-Py/wiki) on Github
1. The [RMG User Guide](http://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/index.html)
1. The [RMG Theory Guide](http://reactionmechanismgenerator.github.io/RMG-Py/theory/index.html)

Philosophically, I can understand the difference between the User and Theory Guides. In reality, a lot of theory about algorithms seem intermingled in the User Guide. Some examples: [thermo](http://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/thermo.html#symmetry-and-chirality), [kinetics](http://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/kinetics.html), [diffusion limited kinetics](http://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/liquids.html#diffusion-limited-kinetics). The theory guide itself seems sort of sparse with only 

I think the wiki page contains mostly notes to help developers, though it also contains some user guides: [reduction](https://github.com/ReactionMechanismGenerator/RMG-Py/wiki/Reaction-centered-reduction-of-RMG-generated-kinetic-models) and [reaction family addition](https://github.com/ReactionMechanismGenerator/RMG-Py/wiki/Adding-reaction-families). 

My preference would be to have one repository, though having clear separation is also preferable. If anyone has any preferences, feel free to share. I might try to merge the theory into the user guide, and I'd appreciate other ideas for improvement.

rwest:
Although this is contrary to your aim of reducing the number of documentation sources, there also used to be, and perhaps ought to be, a "Developers Guide", giving an overview, introduction, tips like what classes are where, why we have two Species classes, what's the difference between a Molecule and a Species, etc. That type of thing. Things that *users* don't need to know, but new developers do, and can't currently get by reading anything.  Something to consider.  This could be in the wiki.  I think the wiki should also contain things like how to package a release, how to configure Travis regression tests, how to republish the documentation.

I think probably the chunks of theory could be appendices to the user guide. At least should be well linked from the user guide. I wouldn't want to force people to read the theory in order to find the spelling of a command or the choices available for an option, for example, but should be available to anyone reading the user guide.

goldmanm:
[...]

PR #847, opened 2016-12-16
Documentation Consolidation

goldmanm:
Consolidated the user guide from the RMG Wiki into the RMG User Guide & added a table in the main page to explain to users what to do in a more visual format (since one person mentioned getting confused and tried installing RMG when he just needed to look up thermo chemistry). This addresses issue #840 .

### DIFF
--- a/documentation/source/_static/rmg.css
+++ b/documentation/source/_static/rmg.css
@@ -270,18 +270,25 @@ p.caption {
 
 
 /* -- table styles ----------------------------------------------------------- */
-
-table.docutils,
-table.docutils colgroup, 
-table.docutils th, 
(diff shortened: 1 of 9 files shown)


KEHANG:
good catch! will add it to index.rst

goldmanm:
Thanks so much for the catch. I added the reduction script to index.rst, and I wrote about updated the index.rst in the RMG developer's wiki, so we can reduce a potential problem in the future.

KEHANG:
I cleaned up the commit history a bit and add a new commit for developer info update. @goldmanm  please `make html` and take a look and see if everything is fine to you.

Issue #831, opened 2016-12-02
Redundant Species in a Reaction Mechanism


rgillis8:
I have encountered a strange error where the core contains multiple copies of the same species. The species are all polycyclic aromatic radicals and the error seems to only occur when the saturated version of these species are included in a thermodynamic library (adding the saturated species to a thermo library seemed to trigger the problem).

Consider these consecutive "enlarge" steps recorded in the RMG.log file.

Summary of Secondary Model Edge Enlargement
---------------------------------
Added 0 new core species
Created 8 new edge species
    [CH] (Cc1ccccc1)c1ccccc1(11694)
    [c]1ccccc1CCc1ccccc1(11695)
    [c]1ccc(cc1)CCc1ccccc1(11696)
    [c]1cccc(c1)CCc1ccccc1(11697)
    [CH] (SCc1ccccc1)c1ccccc1(11698)
    [c]1ccccc1CSCc1ccccc1(11699)
    [c]1cccc(c1)CSCc1ccccc1(11700)
    [c]1ccc(cc1)CSCc1ccccc1(11701)
Added 0 new core reactions
Created 8 new edge reactions
    [CH] (Cc1ccccc1)c1ccccc1(11694) + c1ccc(cc1)CSCc1ccccc1(34) <=> [CH]1C=CC=CC1=CSCc1ccccc1(271) + c1ccc(cc1)CCc1ccccc1(76)
    [c]1ccccc1CCc1ccccc1(11695) + c1ccc(cc1)CSCc1ccccc1(34) <=> [CH]1C=CC=CC1=CSCc1ccccc1(271) + c1ccc(cc1)CCc1ccccc1(76)
    [c]1ccc(cc1)CCc1ccccc1(11696) + c1ccc(cc1)CSCc1ccccc1(34) <=> [CH]1C=CC=CC1=CSCc1ccccc1(271) + c1ccc(cc1)CCc1ccccc1(76)
    [c]1cccc(c1)CCc1ccccc1(11697) + c1ccc(cc1)CSCc1ccccc1(34) <=> [CH]1C=CC=CC1=CSCc1ccccc1(271) + c1ccc(cc1)CCc1ccccc1(76)
    [CH]1C=CC=CC1=CSCc1ccccc1(271) + c1ccc(cc1)CSCc1ccccc1(34) <=> [CH](SCc1ccccc1)c1ccccc1(11698) + c1ccc(cc1)CSCc1ccccc1(34)
[...]

Issue #577, opened 2016-02-17
NetworkError: Unexpected type of path reaction


alongd:
Running RMG-Py with PDep for ethylamine oxidation, I get these NetwokErrors, for example:

```
  File "/home/alongd/ws/RMG-Py/rmgpy/pdep/network.py", line 569, in calculateMicrocanonicalRates
    raise NetworkError('Unexpected type of path reaction "{0}"'.format(rxn))
rmgpy.pdep.network.NetworkError: Unexpected type of path reaction "HCNH(263) + CH3(10) <=> CH3CHNH(294)"
```

or

```
  File "/home/alongd/ws/RMG-Py/rmgpy/pdep/network.py", line 569, in calculateMicrocanonicalRates
    raise NetworkError('Unexpected type of path reaction "{0}"'.format(rxn))
rmgpy.pdep.network.NetworkError: Unexpected type of path reaction "CC(=NC=[CH])[N](12189) <=> N1([C](C)[N])C=C1(12483)"
```

Seems to be similar to an error reported by @rwest in issue #41, quite some time ago, but could also have a different cause.

bslakman:
Hi @alongd, do you have an input file that produces this error?

alongd:
Yes, but the input file also used non-common libraries. I tried to reproduce this error with a minimal input file (so sorry for taking long to response), but only the original library combination resulted in that error.
The reactions RMG crashed on are not present in these libs.

If it's any help, here are the original input file ([input.py](https://github.com/ReactionMechanismGenerator/RMG-Py/files/192231/input.py.txt)), and libraries:
https://github.com/alongd/RMG-database/blob/nitrogen/input/thermo/libraries/1Cu4.py
https://github.com/alongd/RMG-database/blob/nitrogen/input/thermo/libraries/EA.py
https://github.com/alongd/RMG-database/tree/nitrogen/input/kinetics/libraries/A20

alongd:
I get simulations crashing with the "Unexpected type of path reaction" error, e.g.:
```
rmgpy.pdep.network.NetworkError: Unexpected type of path reaction "[CH2][CH2](103) + CO2(30) <=> [CH2]CO[C]=O(4145)"
```
```
rmgpy.pdep.network.NetworkError: Unexpected type of path reaction "[CH2]CC([O])=O(4145) <=> [O]C1([O])CC1(4663)"
```
```
[...]

PR #850, opened 2016-12-20
Fix bug when reordering molecules after HBI thermo calculation

mliu49:
When estimating radical thermo via HBI, RMG will look for the saturated species in libraries. If only some of the saturated molecules are found in a library, then the ones that were not found get dropped from the list when reordering the molecules based on thermo. This leads to failed isomorphism checks and duplicated species, as seen in #577 and #831.

This fixes the issue by adding any non-matched molecules back into the list.

### DIFF
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -1077,7 +1077,10 @@ class ThermoDatabase(object):
                     for i in range(len(thermo)): 
                         logging.debug("Resonance isomer {0} {1} gives H298={2:.0f} J/mol".format(i+1, thermo[i][1].toSMILES(), thermo[i][0]))
                     # Save resonance isomers reordered by their thermo
-                    species.molecule = [item[1] for item in thermo]
+                    newMolList = [item[1] for item in thermo]
+                    if len(newMolList) < len(species.molecule):
+                        newMolList.extend([mol for mol in species.molecule if mol not in newMolList])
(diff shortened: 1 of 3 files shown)


KEHANG:
Good catch! Could you add some unit test to make sure `getThermoData` doesn't change species object except only alter order of isomers?

mliu49:
I added two unit tests to make sure `Species.molecule` is not changed (besides reordering) by thermo generation. The HBI+Library one used to fail, and passes with this PR. The HBI+GAV one has always passed, but I thought it might be good to include.

I also put in a small commit fixing `Molecule.getFormula()`, which I discovered while debugging this issue. Basically, we weren't really following the Hill system to generate the formulas, and were just ordering elements by alphabetical order. The commit fixes the if statements in the method.

KEHANG:
The RMG-tests shows it seems still to generate lots of identical species in job `NC` (line1605-line1632 are same species): https://travis-ci.org/ReactionMechanismGenerator/RMG-tests/builds/185600427

mliu49:
I think the RMG-tests output is mislabeled, as @nyee mentioned previously. Those identical species are actually in the benchmark version, but not in the new version.

KEHANG:
RMG-tests has just been fixed. The new [log](https://travis-ci.org/ReactionMechanismGenerator/RMG-tests/builds/185671895) indeed shows lots of identical species in the original RMG job.

Issue #848, opened 2016-12-18
Cantherm not properly loading Gaussian files for TS's anymore


zjburas:
Using the latest  version of the RMG-Py master branch (commit 65c8cc4) I can no longer run certain Cantherm jobs that were fine before. Specifically, I get the following error if the Cantherm job contains a transition state that  references a Gaussian file:

Traceback (most recent call last):
  File "/home/zjburas/RMG-Py/cantherm.py", line 61, in <module>
    cantherm.execute()
  File "/home/zjburas/RMG-Py/rmgpy/cantherm/main.py", line 237, in execute
    job.execute(outputFile=outputFile, plot=self.plot)
  File "/home/zjburas/RMG-Py/rmgpy/cantherm/statmech.py", line 180, in execute
    self.load()
  File "/home/zjburas/RMG-Py/rmgpy/cantherm/statmech.py", line 226, in load
    self.species.props['elementCounts'] = atoms
AttributeError: 'rmgpy.species.TransitionState' object has no attribute 'props'

So it seems that the transition state is not being properly loaded from the Gaussian file. If I go back to the RMG-Py master branch of about a month ago  (commit cef5c30), then the same Cantherm job runs just fine. So it must have been a recent change that caused this bug, but I'm not sure which out of the many commits that have been merged over the last month.

nyee:
I haven't fully fleshed out the problem, but it's definitely commit c57a9654401a301e70f85352680eaaca2c6f519 that is causing the issue as that's where the line that the error takes place on was added. 

It looks like @mliu49 added a line to save the atoms of the species so that we could use pre-existing code to write a proper chemkin thermo entry. For some reason from the error you listed out, it is trying to apply that line of code to a transition state, which doesn't necessarily have (or need) all the atoms listed.

KEHANG:
@zjburas can you use [this branch (dynamic_attr)](https://github.com/ReactionMechanismGenerator/RMG-Py/tree/dynamic_attr) to see it can resolve the issue?

zjburas:
@KEHANG I get the same error using the dynamic_attr branch. You should be able to replicate the error if you try running any Cantherm job with a transition state that references a  Gaussian file (unfortunately none of the RMG-Py examples fit this criteria).

rwest:
We should add tests that catch this.

KEHANG:
@zjburas you are right, I just updated the branch (`dynamic_attr`) again, this time should be working.

[...]

PR #854, opened 2016-12-21
Fix element count handling in Cantherm

mliu49:
This PR addresses #848. The feature of saving element counts was added in #803, but did not consider all use cases, some of which led to errors.

This fix only saves element counts for species objects. They are not necessary for transition states because it's not possible to run thermo jobs with transition states.

Additionally, there is better handling when writing Chemkin thermo blocks, to catch cases where the elementCounts property may not exist, and in cases where the species actually has a molecule object.

It would be best to add a few unit tests before merging, but I haven't come up with a good way to test this feature. Suggestions or contributions are welcome.


mliu49:
I've added some unit tests, and I think this is ready to be merged, if someone could review it.

PR #860, opened 2016-12-31
Added more features in mechanism analyzer

yunsiechung:
I added the following features / changes in mechanism analyzer:

1) Previously, when the reaction list included pressure-dependent reactions, because PDepReaction object does not have attribute 'family', AttiributeError was raised in the following line:
'if rxn.family == "H_Abstraction": '
To prevent this, when visualizing the pathway, it first checks whether the rxn is the instance of PDepReaction and add to Pdep_rxns list if it is.

2) A new method 'showTheMostDominantSpecies' is added to find the most dominant species and display them in descending order. Users can specify which most dominant species they want to see. Also, users can specify whether they want to see the molecular weight of each species.

3) For each example reaction found, the forward and reverse rate coefficients at the user-specified temperature and pressure are displayed.

4) Users can choose to use either instantaneous flux or time integrated flux for analysis.


rmgpy/tools/extractInfoFromckcsv.py
```diff
@@ -192,4 +194,70 @@ def getROPFlux(spc_rop_dict, species_string, rxn_index):
             if rxnNum == rxn_index:
                 flux = flux_tup[1]
                 return flux
-    return []
\ No newline at end of file
+    return []
+
+def showTheMostDominantSpecies(dominant_list, moleFraction_dict, species_identifier_dict, displayMW=False):
```
KEHANG: I suggest put this block of code in the ipython notebook, just like other functionalities so that users can have the freedom to customize the printing and logic

rmgpy/tools/extractInfoFromckcsv.py
```diff
@@ -192,4 +194,70 @@ def getROPFlux(spc_rop_dict, species_string, rxn_index):
...
+        elif new_order == 2:
+            print "****The 2nd most dominant species****"
+        elif new_order == 3:
+            print "****The 3rd most dominant species****"
+        elif new_order % 10 is 1 and new_order is not 11:
+            print "****The {0}st most dominant species****".format(new_order)
+        elif new_order % 10 is 2 and new_order is not 12:
+            print "****The {0}nd most dominant species****".format(new_order)
+        elif new_order % 10 is 3 and new_order is not 13:
+            print "****The {0}rd most dominant species****".format(new_order)
```
KEHANG: instead of printing a long log, can you convert it into a table in the same format as other functionalities adopt?

ipython/mechanism_analyzer.ipynb
```diff
@@ -17,7 +17,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
```
KEHANG: please clean up the output, usually we don't include the output of ipython notebook on master

rmgpy/tools/extractInfoFromckcsv.py
```diff
@@ -1,6 +1,8 @@
 import csv
 import numpy
 from rmgpy.chemkin import getSpeciesIdentifier
+from IPython.display import display
+import numpy as np
```
KEHANG: numpy has been imported two lines above

rmgpy/tools/extractInfoFromckcsv.py
yunsiechung: Okay. I'll do that!

rmgpy/tools/extractInfoFromckcsv.py
yunsiechung: That's a good point. I'll do that.

ipython/mechanism_analyzer.ipynb
yunsiechung: Okay!

yunsiechung:
I made all the changes you requested. I also got rid of relative fluxes to avoid confusion.
Now, the units for rate coefficients, instantaneous flux, and integrated flux are all displayed. 
Also, every user input lines are boxed.

KEHANG:
Good work!

PR #863, opened 2017-01-06
Correcting updateLonePairs

alongd:
The calculation of lone pairs was wrong for the N1s atom type,
and could potentially be worng for some of the new sulfur atom types
which will soon be introduced in RMG.

This PR fixes the calculation of the number of lone pairs by using
the valence electrons of an element. An error will be printed
in case the number of lone pairs isn't an integer.

The previous calculation resulted in 3 lone pairs for N1s (where it should only have 2):
atom1.lonePairs = 4 - atom1.radicalElectrons - int(order) = 4 - 1 = 3
The new calculation will give (5-1)/2 = 2 for N1s.
The new calculation will to be identical for carbon \ oxygen \ silicon.

### DIFF
--- a/rmgpy/molecule/molecule.py
+++ b/rmgpy/molecule/molecule.py
@@ -1552,11 +1552,9 @@ class Molecule(Graph):
                         order = order + 3
                     if bond12.isBenzene():
                         order = order + 1.5
-
-                if atom1.isSilicon() or atom1.isCarbon():
-                    atom1.lonePairs = (4 - atom1.radicalElectrons - int(order)) / 2
-                else:     
-                    atom1.lonePairs = 4 - atom1.radicalElectrons - int(order)
+                atom1.lonePairs = (elements.PeriodicSystem.valence_electrons[atom1.symbol] - atom1.radicalElectrons - atom1.charge - int(order)) / 2.0
+                if atom1.lonePairs % 1 > 0 or atom1.lonePairs > 4:
+                    logging.error("Unable to determine the number of lone pairs for element {0} in {1}".format(atom1,self))
             else:
                 atom1.lonePairs = 0
                 
--- a/rmgpy/molecule/molecule.py
+++ b/rmgpy/molecule/molecule.py
@@ -1541,17 +1541,8 @@ class Molecule(Graph):
         """
         cython.declare(atom1=Atom, atom2=Atom, bond12=Bond, order=float)
         for atom1 in self.vertices:
-            order = 0
             if not atom1.isHydrogen():
-                for atom2, bond12 in atom1.edges.items():
-                    if bond12.isSingle():
-                        order = order + 1
-                    if bond12.isDouble():
-                        order = order + 2
-                    if bond12.isTriple():
-                        order = order + 3
-                    if bond12.isBenzene():
-                        order = order + 1.5
+                order = atom1.getBondOrdersForAtom()
                 atom1.lonePairs = (elements.PeriodicSystem.valence_electrons[atom1.symbol] - atom1.radicalElectrons - atom1.charge - int(order)) / 2.0
                 if atom1.lonePairs % 1 > 0 or atom1.lonePairs > 4:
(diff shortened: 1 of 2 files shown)


rmgpy/molecule/molecule.py
```diff
@@ -1552,11 +1553,9 @@ def updateLonePairs(self):
                         order = order + 3
                     if bond12.isBenzene():
                         order = order + 1.5
-
-                if atom1.isSilicon() or atom1.isCarbon():
-                    atom1.lonePairs = (4 - atom1.radicalElectrons - int(order)) / 2
-                else:     
-                    atom1.lonePairs = 4 - atom1.radicalElectrons - int(order)
+                atom1.lonePairs = (PeriodicSystem.valence_electrons[atom1.symbol] - atom1.radicalElectrons - int(order)) / 2
+                if atom1.lonePairs % 1 > 0:
```
mliu49: Since the calculation above is with all integer types, won't this condition never be triggered? Should you change the equation to divide by 2.0?

alongd:
Absolutely right.
I changed to 2.0

nyee:
Looks good overall, but could you add a few unit tests. Maybe one for N1s, N3s, CH2 singlet, and an oxygen on a carbonyl?

alongd:
Travis failed with:
```

  File "/home/travis/build/ReactionMechanismGenerator/RMG-Py/rmgpy/molecule/moleculeTest.py", line 511, in testUpdateLonePairs

    self.assertEqual(mol_N5d.atoms[2].lonePairs, 3)

AssertionError: 2 != 3
```
Since my commit did not take into account charge. Will push a fix.

alongd:
A note on future heteroatoms incorporation in aromatic rings:
Currently RMG does not incorporate heteroatoms in aromatic rings (to my understanding).
However, once they are incorporated we should consider the bond order of a heteroatom that donates its lone pair to the aromatic ring (as in `N5b` e.g. [pyrrole](https://en.wikipedia.org/wiki/Pyrrole), as opposed to `N3b` e.g. [pyridine](https://en.wikipedia.org/wiki/Pyridine)).
Since the said lone pair is delocalized, `N5b` has 0 lone pairs (at least that's how we're currently defining it on the nitrogen branch).
For the updateLonePair calculation I added an additional `+1` to the bond order of `N5b` (and we should do the same for `S4b` and probably aromatic oxygen in the future). Therefore the calculation for the nitrogen atom in pyrrole
```python
atom1.lonePairs = (elements.PeriodicSystem.valence_electrons[atom1.symbol] - atom1.radicalElectrons - atom1.charge - int(order)) / 2.0
```
Results in:
```python
[5 - 0 - 0 - (1.5+1.5+1+1)] / 2.0 = 0
```

Currently this has no effect since `N5b` is not in use.

nyee:
@alongd, in the [RMG-tests line 1701](https://travis-ci.org/ReactionMechanismGenerator/RMG-tests/builds/194596235). It looks like [N+]#[C-] thermo is no longer being recognized under the thermo_DFT_CCSDTF12_BAC library. There's a bug where the log lists the comments in the wrong order, so it was previously matching with the DFT library, but now it completely uses group additivity. 

Could you check to see what is happening there and determine which one is more correct? I think this is the last discrepancy.

alongd:
Thanks, @nyee .
The only species in thermo_DFT_CCSDTF12_BAC which might resemble `[N+]#[C-]` is `CNJ` (index 264). But there's no charge separation there. Strange. I'll continue investigating, let me know if you might have a lead.

nyee:
@alongd, turns out I was wrong about the order on RMG-tests, so actually it is now using the library value referencing H[N+]#[C-] to calculate [N+]#[C-] via HBI. It looks like previously it was not doing lone pairs correctly, so it could not recognize that the library entry was the saturated base it needed. I am going to pull once tests finish.

alongd:
@nyee, I guess we can also close issue #677 now, right?

PR #868, opened 2017-01-17
Fixed bug in __gainPair and __losePair function of class GroupAtom

zjburas:
In previous implementation, if a group atom did not already have at least one pair, the __gainPair function would do nothing (no lone pair was incremented). Inserted a simple if...else... statement to cover both cases where group atom either does or does not already have at least one lone pair.

Without this fix, if a reaction family includes GAIN_PAIR on a carbon atom as an action (to make a singlet carbene), the reverse reaction template generated will not be correct because it will not include the new lone pair. Therefore, when generateReactions tries to apply the reverse of the reaction family on a reactant the result will not be correct, in most cases causing RMG to crash. At least this is the situation I was encountering  that led me to this bug.


zjburas:
@alongd and  @nyee Since you guys are doing a lot of work with lone pairs, maybe you can review this small pull request,  which should fix a bug in the gain_pair function for groupatoms. The original function was written by @bbuesser , probably with nitrogen mostly in mind, so it is not surprising that it didn't work initially for the new singlet carbene chemistry I'm trying to teach RMG.

alongd:
Looking good, @zjburas , I think it's ready to merge.

rmgpy/molecule/group.py
```diff
@@ -233,10 +233,16 @@ def __gainPair(self, pair):
...
-        # Set the new lone electron pair count
-        self.lonePairs = lonePairs
+
+        #Add a lone pair to a group atom with none
+        if not self.lonePairs:
+            self.lonePairs.append(pair)
+        #Add a lone pair to a group atom that already has at least one lone pair
+        else:
+            for lonePairs in zip(self.lonePairs):
+                lonePairs.append(lonePairs + pair)
```
KEHANG: this line is very confusing (`lonePairs.append(lonePairs + pair)`), are you sure it is doing what you want? Is the following the intended code?
```python
for lonePair in zip(self.lonePairs):
      lonePairs.append(lonePair + pair)
```

zjburas: What you wrote would make more sense, the overlapping variable names are indeed  confusing, but  that is actually how  it is currently written in master (although the blame history points to me,  I only added the if... part of the conditional statement).

zjburas:
After discussing with @nyee , it looks like the bug is bigger than I initially thought, and will require a more comprehensive solution. Basically, it looks like this function was not working at all as intended, but  somehow avoided giving  an error. More investigation needed.

zjburas:
I think @nyee and me have fixed the  larger problem now. There  were really 3 bugs before:

1. The GroupAtom function "copy" was not copying the lonePairs attribute (everything else is copied), so that whenever this function was called it was setting lonePairs = [].

2. The GroupAtom functions __gainPair and __losePair would simply output lonePairs = [] if the input was also lonePairs = [].

3. The GroupAtom functions __gainPair and __losePair would raise an error if lonePairs actually equaled  something, like [1], because it would turn the list of lonePairs into a tuple and then try to add/subtract an integer from it. This error was never raised, however, because the input the function was always lonePairs = [] due to bug 1 above (GroupAtom.copy() is called before either __gainPair or __losePair). 

Overall, I think the cumulative effect of the 3 bugs in GroupAtom was that the reverse of any reaction family involving GAIN_PAIR or LOSE_PAIR was not defined correctly. There might have been other consequences that I'm not aware of though. 

All 3 bugs have  been fixed, and some unit tests are probably in order next.

bslakman:
I'm surprised I never came across this before. I have a couple of families where GAIN_PAIR or LONE_PAIR is in the recipe. For silicon. Will do some testing with my branch

nyee:
@bslakman, were you able to conclude anything related to this issue on your branch? I am thinking about pulling this in soon.

zjburas:
Latest two commits add unit tests for the gainPair and losePair functions of the GroupAtom  class. Previously, neither function was tested at all, which caused the bugs described above to go undetected. 

Also, the lonePairs attribute was included in all of the other GroupAtom unit tests, such as the test of the copy function, which would have also caught the bugs above if it had been implemented.

Finally, @alongd and @nyee , you guys should take a look at the last commit, where the gainPair and losePair functions are made to change atomtype, in addition to the lonePairs attribute. This change was necessary because at one of the Nitrogen atomtypes (N5dd) changes when the number of lone pairs is decremented/incremented. It could just be a typo though.

zjburas:
@alongd and @nyee , here are the two atomtypes I wanted you guys to take a look at:

https://github.com/ReactionMechanismGenerator/RMG-Py/blob/b3b280b956f9b61feb078e3d825c7bd30c0eb0a5/rmgpy/molecule/atomtype.py#L387

https://github.com/ReactionMechanismGenerator/RMG-Py/blob/b3b280b956f9b61feb078e3d825c7bd30c0eb0a5/rmgpy/molecule/atomtype.py#L391

It looks like these are the only two atomtypes that change atomtype when lonePairs are either decremented or incremented. It looks like a typo, but wanted to check.

alongd:
Re nitrogen, we know that some atomType actions are wrong. They should be all fixed on the [nitrogen branch](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/nitrogen/rmgpy/molecule/atomtype.py#L404) (hope to PR soon!). However, in the updated N atomTypes we do have examples of changing atomTypes upon lonePair change.
The `O` example you pointed at does look suspicious. I guess the most general atomType for each element should remain the same throughout all transitions, unless I missed something with `O`. @nyee?

nyee:
I agree that the O atomtype is a typo. @zjburas, if you look at the nitrogen branch, we have many nitrogen atomtypes that will change with decrement/increment lone pair. In that branch we have pretty carefully thought them out, so they should be correct. For this pull request, I suggest we just change the O atomtype and leave the nitrogen to be patched over later.

zjburas:
Ok, O atomType is fixed.

It looks like this branch and the [nitrogen branch](https://github.com/ReactionMechanismGenerator/RMG-Py/tree/nitrogen) are complementary. I'm glad we didn't make redundant changes. 

So both gainPair and losePair functions of GroupAtom class should update atomtype, as implemented in [c57a111](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/868/commits/c57a1117614f80eee883ae558c26e620b26277ae), and reflected in the new unit tests. If that is the case, I think this PR is ready to be reviewed and pulled.

PR #869, opened 2017-01-20
Change product template duplicate testing from isIsomorphic to isIden…

nyee:
…tical

Previously we used isIsomorphic to delete duplicate product templates, when we really should be using isIdentical which does not respect wildcards. This make it so that we could potentially lose product templates like R u1 if C u1 was added first.

This push may make RMG find many more reverse reactions for family's that are not their own reverse.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -851,7 +851,7 @@ class KineticsFamily(Database):
             for i, struct in enumerate(productStructure):
                 for s in productStructureList[i]:
                     try:
-                        if s.isIsomorphic(struct): break
+                        if s.isIdentical(struct): break
                     except KeyError:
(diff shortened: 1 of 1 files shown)


rwest:
Could we use `isSubgraphIsomorphic` i.e. an asymmetric comparison so that:
`C u1` and `C u1` found to be duplicates/redundant and only `C u1` is preserved.
`R u1` and `C u1` found to be duplicates/redundant and only `R u1` is preserved.
`C u1` and `R u1` found to be duplicates/redundant and only `R u1` is preserved.
`R u1` and `C u1` and `R u[0,1]` are found to be duplicates/redundant and only `R u[0,1]` is preserved.

Would have to be careful to put it the right way around, to preserve the most general version in each case.
That way, the resulting template would contain as few things in the LogicOr as possible.

PR #870, opened 2017-01-22
Solvation thermo

yunsiechung:
This branch cleaned up several codes related to liquid reactor systems and added new solvation thermo correction method to predict the temperature dependence of solvation free energy more accurately at high temperature. The detailed documentations with example plots are included in this branch as well.

Main commits include:

1. A wrapper class for solvent is created. This new class Solvent() contains attribute related to solvent species and it is stored under RMG() class object. Previously, these solvent attributes were stored under Species class, which was redundant because solvent information are the same across all the species. The solvent related attributes are now removed from Species class, and the Species class in no longer created in rmgpy/rmg/model.

2. Previously, there was a bug with thermoEstimator, which used gas phase thermo for liquid phase species. The bug is fixed, and now thermoEstimator can be used for liquid phase species as well.

3. If the solvent can be found in CoolProp module, RMG now applies more accurate solvation free energy correction using two correlations for solute species. For solvent species, the gas-liquid partition coefficient is directly computed from CoolProp. Details about the correlations can be found in the Liquid Phase Systems documentation in this branch. After the solvation free energy is added to the gas phase free energy, the NASA model is fitted to the resulting liquid phase free energy by using a new method, fitToFreeEnergy. The fitted NASA model can correctly estimate the free energy. However, because the NASA model is a heat capacity model originally designed for gas phase thermo, the enthalpy and entropy estimated from the fitted NASA model contain some error, and the estimated heat capacity is very off ( outputs negative values for some temperature range). Because only free energy is used in liquid  reactor, the error in enthalpy, entropy, and heat capacity do not actually affect the simulation output. Yet, the new thermo model that can correctly represent the liquid phase thermo should be added in future. 
If the solvent cannot be found in CoolProp module, RMG assumes the linear temperature dependence of solvation free energy as it did before. In this case, because only constant values of enthalpy and entropy are added to gas phase thermo, the fitting process is not needed. 
The plots below compare the experimental data with the free energy estimated by the old method of assuming linear T-dependence (method 1) and the free energy estimated by the new method (method 2).

dgsolvcomparison

(The references for the experimental data can be found in the Liquid Phase Systems documentation)
The plots show that the new method (method 2) display good agreement with the experimental data. 
To use the new method, RMG-database also needs to be updated. The pull request for RMG-database will be made after this branch is merged.


yunsiechung:
@KEHANG I was rushing to finish up this branch and did not catch many bugs. 
I'll debug some error first!

PR #872, opened 2017-01-25
Enable RMG to produce benzene through propargyl radical recombination

zjburas:
Companion PR to RMG-database PR 161. The following changes in RMG-Py are necessary in order for RMG to form benzene from propargyl radical recombination:

1. Because many of the new reaction families use the GAIN_PAIR and LOSE_PAIR actions, in order for them to work the recently discovered bug in these actions needs to be fixed. Therefore, this PR builds on #868.

2. The 2 new reaction families that are their own reverse are hardcoded in rmgpy/data/kinetics/family.py, where the switching of atom labels in reversing the reaction direction are defined.

Also, added the Bond Additivity Corrections (BAC's) of Goldsmith et al. to Cantherm for the following quantum method (Goldsmith, C. F.; Magoon, G. R.; Green, W. H., Database of Small Molecule Thermochemistry for Combustion. J. Phys. Chem. A 2012, 116, 9033-9057.):

RQCISD(T)/cc-PV(infinity)(Q)Z

This method is simply called "Klip_3" in Cantherm currently.


zjburas:
This PR is no longer necessary.

PR #874, opened 2017-01-26
New criterion

mjohnson541:
This creates a new optional reaction mechanism generation criterion that works alongside the original flux criterion but considers how the edge affects the core rather than how the core affects the edge. The new criterion moves edge reactions (the species in a given reaction) to the core when their estimated effect on the steady state concentration of the involved core species is greater than a given tolerance.


mjohnson541:
This pull request is continued in AcceleratorTools

Issue #877, opened 2017-02-01
Restart options are poorly documented


rwest:
Someone just emailed me asking how to restart a job from a restart file. As I looked for the appropriate page in the documentation to refer them to, I couldn't find it mentioned anywhere.

I believe from looking at https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmg.py#L79 that you just add `-r` (or `--restart`) to the command line when running the job, i.e.
```
$ python rmg.py -r input.py
```
but it would be great if someone could check that this works and add it to the documentation, along with the other various command line options!
Perhaps in the section on "Running a Job" https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/documentation/source/users/rmg/running.rst

samimiomid:
Yes, it works by adding -r to the command. 
thanks
Omid

KEHANG:
documentation added via #878

Issue #822, opened 2016-11-16
Deleting .reverse attribute causes problems pickling restart file.


rwest:
tl;dr: Saving restart files is broken.

To demonstrate, start with the minimal example `/examples/rmg/minimal/input.py` and modify the options block so that it saves the restart file every loop, e.g.:
`saveRestartPeriod=(1,'seconds'),` like this: [input.py.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/596120/input.py.txt)

It will crash with something like this:
```python
Saving restart file...
Traceback (most recent call last):
  File "/Users/rwest/Code/RMG-Py/rmg.py", line 152, in <module>
    rmg.execute(**kwargs)
  File "/Users/rwest/Code/RMG-Py/rmgpy/rmg/main.py", line 636, in execute
    self.saveEverything()
  File "/Users/rwest/Code/RMG-Py/rmgpy/rmg/main.py", line 779, in saveEverything
    self.notify()
  File "/Users/rwest/Code/RMG-Py/rmgpy/util.py", line 109, in notify
    observer.update(self)
  File "/Users/rwest/Code/RMG-Py/rmgpy/restart.py", line 93, in update
    save(rmg)
  File "/Users/rwest/Code/RMG-Py/rmgpy/restart.py", line 42, in save
    delay=0 if rmg.done else rmg.saveRestartPeriod.value_si
  File "/Users/rwest/Code/RMG-Py/rmgpy/restart.py", line 65, in saveRestartFile
    cPickle.dump(reactionModel, f, cPickle.HIGHEST_PROTOCOL)
  File "/Users/rwest/Code/RMG-Py/rmgpy/data/kinetics/family.py", line 126, in __reduce__
    self.reverse,
AttributeError: 'TemplateReaction' object has no attribute 'reverse'
```

Running under a debugger I find the reaction in this case, and at least one other example I have tried, (and I hypothesize always,) is `intra_H_migration` which is its own reverse.

Looking at the `__init__` code I couldn't see how it's possible to make a `TemplateReaction` object without a `.reverse` attribute.

However, then notice that in `rmgpy/rmg/model.py` line 654 we have:
```
                         # We're done with the "reverse" attribute, so delete it to save a bit of memory
                        delattr(reaction,'reverse')
```

I suspect this, which predates the TemplateReaction pickling change in fcbb3838cd2113c4a90138cbefba7d08ddaf8f6f, is causing the problem.

[...]

PR #878, opened 2017-02-02
fix restart file generation

KEHANG:
This PR fixes the issue of restart file generation failure caused by `reaction.reverse` being deleted #822 . The patch 

- replaces `delattr(reaction, 'reverse')` with `reaction.reverse = None`

- adds a unit test for restart file generation

- adds restart documentation to address issue #877

### DIFF
--- a/documentation/source/users/rmg/running.rst
+++ b/documentation/source/users/rmg/running.rst
@@ -3,10 +3,20 @@
 *************
 Running a Job
 *************
- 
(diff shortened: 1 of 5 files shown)


rwest:
Great! Thanks Kehang. 
Is there a simple way to also test we can *load* a restart file once we've saved it?

KEHANG:
@rwest will be added via #879

PR #749, opened 2016-08-17
Uncertainty Class Part I: assign uncertainties and extract original model sources

connie:
This is the beginnings of a new `Uncertainty` class within `rmgpy.tools.uncertainty`.  This class will add uncertainty analysis features to RMG.  This pull request is part one in a series of many pull requests to come to add both local and global uncertainty analysis features to RMG.  I decided to break it down into parts for easier review and clarity.

The following commits primarily parse the comments in a RMG-generated annotated chemkin file.  It then reloads the RMG database with verbose comments to extract all the original source objects from the database and store them in the `Uncertainty` class.  The idea is that by retracing the sources, we can find the weighted dependence on correlated parameters such as thermo groups and kinetic rate rules.  Also, it makes it easier in general to see what sources comprise a model, and to manipulate their contributions.  

Two uncertainty engines: `KineticsParameterUncertainty` and `ThermoParameterUncertainty` allow the user to initialize a set of constants used in formulas that assign uncertainties to kinetic and thermo sources of varying types.  It is intended to be used directly with the function `Uncertainty.extractSourcesFromModel`.     Currently, the `getUncertaintyValue` function obtains the total uncertainty based on several formulas.

For thermo group additivity:

image

For kinetics from rate rules:
image

image

Where `dm_match = log10(N+1)*dlnk_nonexact` and N is the number of rate rules being used in the averaged (therefore, this value is larger for averages containing a lot of rate rules rather than just 1 or 2).

An ipython notebook is provided for testing these features on a sample CHEMKIN file with various types of thermo and kinetics.

PR #891, opened 2017-02-12
Uncertainty Class Part III: local and correlated analysis

connie:
This pull request is (the possibly last) part of the series relating to the `Uncertainty` class and module (see [Part I](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/749) and [Part II](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/754)). 

- Adds the `Uncertainty.localAnalysis` function which runs local first order uncertainty analysis using sensitivity analysis through RMG's native solver.
- Adds `getPartialUncertainty` function to both `KineticParameterUncertainty` and `ThermoParameterUncertainty` classes, such that we can obtain the partial uncertainty contribution of a rate rule or Benson group within a RMG-estimated kinetic or thermo parameter, respectively.
- Adds correlated uncertainty analysis functionality to both local uncertainty (by toggling `correlated=True` in the `Uncertainty.localAnalysis`  function), or within global analysis (by setting `correlated=True` in the `rmgpy.tools.muq.ReactorModPiece` class before running).
- Adds uncertainty plotting functions to the `rmgpy.tools.plot` module which creates bar plots showing the most influential thermo or kinetic parameters on a user-specified output variable.
- Updates the IPython notebooks `localUncertainty.ipynb` and `globalUncertainty.ipynb` to demonstrate the functionality and explain the methodology in depth.  
- Changes the way the kinetic or thermo parameter uncertainties are assigned. Bill mentioned that they should be added linearly, rather than squared, because the uncertainties have assumed uniform distributions rather than normal distributions. i.e.

image

instead of 

image


KEHANG:
the reduced PDD model required by`localUncertainty.ipynb` demo is not included in the PR yet

connie:
Ahh I didn't realize that.. Will update this.

KEHANG:
@connie thanks! I saw the added example mechanism. One more question for the muq library, I guess the current status is that linux and mac are fine through anaconda, windows probably still has some issue?

connie:
Yeah, I don't think windows can run it unless somebody makes a package for it.  Also, I think even for linux and mac there are some issues with the conda package such that I don't think it works completely out of the box unless you modify some paths. Since MUQ wasn't written like a typical python program, it just creates some Python .so library files, you have to do some thing to make sure it's on the PYTHONPATH.  So the conda packaging has a hard time with it.  But Bill says it's fine-- people can install it from scratch if they need to.

KEHANG:
@connie I also got `bus error: 10` on Mac (linux is fine), I guess we should re-compile muq binary?

@mliu49 For the` retrieveSaturatedSpeciesFromList` when you don't have a saturated version in the mechanism, since `uncertainty` has already loaded database of the right version, we probably can generate thermo on-the-fly for the desired saturated species.

connie:
Yes, the mac error is known, and I would ignore that for now in terms of the pull request.  

As for the saturated thermo- it seems like most of the time the saturated species is found in the model.  But so far the script hasn't errored on me yet.  The radical species need to use the stable species' thermo before applying the HBI correction, so that's why it uses the saturated species.  It looks like `retrieveSaturatedSpeciesFromList` will just raise an error if it can't find the saturated species for now....

Any plans to merge?

KEHANG:
@connie if so, I guess it's ready to go, I'll go ahead to fix some typo and merge it in soon.

PR #892, opened 2017-02-13
Fix Group._repr_png_()

mliu49:
This method needed to be updated following the switch to floats for bond orders.


mliu49:
It seems that Travis is generating a slightly different png than my computer. I'll work on a different way to write the test.

Issue #899, opened 2017-02-14
Benzene bonds of sulfur (and nitrogen?) cause Travis to crash, probably a string/float issue


alongd:
An atomTypeTest for an aromatic compound containing sulfur crashed. Heteroatoms in aromatics are not fully developed in RMG, but I wrote this test on the sulfur branch to check that the benzene-bonded sulfur atomType (`S4b`) functions as expected.
The molecule tested:
```python
self.mol25 = Molecule().fromAdjacencyList('''
1 S u0 p1 c0 {2,B} {5,B}
2 C u0 p0 c0 {1,B} {3,B} {6,S}
3 C u0 p0 c0 {2,B} {4,B} {7,S}
4 C u0 p0 c0 {3,B} {5,B} {8,S}
5 C u0 p0 c0 {1,B} {4,B} {9,S}
6 H u0 p0 c0 {2,S}
7 H u0 p0 c0 {3,S}
8 H u0 p0 c0 {4,S}
9 H u0 p0 c0 {5,S}''')
```
![image](https://cloud.githubusercontent.com/assets/16158262/22945595/c5245528-f2c2-11e6-99a6-c8a68dde0be7.png)
and the test is:
```python
self.assertEqual(self.atomType(self.mol25, 0), 'S4b')
```
It crashed with the following trace:
```

======================================================================
ERROR: Test that getAtomType() returns appropriate sulfur atom types.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/ReactionMechanismGenerator/RMG-Py/rmgpy/molecule/atomtypeTest.py", line 222, in setUp
    9 H u0 p0 c0 {5,S}''')
  File "rmgpy/molecule/molecule.py", line 1280, in rmgpy.molecule.molecule.Molecule.fromAdjacencyList (build/pyrex/rmgpy/molecule/molecule.c:24509)
    def fromAdjacencyList(self, adjlist, saturateH=False):
  File "rmgpy/molecule/molecule.py", line 1288, in rmgpy.molecule.molecule.Molecule.fromAdjacencyList (build/pyrex/rmgpy/molecule/molecule.c:24084)
    self.vertices, self.multiplicity = fromAdjacencyList(adjlist, group=False, saturateH=saturateH)
  File "/home/travis/build/ReactionMechanismGenerator/RMG-Py/rmgpy/molecule/adjlist.py", line 703, in fromAdjacencyList
[...]

Issue #898, opened 2017-02-14
String\float confusion in ConsistencyChecker.check_partial_charge


alongd:
I had a mistake in an adjList in a new library. Since the actual charge was different than theoretical, an `InvalidAdjacencyListError` was raised, but crashed when it tried to report the atom's bonds:
```
Traceback (most recent call last):
  File "C:\Code\RMG-Py/rmg.py", line 156, in <module>
    rmg.execute(**kwargs)
  File "C:\Code\RMG-Py\rmgpy\rmg\main.py", line 529, in execute
    self.initialize(**kwargs)
  File "C:\Code\RMG-Py\rmgpy\rmg\main.py", line 395, in initialize
    self.loadDatabase()
  File "C:\Code\RMG-Py\rmgpy\rmg\main.py", line 306, in loadDatabase
    depository = False, # Don't bother loading the depository information, as we don't use it
  File "C:\Code\RMG-Py\rmgpy\data\rmg.py", line 101, in load
    kineticsDepositories
  File "C:\Code\RMG-Py\rmgpy\data\rmg.py", line 164, in loadKinetics
    depositories=kineticsDepositories
  File "C:\Code\RMG-Py\rmgpy\data\kinetics\database.py", line 104, in load
    self.loadLibraries(os.path.join(path, 'libraries'), libraries)
  File "C:\Code\RMG-Py\rmgpy\data\kinetics\database.py", line 209, in loadLibraries
    library.load(library_file, self.local_context, self.global_context)
  File "C:\Code\RMG-Py\rmgpy\data\kinetics\library.py", line 231, in load
    speciesDict = self.getSpecies(os.path.join(os.path.dirname(path),'dictionary.txt'))
  File "C:\Code\RMG-Py\rmgpy\data\base.py", line 284, in getSpecies
    species = Species().fromAdjacencyList(adjlist)
  File "rmgpy\species.py", line 193, in rmgpy.species.Species.fromAdjacencyList (build\Release\pyrex\rmgpy\species.c:4932)
  File "rmgpy\species.py", line 200, in rmgpy.species.Species.fromAdjacencyList (build\Release\pyrex\rmgpy\species.c:4583)
  File "rmgpy\molecule\molecule.py", line 1280, in rmgpy.molecule.molecule.Molecule.fromAdjacencyList (build\Release\pyrex\rmgpy\molecule\molecule.c:24509)
[...]

PR #901, opened 2017-02-16
Bond order fix

goldmanm:
I checked for all instances of `bond.order` in rmgpy, as requested by @mjliu and found two discrepancies, one of which was alluded to in #898 and #899. The other one improves the robustness of databaseTest.

### DIFF
--- a/testing/databaseTest.py
+++ b/testing/databaseTest.py
@@ -459,7 +459,7 @@ class TestDatabase():  # cannot inherit from unittest.TestCase if we want to use
                     elif num_of_Dbonds == 1:
                         for ligand, bond in atom.bonds.iteritems():
                             #Ignore ligands that are not double bonded
-                            if 'D' in bond.order:
+                            if any([abs(2-order) < 1e-7 for order in bond.order]):
                                 for ligAtomType in ligand.atomType:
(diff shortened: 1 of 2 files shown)


testing/databaseTest.py
```diff
@@ -459,7 +459,7 @@ def kinetics_checkCdAtomType(self, family_name):
                     elif num_of_Dbonds == 1:
                         for ligand, bond in atom.bonds.iteritems():
                             #Ignore ligands that are not double bonded
-                            if 'D' in bond.order:
+                            if any([2-order < 1e-7 for order in bond.order]):
```
KEHANG: should be `abs(2-order) < 1e-7` ?

goldmanm: good catch. correcting.

KEHANG:
@goldmanm can you confirm the issues mentioned are resolved by this patch? would be possible or worthwhile to add a unit test?

goldmanm:
@KEHANG, I did change the previous commit to address your recommendations. The changes in this PR are slightly secondary issues, as they are only improving the error catching and testing. I don't think these need unit tests because they aren't looking at any RMG functionality. The only thing I potentially see is ensuring that the unittests that I altered can fail when not finding double bonds, since they weren't throwing an error before (and probably should have been).

Issue #904, opened 2017-02-17
Module 'networkx' not in rmg_env by default (Very Low Priority)


ghost:
The ipython script mechanism_analyzer depends on the 'networkx' module, although it is not installed in the default rmg_env environment created by the configuration file 'environment_linux.yml'. It is an easy fix for the user (simply type 'pip install networkx' when the environment is activated) but should this be included by default?

mliu49:
Thanks for noticing! It seems that networkx is available from anaconda, so it would be as simple as adding it to the environment_[os].py files. Would you like to make the addition and submit a pull request?

amarkpayne:
Can do!

PR #905, opened 2017-02-17
The conda configuration files for the default rmg_env environment hav…

amarkpayne:
…e been

updated to include the module 'networkx' as a dependency. This module is
required by the mechanism_analyzer.ipynb script. This addition will prevent
the user from having to use pip install the first time they encounter this
error.


mliu49:
Addresses #904.

PR #907, opened 2017-02-22
Miscellaneous changes (from heterogeneous catalysis branch).

rwest:
@cfgoldsmith and I have been working on a version of RMG for heterogenous catalysis for the last ~18 months. During this process we made various small tweaks to RMG, minor optimizations, printing more debugging information, fixing docstrings, etc., that helped us along the way. 
I have cherry-picked (almost) all the commits that do not actually relate to heterogeneous catalysis into this "miscellaneous" pull request. Some changes may not be needed for exclusively gas-phase work, but I don't think any of them do it harm.  They should be reviewed independently of the RMG-Cat work, and merged into master.

We'll then try to merge the latest master back into RMG-cat -- which will surely be fun...

### DIFF
--- a/rmgpy/data/kinetics/database.py
+++ b/rmgpy/data/kinetics/database.py
@@ -420,7 +420,12 @@ library instead, depending on the main bath gas (N2 or Ar/He, respectively)\n"""
         reactionList = []
         for label, family in self.families.iteritems():
             if only_families is None or label in only_families:
-                reactionList.extend(family.generateReactions(reactants))
+                try:
+                    reactionList.extend(family.generateReactions(reactants))
+                except:
+                    logging.error("Problem family: {}".format(label))
+                    logging.error("Problem reactants: {}".format(reactants))
+                    raise
         if products:
             reactionList = filterReactions(reactants, products, reactionList)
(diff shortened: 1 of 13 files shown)


rwest:
I keep rebasing this PR. Would someone like to review and merge it? Mabye @kehang @goldmanm or @mliu49 would like to nominate someone...
I think the changes are pretty harmless.

rwest:
Good point in general (the problem with a broad "except" being that you catch and therefore hide an exception that you weren't targeting). However I think in all *these* cases I re-raise the caught exception without altering it, so it will never squash or hide an exception that it wasn't meant to catch by being too broad.  In effect it doesn't actually catch the exception at all, it just logs a few bits of debugging information as the exception flies past.

goldmanm:
It does seem like all the try/except statements do eventually raise the original exceptions, they just print off logging information (like which species/families/reactions were responsible) before the exception is finally raised. Seems like a decent method for better bug-tracking, though definitely looks suspicious at first glance.

None of the other code seemed alarming to me.

rwest:
Thanks for the code reviews, Max and Mark.

PR #911, opened 2017-02-24
refactored Reaction.degeneracy property

goldmanm:
modified the protocol for setting degeneracy to get RMG to update
the kinetics rate constant when degeneracy is changed to keep the degeneracy and
reaction rates in sync.

### DIFF
--- a/environment_linux.yml
+++ b/environment_linux.yml
@@ -8,7 +8,7 @@ dependencies:
   - numpy >=1.10.0
   - cantera >=2.2
   - coolprop
-  - cython ==0.21
+  - cython ==0.25.2
   - matplotlib >=1.5
   - rdkit >=2015.09.2
   - pydas >=1.0.1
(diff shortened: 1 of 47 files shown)


goldmanm:
This is just refactoring the code, so I think it doesn't need to be in the Feb 2017 release of RMG, though it is a small pull request and get placed in easily.

rwest:
Are you sure https://github.com/ReactionMechanismGenerator/RMG-Py/pull/911/commits/d6b26cc975f8d1db4ee69110cf53fd6325f73bfc is a good idea? See https://github.com/ReactionMechanismGenerator/RMG-Py/issues/640

goldmanm:
Thanks for the reference, I am sure that d6b26cc is a bad implementation. I am fixing it to make the other classes storing Arrhenius lists to behave like PDepArrhenius already does.

goldmanm:
This PR will be put in the larger degeneracy pull request.

PR #913, opened 2017-02-24
Run as Admin Instructions

amarkpayne:
Adds a description about running the command prompt as an administrator for windows users


rwest:
Merged with #912

PR #916, opened 2017-02-24
Split enlarge() into smaller pieces

rwest:
*Not ready to merge. Don't include in upcoming release.*

This is based on the discussion from the hackathon on February 24, 2017.
The effort was to reduce the size of the `.enlarge()` model, so that  each part does one thing and not several. This may not actually reduce complexity, but should at least make unit testing easier (as the units are smaller).  One downside is it may increase code duplication.
Here is a PR to work on it (please contribute commits - although I'm opening, I am not staking ownership of this PR!)


mliu49:
Closing because this is way too out of date. Keeping issue #916 open since it's still something we might want to do.

Issue #924, opened 2017-02-28
CanTherm: Spin orbit correction reference


bslakman:
The links for the references to SOC no longer work:
```
 # Spin orbit correction (SOC) in Hartrees
    # Values taken from note 22 of http://jcp.aip.org/resource/1/jcpsa6/v109/i24/p10570_s1 and converted to hartrees
    # Values in millihartree are also available (with fewer significant figures) from http://jcp.aip.org/resource/1/jcpsa6/v106/i3/p1063_s1
    SOC = {'H':0.0, 'N':0.0, 'O': -0.000355, 'C': -0.000135, 'S':  -0.000893, 'P': 0.0, 'Si': -0.000682, 'F': -0.0006135, 'Cl': -0.001338,}
```

I know I have accessed it before but can no longer find it and want to cite it; does anyone have this resource?

alongd:
Try these:
[http://aip.scitation.org/doi/abs/10.1063/1.477794](http://aip.scitation.org/doi/abs/10.1063/1.477794)
[http://aip.scitation.org/doi/abs/10.1063/1.473182](http://aip.scitation.org/doi/abs/10.1063/1.473182)
?

bslakman:
thank you! yeah I just realized that the volume and issue is inside the URL... I can make a pull request to fix the URL.

PR #925, opened 2017-02-28
Update the citations for spin orbit corrections.

bslakman:
Addresses #924


rwest:
Could we use the doi URLs (eg. http://doi.org/10.1063/1.477794 ) so they never need updating again? There's nothing to stop AIP changing their URL format again in a couple of years, but the DOI ones should be stable.

PR #927, opened 2017-03-01
Update the URLs for CanTherm reference for SOC to more stable doi URL

bslakman:
At the suggestion of @rwest updates the references in CanTherm for the spin orbit correction to the more permanent doi URL. Address #924


rwest:
You know you could have (force-)pushed to just update the previous pull request #925?

bslakman:
yeah, but I did the previous one from my master branch and Travis was failing, so I made this one an official branch

PR #938, opened 2017-03-09
Fix signal handling after calling lpsolve

mliu49:
The lpsolve module redirects interrupt signals to a custom handler, and this change is persistent in python until another change is made. This fix resets the handler to the one that was in use prior to calling lpsolve.

This restores the ability to send KeyboardInterrupt signals to stop an RMG job.


rwest:
*that* explains it! I thought I was going crazy. Please can we have a minor release 2.1.1 soon? ;)

mliu49:
Yeah, we all thought that it was something wrong with our computers until we realized that we were all experiencing it.

If someone could double check that it works now, I think this can be merged.

rwest:
Works for me.

PR #941, opened 2017-03-10
Nodal distances

mjohnson541:
this enables the assignment of nodalDistance values to the distance between a given node and its parent with the rate rule trees.  Default these values are 1, but in a group file default values for each tree can be set by setting the treeDistances variable 
ex: treeDistances = [1,2,5]
and the values can be directly assigned as inputs to any individual entry
ex:
Entry( .......
nodalDistance=5,
)

Kinetics averaging and estimation have been extended to work with trees where the nodes are different distances apart. 
Nodal averaging is only done between the nodes that are closest to the parent.  But this may be changed later.  

This pull also fixes some issues with the original implementation of the 2-norm (it only compared distances one step out at a time so it would pick [2,0,0] over [1,1,1])

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -393,8 +393,11 @@ class KineticsFamily(Database):
         self.ownReverse = forwardTemplate is not None and reverseTemplate is None
         self.boundaryAtoms = boundaryAtoms
         self.treeDistances = treeDistances
+        
         if self.treeDistances is None:
             self.treeDistances = np.ones(len(self.top))
+        else:
+            assert len(self.treeDistances) == len(self.top), 'treeDistances must be the same length as the number of trees in the family'
         # Kinetics depositories of training and test data
         self.groups = None
         self.rules = None
@@ -576,13 +579,13 @@ class KineticsFamily(Database):
         local_context['False'] = False
         local_context['reverse'] = None
         local_context['boundaryAtoms'] = None
-
+        local_context['treeDistances'] = None
         self.groups = KineticsGroups(label='{0}/groups'.format(self.label))
         logging.debug("Loading kinetics family groups from {0}".format(os.path.join(path, 'groups.py')))
         Database.load(self.groups, os.path.join(path, 'groups.py'), local_context, global_context)
         self.name = self.label
         self.boundaryAtoms = local_context.get('boundaryAtoms', None)
-
+        self.treeDistances = local_context.get('treeDistances',np.ones(len(self.top)))
         # Generate the reverse template if necessary
         self.forwardTemplate.reactants = [self.groups.entries[label] for label in self.forwardTemplate.reactants]
         if self.ownReverse:
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -563,7 +563,11 @@ class KineticsFamily(Database):
         if not already entered with the value from treeDistances associated
         with the tree the entry comes from
         """
+        treeDistances = self.treeDistances
         toplabels = [i.label for i in self.groups.top]
+        
+        assert len(toplabels) == len(treeDistances), 'treeDistances does not have the same number of entries as there are top nodes in the family'
+        
         for entryName,entry in self.groups.entries.iteritems():
             topentry = entry
             while not (topentry.parent is None): #get the top for the tree entry is in
@@ -571,7 +575,7 @@ class KineticsFamily(Database):
             if topentry.label in toplabels: #not exactly sure where the ones that don't match come from, they aren't in the trees, so this should be ok
                 ind = toplabels.index(topentry.label)
                 if entry.nodalDistance is None:
-                    entry.nodalDistance = self.treeDistances[ind]
+                    entry.nodalDistance = treeDistances[ind]
                 
     def load(self, path, local_context=None, global_context=None, depositoryLabels=None):
         """
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -394,10 +394,6 @@ class KineticsFamily(Database):
         self.boundaryAtoms = boundaryAtoms
         self.treeDistances = treeDistances
         
-        if self.treeDistances is None:
-            self.treeDistances = np.ones(len(self.top))
-        else:
-            assert len(self.treeDistances) == len(self.top), 'treeDistances must be the same length as the number of trees in the family'
         # Kinetics depositories of training and test data
         self.groups = None
         self.rules = None
@@ -560,7 +556,23 @@ class KineticsFamily(Database):
         ftemp.write('\n')
         
         ftemp.close()
-    
+        
+    def distributeTreeDistances(self):
+        """
+        fills in nodalDistance (the distance between an entry and its parent)
+        if not already entered with the value from treeDistances associated
+        with the tree the entry comes from
+        """
+        toplabels = [i.label for i in self.groups.top]
+        for entryName,entry in self.groups.entries.iteritems():
+            topentry = entry
+            while not (topentry.parent is None): #get the top for the tree entry is in
+                topentry = topentry.parent
+            if topentry.label in toplabels: #not exactly sure where the ones that don't match come from, they aren't in the trees, so this should be ok
+                ind = toplabels.index(topentry.label)
+                if entry.nodalDistance is None:
+                    entry.nodalDistance = self.treeDistances[ind]
+                
     def load(self, path, local_context=None, global_context=None, depositoryLabels=None):
         """
         Load a kinetics database from a file located at `path` on disk.
@@ -585,7 +597,8 @@ class KineticsFamily(Database):
         Database.load(self.groups, os.path.join(path, 'groups.py'), local_context, global_context)
         self.name = self.label
         self.boundaryAtoms = local_context.get('boundaryAtoms', None)
-        self.treeDistances = local_context.get('treeDistances',np.ones(len(self.top)))
+        self.treeDistances = local_context.get('treeDistances',None)
+        
         # Generate the reverse template if necessary
         self.forwardTemplate.reactants = [self.groups.entries[label] for label in self.forwardTemplate.reactants]
         if self.ownReverse:
@@ -605,6 +618,7 @@ class KineticsFamily(Database):
         self.rules = KineticsRules(label='{0}/rules'.format(self.label))
         logging.debug("Loading kinetics family rules from {0}".format(os.path.join(path, 'rules.py')))
         self.rules.load(os.path.join(path, 'rules.py'), local_context, global_context)
+        
         # load the groups indicated in the entry label
         for label, entries in self.rules.entries.iteritems():
             nodes = label.split(';')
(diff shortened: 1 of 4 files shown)


mjohnson541:
Documentation has been added to the wiki for this pull.

rmgpy/data/kinetics/rules.py
```diff
@@ -569,105 +583,144 @@ def __getAverageKinetics(self, kineticsList):
...
-                return kinetics, entry if 'Exact' in kinetics.comment else None
+def removeIdenticalKinetics(kList):
+    """
+    doing this based on strings, which should be fine for this specifically, since we shouldn't have any
+    identical kinetics entries in the families all of the identical kinetics should look exactly the same
+    faster than the alternative
+    """
+    outSet = set()
+    outList = []
+    for i in xrange(len(kList)):
```
rwest: As you never use `i` for anything except to look up the `k`, and you don't modify the `kList` inside the loop,  I suggest you could replace this with `for k in kList:`?

rmgpy/data/kinetics/rules.py
```diff
@@ -569,105 +583,144 @@ def __getAverageKinetics(self, kineticsList):
...
+        kinetics.A.value_si *= degeneracy
+        if degeneracy > 1:
+            kinetics.comment += "\n"
+            kinetics.comment += "Multiplied by reaction path degeneracy {0}".format(degeneracy)
+
+        return kinetics, entry if 'Exact' in kinetics.comment else None
 
-                return kinetics, entry if 'Exact' in kinetics.comment else None
+def removeIdenticalKinetics(kList):
+    """
```
rwest: This docstring doesn't describe what the method does (only discusses how it does it) or what it returns.

rmgpy/data/kinetics/rules.py
mjohnson541: Ok, this should be fixed now.

rmgpy/data/kinetics/rules.py
mjohnson541: Ok, this should be fixed now.

mjohnson541: Ok, this should be fixed now.

chiyw13:
What's the status of this pull request? Is it thought to improve estimates of "averaged" rules? Is it ready to merge? (asked by @rwest)

mjohnson541:
It's ready to merge.  The immediate consequence is that we will use the 2-norm (euclidean distance along each tree in the family) to determine the closest rate rules to a given node (what we currently use is the 1-norm (just the sum of the distances along each tree) and then filter everything with the same 1-norm by the 2-norm), which was the intended solution to #712.  Our release notes actually somewhat imply we do this currently.

mjohnson541:
Everything change I've looked at in depth in RMG-tests was related to the 2-norm change, although as you can imagine this produced a whole lot of changes.  Since the distances default to 1 no other changes should be expected when this is pulled.

mjohnson541:
The Entry and Family function headers should be fixed now.  nodalDistances is a float while treeDistances is a numpy.ndarray

chiyw13:
The wiki will need updating:
> Currently the closest rules are determined by taking the rules that minimize the 1-norm distance and then taking the rules among those that minimize the 2-norm distance. After the nodal distances update this will be done by minimizing the 2-norm distance among all rules.

https://github.com/ReactionMechanismGenerator/RMG-Py/wiki/kinetics-estimation#family-rate-estimation-process

and perhaps we should link to it from the main documentation?

rmgpy/data/kinetics/rules.py
```diff
@@ -596,82 +592,141 @@ def estimateKinetics(self, template, degeneracy=1):
...
-                            kinetics.comment += 'Estimated using an average'
-                        else:
-                            kinetics.comment += 'Exact match found' 
-                    else:
-                    # Using a more general node to estimate original template
-                        kinetics.comment +='Estimated using template ' + matchedLeaves
-                            
+            templateList0 = templateList
+            distanceList0 = distanceList
+            distanceList = []
```
nyee: Might be a good idea to comment what each of these lists are supposed to be.

rmgpy/data/kinetics/rules.py
```diff
@@ -596,82 +592,143 @@ def estimateKinetics(self, template, degeneracy=1):
...
+    """
+    removes all identical kinetics entries in kList
+    takes in a list of kinetics entries
+    returns the list with the identical kinetics entries removed
+    
+    does this based on strings, which should be fine for this specifically, since we shouldn't have any
+    identical kinetics entries in the families and all of the identical kinetics should look exactly the same
+    """
+    outSet = set()
+    outList = []
```
nyee: Don't keep track of outList. at the end just return list(outSet)

rmgpy/data/kinetics/family.py
```diff
@@ -347,7 +348,8 @@ def __init__(self,
                  reverseTemplate=None,
                  reverseRecipe=None,
                  forbidden=None,
-                 boundaryAtoms = None
+                 boundaryAtoms = None,
+                 treeDistances = None
```
mjohnson541: Added, the datatype is a float, default nodalDistance is 1 for every entry.

rmgpy/data/kinetics/rules.py
```diff
@@ -460,13 +464,22 @@ def fillRulesByAveragingUp(self, rootTemplate, alreadyDone, verbose=False):
...
             # Start with the root template, and replace the ith member with its children
             if parent.children:
                 childrenSet = [[group] for group in rootTemplate]
                 childrenSet[i] = parent.children
                 childrenList.extend(getAllCombinations(childrenSet))
-
+                distanceList.extend([k.nodalDistance for k in parent.children])
+                
+        if distanceList != []:
+            minDist = min(distanceList) #average the minimum distance neighbors
```
mjohnson541: It's the beginning of the averaging process, I've moved the comment up.

rmgpy/data/kinetics/rules.py
mjohnson541: We need outList because outList has the actual object outSet only has the strings

rmgpy/data/kinetics/rules.py
```diff
@@ -596,82 +592,141 @@ def estimateKinetics(self, template, degeneracy=1):
...
+                for i in range(len(templateList0)):
+                    if norms[i] > minNorm:
+                        toDelete.append(i)
+                toDelete.reverse()
+                for k in toDelete:
+                    del templateList0[k]
+                    del distanceList0[k]
+                        
+            
+            for i in xrange(len(templateList0)):
```
mjohnson541: should be fixed

rmgpy/data/kinetics/rules.py
mjohnson541: added

rmgpy/data/kinetics/rules.py
```diff
@@ -596,82 +592,141 @@ def estimateKinetics(self, template, degeneracy=1):
...
+        if not(entry is None) and entry.data:
+            savedKinetics = [[deepcopy(entry.data),template]]
+            templateList = []
+            
         while len(templateList) > 0:
             
             kineticsList = []
-            for t in templateList:
+            distances = []
+            for i in xrange(len(templateList)):
```
mjohnson541: fixed

rmgpy/data/kinetics/groups.py
```diff
@@ -71,7 +71,7 @@ def __init__(self,
     def __repr__(self):
         return '<KineticsGroups "{0}">'.format(self.label)
 
-    def loadEntry(self, index, label, group, kinetics, reference=None, referenceType='', shortDesc='', longDesc=''):
+    def loadEntry(self, index, label, group, kinetics, reference=None, referenceType='', shortDesc='', longDesc='',nodalDistance=None):
```
mjohnson541: added

rmgpy/data/kinetics/family.py
```diff
@@ -520,7 +524,27 @@ def saveOldTemplate(self, path):
...
+        toplabels = [i.label for i in self.groups.top]
+        
+        assert len(toplabels) == len(treeDistances), 'treeDistances does not have the same number of entries as there are top nodes in the family'
+        
+        for entryName,entry in self.groups.entries.iteritems():
+            topentry = entry
+            while not (topentry.parent is None): #get the top for the tree entry is in
+                topentry = topentry.parent
+            if topentry.label in toplabels: #not exactly sure where the ones that don't match come from, they aren't in the trees, so this should be ok
+                ind = toplabels.index(topentry.label)
```
mjohnson541: done

rmgpy/data/kinetics/family.py
```diff
@@ -520,7 +524,27 @@ def saveOldTemplate(self, path):
...
+        treeDistances = self.treeDistances
+        toplabels = [i.label for i in self.groups.top]
+        
+        assert len(toplabels) == len(treeDistances), 'treeDistances does not have the same number of entries as there are top nodes in the family'
+        
+        for entryName,entry in self.groups.entries.iteritems():
+            topentry = entry
+            while not (topentry.parent is None): #get the top for the tree entry is in
+                topentry = topentry.parent
+            if topentry.label in toplabels: #not exactly sure where the ones that don't match come from, they aren't in the trees, so this should be ok
```
mjohnson541: fixed

nyee:
@mjohnson541, I've been running RMG-tests a bit since we last talked about it. Latest one can be seen [here](https://travis-ci.org/ReactionMechanismGenerator/RMG-tests/builds/247618752).

mjohnson541:
I know we'd like to avoid cythonizing stuff in data, but the code could be numba-ized relatively painlessly see #1001.  I'll try a few profiling runs today and try to see what the bottleneck is.

mjohnson541:
Actually on second thought cythonizing might be a lot easier.  Are there any objections to cythonization of data/kinetics/rules.py?

mjohnson541:
So now that I've looked at it, numba-izing it would require me to do very nasty horrible things to the code that would devastate readablility (and the sections we need faster are hard enough to read without that).  Are there objections to cythonizing data/kinetics/rules.py?

rgillis8:
Sure. I'll look into it.

rgillis8:
Hey, sorry for the delayed reply. In the cases that I ran there was not a significant time increase associated with this branch compared to master (I ran diesel oxidation and PDD pyrolysis). 

However, I did notice that even master was running quite a bit slower than before (i.e. these models were generated much slower than models based on identical inputs from say a year ago). Perhaps there was a problem with my run, and I'll double-check.

mjohnson541:
To my understanding the issue with master is known, but we're not yet sure why it's happening.  

In running these the primary slowdown I've noticed has been in the initial tree averaging process when we fill the trees, which makes me nervous about having the same issue when searching for kinetics data in the tree (since we have to do that all the time).  If you're not observing much slowdown overall that says to me that searching the tree isn't a bottleneck and we should be good to push this.

nyee:
@mjohnson541, I agree. Is there any more rebasing you want to do?

mjohnson541:
Nope, I rebased it this morning already.  Should be good to go.

PR #943, opened 2017-03-13
Allow users to input lists of strings for reactionLibraries

goldmanm:
This PR converts a lists of reaction libraries strings to the
`('text',False)` tuples in the `rmgpy/rmg/input.py/database` method.
This allows more intuitive input files to be written and used.


KEHANG:
also modify this [file](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/examples/rmg/commented/input.py#L12)? possibly add some unittest?

goldmanm:
added both to the commented input file and created a framework to unittest `rmgpy/rmg/input.py` (which ironically isn't an input file, it is the file to process an input file) oy vey. Let me know what you think

PR #947, opened 2017-03-17
added tests for rmgpy.rmg.main

goldmanm:
tested that RMG.execute returns a model with more species and
reactions than originally input, and tested that adding cantera
file works and raises exceptions as expected


goldmanm:
@KEHANG, I modified the unittest to end up in the same directory it started in to avoid the errors in other tests.

PR #948, opened 2017-03-17
Unit tests for the diffusion limited module.

bslakman:
These two tests are for the `getEffectiveRate()` method in `rmgpy/kinetics/diffusionLimited.py`. Some associated changes were made in the module itself.

### DIFF
--- a/rmgpy/kinetics/diffusionLimited.py
+++ b/rmgpy/kinetics/diffusionLimited.py
@@ -48,13 +48,10 @@ class DiffusionLimited():
 
     def getSolventViscosity(self, T):
         return self.solventData.getSolventViscosity(T)
-              
+
     def getEffectiveRate(self, reaction, T):
         """
-        Return the ratio of k_eff to k_intrinsic, which is between 0 and 1.
-        
-        It is 1.0 if diffusion has no effect.
-        
+        Returns the effective rate of reaction, accounting for diffusion.
         For 1<=>2 reactions, the reverse rate is limited.
         For 2<=>2 reactions, the faster direction is limited.
         For 2<=>1 or 2<=>3 reactions, the forward rate is limited.
@@ -66,7 +63,7 @@ class DiffusionLimited():
         Keq = reaction.getEquilibriumConstant(T) # Kc
         k_reverse = k_forward / Keq
         k_eff = k_forward
-        
+
         if reactants == 1:
             if products == 1:
                 k_eff = k_forward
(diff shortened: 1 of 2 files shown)


goldmanm:
Thanks for writing the tests to ensure diffusion limited kinetics functions properly. To reduce rebuilding the database multiple times, I think it may be beneficial to use setUpClass/tearDownClass method used in #953 for anything the `diffusionLimiter.getEffectiveRate` does not modify, like the database creating and loading. What are your thoughts on this?

goldmanm:
This looks like a good test for diffusion limited design. comments:

1. do you think it would be beneficial to set some conservative ratio for the comparison, so the test doesn't possibly pass just due to numerical errors? For example, in the diffusion limited case, would we expect the value to always be above 50% of the original (or more conservative, 90%)? 
2. I think we could avoid multiple database loading if the we moved parts of the `setUp` method to a `setUpClass` or `setUpModule` method similar to #953. 
3. It may be possible to use a test database instead of the full one...though we can wait till the solvent database gets bigger to make a toy database for it.

Let me know your thoughts on these ideas.

bslakman:
1. I just ran example eg3 and there was at least one rate encountered during the generation that had a diffusion factor of around 0.5, with a few more around the 0.6-0.8 range, and most above 0.9.  So I would suggest setting it lower maybe... any thoughts @yunsiechung ?
2. for this class there are only two tests so far, so I don't think it takes as long as thermo or kinetics. But it could be a good idea for the other solvation (thermo) tests.
3. Same comment as 2, basically. we could load just the solvent library instead of all of the solvent/solute stuff, but I don't think it takes that long.

amarkpayne:
The "Update Copyright" commit should be squashed with the "Created a test file" commit before merging. I have held off from doing this just in case it messes up the above comments.

goldmanm:
I ran into an issue a while ago regarding the tearing down of unittests. It seemed that some attributes in rmgpy.rmg.model.Species were being set (Species.solventData ,Species.solventName , Species.solventStructure ,Species.solventViscosity) when creating a solvation database and using it in the `RMG.initialize` method. This method doesn't seem to utilize an RMG object, which indicates that your methods were written modularly (Yay!), so we don't need to add extra tearDown tasks.

goldmanm:
It looks good to me. I am just going to rebase it onto master just to ensure it is passing.

amarkpayne:
It looks like it passed, so I am going to squash my copyright commit before it gets merged

amarkpayne:
This branch can be merged, right? Are we waiting on anything?

goldmanm:
I'll update and merge it

PR #949, opened 2017-03-17
Test and fix SMILES generation and parsing

rwest:
This started out being just unit tests for SMILES, but I made some unit tests that were failing
(always the best kind) and then fixed them, so this PR now has bug fixes in as well as new unit tests.

### DIFF
--- a/rmgpy/molecule/parserTest.py
+++ b/rmgpy/molecule/parserTest.py
@@ -269,6 +269,26 @@ class ParserTest(unittest.TestCase):
         self.compare(adjlist, smiles)
 
 
+        # Test CH, methylidyne.
+        # Wikipedia reports:
+        # The ground state is a doublet radical with one unpaired electron,
+        # and the first two excited states are a quartet radical with three
+        # unpaired electrons and a doublet radical with one unpaired electron.
+        # With the quartet radical only 71 kJ above the ground state, a sample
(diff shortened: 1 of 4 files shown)


rmgpy/molecule/parserTest.py
```diff
@@ -48,6 +59,264 @@ def test_fromSMILES(self):
...
+        #
+        # Should we make the doublet? For now, this is a regression test,
+        # because we currently make the quartet.
+        adjlist = '''
+        multiplicity 4
+        1 C u3 p0 c0 {2,S}
+        2 H u0 p0 c0 {1,S}
+        '''
+        smiles = '[CH]'
+        self.compare(adjlist, smiles)
```
KEHANG: since this PR fixes `[C]`, maybe we should fix this `[CH]` as well? I'll add one commit shortly.

KEHANG:
I've just checked what we have in database for [CH]. 

For related reaction families, they always have both doublet and quartet versions, and almost no rules. 

For thermo group, we don't have [CH]. 

For the thermo libraries, we always use quartet and no doublet data (I'll change quartet into doublet in a database PR).

KEHANG:
after discussing with Alon, this is more complicated than I thought. I think we should merge this PR without [CH] fix at this moment and thoroughly fix [CH] in a new PR.

PR #952, opened 2017-03-18
Add unit tests to chemkinTest

mliu49:
I started out trying to write tests for each untested line, but I realized that it wasn't really helping. A major untested part was writing chemkin files, so I added some additional checks to existing tests. The end result was only +6% coverage within chemkin.py. To increase coverage more, we need to test reading and writing the other kinetics formats besides Arrhenius and Chebyshev.

### DIFF
--- a/environment_linux.yml
+++ b/environment_linux.yml
@@ -37,3 +37,4 @@ dependencies:
   - ffmpeg
   - jupyter
   - networkx
+  - mock
--- a/environment_mac.yml
+++ b/environment_mac.yml
@@ -36,3 +36,4 @@ dependencies:
   - ffmpeg
   - jupyter
   - networkx
+  - mock
--- a/environment_windows.yml
+++ b/environment_windows.yml
@@ -35,3 +35,4 @@ dependencies:
   - ffmpeg
   - jupyter
   - networkx
+  - mock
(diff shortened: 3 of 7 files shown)


KEHANG:
@mliu49 not sure what `mock` is doing here. Can you explain why we need add as a new RMG dependency?

mliu49:
mock allows creating fake instances of python modules. In this case, mock is used to create a fake logging module to intercept calls to logging.warning. This allows unit testing that specific logging statements are actually created.

In this case, the unit tests I added aren't particularly important, but I had written them in my initial attempt to cover line-by-line. I quickly realized that that wasn't a particularly good way to go about it, but I kept the tests anyway.

If you don't think those tests are worth keeping, then I'm fine with not adding them. However, I think mock does have future value in unit testing.

rwest:
I like it. But perhaps add a few docstrings or comments in the unit test to help others understand what on earth is going on?  To someone who's not worked with `mock` before, it's not particularly obvious.  And it might be something we want to replicate elsewhere.

mliu49:
I will add more info about `mock` to the docstrings.

KEHANG:
@mliu49 ready to merge?

Issue #887, opened 2017-02-09
Failing Thermo/Symmetry Tests


mliu49:
I noticed a few unit tests that were marked @work_in_progress by @rwest almost 3 years ago that seem fairly important in rmgpy.data.thermoTest.

I made some changes to have the test work properly again, and found that they are now failing due to differences in our calculations. Please see the [fixThermoTest](https://github.com/ReactionMechanismGenerator/RMG-Py/commits/fixThermoTest) branch for changes. It's possible that this was the original reason that the tests were marked @work_in_progress, and changes to RMG caused the error to shift to syntax problems.

Specifically, the S298 values for cyclopentadiene and naphthalene are different from the benchmark values, and their symmetry numbers are also wrong. For S298, I'm not sure if the benchmarks need to be updated, or if we're just getting it wrong.

rwest:
Just noticed my commit message from that commit:
```
Add @work_in_progress annotations to some failing thermoTests

We mustn't get lazy and forget to fix these...
```

Erm..., hahaha?

goldmanm:
I have looked at the commented out symmetry tests. I think most of them exist there since we currently don't have an algorithm that correctly finds the symmetry. Keeping them there as a sort of 'reminder' might have been the intent...though on another level it also hides the errors. Some issues are also posted which related directly to commented out tests (#858, #844, #332, #119). I also wrote more tests for cyclics on my branch, but since I wasn't revamping the algorithm, I commented them out.

Do we have a preferred way to identify these molecules as issues in our symmetry algorithm?

goldmanm:
#1014 fixes some of the commented out tests (and adds a few more for special cases RMG still doesn't get properly).

PR #953, opened 2017-03-18
Fix database loading in RMG-Py unit tests

mliu49:
This PR changes database loading in kineticsTest to be done once per class and thermoTest to be done once for the entire module. Warnings about multiple database instances no longer appear.

I was not able to change the loading in kineticsTest to be done once for the entire module because the classes now load different databases. TestReactionDegeneracy loads the actual RMG-database, while TestKineticsParsing loads a toy database located in RMG-Py.

I've also included an old fix for some work-in-progress unit tests in thermoTest so that they work properly, although they still fail. See #887 for more details.

### DIFF
--- a/rmgpy/data/kinetics/kineticsTest.py
+++ b/rmgpy/data/kinetics/kineticsTest.py
@@ -11,13 +11,46 @@ from rmgpy.species import Species
 from rmgpy.data.rmg import RMGDatabase
 ###################################################
 
+def setUpModule():
+    """A function that is run ONCE before all unit tests in this module."""
+    global database
+    database = RMGDatabase()
+    database.load(
+        path=os.path.join(settings['test_data.directory'], 'testing_database'),
+        thermoLibraries=['primaryThermoLibrary'],
+        reactionLibraries=[],
+        kineticsFamilies=[
+            'R_Recombination',
+            'Disproportionation',
+            'R_Addition_MultipleBond',
+        ],
+        testing=True,
+        depository=False,
+        solvation=False,
+    )
+    database.loadForbiddenStructures()
+
+    # Prepare the database by loading training reactions and averaging the rate rules
+    for family in database.kinetics.families.values():
+        family.addKineticsRulesFromTrainingSet(thermoDatabase=database.thermo)
+        family.fillKineticsRulesByAveragingUp(verbose=True)
+
+
+def tearDownModule():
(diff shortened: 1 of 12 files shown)


goldmanm:
I like the usage of the setUpClass / setUpModule for reducing database loading. Do you think this would be beneficial to implement in other unittests too (maybe for another PR)? I am thinking of modifying the kineticsTest to only use disproportionation or creating single-node instances of r-recombination and R_addition_multiple_bond, so that we don't need to load multiple databases or load from training reactions. Do you think adding test-databases & refactoring this method should be part of this PR or a separate one?

rwest:
This is being discussed in a number of pull requests, and so perhaps a comment here is the wrong place, but it just occurred to me: because the database, once loaded, is (often) stored as a module level variable that can be accessed via `import`, we could probably put a `setUp` method in many of these classes that detects if it's already been loaded and only re-loads it if needed. That way, if you run an individual unit test that needs the database, it loads it, but otherwise you could run all the unit tests (kinetics, thermo, diffusion, etc. etc.) and only load the database once. 
Just a thought. (And not intended to delay merging of this particular pull request)

mliu49:
@goldmanm I think changing degeneracyTest to also use the toy database is a good idea. It could probably go in this PR as well.

@rwest I think that's a good idea, but might be a bit challenging to do right now. Since some tests are loading the toy database and others are loading the real database, we would also need to determine which one is the "currently loaded" one.

KEHANG:
@mliu49 @goldmanm is anyone still working on this PR (I guess, only missing piece is to make TestDegeneracy use toy db as suggested)?

goldmanm:
@KEHANG, I'm working on this PR. This will also be adding families to the test_database so I can study some reactions in the class as well.

goldmanm:
@KEHANG, I made the changes which were requested

mliu49:
@goldmanm, do you have more plans for this PR? If not, I think I'll squash the two kineticsTest commits.

goldmanm:
I think it should be ready for code review. We could possibly 
incorporate more unittests into the framework, but then again, that 
could be for another PR.

KEHANG:
@mliu49 @goldmanm I'll review and merge it today.

mliu49:
Let me know when you're done reviewing. I want to rebase and squash before merging.

KEHANG:
@mliu49 It looks good to me. You can prepare the commits and merge it directly.

PR #968, opened 2017-03-24
Benzene Bond Reactivity (Resonance Part 3?)

mliu49:
The main feature that this PR brings the the ability to react benzene bonds. This makes writing organizing groups in kinetics families easier, and also reduces the number of resonance structures needed to represent  the reactivity of an aromatic molecule.

The difficulty in reacting benzene bonds is the fact that modifying one bond requires changes to all bonds in the ring, which means that each change would hypothetically need to be included in the reaction recipe. However, those changes can instead be determined on the fly, only based on knowing which benzene bonds were originally incremented or decremented. This allows full compatibility with existing reaction recipes and kinetics families.

To accomplish this, a new kekulization module has been introduced. Previously, RMG relied on RDKit for kekulizing molecules. This new module removes that dependency and also enables handling of modified benzene bonds. Thus, if a benzene bond is modified, the kekulization method will automatically take that into account. The kekulization algorithm is primarily based on atom valency calculations, with additional heuristics for situations where there are multiple valid choices.

Until benzene bond groups are added to the appropriate kinetics families, this change should have no effect on model generation. However, I've already been running models with those changes, so I know that it works.

Other changes:
- Functionality to provide foundation for degeneracy changes
    - Add atom ID attribute for atom tracking purposes
    - Add isIdentical method for molecule comparison (require matching atom IDs)
    - Add keepIsomorphic method for generateResonanceStructures which uses isIdentical as the criteria for removing duplicates
- Resonance fixes
    - Use `getAllCyclesOfSize()` instead of `getSSSR()` for ring retrieval to improve robustness.
        - For bridged polycyclics, the SSSR is not deterministic and can return different sets of rings. Using `getAllCyclesOfSize()` gives more consistent output.
    - Improve generation of aromatic resonance structures when one ring in a polycyclic aromatic species is falsely identified as aromatic.
        - Previously, if one ring was falsely identified as aromatic, then the entire molecule would not be made aromatic. Now, the algorithm tries to check ring by ring.
    - Also apply other resonance algorithms to aromatic species (lone pair/radical and N5dd/N5t).
        - Aromatics were not being checked for these other resonance types after special processing.
    - Add OpenBabel as fallback option for aromaticity perception if RDKit fails.
        - Aromaticity detection would fail when encountering an aromatic species with nitrogen with valence>3 since RDKit cannot handle it.


rmgpy/molecule/kekulize.pyx
```diff
@@ -0,0 +1,290 @@
...
+#   DEALINGS IN THE SOFTWARE.
+#
+################################################################################
+
+"""
+This module contains functions for kekulization of a aromatic molecule.
+
+The basic algorithm is as follows:
+1. Identify all aromatic rings in the molecule, based on bond types.
+2. For each ring, identify endocyclic and exocyclic bonds.
```
goldmanm: I am a little confused about the usage of endocyclic and exocyclic in this file. [this source](http://encyclopedia2.thefreedictionary.com/endocyclic+double+bond) mentions that endocyclic are always in a ring. This file seems to state that benzene bonds can be exocyclic, which seems contradictory to the source. Maybe I am mistaken about the code functionality, or maybe you're using a different definition (like part of 2 rings). Eitherway,  it might be good to elaborate on this.

rmgpy/molecule/kekulize.pxd
```diff
@@ -0,0 +1,30 @@
...
+#   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+#   FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+#   DEALINGS IN THE SOFTWARE.
+#
+################################################################################
+
+from .molecule cimport Atom, Bond, Molecule
+
+
+cpdef kekulize(Molecule mol)
```
goldmanm: Maybe I am confused about cython. Does not listing the other methods of `kekulize.pyx` in `kekulize.pxd` mean that the other methods are private? If so, I am wondering should this private status be made clear in `kekulize.pyx`, so readers don't get confused into thinking that they can find aromaticity with `isinstance(bond, AromaticBond)` instead of `bond.isBenzene()` in other parts of the code?

rmgpy/molecule/molecule.py
```diff
@@ -1829,4 +1830,13 @@ def getDeterministicSmallestSetOfSmallestRings(self):
...
 
-        return cycleList
\ No newline at end of file
+        return cycleList
+
+    def kekulize(self):
+        """
+        Kekulizes an aromatic molecule.
+        """
+        try:
```
goldmanm: I like how you wrote this to avoid any deep copying in this method.

rmgpy/molecule/kekulize.pyx
mliu49: Those terms are defined relative to a particular ring: the bonds that form the specified ring are endocyclic, and all the remaining bonds are exocyclic. For example, in naphthalene, each ring has its own set of 6 endocyclic bonds and 6 exocyclic bonds, and two of the exocyclic bonds of one ring are actually endocyclic bonds of the other ring.

rmgpy/molecule/kekulize.pxd
mliu49: Yes, the unlisted methods cannot be imported into another module. Most of the helper classes and functions are also defined with `cdef`, which means they can't be imported into a Python module at all, although I think they could be imported into a cython module via `cimport`. I could make it clearer that `kekulize()` is the only method that should be used outside of the module.

goldmanm:
For pieces of code that I am working on, I am a little concerned about keeping the robustness of the `keepIsomorphic` option when using `species.generateResonanceStrucutres`. Currently, the generation algorithm just checks to ensure than `len(spec.molecule) == 1` before running the generation method in `molecule.generateResonanceStructures`. For example, if someone ran `species.generateResonanceStrucutres(keepIsomorphic = False)` followed by `species.generateResonanceStrucutres(keepIsomorphic = True)`, the second method would not get isomorphic resonance structures. I guess not a concern in this PR, but if you have an idea of how to make that more robust, that would be helpful. I was thinking about adding two boolean variables to the Species class (one for whether structures were already generated and another for whether those structures were isomorphic, and I am open to improvements on this if you have any.

rmgpy/molecule/resonance.py
```diff
@@ -482,14 +484,52 @@ def generateAromaticResonanceStructures(mol, features=None):
...
+        # Change bond types to benzene bonds for all aromatic rings
         for ring in aromaticBonds:
             for bond in ring:
                 bond.order = 1.5
 
         try:
             mol0.updateAtomTypes(logSpecies=False)
         except AtomTypeError:
-            continue
+            # If this didn't work the first time, then there might be a ring that is not actually aromatic
```
goldmanm: Do you know if there is a robust way to test this exception for a structure that rdkit always fails on? Are we already testing this somewhere? I didn't notice it explicitly in other commits?

mliu49:
With respect to the `keepIsomorphic` option for species objects, it hasn't been included in this PR because it wasn't necessary. See the following commits for my implementation: [c34059](https://github.com/mliu49/RMG-Py/commit/c34059ea064c7983246082fa2d2a48921f7c0d2c) and [859014](https://github.com/mliu49/RMG-Py/commit/859014b29dffac0bdd5d80b83ec7b5c513d2e01b). Is this what you were thinking about?

rmgpy/molecule/kekulize.pyx
```diff
@@ -0,0 +1,290 @@
...
+    cdef AromaticBond item, x
+    for item in aromaticList:
+        item.update()
+    return aromaticList.sort(key=lambda x: (x.endoDOF, x.exoDOF), reverse=True)
+
+cdef class AromaticRing:
+    """Helper class containing information about a single aromatic ring in a molecule."""
+    cdef list atoms
+    cdef set endoBonds, exoBonds
+    cdef public int endoDOF, exoDOF
```
mliu49: Again, endo refers to bonds that comprise the ring, and exo refers to bonds that are connected to the ring, but not a part of it. DOF stands for degrees of freedom, and refers to the number of bonds that do not have a set bond order yet.

rmgpy/molecule/kekulize.pyx
```diff
@@ -0,0 +1,290 @@
...
+            # We've hit the iteration limit, but could not solve the ring
+            return False
+
+        return True
+
+cdef class AromaticBond:
+    """Helper class containing information about a single aromatic bond in a molecule."""
+    cdef Bond bond
+    cdef set ringBonds
+    cdef public int endoDOF, exoDOF
```
mliu49: I'll add more documentation to the docstrings.

rmgpy/molecule/group.py
```diff
@@ -858,16 +858,17 @@ class Group(Graph):
     Corresponding alias methods have also been provided.
     """
 
-    def __init__(self, atoms=None, multiplicity=None):
+    def __init__(self, atoms=None, props=None, multiplicity=None):
         Graph.__init__(self, atoms)
-        self.multiplicity = multiplicity if multiplicity else []
+        self.props = props or {}
+        self.multiplicity = multiplicity or []
```
mliu49: It mirrors the `props` attribute of `Molecule()`, which I just found was added by Nick to mirror the `props` attribute of `Species()`. It's just a placeholder for arbitrary properties/flags that would be suitable for a dictionary format.

rmgpy/molecule/moleculeTest.py
```diff
@@ -2007,6 +2007,59 @@ def testArylRadicalBirad(self):
         mol = Molecule(SMILES='[CH2]c1c[c]ccc1')
         self.assertFalse(mol.isArylRadical())
 
+    def testIdenticalTrue(self):
+        """Test that the isIdentical method works properly"""
```
mliu49: Personally, I prefer short method names and detailed docstrings, since docstrings are much easier to read. I don't know if there's a recommended practice.

rmgpy/data/kinetics/familyTest.py
```diff
@@ -1,22 +1,24 @@
...
+from rmgpy.data.kinetics.database import KineticsDatabase
+from rmgpy.data.kinetics.family import ReactionRecipe
+from rmgpy.molecule import Molecule
 ###################################################
 
 class TestFamily(unittest.TestCase):
 
-
-    def setUp(self):
+    @classmethod
```
mliu49: I don't think there's a need to tear down, since this test class only loads a `KineticsDatabase()` instance, which does not interact with the global `Database` variable. Let me know if you think otherwise though.

rmgpy/molecule/resonance.py
mliu49: The last two tests in 7cc13aa6f2e8edcbd98d561554d8c3e8ec4faf27 are cases where the first `updateAtomTypes()` call fails with an `AtomTypeError`, and are designed to test what happens after. There are also [three tests](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/molecule/resonanceTest.py#L81) that are already in Master which are cases which hit an `AtomTypeError` and ultimately return an empty list.

rmgpy/molecule/kekulize.pyx
goldmanm: Ok. thanks for the explanation about degrees of freedom.

rmgpy/molecule/kekulize.pyx
goldmanm: Sounds good.

rmgpy/molecule/group.py
nickvandewiele: well said, @mliu49 !

rmgpy/molecule/kekulize.pxd
goldmanm: Is it possible to still write unittests for specific methods within `kekulize.pyx`?

rmgpy/molecule/group.py
goldmanm: thanks for the explanation.

rmgpy/data/kinetics/familyTest.py
goldmanm: That makes sense. Thanks for clarifying that.

goldmanm:
I was thinking about possibly using the species.props attribute to store whether or not we have already generated resonance isomers (and at what conditions for `keepIsomorphic` and `keepInitial`. Regenerating the structures need only be done when we ask for different parameters in the method.

mliu49:
I suppose it won't be useful for typical users. I can change it to debug level instead.

I don't know any reactions that will throw this error right now. I've already addressed the cases that I've encountered, which are already reflected in the unit tests.

PR #969, opened 2017-03-27
Surface algorithm (NewCriterion part 2)

mjohnson541:
This extends and improves the NewCriterion framework by adapting the RMG core-edge algorithm.  

The primary issue with NewCriterion was that it treated all species equally.  So while NewCriterion pulled in new species needed to get the concentrations of important species correct it also brought in the species needed to get the concentrations of these new less important species correct and so on.  This feedback loop made it infeasible to set the reaction tolerance criterion to anything much smaller than 10.0.  

This pull extends the core-edge framework with an additional partition referred to as the "surface".  The surface is entirely contained within the core (any species or reaction in the surface is also in the core).  Species that are in the core, but not in the surface are referred to as being part of the "bulk core".  

The original species-flux criterion operates the same as before moving species between the edge and the bulk core.  As before there is also a reaction criterion for moving reactions from edge to bulk core.  However, now a lower tolerance can be set for moving reactions to the surface under a "no layering constraint".  The "no layering constraint" requires that any reaction being moved to the surface must either have all reactants in the bulk core or all products in the bulk core.  This prevents species only connected by the less important surface species from being pulled in.  

There are also additional criteria describing requirements for movement from the surface to the bulk core.  However, they operate the same as those for moving species/reactions from the edge to the bulk core.  

The idea is that this pull will both allow use of lower tolerances for the new reaction criterion and improve the species/reaction selectivity of RMG.  

This pull is branched off the NewCriterion pull.


goldmanm:
To add a note, the pull request this is off of is #874. 

A few questions/comments:

1. I don't see any unittests that check to ensure the surface is working. I would love a better, robust convergence system, which may require unittests to ensure it functions as desired.
2. We might want to have documentation on the website, both about making an input file and about the theory behind it. 
3. Does this change require any modifications to input files?
4. Do we need both the old criteria and the new criteria? Is your method sufficient on it's own. 

I'd like to better understand the insight behind the new algorithm. This PR seems to add two more tolerances, which makes me think it could complicate making mechanisms even more than the current system, which I don't understand functionally. Is there any other place a similar algorithm is being used, which I can read as a reference?

mjohnson541:
It absolutely needs unittests for the surface initialization and and the maximum surface index location functions.  While I have done extensive logging tests to determine that they work currently I haven't made the unittests yet, because I don't need them right now and the pull this is based on, NewCriterion, hasn't been pushed yet and I'd rather do that once NewCriterion is finalized.  Tests for those will absolutely be added before I would recommend pushing this.  

This will certainly need some regression tests, is there anything else you think needs a unittest?

Yes, technically there are actually 6 new tolerances, and I will need to write at least a page of text on how to use it properly.  I just don't feel like it'd be responsible to add it right now when I don't anticipate this change being pushed soon.  

On this branch any input file that works on the current master should run in the same way.  Although, I did make one name change in input.py from NewCriterion to making the naming convention make more sense.  Essentially the defaults leave you with something equivalent to the original algorithm.  

We need both, they each catch chemistry/dynamics the other one doesn't.  

Sorry, this isn't based on any reference.  While it does work, the concept is incredibly confusing, and it will absolutely complicate mechanism generation for those that use it.  I'm planning spend my talk in May explaining how it works and how to use it, but if you're interested just come by and I can explain it, doing it justice over Github is going to be difficult.

mjohnson541:
Just added unitttests for surface initialization and the surface layering constraint.

mjohnson541:
This pull request is continued within AcceleratorTools

PR #970, opened 2017-03-27
Add script to generate html from Chemkin input

mliu49:
Currently users can either set `generateOutputHTML=True` to generate HTML files during the RMG run, or use the website to do so after model completion. This provides a third option to generate the file locally after model completion.

Using `generateOutputHTML=True` slows down model generation, and is inconvenient when the user has to transfer the result from the server to their local machine. Using the website can also be slow or unreliable. These issues are exacerbated for extremely large models.


goldmanm:
Is this code completely separate from the website code? I am of the inclination that the website and RMG could share many more methods than currently shared. Though I am not that knowledgeable about the downsides to doing this. It just seems unsustainable to have multiple methods that seem to accomplish arguably similar tasks.

mliu49:
The website calls [`chemkin.saveHTMLFile`](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/chemkin.py#L1261), which this script copies in order to allow more flexibility with file naming. The original method seems to be for working with RMG-Java output, so I didn't want to modify it since I don't know what else uses it.

goldmanm:
It seems like you've already looked that possibility quite a bit. There is a philosophy 'if it ain't broken, don't fit it'. Maybe you might want to write in the RMG-Website code that an almost duplicate method exists, so that when it does break on the website side, the person in charge will have the knowledge that they could implement it here.

Also, do you think its' worth the effort having a functional test for this script, essentially ensuring that things are being written correctly to folders? This method doesn't seem core to RMG's functionality, so it seems less vital.

KEHANG:
@mliu49 can you add some command examples to use this script in the docstring, like `python $RMGPy/scripts/generateChemkinHTML.py /path/to/your/mechanism/chem.inp /path/to/your/mechanism/species_dictionary.txt /path/to/output`

rwest:
What about adding documentation in http://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/modules/index.html ?

mliu49:
Good point! I'll add a documentation page.

Issue #965, opened 2017-03-24
Representation of singlet\triplet bi-rads


alongd:
The  [N]=C=[N] bi-rad is an important species for Prompt NO models.

The ground state of [N]=C=[N] is a triplet, as expected: [↑N]=C=[↑N]
However, another state would be a **singlet bi-rad**: [↑N]=C=[↓N]
(where the different electron spins cancel each other out)

Sources: [M.C. Lin 2010](http://www.sciencedirect.com/science/article/pii/S0082078400806529), [M.C. Lin 2007](http://pubs.acs.org/doi/abs/10.1021/jp069038%2B).

We currently don't have a way to represent this difference in RMG. Surprisingly, RMG doesn't give an error for the following AdjList:
```
multiplicity 1
1 N u1 p1 c0 {2,D}
2 C u0 p0 c0 {1,D} {3,D}
3 N u1 p1 c0 {2,D}
```
which might suggest that it is possible to represent these two species uniquely. But as was pointed out by @mliu49, there's a concern that the multiplicity line could be overwritten when running RMG.

Running a test with both species in the input file resulted in the following error:
```
Traceback (most recent call last):
  File "D:\Code\RMG-Py\/rmg.py", line 156, in <module>
    rmg.execute(**kwargs)
  File "D:\Code\RMG-Py\rmgpy\rmg\main.py", line 529, in execute
    self.initialize(**kwargs)
  File "D:\Code\RMG-Py\rmgpy\rmg\main.py", line 371, in initialize
    self.loadInput(self.inputFile)
  File "D:\Code\RMG-Py\rmgpy\rmg\main.py", line 208, in loadInput
    readInputFile(path, self)
  File "D:\Code\RMG-Py\rmgpy\rmg\input.py", line 411, in readInputFile
    reactionSystem.convertInitialKeysToSpeciesObjects(speciesDict)
  File "rmgpy\solver\simple.pyx", line 117, in rmgpy.solver.simple.SimpleReactor.convertInitialKeysToSpeciesObjects (build\Release\pyrex\rmgpy\solver\simple.c:2349)
KeyError: 'NCN'
```

mliu49:
I looked into the multiplicity checking, and it seems that we simply raise an error if the multiplicity is not valid. However, multiplicity checking for species with more than 2 unpaired electrons has not been implemented, which is probably why you were able to visualize them without any issues on the website.

[...]

PR #973, opened 2017-03-28
Elaborating the multiplicity check for nRad = 3 or 4

alongd:
This PR relates to https://github.com/ReactionMechanismGenerator/RMG-Py/issues/965
The multiplicity check was elaborated for up to 4 radicals in a molecule.

### DIFF
--- a/rmgpy/molecule/adjlist.py
+++ b/rmgpy/molecule/adjlist.py
@@ -102,10 +102,8 @@ class ConsistencyChecker(object):
     @staticmethod
     def check_multiplicity(nRad, multiplicity):
         '''
-        Check if the parameter multiplicity is an odd number,
-        since the multiplicity should comply with the formula
-        
-        m = 2s + 1, with s the sum of the spin [+/- 1/2) ] of the unpaired electrons
+        Check that the multiplicity complies with the formula: m = 2s + 1,
+        where s is the sum of the spin [+/- (1/2) ] of the unpaired electrons
         
         For a simple radical (nRad = 1): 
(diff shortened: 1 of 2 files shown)


goldmanm:
This looks good to me. I especially like the simplification of `2*nRad/2` to `nRad` and the updated doc string.

mliu49:
Would you like to add a couple of unit tests for molecules with 3 or 4 radicals?

alongd:
I added a test for check_multiplicity.
I was looking for the opposite of assertRaises(), and I understood that simply not asserting anything is the way to go, and in case an error is raised, it will show up.
Let me know if there's a better implementation.

mliu49:
Could you move the test into the `TestConsistencyChecker` class instead of being under `if __name__ == '__main__':`?

I think the best way might be to replicate `test_check_hund_rule_success()` and do
```
try:
    rad1 = ...
except InvalidAdjacencyListError:
    self.fail('InvalidAdjacencyListError thrown unexpectedly for {0}!'.format(rad1))
```
for each of your cases separately.

mliu49:
Ahh, sorry! The code sample I gave doesn't work because if an error is encountered, then you won't be able to use the molecule in the failure message. You can either leave the message as 'InvalidAdjacencyListError thrown unexpectedly!' or add an explicit description of the molecule.

However, that still means that the nitrogen case is raising an InvalidAdjacencyListError for some reason.

alongd:
Thanks for the comments, @mliu49 !
I think this is ready now.

PR #974, opened 2017-03-28
Resonance Fixes

mliu49:
I'm reopening this pull request, since I still think these resonance fixes have higher priority than the react benzene changes, so they should be reviewed separately and merged first. I've included the new kekulizr module in this pull request because some of the fixes rely on it.

Changelog:
- Add kekulize module for generating Kekule structures.
- Add `getAllSimpleCyclesOfSize()` method to only retrieve monocyclic rings.
- Use `getAllSimplesCyclesOfSize()` instead of `getSSSR()` for ring retrieval to improve robustness of aromaticity detection.
    - For bridged polycyclics, the SSSR is not deterministic and can return different sets of rings. Using `getAllSimpleCyclesOfSize()` gives more consistent output.
- Improve generation of aromatic resonance structures when one ring in a polycyclic aromatic species is falsely identified as aromatic.
    - Previously, if one ring was falsely identified as aromatic, then the entire molecule would not be made aromatic. Now, the algorithm tries to check ring by ring.
- Also apply other resonance algorithms to aromatic species (lone pair/radical and N5dd/N5t).
    - Aromatics were not being checked for these other resonance types after special processing.
- Add OpenBabel as fallback option for aromaticity perception if RDKit fails.
    - Aromaticity detection would fail when encountering an aromatic species with nitrogen with valence>3 since RDKit cannot handle it.
- Add some more error handling for generating Clar structures, also add flag to more easily disable Clar structure generation.
    - Anticipated application for the flag is in reaction isomorphism checking, where Clar structures are not particularly helpful.

### DIFF
--- a/rmgpy/molecule/generator.pxd
+++ b/rmgpy/molecule/generator.pxd
@@ -17,7 +17,7 @@ cpdef str toSMARTS(Molecule mol)
 
 cpdef str toSMILES(Molecule mol)
 
-cpdef toOBMol(Molecule mol)
+cpdef toOBMol(Molecule mol, bint returnMapping=*)
 
 cpdef toRDKitMol(Molecule mol, bint removeHs=*, bint returnMapping=*, bint sanitize=*)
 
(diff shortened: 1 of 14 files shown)


mliu49:
I forgot that the current kekulization method loses lone pair information, which is causing the resonance test to fail. I think I'll just leave the reactBenzene pull request as is then, unless anyone has other suggestions.

goldmanm:
These files seem good. There is a slight change in the rmg-tests for methyl-cyclohexane. Though this seems to be caused by issues in RMG-tests, not this branch.

PR #989, opened 2017-04-12
Add Comment Stating that Annotated Chemkin File is Necessary

amarkpayne:
Instructs user that they must use the annotated chemkin file.

### DIFF
--- a/ipython/localUncertainty.ipynb
+++ b/ipython/localUncertainty.ipynb
@@ -34,6 +34,8 @@
    "outputs": [],
    "source": [
     "# Define the CHEMKIN and Dictionary file paths.  This is a reduced phenyldodecane (PDD) model.\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


mliu49:
FYI, you don't need to open a new pull request if you just push to the same branch. You can either add a fixup commit to be squashed later, or directly force push on top of the old commits.

ipython/localUncertainty.ipynb
```diff
@@ -292,7 +294,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython2",
-   "version": "2.7.11"
+   "version": "2.7.13"
```
KEHANG: please don't include this kind of change

amarkpayne: Sorry, iPython notebook must have made this change. I'll fix this

PR #994, opened 2017-04-17
Add fallback for transport search if group additivity fails

mliu49:
Use `getTransportPropertiesViaLennardJonesParameters`, like would be done in a normal RMG job.

This fixes issues with generating transport properties on RMG-website.


rwest:
I know this commit doesn't introduce the broad `except` catch, but can we use it as an opportunity to remove it? i.e. restrict the exceptions we catch to ones that we expect and wish to fix?   Perhaps that's too hard, but would be nice if we could do it. (eg. is it a KeyError? DatabaseError? etc?)

PR #997, opened 2017-04-18
Accelerator Tools (AcceleratorTools part 3)

mjohnson541:
This represents part 3 of the AcceleratorTools branch.  

This pull request contains a number of tools for manipulating the RMG algorithm:

1) Model Concatenation:  model() and simulator() inputs from the input file are now appended to a list and iterated through in main.  This allows model runs to be done one after another (i.e. run it at certain parameters at first and then after the first run terminates start the next run with the model produced by the first run.  

2) Max Number of Species Termination Criterion:  this setting causes a run to terminate when the number of species reaches a set number 

3) Number of Objects Returned from Simulate Setting:  this setting causes simulate runs to continue until they have a set number of objects contained with invalidObject rather than terminating when invalidObject has one object in it, this allows for collection of more objects per simulation speeding up the rmg run.  

4) Additionally settings from model() and simulator() are now held in objects ModelSettings and SimulatorSettings, which makes the syntax for simulator calls much simpler

This was built on surfaceAlgorithm #1054  for convenience since surfaceAlgorithm involves lots of new settings.

### DIFF
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -166,6 +166,7 @@ class RMG(util.Subject):
         self.reactionModel = None
         self.reactionSystems = None
         self.database = None
+        self.reactionSystem = None
         
         self.fluxToleranceKeepInEdge = 0.0
         self.fluxToleranceMoveToCore = 1.0
@@ -591,6 +592,7 @@ class RMG(util.Subject):
             numCoreSpecies = len(self.reactionModel.core.species)
             for index, reactionSystem in enumerate(self.reactionSystems):
                 
+                self.reactionSystem = reactionSystem
                 # Conduct simulation
                 logging.info('Conducting simulation of reaction system %s...' % (index+1))
                 prune = True
@@ -644,25 +646,11 @@ class RMG(util.Subject):
                 # If simulation is invalid, note which species should be added to
                 # the core
                 if obj:
-                    if isinstance(obj, PDepNetwork):
-                        # Determine which species in that network has the highest leak rate
-                        # We do this here because we need a temperature and pressure
-                        # Store the maximum leak species along with the associated network
-                        obj = (obj, obj.getMaximumLeakSpecies(reactionSystem.T.value_si, reactionSystem.P.value_si))
-                        objectsToEnlarge.append(obj)
-                    elif isinstance(obj, Species):
-                        objectsToEnlarge.append(obj)
-                        assert len(objectsToEnlarge)>0
-                    elif isinstance(obj,Reaction):
-                        potentialSpcs = obj.reactants+obj.products
-                        #remove species already in core
-                        neededSpcs = [x for x in potentialSpcs if x not in self.reactionModel.core.species]
-                        for i in range(len(neededSpcs)): #remove duplicate species
-                            if neededSpcs.index(neededSpcs[i]) != i:
-                                del neededSpcs[i]
-                        objectsToEnlarge.extend(neededSpcs)
+                    objectsToEnlarge = self.processToSpeciesNetworks(obj)
+                    if isinstance(objectsToEnlarge[0],PDepNetwork):
+                        assert False
+                        
                     self.done = False
-    
     
             if not self.done: # There is something that needs exploring/enlarging
                 
@@ -917,6 +905,62 @@ class RMG(util.Subject):
             labels.add(potential_label)
         return oldLabels
     
+    ################################################################################
+    def processToSpeciesNetworks(self,obj):
+        """
+        breaks down the objects returned by simulate into Species and PDepNetwork
+        components
+        """
+        
+        if isinstance(obj, PDepNetwork):
+            out = [self.processPdepNetworks(obj)]
+            return out
+        elif isinstance(obj, Species):
+            return [obj]
+        elif isinstance(obj,Reaction):
+            return list(self.processReactionsToSpecies(obj))
+        elif isinstance(obj,list): #list of species
+            rspcs = self.processReactionsToSpecies([k for k in obj if isinstance(k,Reaction)])
+            spcs = list({k for k in obj if isinstance(k,Species)} | rspcs)
+            nworks = list(set(self.processPdepNetworks([k for k in obj if isinstance(k,PDepNetwork)])))
+            return spcs+nworks
+        else:
+            raise TypeError("improper call, obj input was incorrect")
+                
+    def processPdepNetworks(self,obj):
+        """
+        properly processes PDepNetwork objects and lists of PDepNetwork objects returned from simulate
+        """
+        reactionSystem = self.reactionSystem
+        if isinstance(obj, PDepNetwork):
+            # Determine which species in that network has the highest leak rate
+            # We do this here because we need a temperature and pressure
+            # Store the maximum leak species along with the associated network
+            ob = (obj, obj.getMaximumLeakSpecies(reactionSystem.T.value_si, reactionSystem.P.value_si))
+            return ob
+        elif isinstance(obj,list):
+            return [(ob, ob.getMaximumLeakSpecies(reactionSystem.T.value_si, reactionSystem.P.value_si)) for ob in obj]
+        else:
+            raise TypeError("improper call, obj input was incorrect")
+            
+    def processReactionsToSpecies(self,obj):
+        """
+        properly processes Reaction objects and lists of Reaction objects returned from simulate
+        """
+        coreSpecies = self.reactionModel.core.species
+        filterFcn = lambda x: not ((x in coreSpecies)) #remove species already in core
+        if isinstance(obj,Reaction):
+            potentialSpcs = obj.reactants+obj.products
+            potentialSpcs = filter(filterFcn,potentialSpcs)
+        elif isinstance(obj,list) or isinstance(obj,set):
+            potentialSpcs = set()
+            for ob in obj:
+                potentialSpcs = potentialSpcs | set(ob.reactants+ob.products)
+            potentialSpcs = {sp for sp in potentialSpcs if filterFcn(sp)}
+        else:
+            raise TypeError("improper call, obj input was incorrect")
+        return potentialSpcs
+
     def generateCanteraFiles(self, chemkinFile, **kwargs):
         """
         Convert a chemkin mechanism chem.inp file to a cantera mechanism file chem.cti
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -548,8 +548,8 @@ class RMG(util.Subject):
                     edgeSpecies = [],
                     edgeReactions = [],
                     pdepNetworks = self.reactionModel.networkList,
-                    atol = self.absoluteTolerance,
-                    rtol = self.relativeTolerance,
+                    atol = self.simulatorSettingsList[0].atol,
+                    rtol = self.simulatorSettingsList[0].rtol,
                     filterReactions=True,
                 )
                 self.updateReactionThresholdAndReactFlags(
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -42,6 +42,7 @@ import numpy
 import csv
 import gc
 import copy
+from copy import deepcopy
 
 from rmgpy.constraints import failsSpeciesConstraints
 from rmgpy.molecule import Molecule
@@ -73,7 +74,7 @@ from rmgpy.stats import ExecutionStatsWriter
 from rmgpy.thermo.thermoengine import submit
 from rmgpy.tools.sensitivity import plotSensitivity
 from cantera import ck2cti
-
+from rmgpy.rmg.settings import ModelSettings, SimulatorSettings
 ################################################################################
 
 solvent = None
@@ -102,17 +103,8 @@ class RMG(util.Subject):
     `reactionSystems`                   A list of the reaction systems used in this job
     `database`                          The RMG database used in this job
     ----------------------------------- ------------------------------------------------
-    `absoluteTolerance`                 The absolute tolerance used in the ODE/DAE solver
-    `relativeTolerance`                 The relative tolerance used in the ODE/DAE solver
-    `sensitivityAbsoluteTolerance`      The absolute tolerance used in the ODE/DAE solver for the sensitivities
-    `sensitivityRelativeTolerance`      The relative tolerance used in the ODE/DAE solver for the sensitivities
-    `fluxToleranceKeepInEdge`           The relative species flux below which species are discarded from the edge
-    `fluxToleranceMoveToCore`           The relative species flux above which species are moved from the edge to the core
-    `fluxToleranceInterrupt`            The relative species flux above which the simulation will halt
-    `maximumEdgeSpecies`                The maximum number of edge species allowed at any time
-    `minCoreSizeForPrune`               Minimum number of core species before pruning is allowed
-    `minSpeciesExistIterationsForPrune` Minimum number of iterations a species must exist before it can be pruned
-    `filterReactions`                   Specify whether to filter reactions during model enlarging step
+    `modelSettingsList`                 List of ModelSettings objects containing information related to how to manage species/reaction movement
+    `simulatorSettingsList`             List of SimulatorSettings objects containing information on how to run simulations
     `unimolecularThreshold`             Array of flags indicating whether a species is above the unimolecular reaction threshold
     `bimolecularaThreshold`             Array of flags indicating whether two species are above the bimolecular reaction threshold
     `unimolecularReact`                 Array of flags indicating whether a species should react unimolecularly in the enlarge step
@@ -146,6 +138,8 @@ class RMG(util.Subject):
         self.inputFile = inputFile
         self.outputDirectory = outputDirectory
         self.clear()
+        self.modelSettingsList = []
+        self.simulatorSettingsList = []
     
     def clear(self):
         """
@@ -168,20 +162,9 @@ class RMG(util.Subject):
         self.database = None
         self.reactionSystem = None
         
-        self.fluxToleranceKeepInEdge = 0.0
-        self.fluxToleranceMoveToCore = 1.0
-        self.fluxToleranceInterrupt = 1.0
-        self.toleranceMoveEdgeReactionToCore = numpy.inf
-        self.toleranceReactionInterruptSimulation = numpy.inf
-        self.ignoreOverallFluxCriterion=False
+        self.modelSettingsList = []
+        self.simulatorSettingsList = []
         
-        self.absoluteTolerance = 1.0e-8
-        self.relativeTolerance = 1.0e-4
-        self.sensitivityAbsoluteTolerance = 1.0e-6
-        self.sensitivityRelativeTolerance = 1.0e-4
-        self.maximumEdgeSpecies = 1000000
-        self.minCoreSizeForPrune = 50
-        self.minSpeciesExistIterationsForPrune = 2
         self.filterReactions=False
         self.unimolecularReact = None
         self.bimolecularReact = None
@@ -556,7 +539,7 @@ class RMG(util.Subject):
         self.done = False
         
         # Initiate first reaction discovery step after adding all core species
-        if self.filterReactions:
+        if self.modelSettingsList[0].filterReactions:
             # Run the reaction system to update threshold and react flags
             for index, reactionSystem in enumerate(self.reactionSystems):
                 reactionSystem.initializeModel(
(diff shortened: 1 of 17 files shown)


mjohnson541:
I've now added a Model Resurrection feature that attempts to kick the model dynamics by adding a few species/reactions if a DASPK error occurs, in order to make sure the rates information is present when any DASPK error occurs I've set it to skip the stepping function the first time through the while loop in base.pyx, in theory this could cause slight changes in the regression test since it enables objects to be picked up at t=0 rather than forcing it to wait until after the first time step as is currently done.

mjohnson541:
I'm not seeing any differences at all in the regression tests.  This should be ready for review now.

mjohnson541:
Ok, this should be ready for review now.

mjohnson541:

rmgpy/solver/base.pyx
```diff
@@ -856,3 +856,16 @@ class TerminationConversion:
...
         self.species = spec
         self.conversion = conv
+############################
+
+def getReverseSortedIndices(arr):
+    """
+    returns the indices of arr in order of decreasing value of arr[i] such that arr[out[i+1]] < arr[out[i]]
+    making arr[out[-1]] the minimum and arr[out[0]] the maximum value
+    """
+    out = sorted(xrange(0,len(arr)),key=lambda x: arr[x])
```
rwest: is this the same as  `out = np.argsort(arr)` ?

(and the overall function `return np.argsort(arr).reverse()` ?)

mjohnson541: I suppose it is, (although that last command won't work because reverse() doesn't return anything), I'll change that.

rwest: (yeah, best we forget that parenthetical comment...)

rmgpy/solver/base.pyx
```diff
@@ -359,7 +359,39 @@ cdef class ReactionSystem(DASx):
...
+                    boo = False
+            if boo:
+                return index
+            boo = True
+            for j in reactantIndices[index]:
+                if j in surfSpeciesIndices or j > numCoreSpecies:
+                    boo = False
+            if boo:
+                return index
+
```
rwest: Is this for loop the same as...?
```
for index in sortedIndices:
    for j in productIndices[index]:
        if j > numCoreSpecies or j in surfSpeciesIndices:
            break
    else: # (didn't break, so all j are in bulk core)
        return True
    for j in reactantIndices[index]:
        if j > numCoreSpecies or j in surfSpeciesIndices:
            break
    else: # (didn't break, so all j are in bulk core)
        return True
```
and is `index` a reaction index but the `j` are species indices?

rmgpy/solver/base.pyx
```diff
@@ -224,18 +224,57 @@ cdef class ReactionSystem(DASx):
...
+        reactantIndices = self.reactantIndices
+        
+        surfaceSpeciesIndices = []
+        surfaceReactionIndices = []
+        possibleSpeciesIndices = set()
+        
+        for i in range(len(surfaceSpecies)):
+            surfaceSpeciesIndices.append(coreSpecies.index(surfaceSpecies[i]))
+        for i in range(len(surfaceReactions)):
+            surfaceReactionIndices.append(coreReactions.index(surfaceReactions[i]))
```
rwest: Unless you're going to be doing something fancy like manipulating (popping or inserting) the list while iterating over it, it's more common (I think preferable) in Python to do loops like:
```python
for species in surfaceSpecies:
    do_something_with(species)
```
than
```python
for i in range(len(surfaceSpecies)):
    do_something_with(surfaceSpecies[i])
```
This pattern occurs quite a lot.

(unless this is part of some cython optimization I haven't thought of?)

rmgpy/solver/base.pyx
```diff
@@ -181,8 +181,8 @@ cdef class ReactionSystem(DASx):
         return (self.__class__, (self.termination,))
     
      
-    cpdef initializeModel(self, list coreSpecies, list coreReactions, list edgeSpecies, list edgeReactions, list surfaceSpecies,
-                          list surfaceReactions, list pdepNetworks=None, atol=1e-16, rtol=1e-8, sensitivity=False, 
+    cpdef initializeModel(self, list coreSpecies, list coreReactions, list edgeSpecies, list edgeReactions, list surfaceSpecies=[],
```
rwest: I think this may be OK here, and may help the cythonization, but I'm always wary of using mutable objects like lists as default variables (eg https://github.com/ReactionMechanismGenerator/RMG-Py/issues/640). But I don't have time to delve deep here. Can you reassure us this usage is safe?

mjohnson541: That's very interesting, I wasn't aware of that!  I haven't run into any issues with that and I expect I would have if it was an issue, but that doesn't look too difficult to remedy.

rmgpy/solver/base.pyx
mjohnson541: Sorry, no just force of habit, but that's easy to fix.

rmgpy/rmg/main.py
```diff
@@ -1210,6 +1196,62 @@ def readMeaningfulLineJava(self, f):
...
+    properly processes Reaction objects and lists of Reaction objects returned from simulate
+    """
+    filterFcn = lambda x: not ((x in coreSpecies)) #remove species already in core
+    if isinstance(obj,Reaction):
+        potentialSpcs = obj.reactants+obj.products
+        potentialSpcs = filter(filterFcn,potentialSpcs)
+    elif isinstance(obj,list) or isinstance(obj,set):
+        potentialSpcs = set()
+        for ob in obj:
+<<<<<<< HEAD
```
rwest: Can wherever you fix this merge conflict be squashed to this commit when you next rebase?

rmgpy/solver/base.pyx
```diff
@@ -544,7 +545,7 @@ cdef class ReactionSystem(DASx):
         surfaceReactionIndices = self.surfaceReactionIndices
         
        
-        invalidObject = None
+        invalidObject = []
```
rwest: If this is now always a list could it be renamed `invalidObjects`?

rmgpy/solver/base.pyx
```diff
@@ -456,6 +458,34 @@ cdef class ReactionSystem(DASx):
         
         return
     
+    cpdef addReactionsToSurface(self,list newSurfaceReactions,list newSurfaceReactionInds,list surfaceSpecies,list surfaceReactions,list edgeSpecies):
+        """
+        moves new surface reactions to the core
```
rwest: The docstring doesn't seem to match the method name. Does this add  a reaction to the core or to the surface?

rmgpy/solver/base.pyx
mjohnson541: I think there are a few differences, although it certainly should go over sortedIndices and the for else structure is certainly much cleaner, although I personally don't like them very much.  The indices in sortedIndices (index) are edge reaction indices, j indexes species.  

Since they are edge indices you have to add numCoreSpecies to go over the right species indices in productIndices and reactantIndices and it is >= since j == numCoreSpecies will be in the edge.

rmgpy/solver/base.pyx
mjohnson541: Although that may be because it's only called unsafely by tests that aren't testing it explicitly.

rmgpy/solver/base.pyx
mjohnson541: That should definitely say surface

rmgpy/solver/base.pyx
mjohnson541: Yes

rmgpy/rmg/main.py
mjohnson541: Yep

rwest:
Well, I just tried to review this, but 1100 new lines of code is off the end of this chart:
Code review size (see [here](https://smartbear.com/learn/code-review/best-practices-for-peer-code-review/)).

That said, it looks pretty good to me, and I would like to see it merged soon before it grows even bigger.
Two things: Example input files, and documentation!  (Clearer docstrings throughout, but also user guide on what the new settings mean). I think this has already been mentioned on another pull request.

Another thing (I know, I said two). I saw in passing a mention that it's incompatible with pressure dependence? That seems like a big deal. Is that true of the whole pull request, or only some specific features? 

If this can be easily merged as separate smaller pull requests, doing so will help reviewers. 

Did you give a talk on how this all works? Care to give it again at Northeastern?

rwest:
One more thing: If you want a user to help test it, I nominate @chiyw13 who is trying to build some big models and running into all sorts of problems with the old algorithm. (See also #1050). Perhaps she can talk to you about how (best) to use this branch.

rmgpy/solver/base.pyx
mjohnson541: This is fixed now

rmgpy/solver/base.pyx
mjohnson541: This is fixed now

rmgpy/solver/base.pyx
mjohnson541: This particular case is fixed

rmgpy/solver/base.pyx
mjohnson541: This should be fixed now

rmgpy/solver/base.pyx
mjohnson541: This should be fixed

rmgpy/solver/base.pyx
mjohnson541: This should be fixed now

rmgpy/rmg/main.py
mjohnson541: This should be fixed now

mjohnson541:
I have a decent part of the documentation already written I just haven't added it to the wiki yet.  This does definitely need some examples though.  

Sorry, by incompatible, I don't actually mean that surface features can't be used with pressure dependence.  I mean that currently once a reaction has moved into a PDepNetwork the guarantees the surface algorithm tries to provide are no longer valid for that reaction since we can't guarantee that the reactions the surface algorithm thought were important are being included in the mechanism.  At least to my very limited understanding of pdep.  It may not actually be that hard to make it work for pdep, I just need to figure out where to do it and I haven't jumped down that rabbit hole yet.  

I'm in the process trying to break it into two pull requests, but I made some very poor choices in how I patched it and my first two tries left me deathly afraid I left silent errors in main.py.  I'm going to give it another try tomorrow.  

Yep, I gave a group meeting presentation on it about a month ago, of course!

mliu49:
Could you update the actual documentation and not just the wiki? Particularly the [creating input files](http://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/input.html) page.

Do you still have plans to try and split this PR up?

mjohnson541:
Sure.  This is all that's left now (I broke it originally into this one and the two that have already been merged).

mjohnson541:
Ok, I've added the documentation associated with this pull specifically.  There's a lot of stuff missing from the first two and the seed generation pull, but I think it might be best to put that in a separate pull request.

rmgpy/rmg/main.py
```diff
@@ -1348,6 +1356,60 @@ def readMeaningfulLineJava(self, f):
         return line
     
 ################################################################################
+def processToSpeciesNetworks(obj,reactionSystem,coreSpecies):
```
mliu49: Is there a reason these are module level functions rather than methods of the RMG class?

rmgpy/rmg/main.py
```diff
@@ -1348,6 +1356,60 @@ def readMeaningfulLineJava(self, f):
...
+    elif isinstance(obj,Reaction):
+        return list(processReactionsToSpecies(obj,coreSpecies))
+    elif isinstance(obj,list): #list of species
+        rspcs = processReactionsToSpecies([k for k in obj if isinstance(k,Reaction)],coreSpecies)
+        spcs = {k for k in obj if isinstance(k,Species)} | rspcs
+        nworks,pspcs = processPdepNetworks([k for k in obj if isinstance(k,PDepNetwork)],reactionSystem)
+        spcs = list(spcs-pspcs) #avoid duplicate species
+        return spcs+nworks
+    else:
+        assert False, "improper call, obj input was incorrect"
```
mliu49: Could you change this assertion and the ones below to `raise TypeError(...)`?

rmgpy/rmg/model.py
```diff
@@ -233,6 +233,7 @@ def __init__(self, core=None, edge=None):
         self.kineticsEstimator = 'group additivity'
         self.indexSpeciesDict = {}
         self.saveEdgeSpecies = False
+        self.iteration = 0
```
mliu49: Suggestion, feel free to ignore, perhaps this could be called `iterationCounter` or `iterationNum` for clarity.

rmgpy/solver/base.pxd
```diff
@@ -110,16 +110,10 @@ cdef class ReactionSystem(DASx):
...
-        double toleranceMoveEdgeReactionToSurface=?, double toleranceMoveSurfaceSpeciesToCore=?, 
-        double toleranceMoveSurfaceReactionToCore=?,
-        double toleranceMoveEdgeReactionToSurfaceInterrupt=?, 
-        list pdepNetworks=?,ignoreOverallFluxCriterion=?, absoluteTolerance=?, relativeTolerance=?, sensitivity=?, 
-        sensitivityAbsoluteTolerance=?, sensitivityRelativeTolerance=?, sensWorksheet=?,
-        filterReactions=?)
+    cpdef simulate(self, list coreSpecies, list coreReactions, list edgeSpecies, 
+        list edgeReactions,list surfaceSpecies, list surfaceReactions,
+        list pdepNetworks=?, prune=?, sensitivity=?, sensWorksheet=?, modelSettings=?,
+        simulatorSettings=?)
```
mliu49: A lot of these arguments are not typed (not just from this PR, but from before). I'm wondering if there's any reason not to type them.

rmgpy/solver/base.pyx
```diff
@@ -892,48 +916,46 @@ cdef class ReactionSystem(DASx):
                                 bimolecularThreshold[i,j] = True
 
                 # Interrupt simulation if that flux exceeds the characteristic rate times a tolerance
-            if (not ignoreOverallFluxCriterion) and (maxSpeciesRate > toleranceMoveToCore * charRate and not invalidObject):
+            if not (maxSpecies in invalidObjects) and (not ignoreOverallFluxCriterion) and (maxSpeciesRate > toleranceMoveToCore * charRate) and len(invalidObjects) < maxNumObjsPerIter:
```
mliu49: I think `maxSpecies not in invalidObjects` is more readable, but the two methods are completely equivalent so it's up to you whether or not to change it.

rmgpy/tools/sensitivity.py
```diff
@@ -107,24 +108,24 @@ def simulate(rmg):
         pdepNetworks = []
         for source, networks in rmg.reactionModel.networkDict.items():
             pdepNetworks.extend(networks)
-            
+        
+        modelSettings = ModelSettings(toleranceKeepInEdge = 0, toleranceMoveToCore = 1, toleranceInterruptSimulation = 1)
+        simulatorSettings = rmg.simulatorSettingsList[-1]
+        
+        assert isinstance(modelSettings,ModelSettings)
```
mliu49: Is this a helpful assertion?

mjohnson541: Thanks!  Good catch, that's left over from debugging I did several months ago

rmgpy/rmg/RMGSettings.py
```diff
@@ -0,0 +1,91 @@
...
+#   DEALINGS IN THE SOFTWARE.
+#
+################################################################################
+
+"""
+This module contains settings classes for manipulation of RMG run parameters
+"""
+import numpy
+
+class ModelSettings:
```
mjohnson541: Oh, sorry I forgot about the whole new style class issue.

rmgpy/rmg/main.py
mjohnson541: Not particularly, other than that the RMG class itself is unaware of everything but coreSpecies, but I guess making it larger and making these module level will look cleaner.  I also don't expect we'll be calling these outside of the RMG class.

rmgpy/rmg/RMGSettings.py
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+################################################################################
+#
+#   RMG - Reaction Mechanism Generator
+#
+#   Copyright (c) 2002-2010 Prof. William H. Green (whgreen@mit.edu) and the
```
mjohnson541: fixed

rmgpy/rmg/RMGSettings.py
mjohnson541: fixed

rmgpy/rmg/main.py
```diff
@@ -580,171 +575,196 @@ def execute(self, **kwargs):
         
         self.saveEverything()
         
-        self.makeSeedMech(firstTime=True)
+        assert len(self.modelSettingsList) == len(self.simulatorSettingsList), "different number of model and simulator declarations in input file with more than one simulator declaration"
```
mjohnson541: fixed

rmgpy/solver/base.pxd
mjohnson541: Not that I can think of.  Fixed.

rmgpy/rmg/main.py
mjohnson541: fixed

rmgpy/rmg/main.py
mjohnson541: fixed

rmgpy/tools/sensitivity.py
mjohnson541: fixed

rmgpy/solver/base.pyx
mjohnson541: fixed

rmgpy/rmg/main.py
```diff
@@ -580,171 +575,196 @@ def execute(self, **kwargs):
...
+        
+        for q,modelSettings in enumerate(self.modelSettingsList):
+            if len(self.simulatorSettingsList) > 1: 
+                simulatorSettings = self.simulatorSettingsList[q]
+            else: #if they only provide one input for simulator use that everytime
+                simulatorSettings = self.simulatorSettingsList[0]
 
-                if self.generateSeedEachIteration:
-                    self.makeSeedMech()
+            logging.info('Beginning model generation stage {0}'.format(q+1))
```
mjohnson541: fixed

rmgpy/rmg/model.py
mjohnson541: fixed

mjohnson541:
RMGSettings.py has been renamed.  Sorry I'll do that in the future.

mliu49:
Thanks for making the changes. I'll merge this after Travis and RMG-tests finish.

mliu49:
One more thing, I almost forgot. Could you update the docstring for the `RMG` class, particularly regarding the tolerances, and maybe move those lines to the docstrings for `ModelSettings` and `SimulatorSettings`?

rmgpy/rmg/main.py
```diff
@@ -917,6 +924,65 @@ def makeCoreSpeciesLabelsIndependent(self):
...
+            spcs = list(spcs-pspcs) #avoid duplicate species
+            return spcs+nworks
+        else:
+            raise TypeError("improper call, obj input was incorrect")
+
+    def processPdepNetworks(self,obj):
+        """
+        properly processes PDepNetwork objects and lists of PDepNetwork objects returned from simulate
+        """
+        reactionSystem = self.reactionSystem
```
mliu49: RMG-tests is failing because there is no `self.reactionSystem` attribute (only `self.reactionSystems`). I think you still need to pass in `reactionSystem` to these methods.

mjohnson541: I created it at some point, I must've lost it when I cherry-picked

mjohnson541:
Ok, those should all be fixed now.

rmgpy/rmg/main.py
```diff
@@ -564,8 +548,8 @@ def execute(self, **kwargs):
                     edgeSpecies = [],
                     edgeReactions = [],
                     pdepNetworks = self.reactionModel.networkList,
-                    atol = self.absoluteTolerance,
-                    rtol = self.relativeTolerance,
+                    atol = self.modelSettingsList[0].atol,
+                    rtol = self.modelSettingsList[0].rtol,
```
mliu49: Should this be `simulatorSettingsList`?

mjohnson541: ... ... ...yes

PR #1006, opened 2017-04-25
Changed Fourier Fitting in Cantherm

laitcl:
Modified funtion in torsion.pyx, Function name
FitFourierPotentialToData; new function now begins with twelve fourier
parameters, and adds two fitting parameters every single time a negative barrier
height is found to avoid errors.

### DIFF
--- a/rmgpy/statmech/torsion.pyx
+++ b/rmgpy/statmech/torsion.pyx
@@ -511,30 +511,45 @@ cdef class HinderedRotor(Torsion):
         should begin at zero and end at :math:`2 \pi`, with the minimum energy
         conformation having a potential of zero be placed at zero angle.
         """
-        cdef numpy.ndarray[numpy.float64_t,ndim=2] A
-        cdef numpy.ndarray[numpy.float64_t,ndim=1] b
-        cdef double phi
-        cdef int N, i, m
+        cdef numpy.ndarray[numpy.float64_t,ndim=2] A, fourier
+        cdef numpy.ndarray[numpy.float64_t,ndim=1] b, fit
+        cdef double phi, value, V0
+        cdef int N, i, m, numterms
+        numterms = 6
+        cdef bint NegativeBarrier
+        NegativeBarrier=True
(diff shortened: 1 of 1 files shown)


rmgpy/statmech/torsion.pyx
```diff
@@ -510,30 +510,45 @@ cdef class HinderedRotor(Torsion):
...
-                A[i,6+m] = sin(m * phi)
-                b[i] = V[i]
-        # This row forces dV/dangle = 0 at angle = 0
-        for m in range(6):
-            A[N,m+6] = m
-        x, residues, rank, s = numpy.linalg.lstsq(A, b)
-        x *= 0.001
-        
-        self.fourier = ([x[1:6], x[7:12]], "kJ/mol")
+        while NegativeBarrier and numterms<18:
```
mjohnson541: Is there a reason you chose 18?  I'm not exactly sure what value should be used, but 18 seems a bit high to me.

rmgpy/statmech/torsion.pyx
```diff
@@ -510,30 +510,45 @@ cdef class HinderedRotor(Torsion):
...
+            # This checks if there are any negative values in the forier fit.
+            # This part of the algorithm is replicated from the fucntion HinderedRotor(Torsion)
+            NegativeBarrier=False
+            V0 = 0.0
+            self.fourier = ([x[1:numterms], x[numterms+1:2*numterms]], "kJ/mol")
+            fourier = self._fourier.value_si
+            for k in range(fourier.shape[1]):
+                V0 -= fourier[0,k] * (k+1) * (k+1)
+            if V0 < 0:
+                NegativeBarrier=True
```
mjohnson541: I think we should have a warning here that tells the user that it got a NegativeBarrier from a fit of n terms and that it is going to try a fit of m terms
and additionally a warning if numterms reaches the maximum value to tell the user that attempts to avoid having a NegativeBarrier have failed and that an nmax term fourier series is being used.

Issue #1005, opened 2017-04-25
rmgpy.thermo.NASAPolynomial.getFreeEnergy() Returns G in J/molK


amarkpayne:
The API on the RMG website claims that rmgpy.thermo.NASAPolynomial.getFreeEnergy() returns G(T)/RT, however after crunching some numbers it appears that it actually returns G(T) in J/mol-K. I am assuming that there are no errors in RMG because of this, but should we update the API just in case someone accidentally trusts the API?

rwest:
Absolutely the API docs should be accurate!

Where is the error?

My quick search for docstrings shows everything saying "in J/mol"

eg.
https://github.com/ReactionMechanismGenerator/RMG-Py/blob/e42c48ae639da7f5596265b4f9ee152e05f85ab5/rmgpy/thermo/nasa.pyx#L125

mliu49:
Interestingly, the methods in the `NASAPolynomial` class all have the appropriate units, but the methods in the `NASA` class claim to return dimensionless values, although they directly call the methods in the `NASAPolynomial` class.

rwest:
Ah, yes.
https://github.com/ReactionMechanismGenerator/RMG-Py/blob/e42c48ae639da7f5596265b4f9ee152e05f85ab5/rmgpy/thermo/nasa.pyx#L287

That should definitely be fixed. Thanks @amarkpayne

amarkpayne:
This was the webpage I was looking at http://reactionmechanismgenerator.github.io/RMG-Py/reference/thermo/nasa.html#rmgpy.thermo.NASA

amarkpayne:
Is the fix as simple as updating the docstring? If so I can go ahead and do that real quick

rwest:
That is first step, so please go ahead. Then the instructions for rebuilding and publishing the site are on the wiki I believe, but you may need to ask someone with the correct permissions to push it

amarkpayne:
The docstrings have been changed and the pull request is ready to merge. I am looking into the wiki for how to update the documentation, and should hopefully have that done shortly.

PR #1008, opened 2017-04-26
Update NASA Class Docstrings for Returning Thermo in J/mol

amarkpayne:
The NASA class calls the NASAPolynomial methods, which return the thermo in
J/mol or J/mol-K. However, the NASA Class claims erroneously that it returns
the dimensionless versions.

### DIFF
--- a/rmgpy/thermo/nasa.pyx
+++ b/rmgpy/thermo/nasa.pyx
@@ -263,28 +263,28 @@ cdef class NASA(HeatCapacityModel):
     
     cpdef double getHeatCapacity(self, double T) except -1000000000:
         """
-        Return the dimensionless constant-pressure heat capacity
-        :math:`C_\\mathrm{p}(T)/R` at the specified temperature `T` in K.
+        Return the constant-pressure heat capacity
+        :math:`C_\\mathrm{p}(T)` in J/mol*K at the specified temperature `T` in K.
         """
         return self.selectPolynomial(T).getHeatCapacity(T)
     
     cpdef double getEnthalpy(self, double T) except 1000000000:
         """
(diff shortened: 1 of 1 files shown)


amarkpayne:
See #1005

amarkpayne:
I don't have permission to update the API documentation. Whoever merges this pull request please update the documentation, close #1005, and delete this branch. Thanks!

nyee:
@kehang, @amarkpayne, IRRC, the API updates automatically, right? Was there another part of the documentation that needs to be updated?

KEHANG:
@nyee the documentation on website is managed by `gh-page` branch. We have to manually update that. More details on how to do that is [here](https://github.com/ReactionMechanismGenerator/RMG-Py/wiki/Updating-Documentation).

@amarkpayne, I think you have the permission, if you have specific issue let me know I can guide you through.

amarkpayne:
I followed those instructions, but when I tired to do 'Make Publish' it told me I did not have the required permissions, but maybe there was something else that I needed to do.

KEHANG:
fixed.

rwest:
Not sure if anyone noticed, but when you pushed the site to gh-pages you had removed the empty file `.nojekyll` which is required to ensure github pages copies across our css files and images.
The resulting website looked like this:
screen shot 2017-04-28 at 9 15 51 pm

I have added a new `.nojekyll` file (with a comment in so it's more noticeable when you delete it) and pushed.  Once again the site looks like this:

screen shot 2017-04-28 at 9 16 37 pm

Always a good idea to check the site after pushing :)

KEHANG:
Ha, good to know! When I checked yesterday, probably because I only did page refreshing, the css and images were cached in browser so the issue wasn't exposed to me. Thanks @rwest!

PR #1016, opened 2017-05-03
Entry saving

nyee:
This PR has several bug fixes and additions to our saveEntry methods. The motivating change was renumbering the indexes when machine writing the database.

### DIFF
--- a/rmgpy/data/base.py
+++ b/rmgpy/data/base.py
@@ -246,6 +246,8 @@ class Database:
         """
         Return a sorted list of the entries in this database that should be
         saved to the output file.
+
+        Then renumber the entry indexes so that we never have any duplicate indexes.
         """
         entries = self.top[:]
         if len(self.top) > 0:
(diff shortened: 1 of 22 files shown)


mliu49:
I think these changes look fine. Could you briefly explain the `boundaryAtoms` property?

nyee:
Yes, I probably should document it somewhere. 

It only applies to families that use the backbone/ends trees. Each boundary atom is an atom on an end group that is the closest to the other end group, assuming the backbone is linear. For example, if you have a 6 member intra H migration, your atoms would be labeled: 
```
H*3-R*2-{R*6-R*5-R*4}-R*1
```
The portion in closed brackets are labeled atoms that are exclusive to the backbone. Therefore, the boundary atoms would be *2 and *1. 

We need this attribute to merge groups for testing entry accessibility. It tells the family where to hook end groups to the backbone.

PR #1033, opened 2017-05-31
Incude ICCK_Demo ipython script in ipython folder

amarkpayne:


mliu49:
Could you update the RMG executable path to be relative to the ipython folder so that the notebook could be run without any changes? You can keep the comment so people know to change it if they move the file around.

Also, it might be helpful to change the filename to something more descriptive, like `combustion_model_and_ignition_delay_demo` or whatever you think is appropriate.

amarkpayne:
I made the requested changes and squashed it into one commit. Thanks for the help!

Issue #1031, opened 2017-05-30
Updating Copyright Dates in Files


amarkpayne:
I have noticed that some of the RMG files have old dates listed in the copyright notice above the doc-string at the top of the file. For example, draw.py in rmgpy/molecule/ has the dates listed as 2009-2011. Is this correct or does the 2011 need to be changed to 2017? Do we have any sort of policy about this (i.e. "if you edit the file, update the date")? If not I don't mind doing the grunt work to update these files whenever I get free moments throughout the day.

mliu49:
The headers are also slightly inconsistent across different files: there are differences in copyright years and authors and some files don't have a header at all.

I think we should probably update them and make it consistent across all files and the main `COPYING.txt` file. I think we could also consider renaming that file to `LICENSE.txt` so that GitHub recognizes it and puts it in the info bar.

amarkpayne:
I think that this would probably be easy to implement using a python script to update the headers on the files. If not it wouldn't be so terrible to do by hand. I'll make a new branch and see what I can do in my free time.

amarkpayne:
I think I have updated most of the RMG files on my UpdateCopyright branch. Specifically I targeted files in the rmgpy directory and subdirectories, though some data and input files I was unsure about. The major thing I saw was that some scripts had other author information on them. For example, rmgpy/cantherm/__init__.py is copyrighted by Josh Allen. Should this be changed? Files in similar situations are listed below if you are curious:

rmgpy/molecule/adjlist.py
rmgpy/molecule/adjlistTest.py
rmgpy/molecule/atomtype_database.py
rmgpy/molecule/atomtype_databaseTest.py

rmgpy/qm/qmverifier.py
rmgpy/qm/symmetry.py

rmgpy/transport.py

rwest:
Closed by #1036

PR #1036, opened 2017-06-01
Update copyright

amarkpayne:
See #1031


KEHANG:
some documentation files need `COPYING.txt`. A global search of `COPYING.txt` can help identify which files are affected.

amarkpayne:
I only found two documentation files that were affected by this. I have updated them and included the changes in the same commit as the file name change

alongd:
[base.pxd](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/solver/base.pyx), for example, still says
```
Copyright (c) 2002-2010 Prof. William H. Green (whgreen@mit.edu) and the
#   RMG Team (rmg_dev@mit.edu)
```

Is this file different in terms of copyright, or was it overlooked?

alongd:
I see that online it is updated, sorry.
Don't .pxd files automatically update with rebasing?

mliu49:
I just noticed that rmgpy/molecule/adjlist.py did not get updated.

amarkpayne:
I'll take a look at this file and update it. I am thinking of coming up with a way to update these files automatically, so I'll try to put that in one pull request together.

PR #1010, opened 2017-04-26
Cantherm Free Rotors, Reduced Moment of Inertia Calculations and bug fix

mjohnson541:
This pull request adds Free Rotors to Cantherm, adds two additional methods of calculating the reduced moment of inertia.  It also fixes a bug where if the HinderedRotor fit is not specified as 'best' the rotor is ignored.  

Based from pull #942.

Issue #981, opened 2017-04-04
Disagreement between TST calculations from Cantherm versus MESMER with hindered rotors


zjburas:
[MESMER](https://sourceforge.net/projects/mesmer/), like Cantherm, is an open-source package for calculating T,P-dependent kinetics given user-provided quantum calculations. I have recently compared simple high-P TST calculations between the two packages for 4 reactions on the C8H7O PES, and for the majority of cases the two agree. However, in one of the 4 cases the two packages disagree by an order of magnitude when hindered rotors are included.

To get more specific, 2 of the reactions are phenyl radical addition to both the terminal and central Carbon of ketene:

![image](https://cloud.githubusercontent.com/assets/12465817/24683207/b36f60f0-196b-11e7-88cb-3968c756a139.png)

![image](https://cloud.githubusercontent.com/assets/12465817/24683214/ccca1d9c-196b-11e7-9797-78740dbefe1f.png)

In both cases, the transition state (TS) has a hindered rotor that neither reactant has: rotation around then new C-C bond being formed. The plots below compare the TST rates predicted by Cantherm (red) and MESMER (blue) for these 2 reactions, with and without rotors. Also shown is the total predicted TST rate (sum of terminal and central addition),  for comparison against experimental measurements.

![image](https://cloud.githubusercontent.com/assets/12465817/24683263/11aa8d2a-196c-11e7-8ebe-3d9585b098c4.png)

![image](https://cloud.githubusercontent.com/assets/12465817/24683271/195c8bd6-196c-11e7-8def-7e716fb1210e.png)

![image](https://cloud.githubusercontent.com/assets/12465817/24683277/1fde0f70-196c-11e7-9ae0-630c4176adbf.png)

So the two packages give similar results over the entire T range, and both are able to match experiment (further adjusting of the barrier  heights within 1 kcal can explain the remaining discrepancy between prediction and experiment).

The other 2 reactions are OH radical addition to both the terminal and central Carbon of phenylacetylene. For terminal addition there are two hindered rotors (around the forming C-O bond and around the only single C-C bond in phenylacetylene) and for central addition there is one (around the single C-C bond).

![image](https://cloud.githubusercontent.com/assets/12465817/24683394/ca02c48c-196c-11e7-976f-41e97669b8a3.png)

![image](https://cloud.githubusercontent.com/assets/12465817/24683398/d6e1f574-196c-11e7-8e2f-7a0e47a95426.png)

[...]

PR #1040, opened 2017-06-05
Cantherm Free Rotors and Bug fixes

mjohnson541:
This pull request replaces #1010, which became disconnected.  Unit tests have been added.


mjohnson541:
The unit tests, test all of the key FreeRotors functions and test that the proper error is raised if the frequency and geometry files are different in a Gaussian hindered rotor calculation.  I've also added an example (thermo of the benzyl radical) for which this issue originally occurred.  

nyee:
This is also a continuation of #942 for reference.

rmgpy/cantherm/commonTest.py
```diff
@@ -265,7 +266,20 @@ def testTransitionStateStatmech(self):
...
-
-
+        
+class testStatmech(unittest.TestCase):
+    """
+    Contains unit tests of statmech.py
+    """
+    def setUp(self):
+        cantherm = CanTherm()
+        jobList = cantherm.loadInputFile(os.path.join(os.path.dirname(os.path.abspath(__file__)),r'files/Benzyl/input.py'))
```
KEHANG: change `files/Benzyl/input.py` to `test_data/Benzyl/input.py`, just a name convention for test data directory

KEHANG:
Although there's lots of explanations in #981, I want to emphasize what a user should do when dealing with hinder rotor (after this PR merged). Two options to avoid inconsistency in coordinate systems.

- option 1: separate `opt` and `freq` jobs, which generate `opt.log` and `freq.log`. Use `freq.log` for geometry and frequency jobs in cantherm, and only use `opt.log` for energy job.

- option 2: combine `opt` and `freq` in Gaussian input, which generates one log file `comb.log`. Use this log file for all the jobs in cantherm (geometry, frequency, and energy).

@nyee @zjburas @mjohnson541 correct me if I'm summarizing wrongly or if you want to add suggestions for the users.

zjburas:
@KEHANG , there is only one guaranteed way to prepare a cantherm input file if hindered rotors are included: option 1 above. 

Option 2 will not work, because if you combine "opt freq" in one job and try to include the "iop(7/33=1)" keyword, which should output the Hessian, the keyword will only be applied to the first part of the job (opt) but not the second (freq) where it is needed. Therefore, the output of the "opt freq iop(7/33=1)" job will not include the Hessian that is needed for the hindered rotor treatment.

I am working with @alongd on updating the Cantherm documentation to explain all of this, and many other details of using Cantherm that are currently not in the documentation or are out of date. This will likely be in a separate pull request. We should definitely make sure it is clear to future users what the Cantherm input should look like.

rmgpy/cantherm/commonTest.py
mjohnson541: So when I try to do this I think nosetests starts trying to run the methoxy.py file in the test_data folder or at least it's giving me this error when I run unit tests:
====================================================================
ERROR: Failure: NameError (name 'transitionState' is not defined)
  File "/usr/local/Cellar/numpy/1.12.0/libexec/nose/lib/python2.7/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/usr/local/Cellar/numpy/1.12.0/libexec/nose/lib/python2.7/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/usr/local/Cellar/numpy/1.12.0/libexec/nose/lib/python2.7/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/Users/mattjohnson/RMGCODE/RMG-Py/rmgpy/cantherm/test_data/methoxy.py", line 13, in <module>
    transitionState(
NameError: name 'transitionState' is not defined

examples/cantherm/species/Benzyl/benzyl.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python
```
mjohnson541: This should be fixed

rmgpy/cantherm/statmech.py
```diff
@@ -393,13 +393,6 @@ def load(self):
                        
             logging.debug('    Determining frequencies from reduced force constant matrix...')
             frequencies = numpy.array(projectRotors(conformer, F, rotors, linear, TS))
-            
-            # The frequencies have changed after projection, hence we need to recompute the ZPE
-            # We might need to multiply the scaling factor to the frequencies 
-            ZPE = self.getZPEfromfrequencies(frequencies)
```
mjohnson541: This should be fixed now

KEHANG:
Final comment: squash `752c629aba` into `cccc945`, and also squash `c471324` into `afecde8`

mjohnson541:
Ok, not really sure how I ended up squashing the first one into the second to last one instead of the other way around as I intended, it didn't look like this in my local commit log, but I think this should be fine?

KEHANG:
Last commit seems still containing the file we want to get rid of: `.DS_Store`.

mjohnson541:
Ok, that should be fixed now.

Issue #550, opened 2016-01-27
DASPK Error


ghost:
Job crashed with the following error message:
`pydas.daspk.DASPKError: DASPK returned with an IDID = -6, Repeated error test failures occurred on the last attempted step in DDASPK.  A singularity in the solution may be present.  If you are absolutely certain you want to continue, you should restart the integration.  (Provide initial values of Y and YPRIME which are consistent.)`
I was running using the new enlargeReactionFilter. I am going to try playing around with the atol and rtol values to try to solve the issue.

connie:
Can you provide an input.py file so we can see if the problem is replicable?

sherman-sc:
[decane.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/108871/decane.txt)

nickvandewiele:
contents of decane.txt:

```python
# Data sources
database(
	#overrides RMG thermo calculation of RMG with these values.
	#libraries found at http://rmg.mit.edu/database/thermo/libraries/
	#if species exist in multiple libraries, the earlier libraries overwrite the 
	#previous values
    thermoLibraries = ['KlippensteinH2O2','primaryThermoLibrary','DFT_QCI_thermo','CBS_QB3_1dHR'],
	#overrides RMG kinetics estimation if needed in the core of RMG. 
	#list of libraries found at http://rmg.mit.edu/database/kinetics/libraries/
	#input each library as a ('library_name',True/False) where a True means that all 
	#unused reactions will be automatically added to the chemkin file
    reactionLibraries = [],
	#seed mechanisms are reactionLibraries that are forced into the initial mechanism 
	#in addition to species listed in this input file.  
	#This is helpful for reducing run time for species you know will appear in 
	#the mechanism.  
    seedMechanisms = ['KlippensteinH2O2','ERC-FoundationFuelv0.9'],
	#this is normally not changed in general RMG runs.  Usually used for testing with 
	#outside kinetics databases
    kineticsDepositories = 'default', 
	#lists specific families used to generate the model. 'default' uses a list of 
	#families from RMG-Database/input/families/recommended.py
	#a visual list of families is available in PDF form at RMG-database/families
    kineticsFamilies = 'default',
	#specifies how RMG calculates rates.  currently, the only option is 'rate rules'
    kineticsEstimator = 'rate rules',
)

# List of species
#list initial and expected species below to automatically put them into the core mechanism.  
[...]

PR #1046, opened 2017-06-12
Model resurrection / Dynamics Time Scale (AcceleratorTools part 5)

mjohnson541:
This feature attempts to kick the model dynamics by adding a few species/reactions if a DASPK error occurs, in order to make sure the rates information is present when any DASPK error occurs I've set it to skip the stepping function the first time through the while loop in base.pyx, in theory this could cause slight changes in the regression test since it enables objects to be picked up at t=0 rather than forcing it to wait until after the first time step as is currently done.  

This is based from #1082 . 

I have appended the small dynamics time scale feature to this pull request since it is programmatically related:  
This pull request implements a DynamicsTimeScale which is a time before which the dynamics criterion within the surface algorithm will not be able to pull reactions into the model. This is useful because as t->0 the assumptions made in the math behind the dynamics criterion fail making it sometimes more robust to restrict its use until further in the simulation.

### DIFF
--- a/examples/rmg/minimal_surface/input.py
+++ b/examples/rmg/minimal_surface/input.py
@@ -55,6 +55,7 @@ model(
     toleranceMoveEdgeReactionToSurface=1.0,
     toleranceMoveSurfaceReactionToCore=2.0,
     toleranceMoveSurfaceSpeciesToCore=.001,
+    dynamicsTimeScale=(1.0e-10,'sec'),
 )
 
 options(
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -598,6 +598,8 @@ class RMG(util.Subject):
                 
                 allTerminated = True
                 numCoreSpecies = len(self.reactionModel.core.species)
+
+                notResurrectedVec = [True for i in xrange(len(self.reactionSystems))]
                 
                 for index, reactionSystem in enumerate(self.reactionSystems):
                     self.reactionSystem = reactionSystem
@@ -611,7 +613,7 @@ class RMG(util.Subject):
                         # Turn pruning off if we haven't reached minimum core size.
                         prune = False
                         
-                    try: terminated, obj,newSurfaceSpecies,newSurfaceReactions = reactionSystem.simulate(
+                    try: terminated,resurrected,obj,newSurfaceSpecies,newSurfaceReactions = reactionSystem.simulate(
                         coreSpecies = self.reactionModel.core.species,
                         coreReactions = self.reactionModel.core.reactions,
                         edgeSpecies = self.reactionModel.edge.species,
@@ -632,7 +634,9 @@ class RMG(util.Subject):
                             logging.error(prettify(repr(self.reactionModel.core.reactions)))
                         self.makeSeedMech()
                         raise
-
+                    
+                    notResurrectedVec[index] = not resurrected
+                    
                     if self.generateSeedEachIteration:
                         self.makeSeedMech()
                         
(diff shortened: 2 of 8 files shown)


chiyw13:
The feature here sounds helpful (we just ran into #550 again) but this PR contains 39 other commits. Can it be cherry-picked or back-ported to Master, or do we need to wait for all the dependent PRs to be merged?

mjohnson541:
It shouldn't be too hard to cherry-pick, everything for model resurrection should be associated with e55f387, I believe you need to delete the reaction object if statements and map maxSpecies into invalidObject instead of the list, also remove the surface species references.  It should be noted that we've had significant difficulties reproducing the problem after the fact so this isn't tested well.  However, since it only triggers when you're model is about to die anyway that probably isn't too much of a concern.

rmgpy/rmg/main.py
```diff
@@ -684,20 +688,23 @@ def execute(self, **kwargs):
...
-                                    surfaceSpecies = self.reactionModel.surface.species,
-                                    surfaceReactions = self.reactionModel.surface.reactions,
-                                    pdepNetworks = self.reactionModel.networkList,
-                                    modelSettings = tempModelSettings,
-                                    simulatorSettings = simulatorSettings,
-                                )
-                                self.updateReactionThresholdAndReactFlags(
-                                    rxnSysUnimolecularThreshold = reactionSystem.unimolecularThreshold,
-                                    rxnSysBimolecularThreshold = reactionSystem.bimolecularThreshold)
+                                if notResurrectedVec[index]:
```
alongd: This causes reaction thresholds/flags to never be updated for reactors that were resurrected.
Shouldn't we reset the `notResurrectedVec[index]` flag at a certain point?
(What if there's only one reactor in the input file, and it got resurrected?)

rmgpy/solver/base.pyx
```diff
@@ -589,7 +592,7 @@ cdef class ReactionSystem(DASx):
         surfaceSpeciesIndices = self.surfaceSpeciesIndices
         surfaceReactionIndices = self.surfaceReactionIndices
         
-       
+        totalDivAccumNums = None
```
alongd: Could you add a short comment describing this new variable?

rmgpy/solver/base.pyx
```diff
@@ -628,19 +631,51 @@ cdef class ReactionSystem(DASx):
...
-                logging.error("Core species net rates: {!r}".format(self.coreSpeciesRates))
-                logging.error("Edge species net rates: {!r}".format(self.edgeSpeciesRates))
-                logging.error("Network leak rates: {!r}".format(self.networkLeakRates))
-                raise e
+            
+            if not firstTime:
+                try:
+                    self.step(stepTime)
+                except DASxError as e:
+                    logging.error("Trying to step from time {} to {}".format(prevTime, stepTime))
```
alongd: "Trying to step from time {} to {} resulted in a solver (DASPK) error."

rmgpy/solver/base.pyx
```diff
@@ -628,19 +631,51 @@ cdef class ReactionSystem(DASx):
...
+                        if totalDivAccumNums and len(totalDivAccumNums) > 0: #if dynamics data available
+                            ind = numpy.argmax(totalDivAccumNums)
+                            obj = edgeReactions[ind]
+                            logging.info('At time {0:10.4e} s, Reaction {1} at dynamics number {2} was added to model core in model resurrection process'.format(self.t, obj,totalDivAccumNums[ind]))
+                            invalidObjects.append(obj)
+                        
+                        if pdepNetworks != [] and networkLeakRateRatios != []:
+                            ind = numpy.argmax(networkLeakRateRatios)
+                            obj = pdepNetworks[ind]
+                            logging.info('At time {0:10.4e} s, PDepNetwork #{1:d} at {2} was sent for exploring during model resurrection process'.format(self.t, obj.index, networkLeakRateRatios[ind],toleranceMoveToCore))
```
alongd: `toleranceMoveToCore` in `.format()` isn't used.

instead of `at {2}` consider `at network leak rate {2}` or something of the sort that is more descriptive.

rmgpy/solver/base.pyx
```diff
@@ -707,15 +742,15 @@ cdef class ReactionSystem(DASx):
...
                 logging.info('At time {0:10.4e} s, species {1} was added to model core to avoid singularity'.format(self.t, maxSpecies))
-                self.logRates(charRate, maxSpecies, maxSpeciesRate, numpy.inf, maxNetwork, maxNetworkRate)
-                self.logConversions(speciesIndex, y0)
                 invalidObjects.append(maxSpecies)
                 break
             
-            if useDynamics:
                 #######################################################
                 # Calculation of dynamics criterion for edge reactions#
                 #######################################################
```
alongd: this comment should be indented like the if statement below

rmgpy/rmg/main.py
mjohnson541: I'm not sure if I understand, notResurrectedVec is reset to all True every iteration of the while loop, so notResurrectedVec[index] should always equal the value of not resurrected during the iteration associated with reactionSystems[index].

examples/rmg/minimal_surface/input.py
```diff
@@ -55,6 +55,7 @@
     toleranceMoveEdgeReactionToSurface=1.0,
     toleranceMoveSurfaceReactionToCore=2.0,
     toleranceMoveSurfaceSpeciesToCore=.001,
+    dynamicsTimeScale=(1.0e-10,'sec'),
```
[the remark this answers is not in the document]
mjohnson541: If it isn't already, I'll make it part of my PR updating the documentation.

rmgpy/rmg/main.py
alongd: Yes, I see now. OK

mjohnson541:
So the difficulty with testing this is that I have to run an RMG job to trigger it properly and for that RMG job I need to have a large enough database that I can get to the point I can have a DASPK error.  So I think this is best tested using RMG-tests.  I think I've fixed the other issues you had with this PR.

mjohnson541:
@alongd can you merge this?

mliu49:
Could you double-check eg5 runtime by running it with profiling on? RMG-tests reported that this branch took about a minute longer than master. The other examples looked fine though.

mjohnson541:
I didn't see any difference locally:  
ModelResurrection with profiling:  53 sec
Master with profiling:  53 sec

PR #1047, opened 2017-06-12
Thermodynamics based pruning

mjohnson541:
This feature enables the user to prune based on the Gibbs energy of formation:  

The scaling used is Gn = (G_fi - G_max) / (G_max - G_min) where G_max and G_min are the minimum over all core species at the maximum reactor temperature.  

This is compared with a new parameter:  toleranceThermoKeepSpeciesInEdge.  For this parameter 0.0 denotes the maximum value of Gn within the core so 0.1 denotes 10% of the core Gibbs range beyond the maximum Gibbs value in the core.

### DIFF
--- a/rmgpy/rmg/input.py
+++ b/rmgpy/rmg/input.py
@@ -234,7 +234,7 @@ def model(toleranceMoveToCore=None, toleranceMoveEdgeReactionToCore=numpy.inf,to
           toleranceMoveEdgeReactionToSurfaceInterrupt=None,
           toleranceMoveEdgeReactionToCoreInterrupt=None, maximumEdgeSpecies=1000000, minCoreSizeForPrune=50, 
           minSpeciesExistIterationsForPrune=2, filterReactions=False, ignoreOverallFluxCriterion=False,
-          maxNumSpecies=None,maxNumObjsPerIter=1):
+          maxNumSpecies=None,maxNumObjsPerIter=1,toleranceThermoKeepSpeciesInEdge=numpy.inf):
     """
     How to generate the model. `toleranceMoveToCore` must be specified. 
     toleranceMoveReactionToCore and toleranceReactionInterruptSimulation refers to an additional criterion for forcing an edge reaction to be included in the core
@@ -251,7 +251,7 @@ def model(toleranceMoveToCore=None, toleranceMoveEdgeReactionToCore=numpy.inf,to
     rmg.modelSettingsList.append(ModelSettings(toleranceMoveToCore, toleranceMoveEdgeReactionToCore,toleranceKeepInEdge, toleranceInterruptSimulation, 
           toleranceMoveEdgeReactionToSurface, toleranceMoveSurfaceSpeciesToCore, toleranceMoveSurfaceReactionToCore,
           toleranceMoveEdgeReactionToSurfaceInterrupt,toleranceMoveEdgeReactionToCoreInterrupt, maximumEdgeSpecies, minCoreSizeForPrune, 
-          minSpeciesExistIterationsForPrune, filterReactions, ignoreOverallFluxCriterion, maxNumSpecies, maxNumObjsPerIter))
+          minSpeciesExistIterationsForPrune, filterReactions, ignoreOverallFluxCriterion, maxNumSpecies, maxNumObjsPerIter,toleranceThermoKeepSpeciesInEdge))
 
     
 def quantumMechanics(
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -557,8 +557,9 @@ class RMG(util.Subject):
                 self.updateReactionThresholdAndReactFlags(
                     rxnSysUnimolecularThreshold=reactionSystem.unimolecularThreshold, 
                     rxnSysBimolecularThreshold=reactionSystem.bimolecularThreshold)
-        
-        self.reactionModel.setThermodynamicFilteringParameters(self.Tmax,toleranceThermoKeepSpeciesInEdge=self.modelSettingsList[0].toleranceThermoKeepSpeciesInEdge,
+                
+        if not numpy.isinf(self.modelSettingsList[0].toleranceThermoKeepSpeciesInEdge):
+            self.reactionModel.setThermodynamicFilteringParameters(self.Tmax,toleranceThermoKeepSpeciesInEdge=self.modelSettingsList[0].toleranceThermoKeepSpeciesInEdge,
                                                               minCoreSizeForPrune=self.modelSettingsList[0].minCoreSizeForPrune, 
                                                               maximumEdgeSpecies =self.modelSettingsList[0].maximumEdgeSpecies,
                                                               reactionSystems=self.reactionSystems)
@@ -724,8 +725,9 @@ class RMG(util.Subject):
                             logging.info('')    
                         else:
                             self.updateReactionThresholdAndReactFlags()
-                        
-                        self.reactionModel.setThermodynamicFilteringParameters(self.Tmax, toleranceThermoKeepSpeciesInEdge=modelSettings.toleranceThermoKeepSpeciesInEdge,
+                            
+                        if not numpy.isinf(modelSettings.toleranceThermoKeepSpeciesInEdge):
+                            self.reactionModel.setThermodynamicFilteringParameters(self.Tmax, toleranceThermoKeepSpeciesInEdge=modelSettings.toleranceThermoKeepSpeciesInEdge,
                                                               minCoreSizeForPrune=modelSettings.minCoreSizeForPrune, 
                                                               maximumEdgeSpecies=modelSettings.maximumEdgeSpecies,
                                                               reactionSystems=self.reactionSystems)
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -538,6 +538,8 @@ class RMG(util.Subject):
 
         self.done = False
         
+        self.Tmax = max([x.T for x in self.reactionSystems]).value_si
+        
         # Initiate first reaction discovery step after adding all core species
         if self.modelSettingsList[0].filterReactions:
             # Run the reaction system to update threshold and react flags
@@ -555,11 +557,18 @@ class RMG(util.Subject):
                 self.updateReactionThresholdAndReactFlags(
                     rxnSysUnimolecularThreshold=reactionSystem.unimolecularThreshold, 
                     rxnSysBimolecularThreshold=reactionSystem.bimolecularThreshold)
-
+        
+        self.reactionModel.setThermodynamicFilteringParameters(self.Tmax,toleranceThermoKeepSpeciesInEdge=self.modelSettingsList[0].toleranceThermoKeepSpeciesInEdge,
+                                                              minCoreSizeForPrune=self.modelSettingsList[0].minCoreSizeForPrune, 
+                                                              maximumEdgeSpecies =self.modelSettingsList[0].maximumEdgeSpecies,
+                                                              reactionSystems=self.reactionSystems)
+        
         self.reactionModel.enlarge(reactEdge=True, 
             unimolecularReact=self.unimolecularReact, 
             bimolecularReact=self.bimolecularReact)
-
+        
+        self.reactionModel.thermoFilterDown(maximumEdgeSpecies=self.modelSettingsList[0].maximumEdgeSpecies)
+        
         logging.info('Completed initial enlarge edge step...')
         
         self.saveEverything()
@@ -716,10 +725,16 @@ class RMG(util.Subject):
                         else:
                             self.updateReactionThresholdAndReactFlags()
                         
+                        self.reactionModel.setThermodynamicFilteringParameters(self.Tmax, toleranceThermoKeepSpeciesInEdge=modelSettings.toleranceThermoKeepSpeciesInEdge,
+                                                              minCoreSizeForPrune=modelSettings.minCoreSizeForPrune, 
+                                                              maximumEdgeSpecies=modelSettings.maximumEdgeSpecies,
+                                                              reactionSystems=self.reactionSystems)
+        
                         self.reactionModel.enlarge(reactEdge=True, 
                                 unimolecularReact=self.unimolecularReact, 
                                 bimolecularReact=self.bimolecularReact)
-                    
+                        
+                        self.reactionModel.thermoFilterDown(maximumEdgeSpecies=modelSettings.maximumEdgeSpecies)
                     #Adjust Surface
                     #we add added species and remove any species moved out of the core
                     #for now we remove reactions that become part of a PDepNetwork by intersecting with the core
--- a/rmgpy/rmg/model.py
+++ b/rmgpy/rmg/model.py
@@ -800,33 +800,8 @@ class CoreEdgeReactionModel:
                         self.edge.reactions.remove(rxn)
             
             if not numpy.isinf(self.toleranceThermoKeepSpeciesInEdge) and spcs != []: #do thermodynamic filtering
-            
-                Tmax = self.Tmax
-                for spc in spcs:
-                    G = spc.thermo.getFreeEnergy(Tmax)
-                    if G > self.Gfmax:
-                        Gn = (G-self.Gmax)/(self.Gmax-self.Gmin)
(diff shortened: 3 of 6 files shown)


mjohnson541:
This now does pruning on the fly (as species are created) and filters down to maxEdgeSpecies every time the edge is reacted.

rmgpy/rmg/model.py
```diff
@@ -795,7 +798,35 @@ def processNewReactions(self, newReactions, newSpecies, pdepNetwork=None):
                         self.core.reactions.remove(rxn)
                     if rxn in self.edge.reactions:
                         self.edge.reactions.remove(rxn)
-
+            
+            if not numpy.isinf(self.toleranceThermoKeepSpeciesInEdge) and spcs != []: #do thermodynamic filtering
+
```
goldmanm: I am wondering if this could be a separate, possibly private (or not depending on your preference) method. Thoughts?

rmgpy/rmg/model.py
```diff
@@ -795,7 +798,35 @@ def processNewReactions(self, newReactions, newSpecies, pdepNetwork=None):
...
-
+            
+            if not numpy.isinf(self.toleranceThermoKeepSpeciesInEdge) and spcs != []: #do thermodynamic filtering
+            
+                Tmax = self.Tmax
+                for spc in spcs:
+                    G = spc.thermo.getFreeEnergy(Tmax)
+                    if G > self.Gfmax:
+                        Gn = (G-self.Gmax)/(self.Gmax-self.Gmin)
+                        logging.info('Removing species {0} from edge because it\'s Gibbs number {1} is greater than the toleranceThermoKeepSpeciesInEdge of {2} '.format(spc,Gn,self.toleranceThermoKeepSpeciesInEdge))
```
goldmanm: I think it might be more useful to print the actual Gibbs free energy and the Gfmax values. Do you think this should belong in the regular RMG mode or only `debug` mode? Does the other pruning print out each reaction removed in regular mode?

rmgpy/rmg/input.py
```diff
@@ -234,7 +234,7 @@ def model(toleranceMoveToCore=None, toleranceMoveEdgeReactionToCore=numpy.inf,to
           toleranceMoveEdgeReactionToSurfaceInterrupt=None,
           toleranceMoveEdgeReactionToCoreInterrupt=None, maximumEdgeSpecies=1000000, minCoreSizeForPrune=50, 
           minSpeciesExistIterationsForPrune=2, filterReactions=False, ignoreOverallFluxCriterion=False,
-          maxNumSpecies=None,maxNumObjsPerIter=1):
+          maxNumSpecies=None,maxNumObjsPerIter=1,toleranceThermoKeepSpeciesInEdge=numpy.inf):
```
goldmanm: I like the idea of using infinity as the default. Clever!

rmgpy/rmg/model.py
mjohnson541: It might be useful to print both out.  The other pruning prints out each species removed in regular mode, so I thought this should do that as well.

rmgpy/rmg/model.py
mjohnson541: I agree, I like the idea of separating it as it's own function: thermoFilterSpecies.  I'm a bit torn on the public vs private, but I think it would come out clean enough that it could be called separately in a different context so I would be inclined to make it public.  However, I'm not really sure what other context that would be.

rmgpy/rmg/model.py
```diff
@@ -995,7 +1004,80 @@ def addSpeciesToEdge(self, spec):
...
+        self.Gmax = max(Gs)
+        self.Gmin = min(Gs)
+        
+        self.Gfmax = toleranceThermoKeepSpeciesInEdge*(self.Gmax-self.Gmin)+self.Gmax
+        self.toleranceThermoKeepSpeciesInEdge = toleranceThermoKeepSpeciesInEdge
+        self.minCoreSizeForPrune = minCoreSizeForPrune
+        self.reactionSystems = reactionSystems
+        self.maximumEdgeSpecies = maximumEdgeSpecies
+        
+    def thermoFilterDown(self,maximumEdgeSpecies,minSpeciesExistIterationsForPrune=0):
```
goldmanm: should this be called pruning or thermoPrune or something similar? What name convention should we use that would be intuitive for people to understand?

rmgpy/rmg/model.py
```diff
@@ -995,7 +1004,80 @@ def addSpeciesToEdge(self, spec):
         Add a species `spec` to the reaction model edge.
         """
         self.edge.species.append(spec)
-
+    
+    def setThermodynamicFilteringParameters(self,Tmax, toleranceThermoKeepSpeciesInEdge,minCoreSizeForPrune,maximumEdgeSpecies,reactionSystems):
+        """
+        sets parameters for thermodynamic filtering based on the current core
```
goldmanm: You might want to describe what the inputs are (name, datatype, units, purpose) or something similar to other methods that fully describe the function. Maybe also mention that it returns nothing. It doesn't make much sense here, but may be more useful whenever the API gets updated.

rmgpy/rmg/model.py
```diff
@@ -995,7 +1004,80 @@ def addSpeciesToEdge(self, spec):
...
+        self.toleranceThermoKeepSpeciesInEdge = toleranceThermoKeepSpeciesInEdge
+        self.minCoreSizeForPrune = minCoreSizeForPrune
+        self.reactionSystems = reactionSystems
+        self.maximumEdgeSpecies = maximumEdgeSpecies
+        
+    def thermoFilterDown(self,maximumEdgeSpecies,minSpeciesExistIterationsForPrune=0):
+        """
+        removes species from the edge based on their Gibbs energy until maximumEdgeSpecies
+        is reached under the constraint that all removed species are older than
+        minSpeciesExistIterationsForPrune iterations
```
goldmanm: The docstring might want to specify whether `maximumEdgeSpecies` is a list of species, species object, or integer.

rmgpy/rmg/model.py
mjohnson541: I've given that some thought.  I hesitate to call it pruning, because I feel like filtering is a better word since fundamentally beyond all of the scaling we are determining a maximum Gibbs energy and filtering anything larger than that.  However, this is an alternative to flux based pruning and for that reason it may be best to call it pruning.  What do you think?

rmgpy/rmg/model.py
goldmanm: I am curious about the parameter `minSpeciesExistIterationsForPrune`. I like how it contains specific keywords that describe what it does, but the order threw me off. I initially thought it was something about the smallest species.  If I had to describe it, I would use 'minimum iterations that a species must exist to be pruned', which would shorten to `minIterationsSpeciesExistForPrune`. How did you decide on the order of the words? Maybe I am am missing some way to describe it or its purpose

rmgpy/rmg/main.py
```diff
@@ -716,10 +725,16 @@ def execute(self, **kwargs):
                         else:
                             self.updateReactionThresholdAndReactFlags()
                         
+                        self.reactionModel.setThermodynamicFilteringParameters(self.Tmax, toleranceThermoKeepSpeciesInEdge=modelSettings.toleranceThermoKeepSpeciesInEdge,
```
goldmanm: Is it possible to have a check that prevents calling `setThermodynamicFilteringParameters` and `thermoFilterDown` if the user does not specify thermodynamic pruning. This could prevent any errors encountered in thermo filtering if the user does not desire to use that functionality.

rmgpy/rmg/model.py
```diff
@@ -1072,15 +1073,18 @@ def thermoFilterDown(self,maximumEdgeSpecies,minSpeciesExistIterationsForPrune=0
...
                 i = inds[ind]
                 spc = spcs[i]
-                if iteration - spc.creationIteration > minSpeciesExistIterationsForPrune:
+                if iteration - spc.creationIteration >= minSpeciesExistIterationsForPrune:
                     removeInds.append(i)
                     numToRemove -= 1
                 ind += 1
-                    
+            
+            logging.debug('found {0} eligible species for filtering'.format(len(removeInds)))
```
goldmanm: maybe the commit with these debug statements could be squashed with the thermoFilterDown method itself?

rmgpy/rmg/model.py
mjohnson541: This isn't a new parameter, it's from older pruning.  Originally I was going to accommodate that parameter for thermodynamic pruning, but I decided that one of the primary advantages of thermodynamic pruning is that you don't have to run a simulation to filter things, which you lose if you have a minIterationsSpeciesExistForPrune.  I left support for it in thermoFilterDown because I can see situations where it might be desirable to implement that feature for thermodynamic pruning, but it isn't currently used.

rmgpy/rmg/modelTest.py
```diff
@@ -182,7 +190,64 @@ def testMakeNewReaction(self):
...
+            family = 'H_Abstraction',
+        )
+        
+        cerm.setThermodynamicFilteringParameters(Tmax=300.0, 
+                                                 toleranceThermoKeepSpeciesInEdge=0.0,
+                                                 minCoreSizeForPrune=0,
+                                                 maximumEdgeSpecies=1,
+                                                 reactionSystems=[])
+        
+        cerm.processNewReactions(newReactions=[reaction],newSpecies=[])
```
goldmanm: when the method is separated from processNewReactions, maybe you could call that specific method instead.

goldmanm: when the method is separated from processNewReactions, maybe you could call that specific method instead.

rmgpy/rmg/modelTest.py
```diff
@@ -182,7 +190,64 @@ def testMakeNewReaction(self):
...
+        
+        edgeSpcLabels = [x.label for x in cerm.edge.species]
+        self.assertNotIn(spcs[-2].label, edgeSpcLabels)
+        
+        cerm.setThermodynamicFilteringParameters(Tmax=300.0,
```
goldmanm: Maybe these different parameters should be a different test method?

rmgpy/rmg/modelTest.py
```diff
@@ -118,7 +120,13 @@ def setUpClass(cls):
         for family in rmg.database.kinetics.families.values():
             family.forbidden = ForbiddenStructures()
         rmg.database.forbiddenStructures = ForbiddenStructures()
-
+        
+        rmg.database.thermo = ThermoDatabase()
+        database = ThermoDatabase()
+        libraries = ['primaryThermoLibrary', 'GRI-Mech3.0']
+        path = os.path.join(settings['database.directory'], 'thermo')
```
goldmanm: If you want to use a database, which isn't typically recommended unless testing the database, you should use the testing database, not the main database. Otherwise, changes to primaryThermoLibrary might make it appear that Thermo pruning no longer works. I think it is better just to manually input thermo in this file for the tests you are interested in to avoid any issues with later database changes (main or test).

rmgpy/rmg/modelTest.py
```diff
@@ -182,7 +190,64 @@ def testMakeNewReaction(self):
...
 
         self.assertEquals(counter, 3)
+    
+    def testThermodynamicFiltering(self):
+        """
+        test that Thermodynamic pruning functionalities work
+        """
+        cerm = CoreEdgeReactionModel()
 
+        spcs = [Species().fromSMILES('[OH]'),
```
goldmanm: Is it possible to specify the thermo values here?

rmgpy/rmg/modelTest.py
```diff
@@ -182,7 +190,64 @@ def testMakeNewReaction(self):
...
+        cerm.processNewReactions(newReactions=[reaction],newSpecies=[])
+        
+        edgeSpcLabels = [x.label for x in cerm.edge.species]
+        self.assertIn(spcs[-1].label, edgeSpcLabels)
+        self.assertIn(spcs[-2].label, edgeSpcLabels)
+        
+        cerm.thermoFilterDown(maximumEdgeSpecies=1)
+            
+        edgeSpcLabels = [x.label for x in cerm.edge.species]
+        self.assertNotIn(spcs[-2].label,edgeSpcLabels)
```
goldmanm: I don't understand exactly what is being checked here and how it is different than the first one. Wasn't 'CH2' already removed during the first statement? Could you describe what is going on in the docstring for the method?

rmgpy/rmg/modelTest.py
```diff
@@ -182,7 +190,64 @@ def testMakeNewReaction(self):
                     counter += len(rxnList)
 
         self.assertEquals(counter, 3)
+    
+    def testThermodynamicFiltering(self):
```
goldmanm: This is a good functional test since it tests that the methods obtain the expected outcome. It is not necessarily a unittest, since it does not ensure the output of each method is correct. Would it be possible to write unittests for thermoFilterDown and possibly the section of processNewReactions?

rmgpy/rmg/model.py
goldmanm: Sounds good. As long as we are consistent :)

rmgpy/rmg/model.py
goldmanm: I think public is fine.

rmgpy/rmg/model.py
goldmanm: I like the description filtering better (since pruning seems like gardening jargon), though I feel that since this is similar to the function of what we already called pruning, we could keep it consistent. maybe have `thermoPrune` and `fluxPrune` or `kineticsPrune` for the other method.

rmgpy/rmg/model.py
goldmanm: I guess I am still confused about this parameter's function. Are you saying that if you set `minSpeciesExistIterationsForPrune` >0 in the method, then we have to run a simulation to prune thermodynamically?

rmgpy/rmg/model.py
```diff
@@ -1209,21 +1209,23 @@ def removeSpeciesFromEdge(self, reactionSystems, spec):
...
         self.edge.species.remove(spec)
         self.indexSpeciesDict.pop(spec.index)
 
         # clean up species references in reactionSystems
         for reactionSystem in reactionSystems:
-            reactionSystem.speciesIndex.pop(spec)
+            try:
+                reactionSystem.speciesIndex.pop(spec)
+            except:
+                pass
```
mjohnson541: If spec is None it will fail on the first line of the function.  I'll make it only ignore KeyError's.

rmgpy/rmg/model.py
mjohnson541: done

rmgpy/rmg/model.py
mjohnson541: done

rmgpy/rmg/main.py
mjohnson541: done

rmgpy/rmg/modelTest.py
mjohnson541: I've split it into two unittests, one tests the new thermoFilterSpecies and one tests thermoFilterDown.

rmgpy/rmg/modelTest.py
mjohnson541: fixed

rmgpy/rmg/modelTest.py
mjohnson541: It's really difficult to do that because calling processNewReactions is by far the easiest way I've found to move a species to the edge in a way that won't cause removeSpeciesFromEdge to error.

rmgpy/rmg/model.py
mjohnson541: done

rmgpy/rmg/model.py
mjohnson541: Sorry, I didn't say that in a way that made sense.  Unlike flux pruning we don't need a complete simulation to calculate our criteria, furthermore, running a simulation doesn't affect our criteria very much (since it's just calculated from the thermo of that edge species and the core thermo values).  Since thermodynamic filtering isn't strongly related to the simulation I don't see a compelling reason to make it dependent on how many simulations we've run.

rmgpy/rmg/input.py
mjohnson541: Thanks!

mjohnson541:
The commit log should be fixed now.  I've run the example I included in the pull request.  I haven't run any long jobs with this feature, but I don't anticipate much different behavior since the filtering values are only loosely dependent on model growth.  You would also probably expect the most worrying behavior from this feature for small models.

mjohnson541:
Anything else before I push this?

Issue #876, opened 2017-01-30
Reaction Degeneracy Calculation


mliu49:
Working on making benzene bonds reactive in RMG led me to look at how RMG calculates reaction degeneracy, which is currently somewhat questionable.

There have been many previous issues posted regarding degeneracy:
- #140 regarding asymmetrical groups
- #142 regarding kinetics for multiple transition states
- #552 pointing out that this is a problem
- #833 regarding effect of symmetry on reaction rate
- #857 on degeneracy of reactions with identical reactants
- and numerous reports of "unable to calculate degeneracy" which is typically due to incomplete resonance structure generation, leading to non-recognition of duplicate reactions

### Current Method
Currently, the degeneracy of any reaction in RMG is equal to the number of subgraph isomorphic matches between top level groups in reaction family trees and the reactant species. As a result, there are many cases where the degeneracy is affected by the code with no basis in chemistry.
- Top nodes that are not mutually exclusive
    - Eg. bug in Disproportionation tree that was recently found/fixed
    - If the top level nodes are not mutually exclusive, then the degeneracy will be multiplied by the number of nodes that a species matches.
- Groups that include adjacent atoms not involved in reaction
    - Eg. in 1,2_Insertion_carbene, one of the reactants is R_R, and one of the top level nodes requires 3 H atoms be attached to one end
    - With three H atoms defined, there are six ways for that to match a methyl group through subgraph isomorphism, multiplying the degeneracy of that particular match by 6.
- Labeled atoms for electron accounting
    - Eg. in R_Addition_MultipleBond, *2 labels the atom that gains a radical following the addition
    - In reactions where the products have resonance structures, there can be multiple (most commonly 2) ways to label where the electron goes, which are identical via resonance, so the degeneracy is doubled.
    - This is the most relevant to aromatics, and is the reason I started looking into this

### Proposed Changes
I've been working on a potential solution based on atom-tracking which solves some of these issues but not all of them. The underlying idea is there is only one transition state to obtain products with a given atom configuration, in the vast majority of cases. Reaction degeneracy would then be defined as the number of different but equivalent transition states that gives the same product. There are some deficiencies with this idea, which I want to get feedback on. Work-in-progress can be seen in the reactBenzeneTest branch.

[...]

Issue #619, opened 2016-03-18
KineticsError: Unable to calculate degeneracy in reaction family 1,2_Insertion_carbene


DetlevCM:
This error occurred on a forked build - so I then ran it on (an older) regular RMG-Py install and obtained the same error...

At the end of the mechanism, RMG just fails creating an unexpected biradical as well as phenol and a cresol...

```
Error: <Molecule "[CH2]">
Error: <Molecule "OC1C=CC=CC=1">
Error: <Molecule "CC1=CC=CC(O)=C1">
Traceback (most recent call last): 
  File "/home/irsrvshare2/R10/L0277/DCM_RMG/VPYTH_uptd/RMG-Py/rmg.py", line 165, in <module>
    rmg.execute(inputFile, output_dir, **kwargs)
  File "/home/irsrvshare2/R10/L0277/DCM_RMG/VPYTH_uptd/RMG-Py/rmgpy/rmg/main.py", line 541, in execute
    bimolecularReact=self.bimolecularReact)
  File "/home/irsrvshare2/R10/L0277/DCM_RMG/VPYTH_uptd/RMG-Py/rmgpy/rmg/model.py", line 748, in enlarge
    self.processNewReactions(self.react(database, self.core.species[i]), self.core.species[i], None)
  File "/home/irsrvshare2/R10/L0277/DCM_RMG/VPYTH_uptd/RMG-Py/rmgpy/rmg/model.py", line 649, in react
    reactionList.extend(database.kinetics.generateReactionsFromFamilies([moleculeA], products=None, only_families=only_families))
  File "/home/irsrvshare2/R10/L0277/DCM_RMG/VPYTH_uptd/RMG-Py/rmgpy/data/kinetics/database.py", line 419, in generateReactionsFromFamilies
    reactionList.extend(family.generateReactions(reactants))
  File "/home/irsrvshare2/R10/L0277/DCM_RMG/VPYTH_uptd/RMG-Py/rmgpy/data/kinetics/family.py", line 1244, in generateReactions
    reactionList.extend(self.__generateReactions(reactants, forward=False))
  File "/home/irsrvshare2/R10/L0277/DCM_RMG/VPYTH_uptd/RMG-Py/rmgpy/data/kinetics/family.py", line 1474, in __generateReactions
    reaction.degeneracy = self.calculateDegeneracy(reaction)
[...]

Issue #552, opened 2016-01-28
rmgpy.data.kinetics.family.generateReactions needs overhaul


connie:
Issues #545 and #508 are both related.  We aren't properly calculating degeneracy in `__generateReactions` and are sometimes seeing symmetrical reactions because our checks aren't good enough in `__createReaction`.  The function itself also allows both reactants to be either

lists of molecules objects OR molecules objects 

However, our actual usage is only of the latter type, which makes a lot of the code superfluous.  See e9437726e414afc85e927e764f70eee6e9f62a33 for possible update.

Also, we often repeat our usage of generateResonanceIsomers() all over the place to make checks, but still retain molecule objects rather than reusing that information by placing things into Species() objects.  Tackling how to best store this information can potentially speed things up a great deal.  

This function has long been very difficult to understand and debug, so although I will not be tackling this issue in the next release I hope someone will be able to work on it.

goldmanm:
overhauled in #1055

Issue #494, opened 2015-11-16
RMG estimates kinetics non-deterministically


KEHANG:
The order of molecules in `spc.molecule` can eventually lead to same reactions but different templates matched and therefore different kinetics. This difference could be exposed as modelling diverging if newly developed features have a different molecule order than master branch.

I wrote a post using a detailed example at [here](https://github.com/KEHANG/rmg_wiki/blob/master/how_rmg_estimates_kinetics_non-deterministically.ipynb). I want to take it as a chance to discuss what's the possible ways to solve this bug.

One way to make more deterministic might be redefine the equality of reaction; that is treat reactions with different templates as different reaction although all the species in the  reaction are same isomorphically.

connie:
This may be relevant: https://github.com/ReactionMechanismGenerator/RMG-Py/pull/323. This uses the idea of tracking multiple transition states in a single family.  The difficulty lies in generating the reverse direction template.  

I think we already allow duplicate reactions when they are present in separate reaction families.

rwest:
We could:
 1. Take the fastest rate (like we take the lowest energy isomer for thermo). But fastest at what temperature, if they have different activation energies?
 2. Include all the reactions as duplicates (may lead to excessive rate estimates)
 3. Include all the reactions, but weight them by the Boltzmann distribution based on the energy of the reacting isomer, so the reaction of the most stable resonance isomer might be scaled by 0.80 and the rate of reaction of the less stable resonance isomer by 0.20. This gets rather complicated, especially when you have multiple isomers on both reactants.

any other ideas?

KEHANG:
I'm thinking in `processNewReactions()` to decide whether rxn1 == rxn2, we should not only consider whether reactants and products from rxn1 are same as those from rxn2, but also whether template1 from rxn1 is same as template2 from rxn2. We could still keep only rxn1 and eliminate rxn2 but make rxn1.template = [template1, template2].

KEHANG:
And when estimating kinetics, rxn1's rate should be the sum of rate of template1 and rate of template2

nickvandewiele:
@KEHANG your ipython notebook is very valuable as a unit test. Could you convert it into one, and add the WIP decorator onto it so that it passes the travis CI tests.

KEHANG:
Ok, I'll make one.

KEHANG:
[...]

PR #1055, opened 2017-06-16
Degeneracy & Multiple TS

goldmanm:
This PR solves two related problems in RMG: reaction degeneracy values that take into account resonance structures and allows multiple transition states to be stored in RMG's output file. This PR will solve underterminable kinetics problems (#494) , muptiple TS (#323, #142), and proper degeneracy values (#876, #833, #619, #552) . These two problems both involve updates to the degeneracy-finding algorithm, which is why they are in one PR. The major modification for this pull request is the order in which the generate reactions steps take place. The reactions of all resonance structures are now generated before algorithms like degeneracy or finding reverse kinetics are run. In addition, the degeneracy algorithm was redone to take into account which atoms have actually been switched (this correct the excessive degneracy from certain classes # ) and the templates that are matched (to allow multiple transition states). In addition, the adding of new reactions to CoreEdgeReactionModel also checks the family and template, and marks duplicate reactions when they exist (allowing for multiple TS and removing underterminatee kinetics problems)

1. redefines the degeneracy variable as a private float. When degeneracy is updated, the kinetic rate (if set) is too.
2. modularizes parts of generateReactions - finding rxn.reverse and degeneracy are separate methods
3. runs reactions betwween all resonance structures of species before finding degeneracy
1. utilizes reaction templates to find and keep multiple transition states (as separate reactions)
1. uses atomIDs to identify when a reaction path has already been added to degeneracy
1. adding proper cleanup to many test methods (including solvation database cleanup)
1. some code cleanup and other modularization of methods.

This PR is ready for code review. Testing this could involve merging it and running a job of interest, though I think RMG-Tests is already doing that. If it is merged, PR #323 can be removed.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -134,6 +134,36 @@ class TemplateReaction(Reaction):
         TemplateReaction this should be a KineticsGroups object.
         """
         return self.family
+        
+    def copy(self):
+        """
+        creates a new instance of TemplateReaction
+        """
+        other = TemplateReaction.__new__(TemplateReaction)
+        
+        # this was copied from Reaction.copy class
+        other.index = self.index
+        other.label = self.label
+        other.reactants = []
+        for reactant in self.reactants:
+            other.reactants.append(reactant.copy(deep=True))
+        other.products = []
+        for product in self.products:
+            other.products.append(product.copy(deep=True))
+        other.degeneracy = self.degeneracy
+        other.kinetics = deepcopy(self.kinetics)
+        other.reversible = self.reversible
+        other.transitionState = deepcopy(self.transitionState)
+        other.duplicate = self.duplicate
+        other.pairs = deepcopy(self.pairs)
+        
+        # added for TemplateReaction information
+        other.family = self.family
+        other.template = self.template
+        other.estimator = self.estimator
+        other.reverse = self.reverse
+        
+        return other
 
 ################################################################################
 
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1370,60 +1370,72 @@ class KineticsFamily(Database):
         # Forward direction (the direction in which kinetics is defined)
         reactionList.extend(self.__generateReactions(reactants, forward=True))
         
-        if self.ownReverse:
-            # for each reaction, make its reverse reaction and store in a 'reverse' attribute
-            for rxn in reactionList:
-                reactions = self.__generateReactions(rxn.products, products=rxn.reactants, forward=True)
-                if len(reactions) == 0:
-                    logging.error("Expecting one matching reverse reaction, not zero in reaction family {0} for forward reaction {1}.\n".format(self.label, str(rxn)))
-                    logging.error("There is likely a bug in the RMG-database kinetics reaction family involving a missing group, missing atomlabels, forbidden groups, etc.")
-                    for reactant in rxn.reactants:
-                        logging.info("Reactant")
-                        logging.info(reactant.toAdjacencyList())
-                    for product in rxn.products:
-                        logging.info("Product")
-                        logging.info(product.toAdjacencyList())
-                    logging.error("Debugging why no reaction was found...")
-                    logging.error("Checking whether the family's forbidden species have affected reaction generation...")
-                    # Set family's forbidden structures to empty for now to see if reaction gets generated...
-                    # Note that it is not necessary to check global forbidden structures, because this reaction would not have
-                    # been formed in the first place.
-                    tempObject = self.forbidden
-                    self.forbidden = ForbiddenStructures()  # Initialize with empty one
-                    try:
-                        reactions = self.__generateReactions(rxn.products, products=rxn.reactants, forward=True)
-                    finally:
-                        self.forbidden = tempObject
-                    if len(reactions) != 1:
-                        logging.error("Still experiencing error: Expecting one matching reverse reaction, not {0} in reaction family {1} for forward reaction {2}.\n".format(len(reactions), self.label, str(rxn)))
-                        raise KineticsError("Did not find reverse reaction in reaction family {0} for reaction {1}.".format(self.label, str(rxn)))
-                    else:
-                        logging.error("Error was fixed, the product is a forbidden structure when used as a reactant in the reverse direction.")
-                        # Delete this reaction, since it should probably also be forbidden in the initial direction
-                        # Hack fix for now
-                        del rxn
-                elif len(reactions) > 1:
-                    logging.error("Expecting one matching reverse reaction, not {0} in reaction family {1} for forward reaction {2}.\n".format(len(reactions), self.label, str(rxn)))
-                    logging.info("Found the following reverse reactions")
-                    for rxn0 in reactions:
-                        logging.info(str(rxn0))
-                        for reactant in rxn0.reactants:
-                            logging.info("Reactant")
-                            logging.info(reactant.toAdjacencyList())
-                        for product in rxn0.products:
-                            logging.info("Product")
-                            logging.info(product.toAdjacencyList())
-                    raise KineticsError("Found multiple reverse reactions in reaction family {0} for reaction {1}, likely due to inconsistent resonance structure generation".format(self.label, str(rxn)))
-                else:
-                    rxn.reverse = reactions[0]
-
-
-            
-        else: # family is not ownReverse
+        if not self.ownReverse:
             # Reverse direction (the direction in which kinetics is not defined)
             reactionList.extend(self.__generateReactions(reactants, forward=False))
-            
+
         return reactionList
+
+    def addReverseAttribute(self, rxn):
+        """
+        For rxn (with species' objects) from families with ownReverse, this method adds a `reverse`
+        attribute that contains the reverse reaction information (like degeneracy)
+        """
+        from rmgpy.rmg.react import findDegeneracies, getMoleculeTuples
+
+        if self.ownReverse:
+            # make sure to react all resonance structures.
+            tuples = getMoleculeTuples(rxn.products)
+            reactionList = []
+            for tup in tuples:
+                moltup = [mol_and_index[0] for mol_and_index in tup]
+                reactionList.extend(self.__generateReactions(moltup,
+                                                             products=rxn.reactants,
+                                                             forward=True))
+            reactions = findDegeneracies(reactionList)
+            if len(reactions) == 0:
+                logging.error("Expecting one matching reverse reaction, not zero in reaction family {0} for forward reaction {1}.\n".format(self.label, str(rxn)))
+                logging.error("There is likely a bug in the RMG-database kinetics reaction family involving a missing group, missing atomlabels, forbidden groups, etc.")
+                for reactant in rxn.reactants:
+                    logging.info("Reactant")
+                    logging.info(reactant.toAdjacencyList())
+                for product in rxn.products:
+                    logging.info("Product")
+                    logging.info(product.toAdjacencyList())
+                logging.error("Debugging why no reaction was found...")
+                logging.error("Checking whether the family's forbidden species have affected reaction generation...")
+                # Set family's forbidden structures to empty for now to see if reaction gets generated...
+                # Note that it is not necessary to check global forbidden structures, because this reaction would not have
+                # been formed in the first place.
+                tempObject = self.forbidden
+                self.forbidden = ForbiddenStructures()  # Initialize with empty one
+                try:
+                    reactions = self.__generateReactions(rxn.products, products=rxn.reactants, forward=True)
+                finally:
+                    self.forbidden = tempObject
+                if len(reactions) != 1:
+                    logging.error("Still experiencing error: Expecting one matching reverse reaction, not {0} in reaction family {1} for forward reaction {2}.\n".format(len(reactions), self.label, str(rxn)))
+                    raise KineticsError("Did not find reverse reaction in reaction family {0} for reaction {1}.".format(self.label, str(rxn)))
+                else:
+                    logging.error("Error was fixed, the product is a forbidden structure when used as a reactant in the reverse direction.")
+                    # Delete this reaction, since it should probably also be forbidden in the initial direction
+                    # Hack fix for now
+                    del rxn
+            elif len(reactions) > 1 and not all([reactions[0].isIsomorphic(other) for other in reactions]):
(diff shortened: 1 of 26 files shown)


mjohnson541:
Before this branch it was virtually impossible to pick up the propane low T ketohydroperoxide feedback loop without a library due to a multiple TS issue with a key chain branching reaction.  I can confirm that this pull request enables RMG to now pick up this key pathway at appropriate temperatures and feasible tolerances.

rmgpy/molecule/molecule.py
```diff
@@ -1866,12 +1867,34 @@ def kekulize(self):
...
+        If start is not specified, the system will use random integers for each index
+        Otherwise the system will start at `start` and increment one each time.
         """
+
+        from random import randint
         for i, atom in enumerate(self.atoms):
-            atom.id = i
+            if start == -1:
+                # use entire range of integers to label atoms
+                atom.id = randint(-2**15,2**15)
```
mjohnson541: Why do we want to be able to generate entirely random atom ID's rather than just increment them from a set value?

rwest: ...and then if they're not unique (`not self.atomIDValid()`) we make them sequentially increasing from a random start? (Suggesting sequential is fine and it needn't be random?).  Just using sequential sounds simpler to me, unless Matt and I are both missing something.

goldmanm: Good question. Generating IDs from a starting value (like previously going 1 to N) is almost certainly going to cause clashes of atom IDs after two labeled atoms are merged in generateReactions. Generating IDs over the entire range of integer values is much more likely to reduce the clashing IDs when merging two molecules (which is what random IDs seeks to accomplish). When a clash occurs, we would need to reassign IDs properly for all resonance structures (possibly remaking resonance structures which is quite time intensive)

One alternative to completely random numbers could be to label starting from -2^-15 till 2^15 and then loop back over after labeling ~65,000 atoms (not resetting after each call). If you know that this would reduce conflicts and/or computational time, let me know.

The atomIDs cannot be set post-merging since this can result in incorrect degeneracy values when trying to generate reactions when multiple resonance structures are present. Lets say you have two reactants reacting together, and each has multiple resonance structures. In generateReactions, all combinations of resonance structures will be merged and the possible reactions output. Then `findDegeneracies` would use the atomIDs for all the sets of products to find the degeneracy. If different resonance structures are labeled differently (which can happen when relabeling after the merging of structures), the degeneracy could be higher than expected.

goldmanm: I can make a class variable that tracks the entire range of integers to avoid overlap, and then all the atom IDs can be added sequentially. That seems more intuitive than random numbers and probably will save computational time by not checking `atomIDValid` when setting the IDs. Do you think that would be preferable? Any other ideas on how to avoid conflicting IDs when merging without random Numbers?

rwest: Now that I understand what the IDs are doing, I like @goldmanm's last suggestion of a class variable that gradually increments.

rmgpy/data/kinetics/kineticsTest.py
```diff
@@ -215,13 +323,129 @@ def find_reaction_degeneracy(self, reactants_adj_list,rxn_family_str,
...
+            try:
+                reverseTemplates[rxn.template[0]] += 1
+            except KeyError:
+                reverseTemplates[rxn.template[0]] = 1
+
+        self.assertEqual(reverseTemplates, templates,'The reaction output did not output all the transition states in either order of reactants')
+
+    def test_degeneracy_keeps_track_of_both_rate_rules_from_resonance_isomers(self):
+        """
+        rxns that have multiple resonance structures hitting different rate rules should
```
mliu49: Unfinished comment?

rmgpy/data/kinetics/family.py
```diff
@@ -1549,8 +1549,12 @@ def __generateReactions(self, reactants, products=None, forward=True):
...
         # don't generate the given products
         if products is not None:
-            
-            products = [product.generateResonanceIsomers() for product in products]
+            if isinstance(products[0],Molecule):
+                products = [product.generateResonanceIsomers() for product in products]
+            elif isinstance(products[0],Species):
+                products = [product.molecule for product in products]
+            else:
+                raise TypeError('products input to __generateReactions must be Species or Reaction Objects')
```
mliu49: Should this say `Species or Molecule Objects`?

rmgpy/rmg/react.py
```diff
@@ -104,6 +104,112 @@ def reactMolecules(moleculeTuples):
...
+        else:
+            # Loop through each sublist, which represents a unique reaction
+            for rxnList1 in rxnSorted:
+                # Try to determine if the current rxn0 is identical or isomorphic to any reactions in the sublist
+                isomorphic = False
+                identical = False
+                sameTemplate = False
+                for rxn in rxnList1:
+                    isomorphic = rxn0.isIsomorphic(rxn,checkIdentical=False)
+                    identical = rxn0.isIsomorphic(rxn,checkIdentical=True)
```
mliu49: I think you can check identical only if isomorphic is true or check isomorphic only if identical is false. Either way should reduce the number of calls made.

- If two reactions are identical, then they are isomorphic.
- If two reactions are not isomorphic, then they are not identical.

rmgpy/kinetics/arrhenius.pyx
```diff
@@ -423,10 +423,10 @@ cdef class PDepArrhenius(PDepKineticsModel):
     
     """
 
-    def __init__(self, pressures=None, arrhenius=None, highPlimit=None, Tmin=None, Tmax=None, Pmin=None, Pmax=None, comment=''):
+    def __init__(self, pressures=None, arrhenius=[], highPlimit=None, Tmin=None, Tmax=None, Pmin=None, Pmax=None, comment=''):
```
goldmanm: This was an old commit from awhile ago (which I fixed), I am not sure how it got in here. I will remove this commit from the PR.

rmgpy/molecule/molecule.py
```diff
@@ -1866,12 +1867,34 @@ def kekulize(self):
...
+            else:
+                atom.id = i + start
+        if not self.atomIDValid():
+            self.assignAtomIDs(start=randint(-2**15,2**14))
```
goldmanm: I put it lower to ensure it could still count up, but this will be replaced with a more intuitive method.

rmgpy/data/kinetics/family.py
goldmanm: Yup! Fixed in squashed commit.

rmgpy/data/kinetics/kineticsTest.py
goldmanm: fixed comment in squashed commit.

rmgpy/reaction.py
```diff
@@ -160,6 +160,21 @@ def __reduce__(self):
...
+        return self._degeneracy
+    def __setDegeneracy(self, new):
+        # modify rate if kinetics exists
+        if self.kinetics is not None:
+            if self._degeneracy < 2:
+                degeneracyRatio = new
+            else:
+                degeneracyRatio = (new*1.0) / self._degeneracy
+            self.kinetics.changeRate(degeneracyRatio)
+            logging.debug('did not modify A factor when modifying degeneracy since the reaction rate was not set')
```
goldmanm: put an else statement there in squashed commit.

rmgpy/reaction.py
```diff
@@ -1123,3 +990,62 @@ def copy(self):
...
+        if comparison_method(list1[0], list2[0]):
+            return True
+    elif len(list1) == len(list2) == 2:
+        if comparison_method(list1[0], list2[0]) \
+                    and comparison_method(list1[1], list2[1]):
+            return True
+        elif comparison_method(list1[0], list2[1]) \
+                    and comparison_method(list1[1], list2[0]):
+            return True
+    elif len(list1) == len(list2) == 3:
```
mjohnson541: Wouldn't it be much faster and shorter to iterate through the lists and pop out entries that are the same from both lists as you go.  I believe the below code would reduce the maximum calls to comparison_method to 6 rather than the current 18.  
list1Temp = []
list2Temp = []
indexList = []
k = 0
while list1 != []:
   entry = list1[k]
   for i,entry2 in enumerate(list2):
       if comparison_method(entry,entry2):
           temp = list1.pop(k)
           list1temp.append(temp)
           temp = list2.pop(i)
           list2temp.append(temp)
           indexList.append(i)
           break
   else: #loop doesn't break, indexList is missing entries
         for j in xrange(len(indexList),len(list2)):
             indexList.append(j)
         list2 = [list2[i] for i in indexList]
         return False
   #everything is ok, list1 is fine, rebuild list2
   list2 = [list2[i] for i in indexList]
   return True

rmgpy/rmg/react.py
goldmanm: Great idea! This would be an awesome speedup. Added this in a squash commit (for the second case under the assumption that `not isomorphic` is more common than `is identical`)

rmgpy/reaction.py
mjohnson541: Or wait you don't even need to avoid copying the entries do you?
```
list1Temp = copy(list1)
list2Temp = copy(list2)
indexList = []
k = 0
while list1Temp != []:
   entry = list1Temp[k]
   for i,entry2 in enumerate(list2Temp):
       if comparison_method(entry,entry2):
           list1Temp.remove(entry)
           list2Temp.remove(entry2)
           break
   else: #loop doesn't break, something is missing
         return False
   k += 1
#while loop completed
return True
```

rwest: Something must be wrong with this last example. Although I'm not 100% sure what it's trying to achieve, will the `while` loop ever end? You never remove anything from`list1Temp`. In any case, the alternative is not *obviously* faster, with lots of list manipulation, creation, popping, etc.   I have some thoughts on probable efficiency gains, but before sharing them can anyone tell me if this matters? Does it get called a lot? (sorry for not being aware of the context - the discussion caught my eye).

mjohnson541: Oh, yes, sorry lost that in translation, I think you could actually just do "for entry in list1Temp" since you don't need k anymore.  I thought this would be faster under the assumption that most of our time is spent in the isomorphism checks this avoids (maximum for 2-2 lists drops from 4 to 3, maximum for 3-3 lists drops from 18 to 6).  Is that not the case?

goldmanm: @rwest, this method is called every time Reaction isomorphism is checked, which is done in degeneracy calculations, adding reactions to the edge, etc. I think it is relatively commonly called (though this is somewhat subjective without a profiling  comparison). This method modularized what already existed in `reaction.isIsomorphic`, though it would be awesome to get it working faster too.  

@kehang, Is there any listing of previous profiling to see how important the `Reaction.isIsomorphic` method is?

goldmanm: this [notebook](https://github.com/goldmanm/rmg_walkthroughs/blob/master/issues/profile%20efficiency%20of%20reaction%20isomorphic%20checking.ipynb) profiles the recommendations to isisomorphic by @mjohnson541 for H_abstraction reactions. It seems to be faster most of the time by about 10%, though sometimes it is slower (which could be due to other processes demanding resources). There should likely be more benefit if we are looking at 3x3 reactions. Feel free to play around with the notebook, and let me know if you'd like the suggested algorithm implemented here.

rmgpy/rmg/react.py
```diff
@@ -104,6 +104,112 @@ def reactMolecules(moleculeTuples):
...
+    removes degenerate reactions and
+    increments the degeneracy of the reaction object. This method modifies
+    rxnList in place and does not return anything.
+
+    This algorithm used to exist in family.__generateReactions, but was moved
+    here because it didn't have any family dependence.
+    """
+    # These duplicates should be combined and the reaction degeneracy should be increased
+    # Reaction degeneracy is only increased if two reaction are isomorphic but resulted
+    # from different transition states.
```
mjohnson541: So if I came across this comment randomly in the code without reading the function I would have thought the above comments implied that multiple TS's were simply treated as degeneracies for purposes of this function especially considering the docstring doesn't mention multiple TS's being added as duplicate reactions.  Can you make that a bit more clear?

rmgpy/rmg/react.py
```diff
@@ -239,6 +239,36 @@ def convertToSpeciesObjects(reaction):
...
+    This method modifies reactionList in place and returns nothing
+    
+    This does not adjust for identical reactants, you need to use `reduceSameReactantDegeneracy`
+    to adjust for that.
+    """
+    from rmgpy.reaction import _isomorphicSpeciesList
+    from rmgpy.reaction import ReactionError
+
+    for rxn in reactionList:
+        if _isomorphicSpeciesList(rxn.reactants, reactants):
```
mjohnson541: Is there a reason we can't add some boolean property to the reaction object during rate estimation (such as "fromReverse") so we already know if it's the reverse template or not and we don't have to do all of these isomorphism checks?

rmgpy/rmg/react.py
mliu49: This was my comment from earlier changes, before we decided to tackle multiple transition states. In this comment, different transition states refers to ones that are the effectively identical but involving a different set of atoms, which i guess you could also call degenerate transition states.

It should probably be changed now that we're also dealing with actually different transition states.

rmgpy/reaction.py
mjohnson541: You could also avoid defining the function by passing the function pointer instead
```
if checkIdentical:
    comparison_method = molecule.isIdentical
else:
    comparison_method = molecule.isIsomorphic
```
I'm really not sure how significant the time savings would be, but the above timings suggest we do have some noticeable overhead.

mliu49: I like the new approach, but I don't think the time savings would be as significant as you would expect. See the table below for a comparison of the number of function calls. Note that `list.remove` performs an isomorphism comparison with each item in the list until it finds the one it needs to remove. Using `enumerate` along with `del`  instead of `list.remove` might help a little bit.

reactants | case | if/elif method | loop method
---------- | ----- | ---------- | ----------
1 | | 1 comparison call | 1 comparison call <br> 1 list.remove call
2 | best | 2 comparison calls | 2 comparison calls <br> 2 list.remove calls
2 | worst | 4 comparison calls | 3 comparison calls <br> 2 list.remove calls
3 | best | 3 comparison calls | 3 comparison calls <br> 3 list.remove calls
3 | worst | 18 comparison calls | 6 comparison calls <br> 3 list.remove calls

rmgpy/rmg/react.py
goldmanm: updated in the most recent commit.

rmgpy/reaction.py
goldmanm: implemented the updated deletion algorithm, in the `ipynb` and still no improvement in efficiency :(

Issue #1035, opened 2017-05-31
ThermoTests fail without internet or authentication code


goldmanm:
Recently updated RMG, #1030, will fail if you don't have internet connection or don't have special authentication codes. @kehang is working on disseminating the information to allow them to pass.

KEHANG:
RMG developers should have received the auth info by now. If not or have any questions, please let me know by emailing hkh12@mit.edu

KEHANG:
one thing I can add is to attach some decorator to those internet-required tests, developers with limited internet access can run unittests (managed by `nosetests`) with a flag turning off those tests temporarily. By the way, travis should always be able to run all the tests.

KEHANG:
fixed by #1056. Now general users and developers without internet can run `make test` to pass the tests with no auth requirement. To pass all tests, run `make test-all`.

PR #1056, opened 2017-06-16
categorize tests with authentication requirements

KEHANG:
This PR provides a solution for #1035, which categories tests into tests requiring authentication and internet and tests that don't require. It's done using `nose attribute` decorator.

It also reorganizes the test modes in makefile:
- `make test` only runs tests that don't require authentication, designed for general users and develoers with no internet.
- `make test-all` runs all the tests, designed for developers and travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ install:
   - make
 
 script: 
-  - make test
+  - make test-all
   - make test-database
 
 after_success:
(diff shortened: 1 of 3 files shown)


mliu49:
Is it possible to make this the default when running `make test` and have an option to `make test-auth` or `make test-all` or something? It seems easier to tell general users to just run `make test`, while developers can have more options to choose from.

KEHANG:
@mliu49 Thanks, good suggestion!

PR #1058, opened 2017-06-20
Dynamics time scale (AcceleratorTools part 6)

mjohnson541:
This pull request implements a DynamicsTimeScale which is a time before which the dynamics criterion within the surface algorithm will not be able to pull reactions into the model.  This is useful because as t->0 the assumptions made in the math behind the dynamics criterion fail making it sometimes more robust to restrict its use until further in the simulation.


rwest:
Any way the requirement to know a timescale can be removed from the user? Nondimensionalized? Automated? Sometimes one doesn't know, when making a new model, whether things will react in microseconds or hours.

mjohnson541:
That a really good question.  I've actually been thinking about that for a while and I don't have a good answer yet.  However, currently I think it's generally ok to run with dyanmicsTimeScale=0.0 since we take the first time step before calculating (so t0~1e-15 sec).  So far I think you get some extra junk, but from what I've seen so far it isn't particularly overwhelming, just undesirable.  

The values I'm thinking of are more ~1e-10 sec than on the time scale of the actual reaction, just enough time to spread mass around enough that most species are not unreasonably far from steady state.  It's more an option for users who think they're getting a lot of junk from the dynamics criterion that lets them restrict its use to when the scaling is more valid.

mjohnson541:
This has been subsumed into the Model Resurrection pull request

PR #1060, opened 2017-06-20
Update CanTherm User documentation

zjburas:
The current CanTherm User documentation is several years out of date, such that a new user would find it impossible to run Cantherm without the help of an experienced user. This PR aims to update CanTherm's documentation such that a new user can successfully install and run Cantherm on his/her own using only the documentation as a reference.

Specifically, the following changes were made to the documentation:

1. Clarify the two options for specifying a species() or transitionState(): a.) By parsing a QM job, or b.) Directly entering molecular properties (frequencies, etc.). Both options are useful in different situations, but it wasn't clear previously that 2 distinct options exist.

2. Clarify how to correctly prepare QM input files if hindered/free rotors are to be included. This is related to Issue #981 . Previously, this information was only passed by word of mouth.

3. Added table summarizing the various modelChemistry() options available, and whether they include atom, bond and/or spin orbit corrections.

4. Describe how the new FreeRotor() feature should be used.

5. Added paragraph about how optional 'kinetics' parameter in reaction() function is used to get k(E) by Inverse Laplace Tranform (ILT), which is particularly useful for barrierless reactions.

6. Added other useful tidbits (e..g., where to get frequency scaling factors, how exactly is E0 defined, how to input hindered rotor scan as a text file, where to get collisional energy transfer parameters) that I think future users will want to know.

@goldmanm is looking into what exactly the optional 'thermo' parameter of species() and transitionState() does (if it does anything), which would also be helpful to clarify.


documentation/source/users/cantherm/input_pdep.rst
```diff
@@ -285,20 +631,25 @@ be treated as a product channel. An example of the ``network`` block is shown be
...
 ``label``              Use the name for the ``network`` declared previously
-``method``             Method to use for calculating the pdep network. Use either 'modified strong collision', 'reservoir state', or 'chemically-significant eigenvalues'
-``interpolationModel`` Select the output type for the pdep kinetics, either in 'chebyshev' or 'pdeparrhenius' (plog) format
+``method``             Method to use for calculating the pdep network. Use either ``'modified strong collision'``, ``'reservoir state'``, or ``'chemically-significant eigenvalues'``
+``interpolationModel`` Select the output type for the pdep kinetics, either in ``'chebyshev'`` or ``'pdeparrhenius'`` (plog) format
 ``activeKRotor``       A flag indicating whether to treat the K-rotor as active or adiabatic
 ``activeJRotor``       A flag indicating whether to treat the J-rotor as active or adiabatic
 ====================== ====================================================================================================================================================
```
goldmanm: should we add the parameter `rmgmode`?

rmgpy/cantherm/statmech.py
```diff
@@ -678,22 +678,22 @@ def applyEnergyCorrections(E0, modelChemistry, atoms, bonds):
     # 'CCSD(T)-F12/cc-pVDZ-F12' and 'CCSD(T)-F12/cc-pVTZ-F12')
     if modelChemistry == 'CCSD(T)-F12/cc-pVDZ-F12':
         bondEnergies = { 'C-H': -0.46, 'C-C': -0.68, 'C=C': -1.90, 'C#C': -3.13,
-            'O-H': -0.51, 'C-O': -0.23, 'C=O': -0.69, 'O-O': -0.02, 'N-C': -0.67,
-            'N=C': -1.46, 'N#C': -2.79, 'N-O': 0.74, 'N_O': -0.23, 'N=O': -0.51,
+            'O-H': -0.51, 'C-O': -0.23, 'C=O': -0.69, 'O-O': -0.02, 'C-N': -0.67,
```
KEHANG: can we add comments on the order of atoms when creating bond key? like C is always going before N, so that later people (developer and users) will follow the order.

documentation/source/users/cantherm/input_pdep.rst
goldmanm: There are also parameters `Tmin`, `Tmax`, `Tcount`, `Tlist`, `Pmin`, `Pmax`, `Pcount`, `Plist,          `maximumGrainSize`, `minimumGrainCount`

documentation/source/users/cantherm/input.rst
```diff
@@ -135,7 +136,9 @@ Allowed atom symbols for the ``atoms`` parameter are
 
 Allowed bond types for the ``bonds`` parameter are, e.g., ``'C-H'``, ``'C-C'``, ``'C=C'``, ``'N-O'``, ``'C=S'``, ``'O=O'``, ``'C#N'``...
 
-(The element order is according to the order specified under "allowed atom symbols" above, i.e., ``'C=N'`` is correct while ``'N=C'`` is incorrect. Use ``-``/``=``/``#`` to denote a single/double/triple bond, respectively). For example, for formaldehyde we would write::
+``'O=S=O'`` is also allowed.
+
+The order of elements in for the bond correction is important and generally follows the order specified under "allowed atom symbols" above, i.e., ``'C=N'`` is correct while ``'N=C'`` is incorrect. Use ``-``/``=``/``#`` to denote a single/double/triple bond, respectively. For example, for formaldehyde we would write::
```
KEHANG: related to the comment i made before, maybe add atom order list here, [C, N, O, S, H...]

documentation/source/users/cantherm/input_pdep.rst
mliu49: Since that's for internal RMG operation, I don't think it's necessary to include in the user documentation.

documentation/source/users/cantherm/input_pdep.rst
```diff
@@ -619,7 +620,7 @@ An example of the ``network`` function is given below along with a scheme of the
         label = 'acetyl + O2',
         isomers = [
             'acetylperoxy',
-            'hydroperoxylvinoxy',
+            'hydroperoxylvinoxy',min
```
KEHANG: what this `min` stands for?

goldmanm:
It seems like canterm can run kinetics estimation and pdep in the same input code. The documentation is separated out into pdep and kinetics estimation, and it they sometimes explain the same method twice. For example, species thermochemistry is estimated when pdep is used (even if the NASA polynomial is not output). What was the considerations behind keeping two separate files?

documentation/source/users/cantherm/input_pdep.rst
zjburas: Ok, to be consistent we probably should add the missing parameters to the table (Tmin,  Tmax, etc.), except for ``rmgmode`` . The Tlist and Plist options are currently not mentioned at all, so that should also be added. I'll take care of it.

documentation/source/users/cantherm/input_pdep.rst
zjburas: @KEHANG , I will make the suggested changes.

zjburas:
Following the PR review, the following changes have been made to Cantherm documentation:

First, in response to @goldmanm 

1. All of the P-dep algorithm parameters (including ``Tmin``, ``Tmax``, etc., and ``maximumGrainSize``, ``minimumGrainCount``) are included in the relevant table. ``Tlist`` and ``Plist`` are also mentioned as options. The only parameter not mentioned is ``rmgmode``.
2. Sections for thermo() and kinetics() blocks have been added to the P-dep documentation, to make it clear that within a single Cantherm input file thermo(), kinetics() and pressureDependence() jobs can be run (with the one caveat for bimolecular reactant to bimolecular product reactions). 
3. Long term, it makes sense to totally restructure the CanTherm documentation, since there is a lot of repetition as you pointed out. However, I think that should be a separate PR later, and right now it is more important to at least have documentation that is up to date.

In response to @KEHANG 

1. I've added a reference to Goldsmith et al.'s paper for ``Klip_2`` and ``Klip_3`` model chemistries. I don't know what ``Klip_1`` is. I think ``Klip_2_cc`` can be inferred from ``Klip_2``.
2. n in ``n=D,T,Q`` refers  to the n in the basis set (``cc-pVnZ-F12`` for example). This is a common syntax that I don't think requires clarification.
3. Comments added to statmech.py as requested to clarify how bond correction labels should be written.
4. The ``min`` you pointed out was a typo, thanks.

What did you  both think of the new species() section, since that is where most of the changes are? Do you think a new user could successfully run Cantherm on his/her own with only these instructions?

@amarkpayne is also looking at the new documentation, to give a relative newbie perspective.

documentation/source/users/cantherm/input_pdep.rst
goldmanm: I don't see the extra parameters that you input into the table. Did you push all the recent commits?

alongd: I just rebased with master and unintentionally dropped this latest commit. @zjburas will add back

zjburas: My commit was accidentally deleted. I'm adding it back now.

zjburas: Latest commits have been added again.

PR #1061, opened 2017-06-23
Revive Reaction Group Additivity

mjohnson541:
This pull request enables reaction group additivity to be trained on the current training reactions and trees and estimateKineticsUsingGroupAdditivity to be called on a given template.  

Note this does not in anyway activate group additivity kinetics, it just makes it so that the functions needed to do so run.

### DIFF
--- a/rmgpy/data/kinetics/groups.py
+++ b/rmgpy/data/kinetics/groups.py
@@ -196,12 +196,9 @@ class KineticsGroups(Database):
             kinetics.comment += comment_line + '\n'
 
         # Also include reaction-path degeneracy
-        if isinstance(kinetics, KineticsData):
-            kinetics.kdata.value_si *= degeneracy
-        elif isinstance(kinetics, Arrhenius):
-            kinetics.A.value_si *= degeneracy
-        elif kinetics is not None:
-            raise KineticsError('Unexpected kinetics type "{0}" encountered while generating kinetics from group values.'.format(kinetics.__class__))
+
+        kinetics.changeRate(degeneracy)
+
         kinetics.comment += "Multiplied by reaction path degeneracy {0}".format(degeneracy)
         
         return kinetics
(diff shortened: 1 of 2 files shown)


rmgpy/data/kinetics/groups.py
```diff
@@ -196,6 +196,8 @@ def estimateKineticsUsingGroupAdditivity(self, template, referenceKinetics, dege
             kinetics.kdata.value_si *= degeneracy
         elif isinstance(kinetics, Arrhenius):
             kinetics.A.value_si *= degeneracy
+        elif isinstance(kinetics, ArrheniusEP):
+            kinetics.A.value_si *= degeneracy
```
goldmanm: getting rid of the `elif` statements with `kinetics.changeRate(degeneracy)` might be cleaner.

rmgpy/data/kinetics/groups.py
```diff
@@ -218,15 +220,42 @@ class together, returning their product as a new kinetics object of
             )
```
goldmanm: Maybe the entire __multiplyKineticsData method should be moved into the KineticsData class (or at least moved outside the KineticsGroups class), since it doesn't use any methods from that class. Any thoughts on this?

rmgpy/data/kinetics/groups.py
mjohnson541: Ok

rmgpy/data/kinetics/groups.py
mjohnson541: I agree, it certainly doesn't belong where it's at, but I think we should wait on non-essential cleanup until we decide whether it should be brought back into RMG or not.

mjohnson541:
Ok, I've fixed the docstring.  Unless you think we should move __multiplyKineticsData this should be good to push.

rmgpy/data/kinetics/groups.py
goldmanm: makes sense.

rmgpy/data/kinetics/groups.py
goldmanm: The fix looks good. Could you squash this commit with the updated line of code `switch degeneracy adjustments to use changeRate`? Having fewer modifications in the same PR  will make traceback easier. When rewording up the commit message, if you think having the previous squashed commit names and descriptions is not useful, you may consider cleaning up.

mjohnson541: Ok, done

Issue #1065, opened 2017-06-27
better species naming for cantera models


goldmanm:
To the best of my knowledge, cantera does not have the restrictive species naming guidelines that chemkin has. When generating a cantera model, we can apply more descriptive names (like their SMILES) than would be possible with chemkin. 

We currently create the cantera input file by converting from the chemkin input file with `ck2cti`. Getting better naming would require generating the input file directly from RMG. 

@alongd, you wrote the cantera conversion for RMG, do you have thoughts on this?

alongd:
I'm not familiar with the naming convention on Cantera; if it's really that flexible, this could be a nice feature.

goldmanm:
I wrote up [this notebook](https://github.com/goldmanm/rmg_walkthroughs/blob/master/Cantera%20output%20with%20RMG%20names.ipynb) which provides the species, thermo and reaction blocks to input into a cti file so analysis can be conducted in cantera with SMILES as the species names. 

This is kind of a first step to getting better naming for the species in cantera modules.

goldmanm:
I found a way to rename species from an RMG run and place them into a cantera input file.

I also commonly use this method, which renames all species based on SMILES and ensures they don't overlap. This requires the `soln2cti` which is in [PR 451 in cantera](https://github.com/Cantera/cantera/pull/451). You can just put it in the same directory

```python
def make_string_labels_independent(species):
    """
    This method reads in species labels and makes sure none of them conflict
    If a conflict occurs, the second occurance will have '-2' added
    """
    
    labels = set()
    for spec in species:
        duplicate_index = 1
        potential_label = spec.label
        while potential_label in labels:
            duplicate_index += 1
            potential_label = spec.label + '-{}'.format(duplicate_index)
        spec.label = potential_label
        labels.add(potential_label)

# use the method
species, reactions = loadChemkinFile(chemkinPath, speciesDictPath, readComments = False, useChemkinNames=False)
make_string_labels_independent(species)

#convert to cantera objects
ct_species = [spec.toCantera(useChemkinIdentifier = False) for spec in species]
# since this can return a rxn or list of reactions, this allows to make a list based on the returned type
ct_reactions = []
for rxn in reactions:
    ct_rxn = rxn.toCantera(useChemkinIdentifier = False)
    if isinstance(ct_rxn, list):
        ct_reactions.extend(ct_rxn)
    else:
        ct_reactions.append(ct_rxn)

[...]

PR #1066, opened 2017-06-27
allow cantera objects to use RMG names

goldmanm:
Added an option to 'toCantera' methods in Species and Reaction
to allow for the name of the species/reaction to be the RMG name
instead of the chemkin name. The default behavior is still the same.

We can still not create cantera input files with RMG names. This PR is just one step in addressing issue #1065.


goldmanm:
Haha. I completely agree that the most intuitive behavior of the RMG method would be to use RMG's label (instead of chemkin). It was previously written that way, so I wanted to preserve behavior. However, I will add a commit that changes the default behavior and adds the argument to the original methods calling the function. Let me know what you think about this.

Issue #1067, opened 2017-06-28
RMG directory options do not work & are not tested


amarkpayne:
Maybe I am missing something here, but when I try to specify the RMG output directory as shown below:

`python $RMG input.py -o output_dir`

it returns an error about joining the output directory paths. After a quick investigation, I found out that this was because `args.output_directory` in rmg.py is a list of length 1. Maybe I am doing something wrong, but if not I will make a pull request with a fix for this issue.

goldmanm:
This issue was closed with #1068. However, this issue raises another concern with our testing of RMG in regards to the input parameters (which at least currently are not unittested). 
@kehang, is it would be possible to add optional arguments to RMG-tests running scripts. 

For example, one of the tests could utilize one of these commands:

  -o DIR, --output-directory DIR
  -s DIR, --scratch-directory DIR
  -l DIR, --library-directory DIR
  -p, --profile
  -k, --kineticsdatastore

amarkpayne:
Now that I think about this I realize that the scratch directory, and possibly other flags will have the same issue that just got fixed. Let me see if I can figure out these issues and maybe write the necessary unit tests

goldmanm:
We can't unittest __main__ how it is written. [This SX post](https://stackoverflow.com/questions/18160078/how-do-you-write-tests-for-the-argparse-portion-of-a-python-module) describes how the `parseCommandLineArguments` can be restructured for unittesting. In addition, much of the `__main__` method, like the processing of the arguments, could be moved out to a separate function which then can be unittested (making sure we aren't outputting lists, etc.).

KEHANG:
@goldmanm it's possible to do that in RMG-tests. But I think it's better to have simple unittests in RMG-Py for those issue because RMG-tests is more testing domain knowledges such chemistry, thermo, kinetics are right or not.

amarkpayne:
Some of these concerns have been addressed by #1156

PR #1068, opened 2017-06-28
Add indexing to args.output_directory to return a string

amarkpayne:
See #1067


goldmanm:
I reran from master and your branch, and I saw the error and this PR fixes it.

Thanks for the submission!

PR #1069, opened 2017-06-28
added warning to change folder name during installation

goldmanm:
This prevents users from getting 'unable to find module rmgpy' issues when installing.

I made this mistake 2 years ago. Time it got fixed for posterity.


goldmanm:
If you want to be super cautious, you can run 'make documentation' to ensure the formatting is the same. I ran it before submitting this PR and the formatting is consistent.

Issue #789, opened 2016-10-17
Cantherm thermo and Chebyshev output not formatted correctly for Chemkin


zjburas:
This is more of an annoyance than a serious issue, but the chem.inp file generated by Cantherm does not output either the thermo or Chebyshev blocks in the correct Chemkin format.

First, the thermo block does not include the the number of each atom type present in the molecule (e.g., C 4  H 10 for butane). Instead, the correct atom types are identified, but the number of each is simply listed as 0 (e.g., C 0  H 0 for butane). It is probably easiest to see this by running the n-butanol example under RMG-Py/examples/cantherm/networks/n-butanol. The lack of numbers causes pre-processing errors in Chemkin and currently has to be fixed manually.

Second, the Chebyshev block lacks the necessary (+M) on both the reactant and product side. An example of this is at RMG-Py/examples/cantherm/networks/acetyl+O2. Again, this causes pre-processing errors in Chemkin and has to be fixed manually.

Obviously the chem.inp output of an RMG job does not have these formatting errors, so it must just be an issue when using Cantherm as a standalone unit.

mliu49:
I found that Cantherm has a completely independent set of methods for writing Chemkin entries, instead of using the (clearly) more robust methods that RMG uses.

For the thermo problem, Cantherm has a [placeholder](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/cantherm/thermo.py#L173) for counting the number of each element, but no actual code. For the kinetics problem, Cantherm directly [writes](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/cantherm/pdep.py#L451) the reaction string using RMG's representation.

It would seem that the best solution would be to try and import RMG's Chemkin methods to use in Cantherm. It might take some time to compare the two to make sure we aren't losing any functionality, but I can look into it.

mliu49:
I looked at it a bit more, and it seems that the writing kinetics entries can be easily switched over to use `rmgpy.chemkin.writeKineticsEntry`, but it's the issue with thermo entries is not as straightforward. Cantherm does not create Molecule objects at all, so the `retrieveElementCount` function does not work. Thermo jobs also do not store any kind of molecular structure or formula information, which is probably why element counting was not implemented in the first place.

rwest:
If this is for generating thermo blocks from the basis of reading in Gaussian (or some other QChem package) file, then we could grab the number of each atom type from there.

[...]

PR #1071, opened 2017-06-28
allow cantherm to create valid chemkin files

goldmanm:
This PR allows the chemkin input file created during a cantherm run to be ready to use in chemkin/cantera. The major changes are:

1. changed order of running cantherm jobs to run species and thermo before pdep and kinetics.
2. adding headings and labels for chemkin input files.
3. modified the pdep output to not contain both forward and reverse reactions, so it is valid when output. The unused reactions are commented out.
4. updated documentation explaining the chemkin file and changes to pdep output.

When reviewing this PR, it would be helpful if you could run one of your own cantherm jobs to see if you get the intended outcome. Alternatively, you could run n-butanol and H+C2H4 examples to get complete chemkin files or acetyl+O2 to view the pdep changes. You can test the validity of `chem.inp` files by running `ck2cti --input chem.inp` on them.

### DIFF
--- a/documentation/source/users/cantherm/output.rst
+++ b/documentation/source/users/cantherm/output.rst
@@ -19,15 +19,12 @@ The output file contains the entire contents of the input file. In
 addition, the output file contains a block of ``pdepreaction()`` calls. The 
 parameters of each ``pdepreaction()`` block match those of the ``reaction()`` 
 block from the input file, except that no transition state data is given and 
-the ``kinetics`` are by definition pressure-dependent. 
+the ``kinetics`` are by definition pressure-dependent.
 
-A ``pdepreaction()`` item is printed for the forward and reverse direction of 
-every reaction involving isomers and reactant channels only. For reactions 
-involving a product channel, there is only a ``pdepreaction()`` item for the 
-direction in which the product channel is the product of the reaction. To use
-this output, you must either keep all of the reactions and treat them as
-irreversible, or discard the duplicate reverse directions and treat the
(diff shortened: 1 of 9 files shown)


goldmanm:
This addresses issue #789

goldmanm:
The last two commits were added based on a suggestion to comment out the duplicate methods (in case the user wants them). See commit 881c1bb for conversation.

zjburas:
@goldmanm , I just tried out this branch on one of my Phenyl + Propene P-dep PES, and it created a chem.inp file that can be directly pre-processed and run in Chemkin, as you described.

The only problem that I encountered was that there are two species on the PES that cis/trans conformers, which are not distinguished in the isomorphism check. Therefore,  the Chebyshev reactions for one of the conformers was thought to be redundant with the other, and was commented out, which is not desired. Even though RMG cannot distinguish cis/trans, when using Cantherm it is sometimes useful to keep the two conformers distinct.

goldmanm:
@zjburas, Thanks for bringing up this issue. I am trying to rework the comparison to use species' labels instead of isomorphism. This is likely to also be faster. Since the user is inputting names, there's little need to do isomorphism check anyways.

goldmanm:
The rework should be finished.

zjburas:
@goldmanm your change to use labels instead of structures to check for redundant reactions works. I can now distinguish cis- trans- isomers using labels, and the chem.inp created by Cantherm can be directly run in Chemkin. Big time saver, thanks.

PR #1076, opened 2017-07-06
Read templates from chemkin

goldmanm:
Previously, `loadChemkinFile` would not save templates or degeneracy of reactions from chemkin file. This PR stores those attributes in the resulting `TemplateReaction` object.

Also improve testing of these methods.

### DIFF
--- a/rmgpy/chemkin.py
+++ b/rmgpy/chemkin.py
@@ -535,6 +535,26 @@ def _readKineticsLine(line, reaction, speciesDict, Eunits, kunits, klow_units, k
             raise ChemkinError(error_msg)
     return kinetics
 
+def _removeLineBreaks(comments):
+    """
+    This method removes any extra line breaks in reaction comments, so they
+    can be parsed by readReactionComments.
+    """
+    comments = comments.replace('\n',' ')
+    new_statement_indicators = ['Reaction index','Template reaction','Library reaction',
+                                'PDep reaction','Flux pairs',
+                                'Estimated using','Exact match found','Average of ',
+                                'Euclidian distance','Matched node ','Matched reaction ',
+                                'Multiplied by reaction path degeneracy ',
+                                'Kinetics were estimated in this direction',
+                                'dGrxn(298 K) = ','Both directions are estimates',
+                                'Other direction matched ','Both directions matched ',
+                                'This direction matched an entry in ', 'From training reaction',
+                                'This reaction matched rate rule',
+                                ]
+    for indicator in new_statement_indicators:
+        comments = comments.replace(' ' + indicator, '\n' + indicator,1)
+    return comments
 
 def readReactionComments(reaction, comments, read = True):
     """
@@ -561,6 +581,9 @@ def readReactionComments(reaction, comments, read = True):
         
         return reaction  
     
+    # the comments could have line breaks that will mess up reading
+    # we will now combine the lines and separate them based on statements
+    comments = _removeLineBreaks(comments)
     lines = comments.strip().splitlines()
         
     for line in lines:
--- a/rmgpy/chemkin.py
+++ b/rmgpy/chemkin.py
@@ -628,6 +628,23 @@ def readReactionComments(reaction, comments, read = True):
                 reaction.pairs.append((reactant, product))
             assert len(reaction.pairs) == max(len(reaction.reactants), len(reaction.products))
 
+        elif isinstance(reaction,TemplateReaction) and 'rate rule ' in line:
+            bracketed_rule = tokens[-1]
+            templates = bracketed_rule[1:-1].split(';')
+            reaction.template = templates
+            # still add kinetic comment
+            reaction.kinetics.comment += line.strip() + "\n"
+
+        elif isinstance(reaction,TemplateReaction) and\
+                       'Multiplied by reaction path degeneracy ' in line:
+            degen = float(tokens[-1])
+            reaction.degeneracy = degen
+            # undo the kinetic manipulation caused by setting degneracy
+            if reaction.kinetics:
+                reaction.kinetics.changeRate(1./degen)
+            # still add kinetic comment
+            reaction.kinetics.comment += line.strip() + "\n"
+
         elif line.strip() != '':
             # Any lines which are commented out but don't have any specific flag are simply kinetics comments
             reaction.kinetics.comment += line.strip() + "\n"
--- a/rmgpy/chemkinTest.py
+++ b/rmgpy/chemkinTest.py
@@ -31,6 +31,8 @@ import os
 from chemkin import *
 import rmgpy
 from rmgpy.species import Species
+from rmgpy.reaction import Reaction
+from rmgpy.kinetics.arrhenius import Arrhenius
 
 
 ###################################################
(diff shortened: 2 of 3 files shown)


rmgpy/chemkin.py
```diff
@@ -621,6 +621,23 @@ def readReactionComments(reaction, comments, read = True):
                 reaction.pairs.append((reactant, product))
             assert len(reaction.pairs) == max(len(reaction.reactants), len(reaction.products))
 
+        elif isinstance(reaction,TemplateReaction) and 'rate rule ' in line:
+            bracketed_rule = line.split('rate rule ')[-1]
```
mliu49: Could this be `tokens[-1]`?

rmgpy/chemkin.py
```diff
@@ -621,6 +621,23 @@ def readReactionComments(reaction, comments, read = True):
...
+            templates = rule = bracketed_rule[1:-1].split(';')
```
mliu49: Is the `rule` variable used anywhere?

rmgpy/chemkin.py
```diff
@@ -621,6 +621,23 @@ def readReactionComments(reaction, comments, read = True):
...
+            reaction.template = templates
+            # still add kinetic comment
+            reaction.kinetics.comment += line.strip() + "\n"
+
+        elif isinstance(reaction,TemplateReaction) and\
+                       'Multiplied by reaction path degeneracy ' in line:
+            degen = float(line.split('Multiplied by reaction path degeneracy ')[-1])
```
mliu49: Could this be `float(tokens[-1])`?

rmgpy/chemkin.py
```diff
@@ -621,6 +621,23 @@ def readReactionComments(reaction, comments, read = True):
...
+            reaction.degeneracy = degen
+            # undo the kinetic manipulation caused by setting degneracy
+            if reaction.kinetics:
+                reaction.kinetics.changeRate(1./degen)
```
mliu49: This seems inconvenient, but I guess there's no way to avoid it?

rmgpy/chemkinTest.py
```diff
@@ -256,3 +261,88 @@ def testReactantN2IsReactiveAndGetsRightSpeciesIdentifier(self):
         self.assertTrue(n2.reactive)
 
         self.assertEqual(getSpeciesIdentifier(n2), 'N2(35)')
+
+class TestReadReactionComments(unittest.TestCase):
+    def setUp(self):
```
mliu49: Could this be a `setUpClass` method instead?

rmgpy/chemkin.py
goldmanm: One way to avoid it would be to remake the kinetics object or have the kinetics object made all in one statement. The method of reading and executing line by line prevents this from being done in a clean way. That is also why I wrote a test to check the kinetics.

goldmanm:
Thanks so much for the comments @mliu49. I took all your suggestions and commented on potential methods to get the workaround functioning.

rmgpy/chemkinTest.py
```diff
@@ -256,3 +275,88 @@ def testReactantN2IsReactiveAndGetsRightSpeciesIdentifier(self):
         self.assertTrue(n2.reactive)
 
         self.assertEqual(getSpeciesIdentifier(n2), 'N2(35)')
+
+class TestReadReactionComments(unittest.TestCase):
+    def setUpClass(self):
```
mliu49: I think you need a `@classmethod` decorator here.

goldmanm:
looking through RMG's output. I think this type of comment will cause a bug since RMG adds a line break:

```
! Reaction index: Chemkin #69; RMG #171
! Template reaction: intra_H_migration
! Flux pairs: C2H5O(13), C2H5O(12); 
! Estimated using average of templates [R3H_SS;O_rad_out;Cs_H_out_2H] + [R3H_SS_Cs;Y_rad_out;Cs_H_out_2H] for rate rule
! [R3H_SS_Cs;O_rad_out;Cs_H_out_2H]
! Multiplied by reaction path degeneracy 3.0
```

If the method for reading kinetics strings is line-by-line, having line breaks seems to decrease reliability. The code for breaking the lines comes from [this commit](https://github.com/ReactionMechanismGenerator/RMG-Py/commit/d94688ec5ea7b40674ebd3e755a1133815f9b944). Line wrapping is useful for readability, but I am not sure how to reconcile it with reading entries reliably. Honestly, my preference is to remove line wrapping, but there may be a better way. Thoughts?

rwest:
I seem to recall a chemkin preprocessor crashing on very long lines. Some of those reaction source comments end up very long. Even if the most recent chemkin version can cope with infinitely long lines, I expect many potential users of RMG models use older or third party parsers.

mliu49:
At this point, we've already compiled all of the comment lines for the reaction, so I think it would be straightforward to do some additional processing using some unique marker, following up on Alon's suggestion.

For example, use an `@` to indicate the start of each line:
```
! @Reaction index: Chemkin #69; RMG #171
! @Template reaction: intra_H_migration
! @Flux pairs: C2H5O(13), C2H5O(12); 
! @Estimated using average of templates [R3H_SS;O_rad_out;Cs_H_out_2H] + [R3H_SS_Cs;Y_rad_out;Cs_H_out_2H] for rate rule
! [R3H_SS_Cs;O_rad_out;Cs_H_out_2H]
! @Multiplied by reaction path degeneracy 3.0
```

Then, when processing, we could do
```
comment = comment.replace('!', '')
comment = comment.replace('\n', '')
comment = comment.split('@')
```
to get a list with each line of information, ignoring the original line breaks. We could use any character that we're sure will not appear in a species name or template.

goldmanm:
To keep back compatibility with previously generated jobs, keep the `.inp` readable, and not break other stuff that reads these comments, I added a method that removes all the lines and then adds them back in front of keywords like 'Template reaction ', 'Multiplied by reaction...', 'Estimate ...'. The downside of this method is that is slightly slower and sensitive to adding new keywords. The advantages are backwards compatibility, clean chemkin files, and doesn't affect other modules that read comments (like uncertainty analysis).

goldmanm:
This may be put on hold until we figure out if we can make possible reading from an RMG-based file, which could be more robust than parsing comments.

rmgpy/chemkinTest.py
```diff
@@ -118,25 +118,26 @@ def test_readThermoEntry_NoTRange(self):
         self.assertEqual(formula, {'H': 6, 'C': 2})
         self.assertTrue(isinstance(thermo, NASA))
 
-    def testReadAndWriteTemplateReactionFamilyForMinimalExample(self):
+    def testReadAndWriteAndReadTemplateReactionFamilyForMinimalExample(self):
```
mjohnson541: I'm not clear on why this is a better name for this function

rmgpy/chemkinTest.py
```diff
@@ -118,25 +118,26 @@ def test_readThermoEntry_NoTRange(self):
...
 
-        # loadChemkinFile
+        # read original chemkin file
         species, reactions = loadChemkinFile(chemkinPath, dictionaryPath)
 
+        # ensure correct reading
         reaction1 = reactions[0]
         self.assertEqual(reaction1.family, "R_Recombination")
-
+        self.assertEqual(frozenset('C_methyl;C_methyl'.split(';')), frozenset(reaction1.template))
```
mjohnson541: Why can't we compare the templates directly?

goldmanm: The order of the list is not deterministic, so comparing lists with different orders could lead to false Negative.

rmgpy/chemkinTest.py
goldmanm: I added an additional read step, so I thought I'd change the name to mention that. Should I change it back?

rmgpy/data/kinetics/family.py
```diff
@@ -2172,6 +2171,8 @@ def assignWeightsToEntries(entryNestedList, weightedEntries, N = 1):
                 training = {}
                 
                 for tokenTemplateLabel, weight in weightedEntries:
+                    if 'from training reaction' in tokenTemplateLabel:
```
mjohnson541: I don't think these changes will work properly with recent changes in uncertainty parsing.  I think you should probably rebase and take a look at the changes.

rmgpy/data/kinetics/family.py
```diff
@@ -2085,8 +2085,7 @@ def retrieveOriginalEntry(self, templateLabel):
         
         Where the TrainingReactionEntry is only present if it comes from a training reaction
         """
-        
-        templateLabels = templateLabel.split()[0].split(';')
+        templateLabels = templateLabel.split()[-1].split(';')
```
mjohnson541: What improvement does this commit provide?  You don't seem to have changed the comment structure in this PR so if we need this commit to parse properly why does uncertainty parsing work currently?  Is this out of date?

goldmanm: The method of parsing comments here relies on identifying the starting words of lines (the discussion and decision is in the conversation of this PR). Since the template label was at the start of the line, this method wasn't able to work. The uncertainty parsing involves a different scheme to identify line breaks (which is more verbose and has its own quirks), so it is not effected by this change.

rmgpy/data/kinetics/family.py
goldmanm: just rebased. no conflicts. `make test` passed. The quote 'from training reaction' is added to kinetics comments on line 1111 of family.py

rmgpy/chemkinTest.py
```diff
@@ -145,6 +149,19 @@ def testReadAndWriteTemplateReactionFamilyForMinimalExample(self):
         self.assertTrue(os.path.isfile(chemkinSavePath))
         self.assertTrue(os.path.isfile(dictionarySavePath))
 
+        # read newly written chemkin file to make sure the entire cycle works
+        _, reactions2 =loadChemkinFile(chemkinSavePath, dictionarySavePath)
```
mjohnson541: This line does not appear to be necessary.  Codacy points out that reactions2 is unused?

goldmanm: Thanks for bringing that up. Just fixed the issue (caused by not calling reactions2 later in the method). 

Rebased. Let me know if there is anything else you find.

rmgpy/chemkinTest.py
mjohnson541: I guess this is fine.  I find it kind of hard to read with all the and's, but it is a precise name.

goldmanm:
@mjohnson541, I just rebased this to current master. If you have a chance and it looks goood, could your re-approve and merge it?

Thanks for looking it over.

Issue #1077, opened 2017-07-10
`None` is generated instead of a Species during enlarge() step, possibly due to reaction deflation?


alongd:
Running a job on master, which previously ran smoothly, now generates the following trace:

```
Traceback (most recent call last):
  File "/home/alongd/Code/RMG-Py//rmg.py", line 159, in <module>
    rmg.execute(**kwargs)
  File "/home/alongd/Code/RMG-Py/rmgpy/rmg/main.py", line 689, in execute
    bimolecularReact=self.bimolecularReact)
  File "/home/alongd/Code/RMG-Py/rmgpy/rmg/model.py", line 645, in enlarge
    self.processNewReactions([rxn], spc)
  File "/home/alongd/Code/RMG-Py/rmgpy/rmg/model.py", line 774, in processNewReactions
    self.addReactionToUnimolecularNetworks(rxn, newSpecies=newSpecies, network=pdepNetwork)
  File "/home/alongd/Code/RMG-Py/rmgpy/rmg/model.py", line 1395, in addReactionToUnimolecularNetworks
    assert isinstance(newSpecies, Species)
AssertionError
```

where an object sent to `addReactionToUnimolecularNetworks()` is not a Species.

Adding a logging.info statement to [enlarge()](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/rmg/model.py#L639) (which calls `processNewReactions` which then calls `addReactionToUnimolecularNetworks()`) gives prior to the traceback:
```
rxn: C=C=N + [H] <=> 21, spc: CH2CHNH(21)
rxn: 21 <=> C[C]=N, spc: CH2CHNH(21)
rxn: 21 <=> CC=[N], spc: CH2CHNH(21)
rxn: 21 <=> C1[CH]N1, spc: CH2CHNH(21)
rxn: N=[CH] + [CH2] <=> 21, spc: CH2CHNH(21)
rxn: [CH2][C]=N + [H] <=> 21, spc: CH2CHNH(21)
rxn: [H] + C(=[N])[CH2] <=> 21, spc: CH2CHNH(21)
rxn: C=C=N + [H] <=> C(=C)[NH], spc: None
```

I.e., `retrieveNewSpecies()` returned `None`.

[...]

PR #1078, opened 2017-07-10
allow all resonance structures to be used in deflating

goldmanm:
deflate now checks all resonance structures before deflating
reaction objects. This is meant to solve issue #1077, if it is
caused by resonance structures when deflating.

Deflate could use species objects in the future, but we currently
don't have an equals or hash function for species.


goldmanm:
@alongd, let me know if this prevents and/or changes the error. You can also send me the file and I can run it to see if it works.

alongd:
Yes, this did the trick. Thanks @goldmanm!

alongd:
Added an Exception to catch instances when `retrieveNewSpecies()` misbehaves due to the reason in this PR's title or another.

Issue #1079, opened 2017-07-10
Recent Slowdown in RMG


amarkpayne:
I think other people have noticed this as well, but we wanted to post an issue here so that we can have a discussion about it in one place. I believe the slowdown is due to one or more pull requests made on or after our pull-request hackathon.

On my computer, the minimal example has gone from taking 3-4 mins to 7-8 mins. I believe the slowdown becomes more substantial as the model size increases, but I am not entirely sure about this.

I ran the minimal example on a branch I created a few days ago (7-8 min runtime), and a branch I created at the beginning of June(3-4 min runtime). Attached are the RMG profile stats if anyone is interested.
[Slow_RMG.profile.dot.pdf](https://github.com/ReactionMechanismGenerator/RMG-Py/files/1137231/Slow_RMG.profile.dot.pdf)
[Fast_RMG.profile.dot.pdf](https://github.com/ReactionMechanismGenerator/RMG-Py/files/1137232/Fast_RMG.profile.dot.pdf)

mjohnson541:
Turning Dynamics Criterion calculations off makes a difference of ~14% in the minimal example.  This is implemented automatically in later Accelerator Tools pull requests when the dynamics criterion is unused.

goldmanm:
Improving the accuracy of degeneracy and symmetry meant dealing with all the resonance structures, which at least partially contributes to the slowdown. The greatest contribution to slowing down running RMG from my contributions would be in `findDegeneracies`, since we are generating resonance structures to eliminate degenerate reactions (instead of just using one molecular graph for comparison). This increases the resonance structure generation from ~2% runtime to ~8% runtime, which is quite substantial (given that the new case runs for longer). This could be reduced by about half if we could access the species already in the core/edge instead of generating new resonance structures in `findDegeneracies`. However, in parallelization, I am not sure the generate reaction nodes contain information about species in core and edge. 

I think some is also due to just comparing many more structures when dealing with all the resonance isomers. In order to get accurate values, isomorphic resonance structures are generated, which are sort of redundant (since it is the same graph), but I haven't found an alternative to yielding accurate symmetry and degeneracy values without them.

rwest:
If the profile graphics listed the actual (cumulative) time within each function, not just as a percentage of the total time, it would make comparison across different runs easier. (because the total time is different across runs).
[...]

PR #1080, opened 2017-07-11
Superminimal example

mjohnson541:
While there are many jokes about impatience that could be made about this pull request, sometimes in the middle of rebasing you want to know that you haven't broken anything major and in those cases it doesn't really make sense to start the entire minimal example especially since it can now take up to 8 minutes to run see #1079 .  This example based on H2 oxidation runs in ~20 sec.  Other people who wish to remain anonymous referred to this as the "ain't nobody got time for that" pull request.  

Can be run from the makefile with "make eg0"

### DIFF
--- a/examples/rmg/superminimal/input.py
+++ b/examples/rmg/superminimal/input.py
@@ -1,0 +1,56 @@
+# Data sources
+database(
+    thermoLibraries = ['primaryThermoLibrary'],
+    reactionLibraries = [],
+    seedMechanisms = [],
+    kineticsDepositories = ['training'],
+    kineticsFamilies = ['H_Abstraction','Disproportionation','R_Recombination',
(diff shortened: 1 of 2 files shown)


examples/rmg/superminimal/input.py
```diff
@@ -0,0 +1,55 @@
+# Data sources
+database(
+    thermoLibraries = ['primaryThermoLibrary'],
+    reactionLibraries = [],
+    seedMechanisms = [],
+    kineticsDepositories = ['training'],
+    kineticsFamilies = 'default',
```
rwest: I'm sure you could make it even faster by not loading all the reaction families (isn't "averaging up" a slow point?)....   🏃‍♀️💨  🥇

mjohnson541: Good idea, gets us down to ~7 sec!

PR #1083, opened 2017-07-12
Toluene Free Rotor Example

mjohnson541:
We apparently forgot to add a Cantherm example using Free Rotors when Free Rotors was originally implemented.  This adds an appropriate primary input file and species input file treating the Benzyl rotor as a Free Rotor.  Thanks to @rgillis8 for pointing out that we didn't have any Free Rotor examples.


zjburas:
Benzyl is probably not a good free rotor example because it has a high barrier to internal rotation. I think toluene would be a good example (<1 kcal barrier to internal rotation).

mjohnson541:
I only did this for Benzyl because the files were available if someone has the right Gaussian output files for Toulene I can do that instead.

mjohnson541:
Ok, ignore the branch name, we now have a new Toluene example thanks to @laitcl.  In the toluene file it has both a hindered rotor and a free rotor example.

PR #1085, opened 2017-07-12
Add code to set walltime from commandline flag

amarkpayne:
Walltime flag was implemented in rmg.py, but was not utilized in
rmgpy.rmg.main.py


amarkpayne:
@goldmanm  Are we having a problem with the Travis builds? The pull request one went through just fine, but the push build timed out. Do I need to do anything about it?

PR #1088, opened 2017-07-14
added atom energies for m06-2x model chemistry

KEHANG:
Since we've been increasingly using M06-2x for thermo calculations, this PR tries to expands model chemistry set that cantherm allows by adding M06's atom energies.

One might be interested in how we get these numbers; they are from single atom calculation using the level of theory you want. The assumption for the multiplicity of each atom is: H singlet, C triplet, O triplet, N quartet, S triplet, P quartet.


alongd:
Thanks @KEHANG !
Could you add the method you described for getting these numbers to the docstring?

Issue #881, opened 2017-02-03
Desired Feature: automatically create a subdirectory that can be used as a seed mechanism?


cfgoldsmith:
This is not so much an issue as a request.
In the same way that RMG automatically creates a subdirectory ./chemkin/ that includes a chem.inp and a _species_dictionary.txt_, it would be very helpful if RMG would make a new directory ./seed/ that includes _dictionary.txt_ and _reactions.py_, appropriately formatted, so that I could use this directory as a seed mechanism for subsequent runs.
Thanks!

rwest:
See #1093

PR #1093, opened 2017-07-17
automatic seed mechanism generation

mjohnson541:
This pull request causes RMG each iteration to automatically:
1) Generate a seed mechanism based on the current chemkin and species dictionary files
2) save that seed mechanism in a seed folder in the run directory
3) save copies of the associated thermo and kinetic libraries into database under
an input name (new parameter in the options block) (or replace the current thermo and kinetic libraries of that name) 

Makes feature for #881.

### DIFF
--- a/examples/rmg/commented/input.py
+++ b/examples/rmg/commented/input.py
@@ -150,6 +150,16 @@ model(
 )
 
 options(
+    	#provides a name for the seed mechanism produced at the end of an rmg run default is 'Seed'
+    name='SeedName',	
+	#if True every iteration it saves the current model as libraries/seeds
+	#(and deletes the old one)
+	#Unlike HTML this is inexpensive time-wise
+	#note a seed mechanism will be generated at the end of a completed run and some incomplete 
+	#runs even if this is set as False
+    generateSeedEachIteration=True,
+	#If True the mechanism will also be saved directly as kinetics and thermo libraries in the database
+    saveSeedToDatabase=False,
 	#only option is 'si'
     units='si',
 	#how often you want to save restart files.  
--- a/rmgpy/data/kinetics/common.py
+++ b/rmgpy/data/kinetics/common.py
@@ -90,10 +90,6 @@ def saveEntry(f, entry):
         #Write out additional data if depository or library
         #kinetic rules would have a Group object for its reactants instead of Species
         if isinstance(entry.item.reactants[0], Species):
-            if entry.label != str(entry.item):
-                raise KineticsError("Reactions are now defined solely by their labels, "
-                                                    "but reaction {0!s} has label {1!r}".format(
-                                                     entry.item, entry.label))
             # Add degeneracy if the reaction is coming from a depository or kinetics library
             f.write('    degeneracy = {0:.1f},\n'.format(entry.item.degeneracy))
             if entry.item.duplicate:
--- a/rmgpy/reaction.py
+++ b/rmgpy/reaction.py
@@ -150,12 +150,20 @@ class Reaction:
         Return a string representation of the reaction, in the form 'A + B <=> C + D'.
         If a specificCollider exists, the srting representation is 'A + B (+S) <=> C + D (+S)'.
         """
+        return self.toLabeledStr(use_index=True)
+    
+    def toLabeledStr(self, use_index=False):
+        """
+        the same as __str__ except that the labels are assumed to exist and used for reactant and products rather than 
+        the labels plus the index in parentheses
+        """
         arrow = ' <=> '
         if not self.reversible: arrow = ' => '
+        
         if self.specificCollider:
-            return arrow.join([' + '.join([str(s) for s in self.reactants])+' (+'+str(self.specificCollider)+')', ' + '.join([str(s) for s in self.products])+' (+'+str(self.specificCollider)+')'])
+            return arrow.join([' + '.join([str(s) if use_index else s.label for s in self.reactants])+' (+'+str(self.specificCollider)+')', ' + '.join([str(s) if use_index else s.label for s in self.products])+' (+'+str(self.specificCollider)+')'])
         else:
-            return arrow.join([' + '.join([str(s) for s in self.reactants]), ' + '.join([str(s) for s in self.products])])
+            return arrow.join([' + '.join([str(s) if use_index else s.label for s in self.reactants]), ' + '.join([str(s) if use_index else s.label for s in self.products])])
 
     def __reduce__(self):
         """
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -825,8 +825,11 @@ class RMG(util.Subject):
                 while name+str(q) in thermoNames or name+str(q) in kineticsNames:
                     q += 1
                 self.name = name + str(q)
-            
-        if not os.path.exists(os.path.join(currentDir,'seed')): #if seed directory does not exist make it
+        
+        if firstTime and not os.path.exists(os.path.join(currentDir,'seed')): #if seed directory does not exist make it
+            os.system('mkdir seed')
+        else:
+            os.system('rm -rf seed') #otherwise delete the old seed and make a new directory
             os.system('mkdir seed')
             
         speciesList = self.reactionModel.core.species
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -874,15 +874,13 @@ class RMG(util.Subject):
         kineticsLibrary.checkForDuplicates(markDuplicates=True)
         kineticsLibrary.convertDuplicatesToMulti()
     
-        # Save in Py formatp
-        databaseDirectory = settings['database.directory']
-        try:
-            os.makedirs(os.path.join(databaseDirectory, 'kinetics', 'libraries',name))
-        except:
-            pass
-        
         #save in database
         if self.saveSeedToDatabase:
+            databaseDirectory = settings['database.directory']
+            try:
+                os.makedirs(os.path.join(databaseDirectory, 'kinetics', 'libraries',name))
+            except:
+                pass
             thermoLibrary.save(os.path.join(databaseDirectory, 'thermo' ,'libraries', name + '.py'))
             kineticsLibrary.save(os.path.join(databaseDirectory, 'kinetics', 'libraries', name, 'reactions.py'))
             kineticsLibrary.saveDictionary(os.path.join(databaseDirectory, 'kinetics', 'libraries', name, 'dictionary.txt'))
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -816,16 +816,17 @@ class RMG(util.Subject):
         
         currentDir = os.getcwd()
         
-        if firstTime: #make sure don't overwrite current libraries
-            thermoNames = self.database.thermo.libraries.keys()
-            kineticsNames = self.database.kinetics.libraries.keys()
-            
-            if name in thermoNames or name in kineticsNames: 
-                q = 1
-                while name+str(q) in thermoNames or name+str(q) in kineticsNames:
-                    q += 1
-                name += str(q)
-                self.name = name
+        if self.saveSeedToDatabase:
+            if firstTime: #make sure don't overwrite current libraries
+                thermoNames = self.database.thermo.libraries.keys()
+                kineticsNames = self.database.kinetics.libraries.keys()
+                
+                if name in thermoNames or name in kineticsNames: 
+                    q = 1
+                    while name+str(q) in thermoNames or name+str(q) in kineticsNames:
+                        q += 1
+                    name += str(q)
+                    self.name = name
             
         if not os.path.lexists(os.path.join(currentDir,'seed')): #if seed directory does not exist make it
             os.system('mkdir seed')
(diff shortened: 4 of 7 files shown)


mjohnson541:
I think I'll do some profiling generating a seed mechanism every iteration and see how expensive that is.

alongd:
Does it change the species label at the end?
If we run an RMG output "as is" as a seed/library, we get double indices for species: e.g. `OH(5)(3)`
Shall we leave the double indices, or remove label indices for species with a unique name at the end of the run (just for the seed, or for the final output if we're at it)?

mjohnson541:
Right now it does not, I load it directly from the chemkin file.  I'll take a look at doing it directly.

mjohnson541:
Now it does so, I've also set it to save the seed mechnaism every iteration because it doesn't seem to be expensive, although I intend to do some more formal speed tests.

goldmanm:
Is there a method similar to 'loadChemkinFile' which can take a seed mechanism and output a list of species and reactions?

This might remove the need to save a `chem.inp` file at each iteration as well. 

When loading the seed mechanism, will it identify that a reaction came from kinetics estimation and create a TemplateReaction object with the family, template, degeneracy filled in? Or will it just be a general LibraryReaction object with a loss of the other information?

mjohnson541:
It should be equivalent to a library so we should be able to load it as a library.  I'm pretty sure it will current load it as a library reaction, but it may not be hard to make it load as a template reaction when appropriate if the proper fields are available.  Although we need to be careful how we treat it, because we don't know if the database has changed between when the seed mechanism was made and when the file is loaded.

goldmanm:
If there is an easy way to convert the seed mech into a chemkin file, then we can remove the outputting of all the `chem_034.inp` files. If there is not an easy way to convert these seed mechanisms to a chemkin file, then maybe we could just output the seed mechanism at the end of the simulation (to avoid printing out too many files & folders).

mjohnson541:
I feel like the standard way to do so should be to load the seed mechanism to make an RMG mechanism and then print out the chemkin from that.  I don't think you can do that as of now, but we definitely have the code to print out the mechanism in chemkin format and we can definitely load the seed mechanism into an RMG mechanism so it should just be a matter of putting those two together.

mliu49:
I haven't looked at the changes, but I had a few comments in response to the thread:
1. Could you make turning this on an option in the additional options block of the input file? For most of my jobs, I prefer to minimize file output to reduce speed and storage space.
2. Personally, I would prefer that it does not automatically copy the seed mech to RMG-database, although it could be an option.
3. Why do we need to load a seed mechanism to convert it to a chemkin file? I didn't quite follow what led to this point.
4. I like having chemkin files at each iteration, because I can look at how the mechanism evolved at different points. Having a library written at each iteration would probably accomplish the same thing, but seems more cumbersome, since it won't run in chemkin.

mjohnson541:
Each iteration the old seed is replaced with the new seed currently so it shouldn't have any associated memory problems.  I'll add options for those, but it takes so little time to make the seed files I'm inclined to think that seed file generation every iteration should be default.  

Sorry you're missing context from #1096.  This discussion probably belongs there, I don't think any of that would be part of this pull request.  But we are discussing  the possibility of making the seed mechanism our primary output file instead of heavily commented chemkin files since we have full control over the seed mechanism file and this would heavily reduce our needs to parse things.

goldmanm:
If we are likely to use the seed mechanisms, it may be better to have more flexible or intuitive naming than currently available in chemkin. See #1065 for a way to rename species independently using SMILES notation (thought you don't have to convert to cantera objects).

mjohnson541:
Oh, good idea!  I was already doing a bit better than chemkin labels, but this is better.  Do we want to centralize this functionality?  Or should I just plop it into main.py?

rmgpy/rmg/main.py
```diff
@@ -1193,7 +1327,6 @@ def readMeaningfulLineJava(self, f):
         return line
     
 ################################################################################
-
```
goldmanm: you may want to get rid of the spare line additions and deletions in the history.

goldmanm:
We might want to have a few unittests for `makeSeedMech` to ensure it continues to work after others have merged in other changes.

examples/rmg/commented/input.py
```diff
@@ -150,6 +150,16 @@
 )
 
 options(
+    	#provides a name for the seed mechanism produced at the end of an rmg run default is 'Seed'
+    name='RMGButane',	
+	#if True every iteration it saves the current model as a libraries/seeds
+	#(and deletes the old one)
+	#Unlike HTML this is inexpensive time-wise
+	#note this one will be generated at the end of a completed run and some incomplete runs even
```
alongd: Could you rephrase the comment?
"This one will be generated ... even if False" isn't clear.
If "this one" refers to `generateSeeds`, then let's write it explicitly.
If I did understand correctly, then why is the seed generated even if `generateSeeds=False`?

examples/rmg/commented/input.py
```diff
@@ -150,6 +150,16 @@
...
+	#if False
+    generateSeeds=True,
```
alongd: If we always overwrite the previous seed, perhaps we should rename this handle in the single form `generateSeed`

examples/rmg/commented/input.py
```diff
@@ -150,6 +150,16 @@
...
+	#If True the mechanism will also be saved directly as seeds/libraries in the database
```
alongd: perhaps consider:
"#If True, the final seed mechanism will also be saved directly in the database as thermo and kinetic libraries"

rmgpy/rmg/main.py
```diff
@@ -761,7 +777,125 @@ def execute(self, **kwargs):
         logging.info('The final model edge has %s species and %s reactions' % (edgeSpec, edgeReac))
         
         self.finish()
-
+    
+    def makeSeedMech(self,firstTime=False):
+        """
+        causes RMG to make a seed mechanism out of the current chem_annotated.inp and species_dictionary.txt
+        this seed mechnaism is outputted in a seed folder within the run directory and automatically
```
alongd: typo in word `mechnaism`

rmgpy/rmg/main.py
```diff
@@ -761,7 +777,125 @@ def execute(self, **kwargs):
...
+        """
+        
+        logging.info('Making seed mechanism...')
+        
+        name = self.name
+        
+        currentDir = os.getcwd()
+        
+        if self.saveSeedToDatabase:
+            if firstTime: #make sure don't overwrite current libraries
```
alongd: Instead of double `if` statements, consider adding `and`
(nothing gets done here unless both are True anyway)

rmgpy/rmg/main.py
```diff
@@ -761,7 +777,125 @@ def execute(self, **kwargs):
...
+                thermoNames = self.database.thermo.libraries.keys()
+                kineticsNames = self.database.kinetics.libraries.keys()
+                
+                if name in thermoNames or name in kineticsNames: 
+                    q = 1
+                    while name+str(q) in thermoNames or name+str(q) in kineticsNames:
+                        q += 1
+                    name += str(q)
+                    self.name = name
```
alongd: These two can be written in one line: `self.name = name + str(q)`

rmgpy/rmg/main.py
```diff
@@ -761,7 +777,125 @@ def execute(self, **kwargs):
...
+            
+        if not os.path.lexists(os.path.join(currentDir,'seed')): #if seed directory does not exist make it
```
alongd: `lexists()` vs. `exists()`:
For my understanding, why is it important to return `True` for broken symbolic links?

rmgpy/rmg/main.py
```diff
@@ -761,7 +777,125 @@ def execute(self, **kwargs):
...
+        	    entry.longDesc = reaction.kinetics.comment
+            kineticsLibrary.entries[i+1] = entry
+        
+        # Mark as duplicates where there are mixed pressure dependent and non-pressure dependent duplicate kinetics
+        # Even though CHEMKIN does not require a duplicate flag, RMG needs it.
+        # Using flag markDuplicates = True
+        kineticsLibrary.checkForDuplicates(markDuplicates=True)
+        kineticsLibrary.convertDuplicatesToMulti()
+    
+        # Save in Py formatp
```
alongd: minor typo

rmgpy/rmg/main.py
```diff
@@ -761,7 +777,125 @@ def execute(self, **kwargs):
...
+        databaseDirectory = settings['database.directory']
+        try:
+            os.makedirs(os.path.join(databaseDirectory, 'kinetics', 'libraries',name))
```
alongd: Shouldn't this be under `if firstTime`?

alongd: This currently creates an empty folder with the name I chose to the seed, even if `saveSeedToDatabase=False`.
Should be under `if saveSeedToDatabase:`

examples/rmg/commented/input.py
```diff
@@ -150,6 +150,16 @@
 )
 
 options(
+    	#provides a name for the seed mechanism produced at the end of an rmg run default is 'Seed'
+    name='RMGButane',
```
alongd: Perhaps change this to `libraryName` or `seedName` so it's easily understood and not too generic?
BTW, the terminology `seed` should perhaps be reconsidered since we're saving a library which the user can use either as a library or as a seed.

alongd:
The species names in dictionary isn't the same as in reaction.py. In the later it's with parenthesis. @mjohnson541, could you take a look?

examples/rmg/commented/input.py
mjohnson541: Yeah, it should probably be easier to distinguish from the real libraries, I think I'll just name it 'SeedName'.  

I know, but I think use of these as libraries is highly unlikely (since the rates are either from other libraries or from rate rules) so I still think Seed makes the most sense for the user.

examples/rmg/commented/input.py
mjohnson541: I spent quite a while debating that.  My concern is that the user will think the seed is only generated at the end of the run if I name it generateSeed.  I think I'll name it instead generateSeedEachIteration, which is annoying, but clearer.

examples/rmg/commented/input.py
mjohnson541: done

rmgpy/rmg/main.py
mjohnson541: There isn't a real difference, you could use either.  I'll change it to exists to avoid confusion.

rmgpy/rmg/main.py
mjohnson541: I believe I added this because it made it look more clean, I think it was suppose to be there.

rmgpy/data/kinetics/common.py
```diff
@@ -90,10 +90,6 @@ def sortEfficiencies(efficiencies0):
         #Write out additional data if depository or library
         #kinetic rules would have a Group object for its reactants instead of Species
         if isinstance(entry.item.reactants[0], Species):
-            if entry.label != str(entry.item):
```
mjohnson541: @goldmanm does the fact that I'm deactivating this failsafe bother you?  I believe it's suppose to prevent rmg from outputting a seed with non-unique names, but it does so by requiring that they be the string names, which I would argue is too restrictive.

mjohnson541:
I believe I've fixed all of the above issues, the reaction file strings are fixed now.  If you guys don't have any more concerns I'm going push this.

rmgpy/reaction.py
```diff
@@ -156,7 +156,19 @@ def __str__(self):
...
+        the same as __str__ except that the labels are assumed to exist and used for reactant and products rather than 
+        the labels plus the index in parentheses
+        """
+        arrow = ' <=> '
+        if not self.reversible: arrow = ' => '
+        if self.specificCollider:
+            return arrow.join([' + '.join([s.label for s in self.reactants])+' (+'+str(self.specificCollider)+')', ' + '.join([s.label for s in self.products])+' (+'+str(self.specificCollider)+')'])
+        else:
+            return arrow.join([' + '.join([s.label for s in self.reactants]), ' + '.join([s.label for s in self.products])])
+
```
goldmanm: Do you know if it is possible to add an option to this method to have the index in parenthesis? That way we can just have 

```def __str__(self):
    return toLabeledStr(use_index = True)
```

This would avoid code duplication, so modifications are more robust (for example someone only adds a modfication of the specific collider to __str__ and not toLabeledStr)

mjohnson541: Good idea!

rmgpy/rmg/main.py
goldmanm: It may make the plain code look cleaner, which is an improvement. Unfortunately a consequence is that modifying whitespace mixed in with other changes makes understanding the changes (git blame), (git show() or (git log --follow) more messy. 

If you'd like to improve some formatting, I think it would be best practices to improve the formatting of a whole file (or region) and make the formatting improvement a separate commit (and possibly pull request).

mjohnson541: Ok

rmgpy/rmg/mainTest.py
```diff
@@ -116,6 +117,19 @@ def testRMGExecute(self):
...
+        os.chdir(seedDir)
+        self.assertTrue(os.path.exists(self.rmg.name+'.py')) #thermo library made
+        self.assertTrue(os.path.exists(self.rmg.name)) #kinetics library folder made
+        kDir = os.path.join(seedDir,self.rmg.name)
+        os.chdir(kDir)
+        self.assertTrue(os.path.exists('dictionary.txt')) #dictionary file made
+        self.assertTrue(os.path.exists('reactions.py')) #reactions file made
+        
+        os.chdir(os.path.join(originalPath,self.dir_name)) #return to original directory
+
```
goldmanm: I like that we are testing that something is created. I have had a few cases where tests like this often do not check when the file written no longer functions when being read into RMG. This has in the past led to unexpected broken functionality, which is important if we want to rely more on reading from these files.

I guess there is a benefit to pushing this commit through, but writing specific tests is something to keep in mind.

rmgpy/data/kinetics/common.py
goldmanm: I guess I don't quite understand how reading the entries work if the reaction label does not have the names of species inside it.

rmgpy/rmg/mainTest.py
mjohnson541: Now we load and check that we've properly loaded the seed mech through both the library and seed mechanism functions.

rmgpy/data/kinetics/common.py
mjohnson541: I'm pretty sure this is fine

PR #1094, opened 2017-07-18
add new line to nodal distance comments

goldmanm:
this PR fixes nodal distance comments so they appear on a new line
in the chemkin comments. This allows accurate parsing of comments
where one line corresponds to one piece of information, and is
necessary for PR #1076 to be successfully merged in.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -2139,6 +2139,9 @@ class KineticsFamily(Database):
             training = []
             
             lines = kinetics.comment.split('\n')
+            
+            lines = [line for line in lines if not line.startswith('Euclid')] #remove the Euclidean distance line to help parser
+            
             # Discard the last line, unless it's the only line!
             # The last line is 'Estimated using ... for rate rule (originalTemplate)'
             if len(lines) == 1:
--- a/rmgpy/data/kinetics/rules.py
+++ b/rmgpy/data/kinetics/rules.py
@@ -697,7 +697,7 @@ class KineticsRules(Database):
                     )
                 
         kinetics.comment += ' for rate rule ' + originalLeaves
-        kinetics.comment += ' Euclidian distance = {}.'.format(minNorm)
+        kinetics.comment += '\nEuclidian distance = {}'.format(minNorm)
         kinetics.A.value_si *= degeneracy
         if degeneracy > 1:
             kinetics.comment += "\n"


mjohnson541:
This creates an issue when extracting sources from comments.  I've added two lines to delete the Euclidean distance line before fully parsing the comments.

rmgpy/data/kinetics/family.py
```diff
@@ -2139,6 +2139,10 @@ def assignWeightsToEntries(entryNestedList, weightedEntries, N = 1):
             training = []
             
             lines = kinetics.comment.split('\n')
+            
+            euc = [i for i in xrange(len(lines)) if lines[i].startswith('Euclid')] #remove the Euclidean distance line to help parser
```
rwest: Is this just `euc = [line for line in lines if line.startswith('Euclid')]` ?

Or the two lines together:
`lines = [line for line in lines if not line.startswith('Euclid')]`
?

goldmanm:
@mjohnson541, I don't understand the reasoning behind removing the comments before parsing. The commit I wrote fixes the issue I was having of not being able to use `token[-1]` when parsing comments. Why do we need to remove the line?

goldmanm:
Also, what is the Euclidean distance comment used for? Maybe there's a better way to store it than attaching to comments (which increases memory usage). If it is mostly for debugging which node is hit, then maybe it only needs to be added with verbose comments?

rwest:
We (eg. @chiyw13 and @bslakman ) have been using the Euclidean distance to give some indication of how (un)reliable or (un)suitable the estimate. (Higher nodal distance does seem to mean a worse estimate). Perhaps it counts as a "verbose" comment and could be turned on only when those are in use.

rmgpy/data/kinetics/family.py
mjohnson541: Yeah, I think lines = [line for line in lines if not line.startswith('Euclid')] is probably a cleaner way to write this.

mjohnson541:
One of the unittests fails when trying to "eval" out some of the template lists.  The euclidean distance comment gets in the way so it needs removed before that operation is done.

rmgpy/data/kinetics/family.py
mjohnson541: This should be fixed now

goldmanm:
@mjohnson541 Thanks for getting it working. I guess parsing strings to get out specific information is inherently a finicky process when the string's format is being changed.

mjohnson541:
No problem, I guess we we're just lucky it happened to not get in the way the first time.

PR #1100, opened 2017-07-19
remove unused “specificColliderEfficiencies” variable from simple.pyx

mjohnson541:
This variable is unused and apparently appears to be the source of a compiling issue for @amarkpayne.


alongd:
Thanks for the catch!

PR #1105, opened 2017-07-23
Read externalSymmetry and spinMultiplicity from QM job output

alongd:
Read externalSymmetry and spinMultiplicity from QM job output if
unspecified in CanTherm input.
Previously an error was thrown if either externalSymmetry or
spinMultiplicity were unspecified in the CanTherm input, although we are
able to read both from the gaussian/QChem outputs.
Since externalSymmetry is often incorrect (at least in QChem), a general
warning is given if this value is unspecified.
Documentation changed as well.

### DIFF
--- a/documentation/source/users/cantherm/input.rst
+++ b/documentation/source/users/cantherm/input.rst
@@ -113,22 +113,22 @@ will be referred to as the species input file.
 
 The species input file accepts the following parameters:
 
-======================= =========================== ====================================
-Parameter               Required?                   Description
-======================= =========================== ====================================
-``atoms``               yes                         Type and number of atoms in the species
-``bonds``               optional                    Type and number of bonds in the species
-``linear``              yes                         ``True`` if the molecule is linear, ``False`` if not
-``externalSymmetry``    yes                         The external symmetry number for rotation
-``spinMultiplicity``    yes                         The ground-state spin multiplicity (degeneracy)
-``opticalIsomers``      yes                         The number of optical isomers of the species
-``energy``              yes                         The ground-state 0 K atomization energy in Hartree (without zero-point energy)
-                                                    **or**
-                                                    The path to the quantum chemistry output file containing the energy
-``geometry``            yes                         The path to the quantum chemistry output file containing the optimized geometry
-``frequencies``         yes                         The path to the quantum chemistry output file containing the computed frequencies
-``rotors``              optional                    A list of :class:`HinderedRotor()` and/or :class:`FreeRotor()` objects describing the hindered/free rotors
-======================= =========================== ====================================
+======================= ============================= ====================================
+Parameter               Required?                     Description
+======================= ============================= ====================================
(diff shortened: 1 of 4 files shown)


alongd:
This PR is meant to automate reading the external symmetry for a CanTherm job.
However, I noticed that QChem does a bad job in determining the symmetry. For example, it evaluated the symmetry of [CH3] as 1, where in fact it is 6. Gaussian might do a better job in assessing the external symmetry.
If the user trusts the QM software to determine symmetry, now there's an option read it.

rwest:
QM codes (including Gaussian) are so bad at guessing symmetry I'm tempted to say this is a bad idea and we should force the user to specify it by hand.

But with my other hat on I'm all about automation and removing the humans....

So I'm torn. 
I guess just don't let the default be to read and trust it, even if there's a warning (nobody reads warnings).

alongd:
I feel you...

So instead of accepting an input file w/o externalSymmetry, we can enforce using it, but if it's a negative value or a specific key word, we'll read from the QM output.

BWT, CanTherm users, what is you favorite method to determine symmetry? Anything we can automate?

keceli:
Are the codes really bad, or are they using different 'conventions' for the
symmetry number? I think symmetry number depends on how one calculates the
partition function and I guess there is not a standard for that.

I also remember using a separate code to compute symmetry number for RMG,
is it not used anymore?

alongd:
@keceli , good point.
I'll try to see the definitions/conventions of the different QC software (however, 1 for a planar symmetric `[CH3]` molecule looks like a really bad job by QChem, regardless of convention, but I agree it's worth checking)

Our [calculateSymmetryNumber](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/4bd520abe99072c0b9e6dda82cae658eca6c6e9b/rmgpy/molecule/symmetry.py#L420) function was able to correctly calculate the external symmetry number for [CH3]. It didn't even occur to me to look within RMG, since I though (still think?) a 3D structure is required to calculate the symmetry number for many cases.

@keceli , could you direct me to another place in the code, if you're aware of it?

nickvandewiele:
Back in my days, QM jobs sent the optimized 3d coordinates to a tool called `SYMMETRY` that determined the point group. We converted the point group to a symmetry number through a hard-coded map.
https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/qm/symmetry.py#L136

keceli:
> @keceli , could you direct me to another place in the code, if you're aware of it?

I don't know more than what Nick already pointed out.
https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/qm/symmetry.py#L136

alongd:
@nickvandewiele and @keceli , thanks for the input, very much appreciated!
I'll try to either use this code directly or make a similar method for Cantherm.

alongd:
This PR currently has a low priority.
I'm closing the PR for now since I won't get to continue working on it soon.

Issue #1101, opened 2017-07-20
avoid circular dependencies with separate file for RMG-specific exceptions


goldmanm:
I came across a circular dependency when trying to implement `KineticsError` in arrhenius.pyx. Importing KineticsError required importing all import statements in `kinetics.data.common.py` which imported `reaction.py` which then tried to import objects from `arrhenius.pyx`. 

Not being able to use `KineticsError` in classes descending from `KineticsData` is not intuitive nor functional. 

To avoid this, and other, circular imports, maybe grouping all exceptions in a separate file `rmgpy.exceptions.py` which doesn't require any imports. This could also add descriptions for when and where these Exceptions should be used. Any thoughts on this? Any potentially better organizational strategies?

Mark

amarkpayne:
I think this makes a lot of sense.  Is there a way that this can be done without having to modify all the references to KineticsError in all of RMG?

goldmanm:
I think it would require changing the import statements from `from rmgpy.data.kinetics.common import KineticsError` to `from rmgpy.exceptions import KineticsError` in each file that utilizes the error. I'd likely also want to do this for all the dozen or so exceptions: `UndeterminableKineticsError`, `ThermoError`, `SolverError`, etc. 

It's not an easy task to reorder. Which is why I posted this issue to get some feedback before trying to take on the challenge.

amarkpayne:
Could you create a file `rmgpy.exceptions` to place all the exceptions in, and instead of changing every single reference in RMG simply go to `rmgpy.data.kinetics.common` and define the `KineticsError` class there to inherent from the `KineticsError` class defined in `rmgpy.exceptions`? That way:

1) New references can reference `KineticsError` from `rmgpy.exceptions`.
2) All old references to `KineticsError` from `rmgpy.data.kinetics.common` will still work, without any modification needed.
3) Any future changes to `KineticsError` will happen in one place.

Would this work?

goldmanm:
I think that's an ingenious workaround to avoid changing all import statements.

mjohnson541:
closed with #1106

PR #1106, opened 2017-07-24
moved RMG exception class to specific file

goldmanm:
When importing an RMG exception, before this commit, all the
imported modules for that file are automatically imported, which
can sometimes lead to circular dependency issues (see #1101 ).

This commit centralizes all exceptions within RMG into rmgpy.exceptions
and does not require any imports, to avoid any circular dependency
issues. This also should make it easier for people to find and use
more specific exceptions in RMG in the future.

### DIFF
--- a/rmgpy/data/kinetics/common.py
+++ b/rmgpy/data/kinetics/common.py
@@ -33,27 +33,11 @@ This module contains classes and functions that are used by multiple modules
 in this subpackage.
 """
 
-from rmgpy.data.base import DatabaseError, LogicNode
-from rmgpy.reaction import Reaction, ReactionError
+from rmgpy.data.base import LogicNode
+from rmgpy.reaction import Reaction
 from rmgpy.molecule import Group
 from rmgpy.species import Species
-
-################################################################################
-
-class KineticsError(Exception):
-    """
-    An exception for problems with kinetics. Pass a string describing the problem.
-    """
-    pass
-
-class UndeterminableKineticsError(ReactionError):
-    """
-    An exception raised when attempts to estimate appropriate kinetic parameters
-    for a chemical reaction are unsuccessful.
-    """
-    def __init__(self, reaction, message=''):
-        new_message = 'Kinetics could not be determined. '+message
-        ReactionError.__init__(self,reaction,new_message)
+from rmgpy.exceptions import DatabaseError, KineticsError
 
 ################################################################################
 
--- a/rmgpy/exceptions.py
+++ b/rmgpy/exceptions.py
@@ -1,0 +1,274 @@
+#!/usr/bin/env python
+# encoding: utf-8
+
+################################################################################
+#
+#   RMG - Reaction Mechanism Generator
+#
+#   Copyright (c) 2002-2017 Prof. William H. Green (whgreen@mit.edu), 
+#   Prof. Richard H. West (r.west@neu.edu) and the RMG Team (rmg_dev@mit.edu)
+#
+#   Permission is hereby granted, free of charge, to any person obtaining a
+#   copy of this software and associated documentation files (the 'Software'),
+#   to deal in the Software without restriction, including without limitation
+#   the rights to use, copy, modify, merge, publish, distribute, sublicense,
+#   and/or sell copies of the Software, and to permit persons to whom the
+#   Software is furnished to do so, subject to the following conditions:
+#
+#   The above copyright notice and this permission notice shall be included in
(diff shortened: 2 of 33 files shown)


goldmanm:
Info for reviewer:

Since this is refactoring the file, I think unittests should have caught almost all errors. [rmg-tests](https://travis-ci.org/ReactionMechanismGenerator/RMG-tests/builds/257540825) shows no change in the model.

Look at the code, specifically `rmgpy.exceptions.py` and let me know if you have any suggestions. I copied the descriptions from the original methods as well much as possible and made minimum additions (when merging InputError class). I think the descriptions for the different classes could definitely be clarified or cleaned up, especially now that they are all in one spot. That could be in this PR, but I think it may be better in another PR since they have different objectives (refactoring and adding descriptions). 

I could see two more future PRs related to exceptions:

1. clarifying the scope of various exception classes
2. modifying RMG to use more specific exceptions when available (essentially replace the 20ish existing `raise Exception` statements with more specific error classes.

rmgpy/data/kinetics/common.py
```diff
@@ -33,27 +33,11 @@
 in this subpackage.
 """
 
-from rmgpy.data.base import DatabaseError, LogicNode
-from rmgpy.reaction import Reaction, ReactionError
```
mjohnson541: Is ReactionError not used?

rmgpy/exceptions.py
```diff
@@ -0,0 +1,250 @@
...
+    """
+    An exception to be raised when an error occurs while working with atom
+    types. Pass a string describing the circumstances that caused the
+    exceptional behavior.
+    """
+    pass
+
+class ChemicallySignificantEigenvaluesError(Exception):
+    """
+    An exception raised when the reservoir state method is unsuccessful for
```
mjohnson541: shouldn't this be the "chemically significant eigenvalue method" rather than the "reservoir state method"?

rmgpy/exceptions.py
```diff
@@ -0,0 +1,250 @@
...
+    pass
+
+class ChemkinError(Exception):
+    """
+    An exception class for exceptional behavior involving Chemkin files. Pass a
+    string describing the circumstances that caused the exceptional behavior.
+    """
+    pass
+
+class CollisionError(Exception):
```
mjohnson541: Should there be some sort of comment here?

rmgpy/exceptions.py
```diff
@@ -0,0 +1,250 @@
...
+class ImplicitBenzeneError(Exception):
+    """
+    An exception class when encountering a group with too many implicit benzene
+    atoms. These groups are hard to create sample molecules and hard for users
+    to interpret. Pass a string describing the limitation.
+    """
+    pass
+
+class InchiException(Exception):
+    pass
```
mjohnson541: Should we have a comment here?

rmgpy/exceptions.py
```diff
@@ -0,0 +1,250 @@
...
+
+class NegativeBarrierException(Exception):
+    """This Exception occurs when the energy barrier for a hindered Rotor is negative.
+    This can occur if the scan or fourier fit is poor. """
+    
+    pass
+
+class NetworkError(Exception): 
+    pass
```
mjohnson541: Should we have a comment here?

mjohnson541: The API says:  
NetworkError | Raised when an error occurs while working with a pressure-dependent reaction network

goldmanm:
Modified the strings for these classes. In it, there seems to be grey area in the implementation of:

* KineticsError and ReactionError in family.py
* NetworkError and PressureDependenceError in pdep/network.py and rmg/pdep.py

I also moved in two straggling classes. I have not rebased and updated, so the comments don't become outdated.

rmgpy/data/kinetics/common.py
goldmanm: I don't believe so, which is why i removed it. 

My IDE said 'ReactionError imported bu unused', so down the drain it went.

goldmanm:
Haven't seen [this travis error](https://travis-ci.org/ReactionMechanismGenerator/RMG-Py/builds/259549577?utm_source=github_status&utm_medium=notification) before.
```
ResolvePackageNotFound: 

  - cantera >=2.2

The command "conda env create -f environment_linux.yml" failed and exited with 1 during .
```

mjohnson541:
See #1114

Issue #1109, opened 2017-07-26
Travis Build Error


mjohnson541:
On branches that pass tests locally, on travis I'm getting:

gcc -pthread -shared -L/home/travis/miniconda/envs/rmg_env/lib -Wl,-rpath=/home/travis/miniconda/envs/rmg_env/lib,--no-as-needed build/build/pyrex/rmgpy/statmech/torsion.o -L/home/travis/miniconda/envs/rmg_env/lib -lpython2.7 -o ./rmgpy/statmech/torsion.so

cythoning rmgpy/statmech/translation.pyx to build/pyrex/rmgpy/statmech/translation.c

Error compiling Cython file:

------------------------------------------------------------

...

        cdef double qt, mass

        cdef numpy.ndarray sumStates

        if self.quantum:

            raise NotImplementedError('Quantum mechanical model not yet implemented for IdealGasTranslation.')

        elif sumStates0 is not None:

            sumStates = schrodinger.convolve(sumStates0, self.getDensityOfStates(Elist))

                                  ^

------------------------------------------------------------

rmgpy/statmech/translation.pyx:174:35: cimported module has no attribute 'convolve'

Error compiling Cython file:

------------------------------------------------------------

...

            Elist = Elist / constants.Na

            dE = Elist[1] - Elist[0]

            qt = ((2 * constants.pi * mass) / (constants.h * constants.h))**1.5 / 101325.

            densStates = qt * Elist**1.5 / (sqrt(constants.pi) * 0.75) * dE

            if densStates0 is not None:

                densStates = schrodinger.convolve(densStates0, densStates)

                                       ^

------------------------------------------------------------

rmgpy/statmech/translation.pyx:200:40: cimported module has no attribute 'convolve'

building 'rmgpy.statmech.translation' extension

gcc -pthread -fno-strict-aliasing -g -O2 -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -fPIC -I. -I/home/travis/miniconda/envs/rmg_env/lib/python2.7/site-packages/numpy/core/include -I/home/travis/miniconda/envs/rmg_env/include/python2.7 -c build/pyrex/rmgpy/statmech/translation.c -o build/build/pyrex/rmgpy/statmech/translation.o

build/pyrex/rmgpy/statmech/translation.c:1:2: error: #error Do not use this file, it is the result of a failed Cython compilation.

 #error Do not use this file, it is the result of a failed Cython compilation.

[...]

PR #1110, opened 2017-07-27
force to use older version of cython

mjohnson541:
The latest version of cython 0.26.0 appears to be the cause of #1109.  While I work out the actual source of the issue I think it best that we force RMG to use 0.25.2 so that the travis builds will work correctly.


mjohnson541:
solved issue before pushing this

goldmanm:
solved by #1111

PR #1112, opened 2017-07-28
Small code cleanup

goldmanm:
This PR has some small code improvements. I don't think that anything is particularly controversial. The solvation test database allows for us to use the testing database on unittests with liquids (so we don't have to load the entire database). 

For the reviewer:

make sure unittests pass and you sort of understand the commits. Let me know if you have any questions.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -118,6 +118,29 @@ class TemplateReaction(Reaction):
                                    self.reverse,
                                    ))
 
+    def __repr__(self):
+        """
+        Return a string representation that can be used to reconstruct the
+        object.
+        """
+        string = 'TemplateReaction('
+        if self.index != -1: string += 'index={0:d}, '.format(self.index)
+        if self.label != '': string += 'label={0!r}, '.format(self.label)
+        if self.reactants is not None: string += 'reactants={0!r}, '.format(self.reactants)
+        if self.products is not None: string += 'products={0!r}, '.format(self.products)
+        if self.specificCollider is not None: string += 'specificCollider={0!r}, '.format(self.specificCollider)
+        if self.kinetics is not None: string += 'kinetics={0!r}, '.format(self.kinetics)
+        if not self.reversible: string += 'reversible={0}, '.format(self.reversible)
+        if self.transitionState is not None: string += 'transitionState={0!r}, '.format(self.transitionState)
+        if self.duplicate: string += 'duplicate={0}, '.format(self.duplicate)
+        if self.degeneracy != 1: string += 'degeneracy={0:.1f}, '.format(self.degeneracy)
+        if self.pairs is not None: string += 'pairs={0}, '.format(self.pairs)
+        if self.family: string += "family='{}', ".format(self.family)
+        if self.template: string += "template={}, ".format(self.template)
+        if self.comment != '': string += 'comment={0!r}, '.format(self.comment)
+        string = string[:-2] + ')'
+        return string
+
     def getSource(self):
         """
         Return the database that was the source of this reaction. For a
--- a/rmgpy/test_data/testing_database/solvation/groups/abraham.py
+++ b/rmgpy/test_data/testing_database/solvation/groups/abraham.py
@@ -1,0 +1,202 @@
+#!/usr/bin/env python
+# encoding: utf-8
+
+name = "Abraham Solute Descriptors"
+shortDesc = u""
+longDesc = u"""
+
+"""
+entry(
+    index = -3,
+    label = "R",
+    group = 
+"""
+1 * R u0
+""",
+    solute = None,
+    shortDesc = u"""""",
+    longDesc = 
+u"""
+
+""",
+)
+
+entry(
+    index = -2,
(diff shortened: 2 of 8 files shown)


goldmanm:
[rmg-tests](https://travis-ci.org/ReactionMechanismGenerator/RMG-tests/builds/258585232) run shows no differences before likely timing out.

rmgpy/test_data/testing_database/solvation/groups/abraham.py
```diff
@@ -0,0 +1,202 @@
+#!/usr/bin/env python
+# encoding: utf-8
+
+name = "Abraham Solute Descriptors"
+shortDesc = u""
+longDesc = u"""
+
+"""
+entry(
+    index = -3,
```
mjohnson541: Why do these have negative indices?

rmgpy/data/kinetics/family.py
```diff
@@ -134,6 +134,29 @@ def __reduce__(self):
...
+        object.
+        """
+        string = 'TemplateReaction('
+        if self.index != -1: string += 'index={0:d}, '.format(self.index)
+        if self.label != '': string += 'label={0!r}, '.format(self.label)
+        if self.reactants is not None: string += 'reactants={0!r}, '.format(self.reactants)
+        if self.products is not None: string += 'products={0!r}, '.format(self.products)
+        if self.specificCollider is not None: string += 'specificCollider={0!r}, '.format(self.specificCollider)
+        if self.kinetics is not None: string += 'kinetics={0!r}, '.format(self.kinetics)
+        if not self.reversible: string += 'reversible={0}, '.format(self.reversible)
```
mjohnson541: We don't want to print out that reversible=True if it is reversible?

rmgpy/data/kinetics/family.py
```diff
@@ -134,6 +134,29 @@ def __reduce__(self):
...
+        if self.transitionState is not None: string += 'transitionState={0!r}, '.format(self.transitionState)
+        if self.duplicate: string += 'duplicate={0}, '.format(self.duplicate)
+        if self.degeneracy != 1: string += 'degeneracy={0:.1f}, '.format(self.degeneracy)
```
mjohnson541: Are we sure we don't want to print out the degeneracy if it is one?

rmgpy/data/kinetics/family.py
```diff
@@ -134,6 +134,29 @@ def __reduce__(self):
                                    self.reverse,
                                    ))
 
+    def __repr__(self):
+        """
+        Return a string representation that can be used to reconstruct the
+        object.
+        """
+        string = 'TemplateReaction('
+        if self.index != -1: string += 'index={0:d}, '.format(self.index)
```
mjohnson541: Why shouldn't we print out that index=-1 if it is -1?

rmgpy/test_data/testing_database/solvation/groups/abraham.py
goldmanm: The solvation database was pulled from RMG-database. [It](https://github.com/ReactionMechanismGenerator/RMG-database/commit/2479933de81dc6d14209368043bd6394964c2e5a#diff-8d02a65677f4c2ecf8810f77ebfde7cb) looks like it's been here for years. The reasoning seemed to be to stick things at the top and add a previous number.

rmgpy/data/kinetics/family.py
goldmanm: It's the default value, so it should still create the same object. It also keeps things cleaner.

rmgpy/data/kinetics/family.py
goldmanm: It's the default value, so it should still create the same object. It's possible to print out, but not necessary.

rmgpy/data/kinetics/family.py
goldmanm: True is the default value.

goldmanm:
@mjohnson541. Some background may make sense. Most of the `__repr__` method for TemplateReaction came from `reaction.py`, and I wanted to keep it consistent.  

I think it would be better to not need to write an entirely new method with redundant code. Maybe we can call the parent method, remove the closing parenthesis, add our new variables, and slap on the parenthesis at the end, and `Template` at the beginning. This seems to be doable. I just don't want to have messy and finicky regex expressions everywhere.

rmgpy/data/kinetics/family.py
mjohnson541: My concern with not printing fields that are default values is that when this is used, it isn't obvious whether the object has the default value of a field or the field just wasn't printed and even if they know that the person using it may not know what the default value is suppose to be.

goldmanm: See the post on this PR about where I got the code you're actually commenting about. I wasn't trying to change the status quo in RMG, which is currently to not print default values. The point of this commit was to allow printing of non-default values of TemplateReaction family. Changing status quo of RMG was not the goal.

mjohnson541: Sorry, I was trying to discuss if we should change that, not whether or not you were consistent.

PR #1113, opened 2017-07-28
Only core degeneracy is accurate

goldmanm:
This PR tries to fix part of the issue #1079 by allowing degeneracy in the generate reactions step to be done in a similar way to before PR #1055, except that multiple transition states are kept. When a reaction is added to the core, then the degeneracy is found with `family.calculateDegeneracy`. 

This will help fix a slowdown in RMG while minimizing the impact on having uncertain values.

Thanks to @whgreen, @amarkpayne, @rgillis8, @mjohnson541 for comments and suggestions at the meeting this week, which all led to improvements of imlementation.

For reviewer:

* check to see that nothing (or very little) has changed in the RMG-Tests page
* check to see if any time improvement has occurred in the RMG-Tests page
* look through the commits to see if the code makes sense
* (optional) profile performance of longer runs.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1487,6 +1487,11 @@ class KineticsFamily(Database):
         reaction.degeneracy = 1
         from rmgpy.rmg.react import findDegeneracies, reduceSameReactantDegeneracy, getMoleculeTuples
 
+        # if reactants are different instances, make deep copy of reaction object
+        if len(reaction.reactants) > 1 and \
+                      len(set([id(r) for r in reaction.reactants])) != \
+                      len(reaction.reactants):
+            reaction = reaction.copy()
         # find combinations of resonance isomers
         specReactants = []
         if isinstance(reaction.reactants[0], Molecule):
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1528,10 +1528,8 @@ class KineticsFamily(Database):
             
         # log issues
         if len(reactions) != 1:
-            for reactant in reaction.reactants:
-                logging.error("Reactant: {0!r}".format(reactant))
-            for product in reaction.products:
-                logging.error("Product: {0!r}".format(product))
+            logging.error("The full original reaction object is: {0!r}".format(reaction))
+            logging.error("The reaction objects found are: {0!r}".format(reactions))
             raise KineticsError(('Unable to calculate degeneracy for reaction {0} '
                                  'in reaction family {1}. Expected 1 reaction '
                                  'but generated {2}').format(reaction, self.label, len(reactions)))
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1528,11 +1528,14 @@ class KineticsFamily(Database):
             
         # log issues
         if len(reactions) != 1:
-            logging.error("The full original reaction object is: {0!r}".format(reaction))
-            logging.error("The reaction objects found are: {0!r}".format(reactions))
-            raise KineticsError(('Unable to calculate degeneracy for reaction {0} '
+            logging.warning(('Unable to calculate degeneracy for reaction {0} '
                                  'in reaction family {1}. Expected 1 reaction '
                                  'but generated {2}').format(reaction, self.label, len(reactions)))
+            logging.warning("The full original reaction object is: {0!r}".format(reaction))
+            logging.warning("The reaction objects found are: {0!r}".format(reactions))
+            if len(reactions) == 0:
+                return 1.0
+
         return reactions[0].degeneracy
         
     def __generateReactions(self, reactants, products=None, forward=True):
--- a/rmgpy/data/kinetics/kineticsTest.py
+++ b/rmgpy/data/kinetics/kineticsTest.py
@@ -370,9 +370,9 @@ class TestReactionDegeneracy(unittest.TestCase):
         spcA = Species().fromSMILES('[H]')
         spcB = Species().fromSMILES('CC=C[CH]C')
         spcB.generateResonanceIsomers(keepIsomorphic=True)
-        spcTuples = [(spcA,spcB)]
+        spcTuple = (spcA,spcB)
         
-        reactionList = list(react(*spcTuples))
+        reactionList = reactSpecies(spcTuple, accurateDegeneracies = True)
         
         # find reaction with a specific product
         specific_product = Species().fromSMILES('[CH2]C=C[CH]C')
@@ -397,8 +397,9 @@ class TestReactionDegeneracy(unittest.TestCase):
         """
         spcA = Species().fromSMILES('C[C]=C')
         spcB = Species().fromSMILES('C=C[CH2]')
-        spcTuples = [(spcA,spcB)]
-        reactionList = list(react(*spcTuples))
+        spcTuple = (spcA,spcB)
+
+        reactionList = reactSpecies(spcTuple, accurateDegeneracies = True)
         # find reaction with a specific product
         specific_products = [Species().fromSMILES('C=C=C'),
                              Species().fromSMILES('CC=C'),]
@@ -456,9 +457,9 @@ class TestReactionDegeneracy(unittest.TestCase):
         spcA = Species().fromSMILES('[H]')
         spcB = Species().fromSMILES('CC=C[CH]C')
         spcB.generateResonanceIsomers(keepIsomorphic=True)
-        spcTuples = [(spcA,spcB)]
+        spcTuple = (spcA,spcB)
         
-        reactionList = list(react(*spcTuples))
+        reactionList = reactSpecies(spcTuple, accurateDegeneracies = True)
         
         # find reaction with a specific product
         specific_product = Species().fromSMILES('CC=C[CH][CH2]')
--- a/rmgpy/rmg/model.py
+++ b/rmgpy/rmg/model.py
@@ -46,6 +46,7 @@ from rmgpy.quantity import Quantity
 import rmgpy.species
 from rmgpy.thermo.thermoengine import submit
 from rmgpy.reaction import Reaction
+from rmgpy.data.rmg import getDB
 from rmgpy.exceptions import ForbiddenStructureException
 from rmgpy.data.kinetics.depository import DepositoryReaction
 from rmgpy.data.kinetics.family import KineticsFamily, TemplateReaction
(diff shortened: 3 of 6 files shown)


goldmanm:
The [rmg tests results](https://travis-ci.org/ReactionMechanismGenerator/RMG-tests/builds/258667867) were mostly as expected. It ran eg1, eg3 and eg5. 

* eg1 had two more species added. all the core reaction kinetics were the same (yay!) and many of the edge reactions were different (as expected)
* eg3 had the same core model with different edge kinetics (again as expected)
* eg5, which is pdep, had an extra species, and one of the core, pdep reactions had a difference in kinetics, which I don't think should be the case.

I'm going to look at how rmg modifies the kinetics of a pdep reaction to make sure it stays consistent.

mjohnson541:
It might be worth running a longer model to see if this causes long term changes.  I think the primary fear is that RMG will hit a branching point in model growth and follow a slower reaction (due to not having the degeneracy right) and continue down that path once the build up is lost.  The RMG-tests examples may be too short to properly test the extent these changes affect the results.  I think running something like Hexane oxidation (that has these sort of branching points) would give us a better idea of how much of an effect this has on model construction.  Which, also should help tell us how much we need to worry about finding a way to treat edge degeneracies better.

goldmanm:
The longer model is currently being run by @amarkpayne. The RMG-tests runs are definitely too short to test longer model issues.

For that we'd need to have separate small timed tests instead of an entire RMG run:

* run generateReactions on a large mechanism
* run the solver on a huge mechanism
* etc.

mliu49:
I think there is another speed up opportunity at the cost of a little memory: `react.correctDegeneracyOfReverseReactions` runs species isomorphism for every reaction generated. The could be completely avoided by saving the `forward` attribute to the reaction during `KineticsFamily.__generateReactions` and checking it here.

Also, another reason for the slowdown might be that the new algorithm finds degeneracy for the full reaction list generated from a set of reactants for all reaction families, whereas it previously only did it for one family at a time. This means that isomorphism is being done with a much larger list of reactions now. I'm not sure if this is necessary, or if we can add a step to pre-filter by family.

goldmanm:
The reason to use find degeneracy on all families at once is to identify reactions that come from different families as DUPLICATE, which can cause an error later on. This may not be worth the increase in computational time or there might be another way around it.

mliu49:
I found another change from the old algorithm which contributes to computation time. Previously, only products were compared for degeneracy determination, and therefore resonance structures were only generated for the products. In the new algorithm, both reactants and products are converted to species with resonance structures, then the new `reaction.isIsomorphic` method is called, which compares both reactants and products, and goes a step further to check if the reactions are isomorphic in the other direction if they aren't in the inputted direction.

goldmanm:
I think that change was created since the knowledge of whether the particular reaction was generated in the forward or reverse template. I think the adding of the `forward` to the reaction class would also be beneficial in this case, which could reduce comparisons by between 1-2x. 

These comments might be more relevant for #1079, since they relate to possible causes for the slowdown.

goldmanm:
@amarkpayne, you ran some profiling tests for this addition. I remember it significantly speeding up model growth. What are your thoughts on the effect and merge-ability of this PR.

mliu49:
I guess my point was that it might be better to first consider speed up methods that do not reduce accuracy before merging this pull request.

mliu49:
I wanted to do some profiling, but eg1 fails on this branch with the following error:
```
  File "rmg.py", line 153, in <module>
    cProfile.runctx(command, global_vars, local_vars, stats_file)
  File "/home/mjliu/anaconda/envs/rmg_env/lib/python2.7/cProfile.py", line 49, in runctx
    prof = prof.runctx(statement, globals, locals)
  File "/home/mjliu/anaconda/envs/rmg_env/lib/python2.7/cProfile.py", line 140, in runctx
    exec cmd in globals, locals
  File "/home/mjliu/Code/RMG-Py/rmgpy/rmg/main.py", line 728, in execute
    bimolecularReact=self.bimolecularReact)
  File "/home/mjliu/Code/RMG-Py/rmgpy/rmg/model.py", line 650, in enlarge
    self.processNewReactions([rxn], spc)
  File "/home/mjliu/Code/RMG-Py/rmgpy/rmg/model.py", line 717, in processNewReactions
    rxn, isNew = self.makeNewReaction(rxn)
  File "/home/mjliu/Code/RMG-Py/rmgpy/rmg/model.py", line 467, in makeNewReaction
    reactants = [self.makeNewSpecies(reactant)[0] for reactant in forward.reactants]
  File "/home/mjliu/Code/RMG-Py/rmgpy/rmg/model.py", line 333, in makeNewSpecies
    submit(spec)
  File "/home/mjliu/Code/RMG-Py/rmgpy/thermo/thermoengine.py", line 179, in submit
    spc.thermo = submit_(evaluator, spc)
  File "/home/mjliu/Code/RMG-Py/rmgpy/scoop_framework/util.py", line 171, in submit_
    return func(*args, **kwargs)
  File "/home/mjliu/Code/RMG-Py/rmgpy/thermo/thermoengine.py", line 165, in evaluator
    thermo = generateThermoData(spc)
  File "/home/mjliu/Code/RMG-Py/rmgpy/thermo/thermoengine.py", line 129, in generateThermoData
    thermo0 = thermodb.getThermoData(spc)
  File "/home/mjliu/Code/RMG-Py/rmgpy/data/thermo.py", line 1140, in getThermoData
    thermo0.S298.value_si -= constants.R * math.log(species.getSymmetryNumber())
  File "rmgpy/species.py", line 401, in rmgpy.species.Species.getSymmetryNumber (build/pyrex/rmgpy/species.c:11584)
    resonanceHybrid = self.getResonanceHybrid()
  File "rmgpy/species.py", line 467, in rmgpy.species.Species.getResonanceHybrid (build/pyrex/rmgpy/species.c:12982)
    newBond.applyAction(('CHANGE_BOND',None,oldBondOrder / numResonanceStructures / 2))
  File "rmgpy/molecule/molecule.py", line 650, in rmgpy.molecule.molecule.Bond.applyAction (build/pyrex/rmgpy/molecule/molecule.c:15086)
    self.__changeBond(action[2])
  File "rmgpy/molecule/molecule.py", line 636, in rmgpy.molecule.molecule.Bond.__changeBond (build/pyrex/rmgpy/molecule/molecule.c:14863)
    raise gr.ActionError('Unable to update Bond due to CHANGE_BOND action: Invalid resulting order "{0}".'.format(self.order))
ActionError: Unable to update Bond due to CHANGE_BOND action: Invalid resulting order "3.00000047684".
```
This occurred after adding core species 37, [CH2]C1[CH]C(=C)C=1(3059). Note that the RMG-Py eg1 is different from the RMG-tests eg1 (lower tolerance).

amarkpayne:
I finally tracked down the error in this branch that has been preventing my profiling runs from completing.

Early in my surrogate oil model run, I was getting an error that `calculateDegeneracy` failed for a reaction because it had generated 0 reactions but expected 1. The reaction involved two identical molecules of RAD1 (PDD with a benzylic radical) combining together (R_Recombination). The fact that the molecules are aromatic is key to this error as I found out.

After a lot of playing around with the code, I was able to determine what went wrong. When this reaction was generated in the edge, the aromatic rings contained benzene type bonds, which gave the reaction a template of `['C_rad/H/OneDeC', 'C_rad/H/CdCs']`. When the reaction was generated again in `calculateDegeneracy`, the molecules had double bonds in the aromatic rings, so all reactions generated had the template `['C_rad/H/CdCs', 'C_rad/H/CdCs']`. Because none of the generated reactions had the same template as the original reaction object, all of these reactions were removed, leaving zero reactions. You can see this is the case by running the attached ipynb file `CalculateDegeneracyErrorInvestigation`.

In the long-term, we need to figure out how to overcome errors between the two representations of benzene that RMG uses, but I know that this has been a struggle for a long time.

In the short-term, perhaps we can add a check to `calculateDegeneracy` so that if none of the reactions generated have the same template as the original reaction object that the user is simply warned and the degeneracy of the generated reactions is used.

Here are the attached files:

[Investigation_Files.zip](https://github.com/ReactionMechanismGenerator/RMG-Py/files/1236882/Investigation_Files.zip)


Here is the error traceback:
```
discarding /opt/anaconda/anaconda/bin from PATH
prepending /home/ampayne/.conda/envs/rmg_env_AMP/bin to PATH
DEPRECATION: The default format will switch to columns in the future. You can use --format=(legacy|columns) (or define a format=(legacy|columns) in your pip.conf under the [list] section) to disable this warning.
DEPRECATION: The default format will switch to columns in the future. You can use --format=(legacy|columns) (or define a format=(legacy|columns) in your pip.conf under the [list] section) to disable this warning.
  File "/home/ampayne/RMG/RMG-Py/rmg.py", line 153, in <module>
    cProfile.runctx(command, global_vars, local_vars, stats_file)
  File "/home/ampayne/.conda/envs/rmg_env_AMP/lib/python2.7/cProfile.py", line 49, in runctx
    prof = prof.runctx(statement, globals, locals)
  File "/home/ampayne/.conda/envs/rmg_env_AMP/lib/python2.7/cProfile.py", line 140, in runctx
    exec cmd in globals, locals
  File "/home/ampayne/RMG/RMG-Py/rmgpy/rmg/main.py", line 728, in execute
    bimolecularReact=self.bimolecularReact)
  File "/home/ampayne/RMG/RMG-Py/rmgpy/rmg/model.py", line 650, in enlarge
    self.processNewReactions([rxn], spc)
  File "/home/ampayne/RMG/RMG-Py/rmgpy/rmg/model.py", line 762, in processNewReactions
    self.addReactionToCore(rxn)
  File "/home/ampayne/RMG/RMG-Py/rmgpy/rmg/model.py", line 1177, in addReactionToCore
    rxn.degeneracy = family.calculateDegeneracy(rxn)
  File "/home/ampayne/RMG/RMG-Py/rmgpy/data/kinetics/family.py", line 1534, in calculateDegeneracy
    'but generated {2}').format(reaction, self.label, len(reactions)))
rmgpy.exceptions.KineticsError: Unable to calculate degeneracy for reaction RAD1(14) + RAD1(14) <=> L177(107) in reaction family R_Recombination. Expected 1 reaction but generated 0
discarding /home/ampayne/.conda/envs/rmg_env_AMP/bin from PATH
```

mjohnson541:
I saw a similar bond change error on master in #1157.  However it wasn't in eg1.

mliu49:
Thanks for investigating, @amarkpayne! I will try to look into a solution as soon as I can.

goldmanm:
@amarkpayne Thanks for the investigation. I added a commit to this PR to fix the short-term problem.

goldmanm:
This PR seems to obtain a similar aim as #1209, so it no longer needs to be open.

Issue #1115, opened 2017-08-01
eg5 fails when making seed mechanism


goldmanm:
I was trying to `make eg5` on the most recent master, and got the following error. Which happened on my local machine and pharos. The [travis test](https://travis-ci.org/ReactionMechanismGenerator/RMG-tests/builds/258905114) for the seed mechanisms PR did not error. Is anyone able to replicate this error? 

```
Making seed mechanism...
trying to find duplicates
NB. the entries have not been renumbered, so these indices are missing.
Traceback (most recent call last):
  File "rmg.py", line 162, in <module>
    rmg.execute(**kwargs)
  File "/home/goldmanm/repos/RMG-Py/rmgpy/rmg/main.py", line 583, in execute
    self.makeSeedMech(firstTime=True)
  File "/home/goldmanm/repos/RMG-Py/rmgpy/rmg/main.py", line 892, in makeSeedMech
    thermoLibrary.save(os.path.join(seedDir, name + '.py'))
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

mliu49:
Apparently, eg5 on RMG-tests is different from eg5 in RMG-Py. Notably, eg5 in RMG-Py is missing the options block. If you add the options block in with the same settings as RMG-tests, then the job runs without erroring. This suggests some problems with default values for the options block.

goldmanm:
Seems like previous issues with arguments causing errors #1067 may be related.

mjohnson541:
closed with #1117

PR #1117, opened 2017-08-02
Fix seed generation

### LINKED ISSUE #1116: seed mechanism folder not generated in outputdir
The seed mechanism seems to be generated in the current working directory, which is not based on the `outputDirectory` of the RMG job. 

This can be seen by going into the minimal example folder, creating a subdirectory `output` and running `python $rmg/rmg.py input.py -o output`. 

The folder `seed` is not in the folder `output`.

mjohnson541:
This should fix #1116 and #1115.  The seed is now produced in the output directory rather than the working directory and the seed option variables are now properly initialized by RMG.clear().  

An alternative fix to #1115 would be to run options() right before input generation, which would make it so that options variables' default values would only be set at one location (the values in RMG.clear() would be irrelevant).  I don't think we currently use RMG.clear() for anything other than initialization, however if we were to do so, being dependent on the options() call to fill in the options variables would be undesirable.

### DIFF
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -814,8 +814,6 @@ class RMG(util.Subject):
         
         name = self.name
         
-        currentDir = os.getcwd()
-        
         if self.saveSeedToDatabase and firstTime: #make sure don't overwrite current libraries
             thermoNames = self.database.thermo.libraries.keys()
             kineticsNames = self.database.kinetics.libraries.keys()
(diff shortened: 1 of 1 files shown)


mjohnson541:
RMG.clear initializes a lot of stuff, so I don't think that in itself would be sufficient, but I think we could do so if we adapted rmg.py and input.py sufficiently.

goldmanm:
I guess I don't understand how RMG.clear is different in function from RMG.__init__ if it is only used for initialization.

mjohnson541:
As we use it, it might as well just be tacked on to the end of RMG.init.

mliu49:
`RMG.clear` is already tacked on to the end of `RMG.__init__`.

rmgpy/rmg/main.py
```diff
@@ -826,11 +824,13 @@ def makeSeedMech(self,firstTime=False):
                     q += 1
                 self.name = name + str(q)
         
-        if firstTime and not os.path.exists(os.path.join(currentDir,'seed')): #if seed directory does not exist make it
-            os.system('mkdir seed')
+        seedDir = os.path.join(self.outputDirectory,'seed')
+        
+        if firstTime and not os.path.exists(seedDir): #if seed directory does not exist make it
+            os.system('mkdir '+seedDir)
```
mliu49: You merged it before I could comment, but why did you use `os.system` instead of `os.mkdir` and `os.rmdir`?

mjohnson541: Simply because I forgot those existed.  Do you think it's worth making a new pull request?

mjohnson541: Does os.rmdir work on non-empty directories?

mliu49: Good point, Python documentation recommends `shutil.rmtree()` for non-empty directories. The bigger issue is that I don't think `rm -rf` works on Windows.

mjohnson541: Oh...yeah...good point!

PR #1123, opened 2017-08-03
fix windows seed generation

mjohnson541:
Switch out unix commands for python os/shutil commands.  Ideally would like someone with a windows install to test this.

### DIFF
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -827,10 +827,10 @@ class RMG(util.Subject):
         seedDir = os.path.join(self.outputDirectory,'seed')
         
         if firstTime and not os.path.exists(seedDir): #if seed directory does not exist make it
-            os.system('mkdir '+seedDir)
+            os.mkdir(seedDir)
         else:
-            os.system('rm -rf '+ seedDir) #otherwise delete the old seed and make a new directory
-            os.system('mkdir '+ seedDir)
+            shutil.rmtree(seedDir) #otherwise delete the old seed and make a new directory
+            os.mkdir(seedDir)
             
         speciesList = self.reactionModel.core.species
         reactionList = self.reactionModel.core.reactions


alongd:
I can confirm that on a Windows machine, RMG generated a seed mechanism both when using the current master, and using this fix.
What was the original issue?

mjohnson541:
Originally it used unix commands so we would expect it to crash on windows.

alongd:
Neither crushed for me. Could someone else double check generating the seed mechanism by running something on Windows using master and this branch?

rwest:
Inspecting the code and guessing: if the directory didn't already exist then it would only call `'mkdir' + seedDir`, a command also exists in MSDOS and Windows. However, if the directory exists it would try `'rm -rf '+ seedDir`, which I would expect to crash if running from MSDOS, but probably work if you're using a unix-like shell (like Git Bash, or CygWin, etc.) or bash on Windows 10.

In any case, the new version, using `os` and `shutil` libraries looks preferable to me.

alongd:
This is very strange.
I created an empty `seed` library in the run folder before running.
Running on master, RMG was able to generate the seed correctly. Didn't crush.

Using this fix, however, I get:
```
Making seed mechanism...
  File "../../RMG-Py/rmg.py", line 162, in <module>
    rmg.execute(**kwargs)
  File "C:\Code\RMG-Py\rmgpy\rmg\main.py", line 583, in execute
  File "C:\Code\RMG-Py\rmgpy\rmg\main.py", line 833, in makeSeedMech
    os.mkdir(seedDir)
WindowsError: [Error 5] Access is denied: 'C:\\Code\\runs\\superminimal\\seed'
```

This might be related to another pain on Windows, where if the user doesn't clean the run folder (at least RMG's sub directories such as `pdep` and `chemkin`) RMG crushes with a similar error message.

mliu49:
Interestingly, both master and this branch worked without issues for me. I tested in a Windows 7 virtual machine using both a bash shell and windows cmd.

I think this PR can be merged. @alongd's error seems to be a permissions problem.

PR #1126, opened 2017-08-03
Use thermo library names

mjohnson541:
This causes RMG to use the names associated with thermo library entries rather than smiles when a thermo library entry is available (and if not already assigned by a seed mechanism or reaction library).

### DIFF
--- a/rmgpy/rmg/model.py
+++ b/rmgpy/rmg/model.py
@@ -307,14 +307,6 @@ class CoreEdgeReactionModel:
         # Check that the structure is not forbidden
 
         # If we're here then we're ready to make the new species
-        if label == '': 
-            # Use SMILES as default format for label
-            # However, SMILES can contain slashes (to describe the
-            # stereochemistry around double bonds); since RMG doesn't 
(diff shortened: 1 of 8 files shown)


rmgpy/rmg/model.py
```diff
@@ -331,7 +323,20 @@ def makeNewSpecies(self, object, label='', reactive=True, checkForExisting=True)
         spec.molecularWeight = Quantity(spec.molecule[0].getMolecularWeight()*1000.,"amu")
         
         submit(spec)
-
+        
+        if spec.label == '' and spec.thermo and spec.thermo.label != '':
+            logging.info('Species with SMILES of {0} named {1} based on thermo library name'.format(molecule.toSMILES().replace('/','').replace('\\',''),spec.thermo.label))
+            spec.label = spec.thermo.label
+        
+        if label == '':
```
alongd: If we assign a value to `spec.label`, isn't it still possible for `label` to be empty (`''`)?

mjohnson541: This has been reorganized to be clearer

PR #1128, opened 2017-08-03
Ring Perception

mliu49:
Improved ring handling has been on the to do list for a while, but before adding new features, it seemed important to make sure our ring perception methods are sufficiently robust. This PR introduces a new algorithm for getSmallestSetOfSmallestRings and a new method called getRelevantCycles.

The previous smallest set of smallest rings (SSSR) algorithm was an incomplete implementation of Fan et al.'s algorithm, and failed for molecules with no degree 2 nodes (eg. cubane). Additionally, the SSSR is not always the most useful set of rings. The set of relevant cycles (RC) is often cited as a more "chemically useful" set of rings, and is defined as the union of all possible SSSRs.

Using cubane as an example, intuitively, one would likely say that there are 6 rings. The previous RMG SSSR algorithm would incorrectly return 4 rings. The actual SSSR includes 5 rings, and there are multiple valid choices for which 5 rings are included. The set of RC includes all 6 rings, and is unique.

The new implementation uses the open source [RingDecomposerLib](https://github.com/rareylab/RingDecomposerLib) developed by a group at Hamburg University and [published](http://pubs.acs.org/doi/10.1021/acs.jcim.6b00736?src=recsys&) earlier this year. It implements an extension of a method described by Vismara for enumerating the set of relevant cycles. The package is a C library with a Python wrapper.

This new implementation is about an __order of magnitude faster__ than the previous SSSR algorithm, and we now have access to both the SSSR and RC (as well as the [unique ring families](http://pubs.acs.org/doi/abs/10.1021/ci200629w), if we ever need it).

Notes:
- ~~Conda packages still need to be made for RingDecomposerLib before merging this PR.~~ Done!
- Are there other places where SSSR should be replaced by RC (eg. bicyclic decomposition)?
- Does this eliminate the need for `getDeterministicSmallestSetOfSmallestRings`?

### DIFF
--- a/rmgpy/molecule/graph.pyx
+++ b/rmgpy/molecule/graph.pyx
@@ -991,6 +991,14 @@ cdef class Graph:
                     ordered.append(vertex)
                     vertices.remove(vertex)
                     break
+            else:
+                # No connected vertex was found
+                raise RuntimeError('Could not sort cyclic vertices because '
+                                   'not all vertices are connected to two '
+                                   'other vertices in the input list.')
+
+        if not self.hasEdge(ordered[0], ordered[-1]):
+            raise RuntimeError('Input vertices do not comprise a single cycle.')
 
         return ordered
 
--- a/rmgpy/molecule/graph.pyx
+++ b/rmgpy/molecule/graph.pyx
@@ -236,6 +236,21 @@ cdef class Graph:
         edge.vertex2.edges[edge.vertex1] = edge
         return edge
 
+    cpdef list getAllEdges(self):
+        """
+        Returns a list of all edges in the graph.
+        """
+        cdef set edgeSet
+        cdef Vertex vertex
+        cdef Edge edge
+
+        edgeSet = set()
+        for vertex in self.vertices:
+            for edge in vertex.edges.itervalues():
+                edgeSet.add(edge)
+
+        return list(edgeSet)
+
     cpdef dict getEdges(self, Vertex vertex):
         """
         Return a dictionary of the edges involving the specified `vertex`.
--- a/rmgpy/molecule/graph.pyx
+++ b/rmgpy/molecule/graph.pyx
@@ -34,6 +34,7 @@ are the components of a graph.
 """
 
 import logging
+import py_rdl
 from .vf2 cimport VF2
 
 ################################################################################
@@ -197,6 +198,12 @@ cdef class Edge(object):
 ################################################################################
 
 cdef VF2 vf2 = VF2()
+
+cdef  Vertex _getEdgeVertex1(Edge edge):
+    return edge.vertex1
+
+cdef Vertex _getEdgeVertex2(Edge edge):
+    return edge.vertex2
 
 cdef class Graph:
     """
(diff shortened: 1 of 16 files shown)


rwest:
Sounds very promising. 
Please also maintain the pip requirements.txt file and instructions for people not using anaconda. (I keep coming across people whose clusters cannot run anaconda)

KEHANG:
Very interesting! `getDeterministicSmallestSetOfSmallestRings` was designed for molecules like [this](http://rmg.mit.edu/database/molecule/1%20%20C%20u0%20p0%20c0%20%7B2,S%7D%20%7B6,S%7D%20%7B9,S%7D%20%7B10,S%7D%0A2%20%20O%20u0%20p2%20c0%20%7B1,S%7D%20%7B3,S%7D%0A3%20%20C%20u0%20p0%20c0%20%7B2,S%7D%20%7B4,S%7D%20%7B8,S%7D%20%7B11,S%7D%0A4%20%20C%20u0%20p0%20c0%20%7B3,S%7D%20%7B5,S%7D%20%7B12,S%7D%20%7B13,S%7D%0A5%20%20C%20u0%20p0%20c0%20%7B4,S%7D%20%7B6,S%7D%20%7B14,S%7D%20%7B15,S%7D%0A6%20%20C%20u0%20p0%20c0%20%7B1,S%7D%20%7B5,S%7D%20%7B7,S%7D%20%7B16,S%7D%0A7%20%20C%20u0%20p0%20c0%20%7B6,S%7D%20%7B8,S%7D%20%7B17,S%7D%20%7B18,S%7D%0A8%20%20C%20u0%20p0%20c0%20%7B3,S%7D%20%7B7,S%7D%20%7B19,S%7D%20%7B20,S%7D%0A9%20%20H%20u0%20p0%20c0%20%7B1,S%7D%0A10%20H%20u0%20p0%20c0%20%7B1,S%7D%0A11%20H%20u0%20p0%20c0%20%7B3,S%7D%0A12%20H%20u0%20p0%20c0%20%7B4,S%7D%0A13%20H%20u0%20p0%20c0%20%7B4,S%7D%0A14%20H%20u0%20p0%20c0%20%7B5,S%7D%0A15%20H%20u0%20p0%20c0%20%7B5,S%7D%0A16%20H%20u0%20p0%20c0%20%7B6,S%7D%0A17%20H%20u0%20p0%20c0%20%7B7,S%7D%0A18%20H%20u0%20p0%20c0%20%7B7,S%7D%0A19%20H%20u0%20p0%20c0%20%7B8,S%7D%0A20%20H%20u0%20p0%20c0%20%7B8,S%7D%0A). The old`getSmallestSetOfSmallestRings` gives two rings randomly (sometimes both rings have oxygen, sometimes only one has), not sure about the new one you suggested? (I guess it will still gives two rings, but not sure about if it'll be more deterministic or not)

mliu49:
@rwest Unfortunately, the package is not on pypi. Would it be sufficient to add instructions to the documentation?

@KEHANG For the first case (C1CC2CCC1CO2), the new SSSR algorithm would still return 2 rings non-deterministically, while the RC would return 3 rings. For the second case (C1C2CCC1CC2), the RC is the same as the SSSR -- two 5-membered rings. For the third case (C12CCC1CC2), the RC is also the same as the SSSR -- two 4-membered rings.

Since the RC is the union of possible SSSR, they will differ only when there are multiple choices for the SSSR. In those cases, the RC will be a unique set of all the possible smallest rings.

rmgpy/molecule/graph.pyx
```diff
@@ -972,10 +972,29 @@ cdef class Graph:
...
+        cdef Vertex vertex
+
+        ordered = [vertices.pop()]
+        while vertices:
+            for vertex in vertices:
+                if vertex in ordered[-1].edges:
+                    ordered.append(vertex)
+                    vertices.remove(vertex)
+                    break
+
```
KEHANG: Currently the method accept any list of vertices, if one vertex that is out of a cyclic is accidentally put in, it seems easily to end up in an infinite loop

KEHANG:
Could you please rebase and run RMG-tests, especially thermo validation since polycyclic thermo estimation calls `getSmallestSetOfSmallestRing` in multiple places?

rmgpy/molecule/graph.pyx
mliu49: Thanks, I'll do that.

mliu49:
RMG-tests results:

---|---|---|---|---
eg1|00:00:02:16|00:00:02:13|456.35|456.43
eg3|00:00:04:56|00:00:04:46|399.13|397.93
eg5|00:00:01:21|00:00:01:18|357.09|356.76
eg6|00:00:00:52|00:00:00:50|341.95|341.20
eg7|00:00:00:34|00:00:00:34|339.26|337.33
NC|00:00:01:47|00:00:01:44|342.72|342.87
solvent_hexane|00:00:00:50|00:00:00:49|327.45|329.42
MCH|00:00:03:32|00:00:03:35|691.14|692.52

Thermo validation results:

dataset | benchmark mae (kcal/mol) | new mae (kcal/mol)
--- | --- | ---
large_fused_polycyclic_table | 9.57 | 9.59
large_linear_polycyclic_table | 9.08 | 9.22
small_cyclic_table | 3.42 | 3.42
large_linear_O_only_polycyclic_table | 6.94 | 7.06
large_fused_O_only_polycyclic_table | 10.31 | 10.34
small_O_only_polycyclic_table | 5.30 | 5.30
zb_bicyclics_195_table | 9.59 | 9.59

@KEHANG, any thoughts on why the thermo performance is different? The only difference in SSSR output should be for molecules without degree 2 nodes.

KEHANG:
@mliu49 we probably want to check thermo comments for the molecules estimated differently. RMG-tests can save those comments for all the tested molecules, I'll provide those files for you tomorrow.

mliu49:
I looked into the discrepancies in the large_linear_polycyclic_table, which had the largest difference between the two versions. The results actually raise more questions than they answer.

First, comparing Case 1 and Case 4, the molecules are essentially identical, with the exception of an additional methyl group in Case 4. However, the thermo predictions and groups are completely different. The groups also don't seem to match with the SSSR results in Case 4.

In the benchmark for Case 1, the SSSR includes rings of size 3, 3, 4, and 4, which seems to match with the polycyclic groups that are matched. In the testing case, the SSSR includes an additional ring of size 5, which leads to matching different groups. I would guess that the incorrect SSSR previously might also be why this molecule is in the linear table instead of the fused table.

For Case 4, the comments (and possibly H298 predictions?) seem to be switched between the benchmark and testing versions. The benchmark version has groups with 5 membered rings, while the SSSR does not have any, and the opposite is true for the testing version. Is there any chance that this resulted from a bug while compiling the results?

Then, looking at Case 2 and Case 3, it seems that the testing case fails to subtract one of the overlapped groups. That will need to be looked at in more detail to figure out why.

Molecule | Actual | Benchmark | Testing
----- | ----- | ----- | -----
C12C3C4C3C3C1C2C43 | 112.40476053 | 102.724091778 | 87.0627342256
C1C2C1C1CC2C2CC12 | 63.30899487 | 57.5718307839 | 63.8718307839
C1C2CC3(CC4CC34)C12 | 86.23946529 | 88.7539483748 | 114.953948375
CC12C3C4C3C3C1C2C43 | 104.59964312 | 79.2627342256 | 151.127007648


**Case 1: C12C3C4C3C3C1C2C43**


Benchmark: [3, 3, 4, 4]
polycyclic(s2_3_4_ane) + polycyclic(s2_4_4_ane) + polycyclic(s2_3_4_ane) - ring(Cyclobutane) - ring(Cyclobutane)

Testing: [3, 3, 4, 4, 5]
polycyclic(s2_3_5_ane) + polycyclic(s2_4_5_ane) + polycyclic(s2_3_5_ane) + polycyclic(s2_3_4_ane) - ring(Cyclobutane) - ring(Cyclopentane) - ring(Cyclopentane) - ring(Cyclopropane)

**Case 2: C1C2C1C1CC2C2CC12**


Benchmark: [3, 5, 5, 3]
polycyclic(s2_3_5_ane) + polycyclic(s3_5_5_ane) + polycyclic(s2_3_5_ane) - ring(Cyclopentane) - ring(Cyclopentane)

Testing: [3, 3, 5, 5]
polycyclic(s3_5_5_ane) + polycyclic(s2_3_5_ane) - ring(Cyclopentane) + polycyclic(s2_3_5_ane)

**Case 3: C1C2CC3(CC4CC34)C12**


Benchmark: [3, 4, 4, 3]
polycyclic(s2_3_4_ane) + polycyclic(s1_4_4_ane) + polycyclic(s2_3_4_ane) - ring(Cyclobutane) - ring(Cyclobutane)

Testing: [3, 4, 3, 4]
polycyclic(s1_4_4_ane) + polycyclic(s2_3_4_ane) - ring(Cyclobutane) + polycyclic(s2_3_4_ane)

**Case 4: CC12C3C4C3C3C1C2C43**


Benchmark: [3, 4, 4, 3]
polycyclic(s2_3_5_ane) + polycyclic(s2_4_5_ane) + polycyclic(s2_3_5_ane) + polycyclic(s2_3_4_ane) - ring(Cyclobutane) - ring(Cyclopropane) - ring(Cyclopentane) - ring(Cyclopentane)

Testing: [3, 3, 4, 4, 5]
polycyclic(s2_3_4_ane) + polycyclic(s2_4_4_ane) + polycyclic(s2_3_4_ane) - ring(Cyclobutane) - ring(Cyclobutane) + polycyclic(s2_3_4_ane)

mliu49:
After some debugging with @KEHANG, we discovered that `Graph.getDisparateRings` was giving incorrect results non-deterministically, depending on the order of rings in the SSSR output. To address that issue, the method was rewritten to recursively combine rings that share atoms.

Because the order of vertices is no longer dependable due to use of sets, an additional call to `Graph._sortCyclicVertices` was required in `calculateCyclicSymmetryNumber`.

I am currently running RMG-tests locally, and will post results once it finishes.

mliu49:
To answer the remaining two questions I posed in my original post:
- Bicyclic decomposition does not work properly with the set of relevant cycles. Including the extra ring (in cases where the RC differs from the SSSR) leads to double counting of rings, which resulted in a 2x increase in enthalpy for some of the cases I tested (cubane, barrelene). Therefore, I don't think there's any benefit to switching.
- Similarly, getDeterministicSSSR is still needed because the new SSSR is still non-deterministic in terms of the rings that are returned in cases where there are multiple options. Although this is what the set of relevant cycles is intended to address, it does not work in this application as mentioned above. Of course, because getDeterministicSSSR is based on the old SSSR algorithm, it still returns incorrect ring sets in certain cases. However, I think that is acceptable for now if deterministic behavior is more important. This can be revisited in the future if desired.

I have added a note to the doctring of `getDeterminsticSSSR` recommending use of `getSSSR` and `getRC` for new applications.

mliu49:
Updated RMG-tests results:

---|---|---|---|---
eg1|00:00:02:10|00:00:02:09|435.42|434.64
eg3|00:00:04:59|00:00:04:49|400.60|399.25
eg5|00:00:01:19|00:00:01:16|355.98|356.02
eg6|00:00:00:52|00:00:00:51|345.49|349.77
eg7|00:00:00:34|00:00:00:34|339.53|342.32
NC|00:00:01:27|00:00:01:25|336.71|334.16
solvent_hexane|00:00:00:49|00:00:00:48|331.40|330.59
MCH|00:00:03:27|00:00:03:31|692.00|698.31

Thermo validation results:

dataset | benchmark mae (kcal/mol) | new mae (kcal/mol)
--- | --- | ---
large_fused_polycyclic_table | 9.55 | 9.57
large_linear_polycyclic_table | 9.05 | 9.08
small_cyclic_table | 3.42 | 3.42
large_linear_O_only_polycyclic_table | 6.94 | 6.94
large_fused_O_only_polycyclic_table | 10.31 | 10.32
small_O_only_polycyclic_table | 5.30 | 5.30
zb_bicyclics_195_table | 9.59 | 9.59

mliu49:
This is ready for final review and merging.

PR #1130, opened 2017-08-04
extend docstring for Atom, Bond, Molecule

goldmanm:
the docstrings for Atom, Bond, and Molecule were missing some
attributes, so this commit/PR adds them to the docstring to improve
code readability and usability.

### DIFF
--- a/rmgpy/molecule/molecule.py
+++ b/rmgpy/molecule/molecule.py
@@ -688,7 +688,7 @@ class Molecule(Graph):
     
     
     def __hash__(self):
-        return hash((self.getFingerprint()))
+        return hash((self.fingerprint))
             
     def __richcmp__(x, y, op):
         if op == 2:#Py_EQ
@@ -702,7 +702,7 @@ class Molecule(Graph):
         """Method to test equality of two Molecule objects."""
         if not isinstance(other, Molecule): return False #different type
         elif self is other: return True #same reference in memory
-        elif self.getFingerprint() != other.getFingerprint(): return False
+        elif self.fingerprint != other.fingerprint: return False
         else:
             return self.isIsomorphic(other)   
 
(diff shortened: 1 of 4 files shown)


rmgpy/molecule/molecule.py
```diff
@@ -83,6 +83,10 @@ class Atom(Vertex):
     `coords`            ``numpy array``     The (x,y,z) coordinates in Angstrom
     `lonePairs`         ``short``           The number of lone electron pairs
     `id`                ``int``             Number assignment for atom tracking purposes
+    `bonds`             ``dict``            Dictionary of bond objects with keys being neighboring atoms
+    `mass`              ``int``             atomic mass of element
+    `number`            ``int``             atomic number of element
+    `symbol`            ``str``             atomic symbol of element
```
mliu49: Perhaps add a `(read-only)` note for mass, number, and symbol?

rmgpy/molecule/molecule.py
```diff
@@ -669,6 +675,9 @@ class Molecule(Graph):
     `symmetryNumber`        ``int``     The (estimated) external + internal symmetry number of the molecule
     `multiplicity`          ``int``     The multiplicity of this species, multiplicity = 2*total_spin+1
     `props`                 ``dict``    A list of properties describing the state of the molecule.
+    `InChI`                 ``str``     A string representation of the molecule in InChI
+    `atoms`                 ``list``    A list of Atom objects in the molecule
+    `_fingerprint`          ``str``     A representation for fast comparison, set as molecular formula
```
mliu49: I don't think you should include `_fingerprint` since it's intended to be a private attribute. The `getFingerprint` method should be used to access it.

rmgpy/molecule/molecule.py
```diff
@@ -669,6 +675,9 @@ class Molecule(Graph):
     `symmetryNumber`        ``int``     The (estimated) external + internal symmetry number of the molecule
     `multiplicity`          ``int``     The multiplicity of this species, multiplicity = 2*total_spin+1
     `props`                 ``dict``    A list of properties describing the state of the molecule.
+    `InChI`                 ``str``     A string representation of the molecule in InChI
+    `atoms`                 ``list``    A list of Atom objects in the molecule
```
mliu49: In the paragraph above this table, it says that the `atoms` and `bonds` attributes are aliases for the `vertices` and `edges attributes. If you're going to list `atoms` in the table, I think you can remove it from the text. Also, there is no `bonds` bonds attribute (or `edges` for that matter), so I think you should also correct that.

goldmanm: fixed

goldmanm:
@mliu49, do you think this is ready to merge?

mliu49:
Yes, I think this can be merged. I don't have any more things to add. It needs to be rebased though.

PR #1134, opened 2017-08-07
Categorize tests into unittests and functional tests and only use unittests for coverage report

goldmanm:
This PR separates functional tests from the rest of the tests using the attribute 'functional', so that coverage is testing only unittests. The boundary between unittest and functional test is gray, but I drew the cutoff when an external `input.py` file is used to run RMG. 

Overall, 10 (out of 1277) tests became functional. These ten tests took over a third of the runtime for tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,8 @@ install:
   - make
 
 script: 
-  - make test-all
+  - make test-unittests
+  - make test-functional
   - make test-database
 
 after_success:
--- a/Makefile
+++ b/Makefile
@@ -74,22 +74,31 @@ decython:
 	find . -name *.so ! \( -name _statmech.so -o -name quantity.so -o -regex '.*rmgpy/solver/.*' \) -exec rm -f '{}' \;
 	find . -name *.pyc -exec rm -f '{}' \;
 
-test-all:
+test-unittests:
 ifeq ($(OS),Windows_NT)
-	nosetests --nocapture --nologcapture --all-modules --verbose --with-coverage --cover-inclusive --cover-package=rmgpy --cover-erase --cover-html --cover-html-dir=testing/coverage --exe rmgpy
+	nosetests --nocapture --nologcapture --all-modules -A 'not functional' --verbose --with-coverage --cover-inclusive --cover-package=rmgpy --cover-erase --cover-html --cover-html-dir=testing/coverage --exe rmgpy
 else
 	mkdir -p testing/coverage
 	rm -rf testing/coverage/*
-	nosetests --nocapture --nologcapture --all-modules --verbose --with-coverage --cover-inclusive --cover-package=rmgpy --cover-erase --cover-html --cover-html-dir=testing/coverage --exe rmgpy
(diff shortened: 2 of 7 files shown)


KEHANG:
The `test`, `test-all` and `test-functional` are not doing exactly as their names suggest. I'd propose `test-non-auth`, `test-unittests` and `test-functional`. Some commits will come soon to reflect that.

goldmanm:
I am not sure I agree with the 'local users still want to run functional tests' commit.  I actually would prefer to more often not run functional tests than run functional tests. Two reasons:

1. most errors already should occur on unittests 
1.  running only unittests will cut the time for `make test` by about 40%, 

By not running functional tests with `make test`, I can do more debugging quicker than if I was forced to run functional tests with `make test`.

If I care to run functional tests, then I can always `make test-functional`. 

Maybe we should see what other people think about this though.

goldmanm:
I think the updates to this pull request will no longer allow users to use the command `make test`, which may not be desired behavior.  Maybe `make test` could perform `make test-unittest` but without the `auth` requirements.

mliu49:
I think that `make test` should be kept, and it should run non-auth unit and functional tests with coverage. The intention is that typical users should run `make test` to confirm that their installation works properly. A typical user will not need to run the test suite anymore after installation.

For developers, we could have the following:
- `make test-unittests` for all unit tests
- `make test-unittests-non-auth` for non-auth unit tests
- `make test-functional` for all functional tests

For debugging, you can run individual test modules instead of the entire test suite, so speed should not be a concern. For checking local changes before pushing, run `make test-unittests-non-auth`.

Travis will run `make test-unittests` (with coverage) and `make test-functional` (without coverage).

goldmanm:
I like the division of testing for users and developers. Under this situation, might it be possible to remove coverage from `make test` as that is more meant for developers?

It seems like we could come up with a hundred permutations of test commands in the makefile. Haha.

KEHANG:
I think the concerns from @goldmanm are valid. Thanks @mliu49 for a good summary! We can add a new category `make test-unittests-non-auth` to make it more useful (potentially we can have `make test-functional-non-auth` in the future). 

But we might not want to over-subdivide the tests in the future. The categories in Makefile is more defined for users and Travis. e.g., the creation of `make test-unittests` (with coverage) and `make test-functional` (without coverage) is for Travis and Codecov automation. Developers probably should be able to customize the commands to test whatever they want (of course it'd be convenient for developer if there's a command in Makefile can be directly used).

mliu49:
I agree with @KEHANG. For any further subdivisions, developers can just make aliases for their own use.

goldmanm:
merged the commits and updated to your recommendations. Made the alias 'make test' correspond to 'make test-unittest-non-auth'.

goldmanm:
The main thing a reviewer can do is glance over the code, and run the different unittests locally. 

Also, if you notice any other functional tests accidentally put into unittests, feel free to mention it or add the functional attribute.

goldmanm:
@mliu49, let me know if there's anything else I can help with the reviewal process for this PR.

Ideally, I would like to get this merged before trying to get travis to get coverage of non-cythonized code.

goldmanm:
@mliu49 , rebased and ready to merge.

PR #1139, opened 2017-08-09
print species dictionary with cantherm job

goldmanm:
prints the species dictionary for every species given with a
thermo() function in cantherm. This allows the user to 
more easily add kinetics libraries or do other manipulations
in RMG.

### DIFF
--- a/rmgpy/cantherm/thermo.py
+++ b/rmgpy/cantherm/thermo.py
@@ -164,7 +164,7 @@ class ThermoJob:
         f.write('{0}\n\n'.format(prettify(string)))
         
         f.close()
-        
+        # write chemkin file
         f = open(os.path.join(os.path.dirname(outputFile), 'chem.inp'), 'a')
         if isinstance(species, Species):
             if species.molecule and isinstance(species.molecule[0], Molecule):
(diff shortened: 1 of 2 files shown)


goldmanm:
Note that this only works if the user inputs the species information directly, since the molecular structure is not input when the user uses method 1 to input species into cantherm.

zjburas:
You can actually specify the species structure using either method 1 or 2 and it will be read in either case. I suppose this point should be clarified in the documentation.

goldmanm:
You are completely correct, and it is clear in the documentation. I'll update the part of the documentation that I added to.

rmgpy/cantherm/thermo.py
```diff
@@ -179,7 +179,16 @@ def save(self, outputFile):
...
         f.close()
-    
+
+        # write species dictionary
+        f = open(os.path.join(os.path.dirname(outputFile), 'species_dictionary.txt'), 'a')
+        if isinstance(species, Species):
+            if species.molecule and isinstance(species.molecule[0], Molecule):
+                f.write(species.label)
+                f.write('\n')
+                f.write(species.molecule[0].toAdjacencyList(removeH=False))
```
mliu49: You can pass a `label` argument to `toAdjacencyList` instead of writing the label above.

alongd:
@goldmanm, could you please rebase and review this one comment on this branch?

rmgpy/cantherm/thermo.py
goldmanm: sounds like a much better strategy. I'll fix it.

goldmanm:
@mliu49 @alongd , Just rebased to master, and implemented the suggestion.

Issue #1050, opened 2017-06-13
On the fly model reduction


chiyw13:
When I want to construct a combustion model for a fuel, I usually found there is a huge number of reactions in the model with a small number of species. Too much trivial reactions really makes model analysis, such as sensitivity analysis and pathway analysis in Chemkin, quite difficult. While we still have difficulties in running RMG model reduction tool in parallel with scoop and slurm.  So I want to work on on-the-fly model reduction in RMG so that RMG be used to generate more concise and effective models. Does anyone have experiences on this or have suggestions? Any advice and suggestions will be greatly appreciated.

alongd:
You can look at [rmgpy/reduction](https://github.com/ReactionMechanismGenerator/RMG-Py/tree/master/rmgpy/reduction).
It's not on-the-fly, but could be a good starting point.
Syntax for usage is `python %RMGPy%/rmgpy/reduction/main.py input.py reduction_input.py chem_annotated.inp species_dictionary.txt`
`reduction_input.py` should be something like:
```
"""
A list of the names of the species whose observables will be use to reduce the model.
A species observable may be the mole fraction at the end time of the batch reactor simulation.
The species name is the name of the species that can be found in the chemkin model or species 
dictionary, MINUS the species index in parentheses.
E.g.: ethane(1) --> ethane
"""

targets = ['H']

"""
A value between 0 and 1 that indicates how much percent (1 = 100%) the species observables
may deviate from their original values.
0.05 = 5% relative deviation
"""
tolerance = .3
```

mjohnson541:
I think a much faster way could be to use the dynamics criterion I developed that is a part of #997.  Since the dynamics criterion measures the relative impact a reaction has on related species concentrations measuring this criterion for core species could be used to eliminate unimportant core reactions.  This could take the form of a mechanism by which reactions are removed from the model if their maximum dynamics criterion is less than some set value after a complete simulation.  This would of course not be as robust as sensitivity based reduction, however if applied lightly you could probably greatly reduce the number of reactions you have without too much computational or accuracy cost.  

You could try to do this from a flux approach (scaling the reaction rate with the characteristic flux), but it would be much less direct (since it scales flux rather than the concentrations you're concerned with preserving).  You would also need to be careful to prevent removal of key structural reactions in the mechanism that the dynamics criterion naturally takes into account.

mjohnson541:
[...]

PR #1140, opened 2017-08-10
model core reaction reduction

mjohnson541:
This pull request enables two types of model core reaction reduction based on the dynamics criterion of core reactions:
New Input Parameters:  
toleranceNeglectCoreReaction
removeNeglectedReactions (flag)

1)  if removeNeglectedReactions = True, core reactions will be removed in the same way edge species are pruned-if the simulation goes all the way through all of the core reactions under tolerance for all time and all reaction systems will be removed permanently

2) if removeNeglectedReactions = False, (and toleranceNeglectCoreReaction > 0.0) then a chem_annotated_reduced.inp chemkin file will be produced each iteration containing only core reactions that passed reduction, note that if the simulation didn't go close to all the way through the reduction may be very poor.  

I'm in the process of adding a post-processing script to do this as well.  

This is based from #1082.


mjohnson541:
#1050

mjohnson541:
If we do this I no longer think we will approach it this way so I'm closing this PR.

PR #1141, opened 2017-08-10
generate edge seeds

mjohnson541:
A fundamental problem with using seed mechanisms to restart a run is that an input seed mechanism could be treated one of two ways
1) you don't react them -> you are missing lots of species from the edge that should be there
2) you react them -> you haven't had time to prune so you risk crashing your simulation or being forced to remove lots of edge species that might not have otherwise been pruned

Currently we only treat them as done in (1).  

@amarkpayne suggested that we just save the edge as well as core instead of having to manage all of this.  

This pull request generates an additional seed mechanism named name+"_edge" with an edge thermo library and an edge reaction library.  

Now when effectively restarting a run you can add the core seed mechanism as a seed mechanism and the edge seed mechanism as a reaction library, giving you the edge and core your original run ended with.

### DIFF
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -833,14 +833,16 @@ class RMG(util.Subject):
             
         speciesList = self.reactionModel.core.species
         reactionList = self.reactionModel.core.reactions
+        edgeSpeciesList = self.reactionModel.edge.species
+        edgeReactionList = self.reactionModel.edge.reactions
         
         # Make species labels independent
-        oldLabels = self.makeCoreSpeciesLabelsIndependent()
-            
+        oldLabels = self.makeSpeciesLabelsIndependent(speciesList)
+        edgeOldLabels = self.makeSpeciesLabelsIndependent(edgeSpeciesList)
+        
         # load thermo library entries
         thermoLibrary = ThermoLibrary(name=name)
-        for i in range(len(speciesList)): 
-            species = speciesList[i]
+        for i,species in enumerate(speciesList): 
             if species.thermo:
                 thermoLibrary.loadEntry(index = i + 1,
                                         label = species.label,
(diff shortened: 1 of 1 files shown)


mliu49:
What are the use cases of this "restart" method vs. restart files?

mjohnson541:
This approach requires a tiny fraction of the runtime and memory required to save a restart file, but after some tests I believe this will give you the exact same core and edge (I though it might react species in the input file with the seed mechanism, but it doesn't appear to do this if the included species are in the seed mechanism).  So I think ideally we would phase out restart file generation in favor of this.  I haven't been able to think of any advantage of restart file generation over this yet.

mjohnson541:
I just ran heptane-eg5 and did not get that error when starting with the created libraries.

mjohnson541:
We should also note that we've resolved the case with eg1 in #1147 now.

amarkpayne:
The issue for heptane-eg5 was solved by @mjohnson541 and @mliu49, and the fix will be included on a separate pull request. With the minimal example working with the fix suggested in #1147 this pull request should be ready to merge.

KEHANG:
Did anyone check the RMG-tests result for this branch before merging: https://travis-ci.org/ReactionMechanismGenerator/RMG-tests/builds/265355211?utm_source=github_status&utm_medium=notification? It cannot even finish eg3 (the second example). Before this merge, we at least can finish eg6: https://travis-ci.org/ReactionMechanismGenerator/RMG-tests/builds/262807529?utm_source=github_status&utm_medium=notification

@mjohnson541 could you investigate what leads to the slow-down?

mjohnson541:
Sorry, I checked it, but I'd seen it timeout that early before on other jobs for no apparent reason.  The slowdown appears to come from calling checkForDuplicates on the edge library.  I should have a fix up soon.

Issue #206, opened 2014-04-04
Setting spin multiplicity for pdep jobs while running rmg


shamelmerchant:
It seems we do not set the correct spin multiplicity for mono atomic species like [H] or for TS.

We also use the formula Spin multiplicity = Radical Count + 1 which is not the same as 2S  +1 especially when you consider singlet and triplet (or in general multivalent) species. This will directly affect the partition function calculations as 

Qnet = (Spin multiplicity \* Optical isomer \* Q / Symmetry number) where Q is the partition function.

PR #1142, opened 2017-08-11
Aromatics updates

zjburas:
Very small PR. Just added 5 new reaction families that are their own reverse to the appropriate hard-coded section of RMG-Py. Companion to Aromatics Updates PR's on Database.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1213,9 +1213,14 @@ class KineticsFamily(Database):
                 # reverse all the atoms in the chain between *1 and *2
                 # i.e. swap *4 with the highest, *5 with the second-highest
                 highest = len(atomLabels)
-                if highest>4:
-                    for i in range(4,highest+1):
-                        atomLabels['*{0:d}'.format(i)].label = '*{0:d}'.format(4+highest-i)
+                if highest > 4:
+                    # swap *4 with *5
+                    atomLabels['*4'].label = '*5'
+                    atomLabels['*5'].label = '*4'
+                if highest > 6:
+                    # swap *6 with the highest, etc.
+                    for i in range(6, highest+1):
+                        atomLabels['*{0:d}'.format(i)].label = '*{0:d}'.format(6+highest-i)
                         
             elif label == 'intra_ene_reaction':
                 # Labels for nodes are swapped
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1246,6 +1246,11 @@ class KineticsFamily(Database):
                 atomLabels['*1'].label = '*3'
                 atomLabels['*3'].label = '*1'
 
+            elif label == 'intra_substitutions_isomerization':
+                # Swap *2 and *3
+                atomLabels['*2'].label = '*3'
+                atomLabels['*3'].label = '*2'
+
         if not forward: template = self.reverseTemplate
         else:           template = self.forwardTemplate
 
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1217,12 +1217,29 @@ class KineticsFamily(Database):
                     for i in range(4,highest+1):
                         atomLabels['*{0:d}'.format(i)].label = '*{0:d}'.format(4+highest-i)
                         
-            elif label == 'H_shift_cyclopentadiene':
+            elif label == 'intra_ene_reaction':
                 # Labels for nodes are swapped
                 atomLabels['*1'].label = '*2'
                 atomLabels['*2'].label = '*1'
                 atomLabels['*3'].label = '*5'
                 atomLabels['*5'].label = '*3'
+
+            elif label == '6_membered_central_c-c_shift':
+                # Labels for nodes are swapped
+                atomLabels['*1'].label = '*3'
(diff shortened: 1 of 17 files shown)


rmgpy/data/kinetics/family.py
```diff
@@ -1219,13 +1219,30 @@ def applyRecipe(self, reactantStructures, forward=True, unique=True):
...
+            elif label == '1,2_shiftC':
+                # Labels for nodes are swapped
+                atomLabels['*2'].label = '*3'
+                atomLabels['*3'].label = '*2'
+
+            elif label == 'Intra_R_Add_Exo_scission':
+                # Labels for nodes are swapped
+                atomLabels['*1'].label = '*4'
+                atomLabels['*4'].label = '*1'
+
```
zjburas: Is this section of applyRecipe even necessary? It just looks like atomLabels are defined for reactions that are their own reverse, but are then not used anywhere.

zjburas: Never mind, I didn't realize variables were linked in python.

zjburas:
Added unit tests for the 6 following reaction families that are their
own reverse:

intra_H_migration
H_Abstraction
Intra_ene_reaction
6-membered_central_C-C_shift
1,2_shiftC
Intra_R_Add_Exo_scission

I think this PR is ready to merge, along with RMG Database PR's [#205](https://github.com/ReactionMechanismGenerator/RMG-database/pull/205) and [#206](https://github.com/ReactionMechanismGenerator/RMG-database/pull/206), in that order.

rmgpy/data/kinetics/family.py
```diff
@@ -1219,13 +1219,30 @@ def applyRecipe(self, reactantStructures, forward=True, unique=True):
...
                         
-            elif label == 'H_shift_cyclopentadiene':
+            elif label == 'intra_ene_reaction':
                 # Labels for nodes are swapped
                 atomLabels['*1'].label = '*2'
                 atomLabels['*2'].label = '*1'
                 atomLabels['*3'].label = '*5'
                 atomLabels['*5'].label = '*3'
 
+            elif label == '6_membered_central_C-C_shift':
```
mliu49: This label and the 1,2_shiftC label need to be all lowercase, since `label.lower()` is called higher up.

rmgpy/data/kinetics/familyTest.py
```diff
@@ -180,6 +182,289 @@ def testReactBenzeneBond2(self):
...
+17    H u0 p0 c0 {5,S}
+18    H u0 p0 c0 {10,S}
+19    H u0 p0 c0 {6,S}
+20    H u0 p0 c0 {9,S}
+21    H u0 p0 c0 {7,S}
+""")
+        products = family.applyRecipe(reactants)
+
+        self.assertEqual(len(products), 1)
+        self.assertTrue(expectedProduct.isIsomorphic(products[0]))
```
mliu49: These tests pass despite the labels having been wrong because `isIsomorphic` does not check atom labels. In order to check those, an mapping needs to be passed as a second argument to `isIsomorphic`.

mliu49:
I fixed the items I commented on. It seems that the atom label swapping for intra_H_migration was actually incorrect. Perhaps we changed the way we label backbone atoms at some point?

Could @nyee double-check this?

mliu49:
After looking into why incorrect atom label swapping never raised any errors, it appears that this entire section of code is unnecessary. We never use the labeled product structures afterwards. For generating the reverse reaction, we call `__matchReactantToTemplate` in order to regenerate the atom labels.

My guess is that the atom label swapping was for the original implementation of reaction generation and was never removed. This section of code remains unchanged since Josh's commits 7 years ago.

If my conclusions are correct, then this PR will not be needed anymore. Instead, I'll make a new one removing this section of code. Apologies to @zjburas for making you write these unit tests.

I would appreciate it if someone else could double check. I've read through the code and checked eg1, which remains completely unchanged.

zjburas:
There is a section of the RMG documentation ([12.4.2.1. Adding Reaction Family)](http://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/database/modification.html) that should be changed accordingly.

mjohnson541:
Sorry, I will review this when I finish with the third database aromatics PR.

rmgpy/data/kinetics/family.py
```diff
@@ -1217,7 +1217,7 @@ def applyRecipe(self, reactantStructures, forward=True, unique=True):
                     for i in range(4,highest+1):
                         atomLabels['*{0:d}'.format(i)].label = '*{0:d}'.format(4+highest-i)
                         
-            elif label == 'Intra_ene_reaction':
+            elif label == 'intra_ene_reaction':
```
mjohnson541: Please fix this in the original commit

mliu49: Will do! Must have squashed into the wrong commit.

mliu49:
I cleaned up the items you mentioned. I'm not sure why the MCH example would have found an additional species and more reactions. If anything, I would have expected fewer reactions.

mjohnson541:
Well, if you expected there might be fewer reactions I'm not very surprised by the appearance of more species and reactions.  If you can squash the first three I think this should be good to merge.

mjohnson541:
Sorry, can you rebase this again?

PR #1146, opened 2017-08-14
Canthermchanges2

rgillis8:
This commit adds a few functionalities to cantherm...
-Molpro output files are able to provide geometries, frequencies and zero-point energies for cantherm calculations
-Iodine species are allowed with Cantherm and the appropriate values for their spin orbital correct, MW, etc. are added
-CCSD(T)/aug-cc-pVTZ(-PP) method is added. (This is a high level method that is compatible with iodine species)   

In many ways this is a "cleaned up" version of canthermchanges that was put forward a few weeks ago. Thanks to @phalgunlolur for his work on this.


rmgpy/cantherm/statmech.py
```diff
@@ -537,18 +537,17 @@ def applyEnergyCorrections(E0, modelChemistry, atoms, bonds):
     # Spin orbit correction (SOC) in Hartrees
     # Values taken from ref 22 of http://dx.doi.org/10.1063/1.477794 and converted to hartrees
     # Values in millihartree are also available (with fewer significant figures) from table VII of http://dx.doi.org/10.1063/1.473182 
-    SOC = {'H':0.0, 'N':0.0, 'O': -0.000355, 'C': -0.000135, 'S':  -0.000893, 'P': 0.0} 
+    SOC = {'H':0.0, 'N':0.0, 'O': -0.000355, 'C': -0.000135, 'S':  -0.000893, 'P': 0.0, 'I':-0.011547226,}
     
     # Step 1: Reference all energies to a model chemistry-independent basis
     # by subtracting out that model chemistry's atomic energies
     # Note: If your model chemistry does not include spin orbit coupling, you should add the corrections to the energies here
-    if modelChemistry == 'CBS-QB3':
```
alongd: CBS-QB3 was removed and placed at the bottom between two `CCSD(T)` methods. I think it's less intuitive to look for it there. Could you return it here?

rmgpy/cantherm/statmech.py
```diff
@@ -537,18 +537,17 @@ def applyEnergyCorrections(E0, modelChemistry, atoms, bonds):
     # Spin orbit correction (SOC) in Hartrees
     # Values taken from ref 22 of http://dx.doi.org/10.1063/1.477794 and converted to hartrees
     # Values in millihartree are also available (with fewer significant figures) from table VII of http://dx.doi.org/10.1063/1.473182 
-    SOC = {'H':0.0, 'N':0.0, 'O': -0.000355, 'C': -0.000135, 'S':  -0.000893, 'P': 0.0} 
+    SOC = {'H':0.0, 'N':0.0, 'O': -0.000355, 'C': -0.000135, 'S':  -0.000893, 'P': 0.0, 'I':-0.011547226,}
```
alongd: I didn't see the SOC value for iodine in the ref in the comments above. Please add a new ref or add the method of calculation.

rmgpy/cantherm/statmech.py
```diff
@@ -671,7 +673,8 @@ def applyEnergyCorrections(E0, modelChemistry, atoms, bonds):
         atomEnergies[element] = atomHf[element] - atomThermal[element]
     for symbol, count in atoms.items():
         if symbol in atomEnergies: E0 += count * atomEnergies[symbol] * 4184.
-    
+
+    #raise Exception('{0:f}'.format(E0))
```
alongd: Is this commented line a debug leftover?

rmgpy/cantherm/statmech.py
rgillis8: @phalgunlolur - what is the source of this iodine SOC?

rmgpy/cantherm/molepro.py
```diff
@@ -42,7 +45,174 @@ class MoleProLog:
...
+                        break
+
+                    # Read molecular mass for external translational modes
+                    elif 'Molecular Mass:' in line:
+                        mass = float(line.split()[2])
+                        translation = IdealGasTranslation(mass=(mass,"amu"))
+                        modes.append(translation)
+                    # Read MolPro's estimate of the external symmetry number
+                    elif 'Rotational Symmetry factor' in line and symmetry is None:
+                        symmetry = int(float(line.split()[3]))
```
rgillis8: The next version will use the value put into the input file if there is one. If there is not one it will default to use the symmetry estimated by the software.

alongd:
@rgillis8 and @phalgunlolur , is this PR still relevant?

rgillis8:
I have now added an update version that I will be testing imminently. I believe I cleaned up all comments from the previous version.

rmgpy/cantherm/gaussian.py
```diff
@@ -156,12 +156,14 @@ def loadGeometry(self):
...
                 mass[i] = 35.4527
+            elif number[i] == 53:
+                mass[i] = 126.90447
             else:
                 raise NotImplementedError('Atomic number {0:d} not yet supported in loadGeometry().'.format(number[i]))
         
         return coord, number, mass
 
-    def loadConformer(self, symmetry=None, spinMultiplicity=None, opticalIsomers=1):
+    def loadConformer(self, symmetry=None, spinMultiplicity=None, opticalIsomers=1, symfromlog=None):
```
alongd: The gaussianTest currently fails since it doesn't pass this argument, and symmetry isn't being read.
Could you change `conformer = log.loadConformer()` into `conformer = log.loadConformer(symfromlog=True)` in [gaussianTest.py](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/cantherm/gaussianTest.py#L81)?
While we're there, and if you don't mind, could you also add this parameter to [QChem](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/cantherm/qchem.py#L179) (and relevant tests)?

rmgpy/cantherm/statmech.py
alongd: @phalgunlolur , could you see where did this SOC data come from?

rmgpy/cantherm/molepro.py
```diff
@@ -79,12 +274,47 @@ def loadEnergy(self,frequencyScaleFactor=1.):
...
+
+        f = open(self.path, 'r')
+        line = f.readline()
+        while line != '':
+
+            # Do NOT read the ZPE from the "E(ZPE)=" line, as this is the scaled version!
+            # We will read in the unscaled ZPE and later multiply the scaling factor
+            # from the input file
+
+            if 'Electronic Energy at 0 [K]:' in line:
```
alongd: My test does not have an `Electronic Energy at 0 [K]` line, but rather ends with:
```
 UCCSD(T)-F12A/cc-pVTZ-F12 energy=   -547.974475211996

   UCCSD(T)-F12A        HF-SCF  
   -547.97447521   -547.20094715
```

alongd: My fault, I didn't add the thermo keyword to the frequencies block in the molpro input file.
The above comment can be ignored, however could you add a note to the user in the Exception below to make sure the `thermo` option was added to the `{frequencies}` block?

rgillis8: Ok I'll work on these tonight. You are correct that you need the thermo keyword...
 {frequencies
 thermo
 print,thermo}
I'll makes sure that is clear from the debug note.

rmgpy/cantherm/molepro.py
```diff
@@ -42,7 +45,196 @@ class MoleProLog:
...
+        symbol = []; coord = []
+
+        f = open(self.path, 'r')
+        line = f.readline()
+        while line != '':
+            # Automatically determine the number of atoms
+            if 'Current geometry' in line:
+                symbol = []; coord = []
+                for i in range(4): line = f.readline()
+                while line != '\n':
```
alongd: I have a slightly different syntax for the molpro opt:
```

 END OF GEOMETRY OPTIMIZATION.

 Current geometry (xyz format, in Angstrom)


 Dump information in style XYZ to geometry.xyz                                                                    

    5
 DF-RKS-SCF000/CC-PVTZ  ENERGY=-698.91289988
 S         -0.0000000000       -0.0000000000        0.2325737348
 O         -0.0000000000        0.7683688596       -1.1496180790
 O         -1.2455119069        0.0000000000        0.9137857116
 O         -0.0000000000       -0.7683688596       -1.1496180790
 O          1.2455119069        0.0000000000        0.9137857116
```
Counting 4 lines from the ` Current geometry` phrase leads to an empty line in this case, and the geometry isn't parsed. We need a more robust method here. Maybe search for the word `ENERGY` after the `Current geometry` phrase, the parse the geometry?

alongd:
Thanks for all the changes @rgillis8 ! I think we only have these left:
- [ ] Modify the assertion message in molpro.py to include the `thermo` keyword in the `{frequencies}` block
- [ ] Add citation for the iodine SOC data (@phalgunlolur)
- [ ] Add the `symfromlog` keyword to QChem's loadConformer() and relevant tests

rmgpy/cantherm/molepro.py
```diff
@@ -42,7 +45,197 @@ class MoleProLog:
...
+                    elif 'Rotational Constants' in line and line.split()[-1]=='[GHz]':
+                        inertia = [float(d) for d in line.split()[-4:-1]]
+                        for i in range(3):
+                            inertia[i] = constants.h / (8 * constants.pi * constants.pi * inertia[i] * 1e9) *constants.Na*1e23
+                        rotation = NonlinearRotor(inertia=(inertia,"amu*angstrom^2"), symmetry=symmetry)
+                        modes.append(rotation)
+
+                    elif 'Rotational Constant' in line and line.split()[3]=='[GHz]':
+                        inertia = float(line.split()[2])
+                        inertia[0] = constants.h / (8 * constants.pi * constants.pi * inertia[0] * 1e9) *constants.Na*1e23
```
alongd: Since `inertia` is a float under this condition and not a list as above, we can't call it or assign into it as `inertia[0]`.
I think that all `inertia[0]` instances in this line should be replaced by `inertia`.

rgillis8:
Wasn't that a part of the other pull request that was merged [yesterday?](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1299) I'll rebase this one though.

alongd:
Yes it is! My confusion...

alongd:
@rgillis8 , could you add the following comment to the iodine SOC value, then we'll merge this in? 

`# Calculated as a weighted average of the electronic spin splittings of the lowest energy state. The splittings are obtained from Huber, K.P.; Herzberg, G., Molecular Spectra and Molecular Structure. IV. Constants of Diatomic Molecules, Van Nostrand Reinhold Co., 1979`

(there are also minor rebase conflicts with master to be solved)

Thanks!

rmgpy/cantherm/molepro.py
```diff
@@ -28,10 +28,13 @@
...
 
+import math
+import numpy
 import os.path
-
 import rmgpy.constants as constants
 
+from rmgpy.statmech import IdealGasTranslation, NonlinearRotor, LinearRotor, HarmonicOscillator, Conformer
+
 class MoleProLog:
```
alongd: In statmech.py we import `from rmgpy.cantherm.molpro import MolproLog`, but here the class name is `MoleProLog`. Could you fix this and make sure we only use the `Molpro` naming convention throughout this file?

alongd:
I remember we discussed the `loadNegativeFrequency()` block, which was already written for Molpro, but I don't see it in this PR. Could you add it (or did I get confused)?

rmgpy/cantherm/molepro.py
```diff
@@ -42,7 +45,197 @@ class MoleProLog:
...
+        last is returned.
+        """
+
+        symbol = []; coord = []
+
+        f = open(self.path, 'r')
+        line = f.readline()
+        while line != '':
+            # Automatically determine the number of atoms
+            if 'Current geometry' in line:
```
alongd: Loading the geometry fails when it is read from the freq.log instead of the opt.log (as we advise users to do when using hindered rotors, e.g., CanTherm documentation section 3.4.1.2). We should also be able to parse the Molpro freq.log geometry format:
```
 ATOMIC COORDINATES

 NR  ATOM    CHARGE       X              Y              Z

   1  S      16.00   -9.453662536    4.274754420    0.000189541
   2  H       1.00   -9.898568657    6.789652041   -0.000157075
   3  S      16.00   -5.465181259    4.791095755    0.000139265
   4  H       1.00   -5.020275141    2.276198114    0.000280349

 Bond lengths in Bohr (Angstrom)
```

rgillis8:
I'll work on these tonight. I am giving group meeting on Wednesday and so I can't guarantee I'll have everything ready before then, but I'll make sure to update whatever you mention here.

alongd:
Thanks! This could all wait till after your talk, no worries.

rgillis8:
As always let me know if you notice anything strange. I tested it, but things can always slip through the cracks.

alongd:
@rgillis8 , could you reword two commit messages before we merge?
I think `More robust geometry reading method` and `Bugfix for linear molecules` should be more specific. Perhaps mention that it's for Molpro?

mliu49:
Could you address some of the Codacy warnings before merging? Mainly the four unused variables (not including the two for `i`). Thanks!

rmgpy/cantherm/molepro.py
alongd: @rgillis8, I remembered what I meant in this comment:
When reading the geometry from a Molpro FREQ file, Cantherm cannot find the `Current geometry` line (see example FREQ geometry above), so it skips it. As a consequence, `symbol` remains an empty list, and `loadGeomerty()` cannot return the correct mass (which is latter required in statmech.py `if F is not None and len(mass) > 1 and len(rotors) > 0:` to calc rotors). Could you take a look at improving `loadGeometry()` for these cases?

alongd:
@rgillis8, I think that all we have left here is to enable reading the geometry from a molpro FREQ file, reword two commit messages, and address the Codacy warnings. Let me know if your too tied-up and whether I can assist with anything.

rgillis8:
I think this should be clean and ready to merge. There were quite a few issues merging it with Colin's recent changes to statmech.py. You'll note that I added Iodine to the lists of atomic ethalpy and thermal with a 0 value. This is not true but works because I fit the atomic energies using a zero value.

cgrambow:
Figured I would chime in here before this gets merged. Did you compare your fitted atomic energies to just using the calculated atomic energies? For F12 methods, I typically get excellent results with just the calculated energies.

rmgpy/cantherm/statmech.py
```diff
@@ -613,6 +620,12 @@ def applyEnergyCorrections(E0, modelChemistry, atoms, bonds,
                 atomEnergies = {'H':-0.499946213243 + SOC['H'], 'N':-54.588545831900 + SOC['N'], 'O':-75.065995072347 + SOC['O'], 'C':-37.844662139972+ SOC['C']}
             elif modelChemistry == 'ccsd(t)-f12/cc-pcvqz-f12':
                 atomEnergies = {'H':-0.499994558325 + SOC['H'], 'N':-54.589137594139+ SOC['N'], 'O':-75.067412234737+ SOC['O'], 'C':-37.844893820561+ SOC['C']}
+            elif modelChemistry == 'ccsd(t)-f12/cc-pvtz-f12(-pp)':
+                atomEnergies = {'H':-0.499946213243, 'N':-54.53000909621, 'O':-75.004127673424, 'C':-37.789862146471, 'S':-397.675447487865, 'I':-294.81781766}
+            elif modelChemistry == 'ccsd(t)-f12/aug-cc-pvtz(-pp)':
+                atomEnergies = {'H':-0.499844820798 + SOC['H'], 'N':-54.527419359906 + SOC['N'], 'O':-75.000001429806+ SOC['O'], 'C':-37.788504810868+ SOC['C'], 'S':-397.615032078484, 'I':-294.816178186847}
+            elif modelChemistry == 'ccsd(t)/aug-cc-pvtz(-pp)':
+                atomEnergies = {'H':-0.499821176024 + SOC['H'], 'N':-54.417460988701 + SOC['N'], 'O':-74.97498502 + SOC['O'], 'C':-37.729456757559 + SOC['C'], 'S':-397.6242802 + SOC['S'], 'I':-294.8436651}
```
cgrambow: If these are fitted, can you add a comment briefly describing the procedure? How come some of them have spin-orbit corrections, while others don't? (e.g. no SOC for sulfur for ccsd(t)-f12/aug-cc-pvtz(-pp))

rmgpy/cantherm/statmech.py
```diff
@@ -692,12 +705,12 @@ def applyEnergyCorrections(E0, modelChemistry, atoms, bonds,
...
         atomHf = {'H': 51.63,
                   'Li': 37.69, 'Be': 76.48, 'B': 136.2, 'C': 169.98, 'N': 112.53, 'O': 58.99, 'F': 18.47,
-                  'Na': 25.69, 'Mg': 34.87, 'Al': 78.23, 'Si': 106.6, 'P': 75.42, 'S': 65.66, 'Cl': 28.59}
+                  'Na': 25.69, 'Mg': 34.87, 'Al': 78.23, 'Si': 106.6, 'P': 75.42, 'S': 65.66, 'Cl': 28.59, 'I':0}
         # Thermal contribution to enthalpy Hss(298 K) - Hss(0 K) reported by Gaussian thermo whitepaper
         # This will be subtracted from the corresponding value in atomHf to produce an enthalpy used in calculating the enthalpy of formation at 298 K
         atomThermal = {'H': 1.01,
                        'Li': 1.1, 'Be': 0.46, 'B': 0.29, 'C': 0.25, 'N': 1.04, 'O': 1.04, 'F': 1.05,
-                       'Na': 1.54, 'Mg': 1.19, 'Al': 1.08, 'Si': 0.76, 'P': 1.28, 'S': 1.05, 'Cl': 1.1}
+                       'Na': 1.54, 'Mg': 1.19, 'Al': 1.08, 'Si': 0.76, 'P': 1.28, 'S': 1.05, 'Cl': 1.1, 'I':0}
```
cgrambow: Can you add the true values for iodine here and adjust your atomic energies so that future model chemistries that might want to use iodine have the true values available? You can probably use the values from [here](http://www.science.uwaterloo.ca/~cchieh/cact/tools/thermodata.html) (note that the heats of formation in `atomHf` are at 0 K).

rgillis8:
Sure. I'll get to these probably late tonight or tomorrow.

rmgpy/cantherm/statmech.py
rgillis8: They were all fit to about 5-8 iodine containing molecules. I'll include a more detailed note. I expect the disparity in SOC is whether they were fit with them or not but I'll change them so that the f12 method includes the SOC.

rgillis8:
I think this might finally be ready to pull in. Per Colin's suggestion I just used the calculated atomic energies for the CCSD(T)-F12 method. I did need to use fitted atomic energies for the non-f12 ccsd method, which is based on an admittedly small number of molecules. It could certainly be updated, but provides at least a way to calculate the thermo for organic iodine containing molecules.

rgillis8:
@alongd  I just spoke with Colin who says that the changes to statmech.py look fine. If the previous changes were fine I think it is ready to merge.

Issue #1148, opened 2017-08-15
Using thermo library name for species name bad for radicals with HBI


mliu49:
When estimating radical thermo using an HBI correction on top of a thermo library value for the stable species, the thermo object will carry the name of the stable species, resulting in it being used for the radical as well.

mjohnson541:
closed with #1149

PR #1149, opened 2017-08-16
removes the ThermoData.label for HBI

mjohnson541:
this prevents the pre-HBI corrected thermo label from being used for the HBI corrected thermo


mliu49:
Could you add a comment to document the purpose of this line?

mjohnson541:
added

mjohnson541:
fixes #1148

Issue #1120, opened 2017-08-02
Wall Time Calculation Causes generateReactions.py to Fail to Run


laitcl:
I was attempted to run the generateReactions.py script for a system of Hexylbenzene radicals, using the example file provided as a template.

I've encountered the following error message while doing so:

  File "C:\RMG_Workspace\RMG-Py\rmgpy\rmg\main.py", line 419, in initialize
    self.wallTime = int(data[-1]) + 60 * int(data[-2]) + 3600 * int(data[-3]) + 86400 * int(data[-4])

Upon inspection of main.py, commenting out the following lines makes generateReactions.py run properly (starting line 419)

        data = self.wallTime.split(':')
        self.wallTime = int(data[-1]) + 60 * int(data[-2]) + 3600 * int(data[-3]) + 86400 * int(data[-4])
        if not len(data) == 4:
            raise ValueError('Invalid format for wall time; should be DD:HH:MM:SS.')

It would appear that the variable data is a vector of one element only, causing an out of bounds index error.

PR #1151, opened 2017-08-17
Fix Walltime for generateReactions.py

amarkpayne:
This file set the walltime as '0', but the correct format is '00:00:00:00'. This should fix issue #1120


amarkpayne:
So that we don't have similar issues whenever we update parseCommandLineArguments in rmg.py, I decided to import this module from rmg.py in generateReactions.py in rmgpy.tools

amarkpayne:
This pull request is being combined with #1156

PR #1152, opened 2017-08-17
Seed mechanism adaptions

mjohnson541:
This pull request causes seed mechanism generation to avoid time intensive duplicate checking, by instead doing the duplicate checking when the library is loaded.  This also sets seed mechanism generation to default off.


mjohnson541:
This will be superseded soon

PR #1155, opened 2017-08-17
Proper seed mechanisms

mjohnson541:
This pull request improves seed mechanism generation:
1) created seeds are checked for duplicates when they are loaded rather than when they are created to save time
2) created seed/libraries are loaded as template and original library reactions, (pdep reactions are still treated as library reactions for now)

### DIFF
--- a/documentation/source/users/rmg/input.rst
+++ b/documentation/source/users/rmg/input.rst
@@ -540,6 +540,9 @@ Miscellaneous Options
 Miscellaneous options:: 
 
     options(
+        name='Seed',
+        generateSeedEachIteration=True,
+        saveSeedToDatabase=True,
         units='si',
         saveRestartPeriod=(1,'hour'),
         generateOutputHTML=True,
@@ -548,6 +551,12 @@ Miscellaneous options::
         verboseComments=False,
         saveEdgeSpecies=True,
     )
+
+The ``name`` field is the name of any generated seed mechanisms
+
+Setting ``generateSeedEachIteration`` to ``True`` tells RMG to save and update a seed mechanism and thermo library during the current run
+
+Setting ``saveSeedToDatabase`` to ``True`` tells RMG (if generating a seed) to also save that seed mechanism and thermo library directly into the database
 
 The ``units`` field is set to ``si``.  Currently there are no other unit options.
 
--- a/rmgpy/data/kinetics/library.py
+++ b/rmgpy/data/kinetics/library.py
@@ -35,7 +35,12 @@ This module contains functionality for working with kinetics libraries.
 import os.path
 import logging
 import re
-
+try:
+    from collections import OrderedDict
+except ImportError:
+    logging.warning("Upgrade to Python 2.7 or later to ensure your database entries are read and written in the same order each time!")
+    OrderedDict = dict
+    
 from rmgpy.data.base import DatabaseError, Database, Entry
 
 from rmgpy.reaction import Reaction
@@ -45,6 +50,7 @@ from rmgpy.kinetics import Arrhenius, ThirdBody, Lindemann, Troe, \
 from rmgpy.molecule import Molecule
 from rmgpy.species import Species
 from .common import saveEntry
+import codecs
 
 ################################################################################
 
@@ -117,8 +123,9 @@ class KineticsLibrary(Database):
     A class for working with an RMG kinetics library.
     """
 
-    def __init__(self, label='', name='', solvent=None, shortDesc='', longDesc=''):
+    def __init__(self, label='', name='', solvent=None, shortDesc='', longDesc='', duplicatesChecked=True):
         Database.__init__(self, label=label, name=name, shortDesc=shortDesc, longDesc=longDesc)
+        self.duplicatesChecked=duplicatesChecked
         
     def __str__(self):
         return 'Kinetics Library {0}'.format(self.label)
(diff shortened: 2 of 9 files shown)


rmgpy/rmg/model.py
```diff
@@ -1257,11 +1257,36 @@ def addSeedMechanismToCore(self, seedMechanism, react=False):
 
         seedMechanism = database.kinetics.libraries[seedMechanism]
 
-        for entry in seedMechanism.entries.values():
-            rxn = LibraryReaction(reactants=entry.item.reactants[:], products=entry.item.products[:],\
-             library=seedMechanism.label, specificCollider=entry.item.specificCollider, kinetics=entry.data, duplicate=entry.item.duplicate,\
-             reversible=entry.item.reversible
-             )
+        for entry in seedMechanism.entries.values():                
+            if entry._longDesc and 'Originally from reaction library: ' in entry._longDesc:
```
mliu49: I think you should make a new method for this to reduce code duplication.

rmgpy/data/kinetics/rules.py
```diff
@@ -705,7 +705,10 @@ def estimateKinetics(self, template, degeneracy=1):
         if degeneracy > 1:
             kinetics.comment += "\n"
             kinetics.comment += "Multiplied by reaction path degeneracy {0}".format(degeneracy)
-
+        
+        kinetics.comment += "\n"
+        kinetics.comment += "family: {0}".format(self.label.replace('/rules',''))
```
mliu49: I'm pretty sure the family is already stored in the comments.

mjohnson541: I know, but it isn't in the kinetics comments in the seed so I needed to add it.

rmgpy/rmg/model.py
mjohnson541: done

rmgpy/rmg/model.py
```diff
@@ -1343,31 +1343,13 @@ def getStoichiometryMatrix(self):
...
+    
+    def constructLibraryReactions(self,library):
         """
-        Add all species and reactions from `seedMechanism`, a 
-        :class:`KineticsPrimaryDatabase` object, to the model core. If `react`
-        is ``True``, then reactions will also be generated between the seed
-        species. For large seed mechanisms this can be prohibitively expensive,
-        so it is not done by default.
+        adds reactions from a given kinetics library in preparation to add it to core as seed mechanism
+        or to the edge as a reaction library
```
goldmanm: From the comment, I don't understand:

1. what is the method adding reactions to if not to the core nor edge?
1. Is the attribute `library` a string or a library object?

It seems like only one line of this method requires the CoreEdgeReactionModel attributes. I think it might be better to have this method in library.py, and have it return a list of reactions. Then `addSeedMechanismToCore` can use `makeNewReactions`. This can also reduce the bloating, and subsequently difficulty in testing, of model.py that we have seen recently. 

When it is in library.py, it might be good to have a test method for it, unless you are hoping to scrap it soon and no longer rely on kinetics comments anymore.

mjohnson541: I'm afraid I don't understand 1 it says in the docstring specifically that it is preparing the library to be added to the core as a seed mechanism or to the edge as a reaction library.  

2.  It is a library object.  

This is primarily a helper function for CoreEdgeReactionModel methods addSeedMechanismToCore and addReactionLibraryToEdge, (the only methods that call it).  It also couldn't be called without a CoreEdgeReactionModel to begin with so I really don't think it belongs outside of Model.py.

goldmanm: I don't think I was clear in my explanation earlier. We should discuss this in person before anything happens with this PR.

mjohnson541: Ok those issues should be resolved now.  I've moved the LibraryReaction and TemplateReaction construction to library.py.

goldmanm:
@mjohnson541, could you add any new input file parameters to the documentation in the pull request as well?

mjohnson541:
done

documentation/source/users/rmg/input.rst
```diff
@@ -549,6 +552,12 @@ Miscellaneous options::
         saveEdgeSpecies=True,
     )
 
+The ``name`` field is the name of any generated seed mechanisms
+
+Setting ``generateSeedEachIteration`` to ``True`` tells RMG to save and update a seed mechanism and thermo library during the current run
```
goldmanm: If this is 'off', does this create a seed at the end of the simulation? What is the intended behavior?

rmgpy/rmg/model.py
goldmanm: Could you combine this commit with the commit moving the new method to 'family.py'?

Also, I am not sure what 'no change in pdep handling' indicates in the commit message. When merging, could you reword it to describe that loading does not work properly with jobs that output pdep reactions (and maybe why they don't work properly). This would be super helpful for future development.

goldmanm:
You may want to unittest the load library method, if you are not expecting to replace it anytime soon.

documentation/source/users/rmg/input.rst
mjohnson541: Supposing it is 'off' It will make one seed at the beginning to initialize the folder and I believe that if the job fails in certain ways it will generate one at the end, but otherwise it will not update it.

rmgpy/data/kinetics/library.py
```diff
@@ -230,8 +277,45 @@ def convertDuplicatesToMulti(self):
...
+        local_context['shortDesc'] = self.shortDesc
+        local_context['longDesc'] = self.longDesc
+        # add in anything from the Class level dictionary.
+        for key, value in Database.local_context.iteritems():
+            local_context[key]=value
         
+        # Process the file
+        f = open(path, 'r')
+        try:
+            exec f in global_context, local_context
```
mliu49: Codacy is complaining because you added an `exec` statement here. It got flagged twice, once by Bandit and once by PyLint.

mjohnson541: Yeah, I guess I forgot about that copy and paste operation.

PR #1156, opened 2017-08-17
Add Unit Test for rmg.py

amarkpayne:
Add test for parseCommandLineArguments for proper default values

### DIFF
--- a/rmg.py
+++ b/rmg.py
@@ -42,7 +42,7 @@ from rmgpy.rmg.main import RMG, initializeLog, processProfileStats, makeProfileG
 
 ################################################################################
 
-def parseCommandLineArguments():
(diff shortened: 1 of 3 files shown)


amarkpayne:
This will need to be updated as soon as #1151 get pulled in

mliu49:
I don't think this test actually gets found by nosetests since it's not in the rmgpy directory. Perhaps you could move it to rmgpy/rmg/rmgTest.py as a new test class? The class could be called `TestRMGScript`.

Also, I think you could probably combine this with #1151 as one PR.

amarkpayne:
I made the requested changes. Assuming it passes all the tests this should be ready to merge.

PR #1160, opened 2017-08-18
Completely refactor rmgpy.rmg.mainTest.TestMain

mliu49:
- Create permanent input file so it is not written every time
- Split tests into multiple test methods
- Add tests for edge seed libraries
- Clean up properly after tests

### DIFF
--- a/rmgpy/rmg/mainTest.py
+++ b/rmgpy/rmg/mainTest.py
@@ -43,136 +43,127 @@ originalPath = getPath()
 
 class TestMain(unittest.TestCase):
 
-    def setUp(self):
-        self.dir_name = 'temp_dir_for_testing'
-        os.chdir(originalPath)
-        os.mkdir(self.dir_name)
-        os.chdir(self.dir_name)
-        self.databaseDirectory = settings['database.directory']
-        self.thermoDir = os.path.join(self.databaseDirectory,'thermo','libraries','testSeed.py')
-        self.kineticsDir = os.path.join(self.databaseDirectory,'kinetics','libraries','testSeed')
-        inputFile = """
-database(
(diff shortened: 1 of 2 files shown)


rmgpy/rmg/mainTest.py
```diff
@@ -43,136 +43,143 @@
...
+
+        Note: Since this test modifies the class level RMG instance,
+        it can cause other tests to fail if run out of order."""
+        # Load the seed libraries into the database
+        self.rmg.database.load(
+            path=self.databaseDirectory,
+            thermoLibraries=['testSeed_edge'],
+            reactionLibraries=['testSeed_edge'],
+            seedMechanisms=['testSeed_edge'],
+            kineticsFamilies='none',
```
mjohnson541: After the next seed mech pull we will try to load them as template reactions so we need to have the families loaded so can you load the default families?

goldmanm: What do you mean by 'them'?

I don't understand why you need default families loaded. Which PR are you talking about? A specific number and link would be helpful.

mjohnson541: Sorry, #1155 starts loading reactions as their original sources (template reactions and original library reactions).  If the family or library use to create a given reaction isn't loaded it can't find the original template/library and errors.  I mean I can change it later in that pull request of course, but I think it would be cleaner to do it here.

mliu49: I understand what you mean. I removed them because it cut off 2 minutes from the test run time. I can put it back in.

mliu49:
Updated and ready to merge. Kinetics families are now being loaded again.

PR #1161, opened 2017-08-19
Restructure rmg.py

amarkpayne:
- Move lines that modify parsed args into the arg parse module
- Add unit tests for non-default values of command line arguments
- create `main` function for rmg.py

### DIFF
--- a/rmg.py
+++ b/rmg.py
@@ -57,7 +57,7 @@ def parse_command_line_arguments(command_line_args=None):
     how molecules react.
     """)
     parser.add_argument('file', metavar='FILE', type=str, nargs=1,
-        help='a file describing the job to execute')
+                        help='a file describing the job to execute')
 
     # Options for controlling the amount of information printed to the console
     # By default a moderate level of information is printed; you can either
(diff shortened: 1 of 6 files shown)


amarkpayne:
I forgot to include these commits in #1156 to address some of the issues mentioned in #1067.

amarkpayne:
@mliu49 will the website need to be updated because of this? I can look into it if so

mliu49:
I don't think it will, since the website doesn't run full RMG jobs, and generateReactions is called as a subprocess. It might be good to double check though.

amarkpayne:
I searched for `generateReactions` in the website repository to see if anything was imported. I found one instance where `generateReactions.py` is called and handed an input file, but this should work with my code (from the outside the code still works the same way).

I also searched for `parse` since I changed the name of `parse_command_line_arguments` in rmg.py, but I found no matches to this. I think that is everything that could be affected, so we should be good to go.

rmg.py
```diff
@@ -69,94 +70,105 @@ def parseCommandLineArguments(args):
...
     if args.postprocess:
-        print "Postprocessing the profiler statistics (will be appended to RMG.log)"
         args.profile = True
+
+    return args
+
+
+def main():
+    # Parse the command-line arguments (requires the argparse module)
+    args = parse_command_line_arguments(sys.argv[1:])
```
mliu49: I found out that the `parse_args` method defaults to parsing `sys.argv[1:]`, so I think it would be cleaner to have `parse_command_line_arguments()` take an optional `command_line_args` with a default of `None`. That would still allow testing it the way you are currently, and not require an argument be provided when using it here and in generateReactions.py.

amarkpayne: Thanks for the suggestion! I was unaware that the `parse_args` method worked this way by default. I think this is a great idea. I'll make these changes and rebase.

amarkpayne:
I have made the requested changes and rebased. It looks like all of the checks passed as well.

Issue #1163, opened 2017-08-20
Travis Build Hung-up on 'Test that Mocpac getThermoData() works correctly'


amarkpayne:
This error is present in #1155, #1161, and #1162, so I think it is safe to say that this is an issue with Travis and not with these pull requests.

I think others in the group have seen this happen locally when running unit tests on Windows, such as @mliu49.

amarkpayne:
For what it's worth, some of our dependencies were updated right when this issue occurred. Below is the difference between the latest conda environment and one I had made a couple days before this issue.

```
1c1
< # packages in environment at /home/mark/anaconda2/envs/newest:
---
> # packages in environment at /home/mark/anaconda2/envs/rmg_env:
50a51,52
< markupsafe                1.0                      py27_0  
---
> markupsafe                0.23                     py27_2  
89c91
< prompt_toolkit            1.0.15                   py27_0  
---
> prompt_toolkit            1.0.14                   py27_0  
119c121
< ssl_match_hostname        3.5.0.1                  py27_0  
---
> ssl_match_hostname        3.4.0.2                  py27_1  
133c135
< xz                        5.2.3                         0  
---
> xz                        5.2.2                         1  
135c137
< zlib                      1.2.11                        0  
---
> zlib                      1.2.8                         3  
```

mliu49:
Seems like the MOPAC executable has expired. The Anaconda package was prepared exactly a year ago. The message can be ignored by pressing enter, but the subprocess call does not do that, so it hangs indefinitely. Potential solutions: 1) make new Anaconda packages, 2) press enter to continue. It seems that sending enter to the subprocess allows it to continue running without any problems. Idk why MOPAC makes their executables expire...

```
Your MOPAC executable, Version: 16.175M, has expired.
Please go to web-site: http://openmopac.net/Download_MOPAC_Executable_Step2.html to get a new version of MOPAC.
Do NOT request a new license key.  Academic license keys do not include time limits - the limit is in the program.

Press (enter) to continue.
```

amarkpayne:
It just passed. Thanks @mliu49 for the fix! I'll close out this issue then.

mliu49:
Personally, I think it's better to leave issues open until the fix has been merged into master. I'd also like to test this on Windows to see if that issue is also fixed before merging.

rwest:
[...]

PR #1164, opened 2017-08-21
Wallclock error fixes

goldmanm:
Improves error generation for wallclock updates by 

1. changing when an error is thrown by the wallclock
2. fixing the input file for generateReactions.ipynb to use the correct format.


amarkpayne:
The tests ran just fine locally. Any idea what is causing the issue in #1163 ?

goldmanm:
I probably know as much as you do about it. Thanks for running the local tests on this (and for creating issue #1163).

Issue #1158, opened 2017-08-18
Uncertainty extractSourcesFromModel broken


mjohnson541:
uncertainty.extractSourcesFromModel() is failing to properly parse the kinetics comment in Family.getSourcesForTemplate

Traceback (most recent call last):
  File "QueueGenerator.py", line 243, in <module>
    thermoUnc = getUncertainties(runDir)
  File "QueueGenerator.py", line 176, in getUncertainties
    uncertainty.extractSourcesFromModel()
  File "/Users/mattjohnson/RMGCODE/RMG-Py/rmgpy/tools/uncertainty.py", line 378, in extractSourcesFromModel
    source = self.database.kinetics.extractSourceFromComments(reaction)
  File "/Users/mattjohnson/RMGCODE/RMG-Py/rmgpy/data/kinetics/database.py", line 576, in extractSourceFromComments
    training, dataSource = self.families[reaction.family].extractSourceFromComments(reaction)
  File "/Users/mattjohnson/RMGCODE/RMG-Py/rmgpy/data/kinetics/family.py", line 2256, in extractSourceFromComments
    rules, trainingEntries = self.getSourcesForTemplate(template)
  File "/Users/mattjohnson/RMGCODE/RMG-Py/rmgpy/data/kinetics/family.py", line 2167, in getSourcesForTemplate
    entryNestedList = eval(evalCommentString)
  File "<string>", line 1
    O_sec_rad;HCO from training reaction 11
             ^
SyntaxError: invalid syntax

mliu49:
Any chance you could post the comment that led to this error?

mjohnson541:
O_sec_rad;HCO from training reaction 11
Estimated using template [O_sec_rad;HCO] for rate rule [O_rad/OneDe;HCO]
Euclidian distance = 1.0

mjohnson541:
Never mind, I think this is actually due to some differences when I ran the original model.  But on master I'm still getting an error nearby:  

Traceback (most recent call last):
  File "QueueGenerator.py", line 249, in <module>
    thermoUnc = getUncertainties(runDir)
  File "QueueGenerator.py", line 182, in getUncertainties
    uncertainty.extractSourcesFromModel()
  File "/Users/mattjohnson/RMGCODE/RMG-Py/rmgpy/tools/uncertainty.py", line 378, in extractSourcesFromModel
    source = self.database.kinetics.extractSourceFromComments(reaction)
[...]

PR #1165, opened 2017-08-21
Fix uncertainty

mjohnson541:
The pull request solves a number of current issues with uncertainty:

1.  Degeneracies can now be floats
2.  The parser doesn't break when the 'from training reaction' statement comes before and on a different line than 'estimated using template'
3.  Uncertainties can now deal with radical HBI corrections when the saturated species of the radical is not a species in the model

I've added a primitive test that checks that the uncertainties module can load the parameter uncertainty engines (effectively that it doesn't fail to estimate any raw thermo or kinetic uncertainties).  This test should be sufficient that things like 1-3 won't happen again.  

I had to add files to testing_database to pass the tests, but all of the new files are empty.  I had to update the other.py thermo group file because the uncertainty test needed one of the groups in it.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -2140,8 +2140,12 @@ class KineticsFamily(Database):
             
             # Discard the last line, unless it's the only line!
             # The last line is 'Estimated using ... for rate rule (originalTemplate)'
-            if len(lines) == 1:
-                comment = lines[0]
+            #if from training reaction is in the first line append it to the end of the second line and skip the first line
+            if not 'Average of' in kinetics.comment:
+                if 'from training reaction' in lines[0]:
+                    comment = lines[1]
+                else:
+                    comment = lines[0]
                 if comment.startswith('Estimated using template'):
                     tokenTemplateLabel = comment.split()[3][1:-1]
                     ruleEntry, trainingEntry = self.retrieveOriginalEntry(tokenTemplateLabel) 
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -2161,8 +2161,8 @@ class KineticsFamily(Database):
                 evalCommentString = re.sub(r" \+ ", ",",                        # any remaining + signs
                                     re.sub(r"Average of ", "",                  # average of averages
                                     re.sub(r"Average of \[(?!Average)", "['",   # average of groups
-                                    re.sub(r"(\b|\))]", r"\1']",                # initial closing bracket
-                                    re.sub(r"(?<=\b) \+ (?=Average)", "',",     # + sign between non-average and average
+                                    re.sub(r"(\w|\))]", r"\1']",                # initial closing bracket
+                                    re.sub(r"(?<=[\w)]) \+ (?=Average)", "',",  # + sign between non-average and average
                                     re.sub(r"(?<=]) \+ (?!Average)", ",'",      # + sign between average and non-average
                                     re.sub(r"(?<!]) \+ (?!Average)", "','",     # + sign between non-averages
                                     comment)))))))
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -2243,7 +2243,7 @@ class KineticsFamily(Database):
             elif line.startswith('Estimated'):
                 pass
             elif line.startswith('Multiplied by'):
-                degeneracy = int(line.split()[-1])
+                degeneracy = float(line.split()[-1])
 
         # Extract the rate rule information 
         fullCommentString = reaction.kinetics.comment.replace('\n', ' ')
--- a/rmgpy/tools/uncertainty.py
+++ b/rmgpy/tools/uncertainty.py
@@ -28,24 +28,9 @@
 import os
 import numpy
 import rmgpy.util as util 
+from rmgpy.species import Species
 from rmgpy.tools.plot import *
 from rmgpy.tools. data import GenericData
-
-def retrieveSaturatedSpeciesFromList(species, speciesList):
-    """
-    Given a radical `species`, this function retrieves the saturated species objects from a list of species objects.
-    """
-    molecule = species.molecule[0]
-    assert molecule.isRadical(), "Method only valid for radicals."
-    saturatedStruct = molecule.copy(deep=True)
-    saturatedStruct.saturate()
-    for otherSpecies in speciesList:
-        if otherSpecies.isIsomorphic(saturatedStruct):
-            return otherSpecies
-    else:
-        raise Exception('Could not retrieve saturated species form of {0} from the species list'.format(species))
-
-
 
 class ThermoParameterUncertainty:
     """
@@ -341,6 +326,31 @@ class Uncertainty:
                                                               dictionaryPath=dictionaryPath,
                                                               transportPath=transportPath)
 
+    def retrieveSaturatedSpeciesFromList(self,species):
+        """
+        Given a radical `species`, this function retrieves the saturated species objects from a list of species objects
+        and returns the saturated species object along with a boolean that indicates if the species is not part of the model
+        (True->not in the model, False->in the model)
+        """
+        
+        molecule = species.molecule[0]
+        assert molecule.isRadical(), "Method only valid for radicals."
+        saturatedStruct = molecule.copy(deep=True)
+        saturatedStruct.saturate()
+        for otherSpecies in self.speciesList:
+            if otherSpecies.isIsomorphic(saturatedStruct):
+                return otherSpecies, False
+        
+        #couldn't find saturated species in the model, try libraries
+        newSpc = Species(molecule=[saturatedStruct])
+        thermo = self.database.thermo.getThermoDataFromLibraries(newSpc)
+        
+        if thermo is not None:
+            newSpc.thermo = thermo
+            self.speciesList.append(newSpc)
+            return newSpc, True
+        else:         
+            raise Exception('Could not retrieve saturated species form of {0} from the species list'.format(species))
 
     def extractSourcesFromModel(self):
         """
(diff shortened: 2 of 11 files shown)


mjohnson541:
This should fix #1158

amarkpayne:
@mjohnson541, can you fix the code style issues identified by Codacy and then rebase? I should be able to review this by tomorrow evening if so.

mjohnson541:
I've removed the extraneous import, but I'm of the opinion that the rest of the codacy suggested changes for this PR would make harder to read and less clear.

mjohnson541:
Also fixed the iterator one as well now.

rmgpy/tools/uncertainty.py
```diff
@@ -341,37 +326,65 @@ def loadModel(self, chemkinPath, dictionaryPath, transportPath=None):
...
+            if otherSpecies.isIsomorphic(saturatedStruct):
+                return otherSpecies, False
+        else:
+            for key,value in self.database.thermo.libraries.iteritems():
+                for entry in value.entries.values():
+                    if entry.item.isIsomorphic(saturatedStruct):
+                        newSpc = Species().fromAdjacencyList(saturatedStruct.toAdjacencyList())
+                        newSpc.thermo = entry.data
+                        self.speciesList.append(newSpc)
+                        return newSpc, True
```
amarkpayne: If I understand this correctly, this function is used to return the species object of the saturated molecule in the model to use its sources of thermo estimation to later determine the uncertainty. As you noted, it is possible for a radical species to be present in the model without the corresponding saturated species being present. The idea of your code here is that it is a fall-back for this case, and searches our libraries to see if the molecule exists there.

Specifically, it looks like the uncertainty module loads whichever thermo libraries the user specifies, so as long as the user specifies the same libraries in the same order as they did when they made the RMG model. Assuming the user does this, and we are still unable to find the species in the libraries, could we not assume that the saturated species was estimated by group additivity? If so we should implement a final fall-back method that returns a species object with thermo estimated with group additivity.

tldr: Is a fall-back method for group-additivity needed?

rmgpy/tools/uncertainty.py
```diff
@@ -341,37 +326,65 @@ def loadModel(self, chemkinPath, dictionaryPath, transportPath=None):
...
     def extractSourcesFromModel(self):
         """
         Extract the source data from the model using its comments.
         Must be done after loading model and database to work.
         """
         self.speciesSourcesDict = {}
+        ignoreSpcs = []
         for species in self.speciesList:
-            source = self.database.thermo.extractSourceFromComments(species)
+            if not species in ignoreSpcs:
```
amarkpayne: Is this line necessary? I think the only way for a species to be added to `ignoreSpcs` is for it to have gone through this block once, and since we only loop over each specie in `self.speciesList` once this line should always be true.

rmgpy/tools/uncertainty.py
```diff
@@ -341,37 +326,65 @@ def loadModel(self, chemkinPath, dictionaryPath, transportPath=None):
...
+        """
+        
+        molecule = species.molecule[0]
+        assert molecule.isRadical(), "Method only valid for radicals."
+        saturatedStruct = molecule.copy(deep=True)
+        saturatedStruct.saturate()
+        for otherSpecies in self.speciesList:
+            if otherSpecies.isIsomorphic(saturatedStruct):
+                return otherSpecies, False
+        else:
```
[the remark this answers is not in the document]
mliu49: Codacy is flagging it because you don't need the else. It either returns or continues after the for loop. The same thing would be accomplished by removing the else and reducing the indentation of everything after.

rmgpy/tools/uncertainty.py
mliu49: In `extractSourceFromComments`, HBI and GAV are considered a single source. Retrieving the saturated structure is only done if the thermo was from two sources, ie. HBI on top of a library value or HBI on top of a QM value.

rmgpy/tools/uncertainty.py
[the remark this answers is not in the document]
amarkpayne: That's fair. I do agree with @mjohnson541 though, I like having the else there so that it is clear that the code only runs in the event that the for loop does not succeed. Would you object to changing the else statement with a comment like `# else if no species were found, check for the species in the thermo libraries`

rmgpy/tools/uncertainty.py
```diff
@@ -341,37 +326,65 @@ def loadModel(self, chemkinPath, dictionaryPath, transportPath=None):
...
+            for key,value in self.database.thermo.libraries.iteritems():
```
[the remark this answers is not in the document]
mliu49: It might be better to create the species first, then call `self.database.thermo.getThermoDataFromLibraries()`.

As an additional note, the species can be directly using `newSpc = Species(molecule=[saturatedStruct])` instead of conversion via adjlist.

rmgpy/tools/uncertainty.py
amarkpayne: Gotcha, so a fall-back to GAV is not only unnecessary but would give the wrong result (since we only would get here if GAV was not used)?

mliu49:
Sorry for the late comment, but I just remembered that I made a quick fix for the saturated species problem a long time ago. You can see it [here](https://github.com/mliu49/RMG-Py/commit/2d08944d863aa8215364cd134d7028bea2fc892b), but I think your changes are pretty similar and possibly more robust.

rmgpy/tools/uncertainty.py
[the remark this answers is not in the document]
mliu49: I think the cleanest way might be to return at the very end of the method. Returning early is also fine, but I would prefer not having the else. A comment would be helpful though.
```
for species in self.speciesList:
    if ... :
        result = something
        break
else:
    result = something_else

return result
```

rmgpy/tools/uncertainty.py
mliu49: I think so, if I understood correctly.

rmgpy/tools/uncertainty.py
mjohnson541: So ignoreSpcs are saturated species that are needed for getting thermo uncertainty correct, but aren't in the model, and they get appended to the end of speciesList so it can find them later.  However we do not want the sources for that molecule indexed in the same way species that are actually in the model are to avoid confusion later in processing.

rmgpy/data/kinetics/family.py
```diff
@@ -2140,8 +2140,11 @@ def assignWeightsToEntries(entryNestedList, weightedEntries, N = 1):
             
             # Discard the last line, unless it's the only line!
             # The last line is 'Estimated using ... for rate rule (originalTemplate)'
```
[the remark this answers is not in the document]
mjohnson541: added

rmgpy/tools/uncertainty.py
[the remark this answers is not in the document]
mjohnson541: This should be much cleaner now

rmgpy/tools/uncertainty.py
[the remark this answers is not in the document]
mjohnson541: I've removed the else

mjohnson541:
I've made all of the changes and I added a small fix for some group names that end in parentheses that solves an issue I ran into.

rmgpy/data/kinetics/family.py
```diff
@@ -2140,8 +2140,12 @@ def assignWeightsToEntries(entryNestedList, weightedEntries, N = 1):
...
-            if len(lines) == 1:
-                comment = lines[0]
+            #if from training reaction is in the first line append it to the end of the second line and skip the first line
+            if not 'Average of' in kinetics.comment:
+                if 'from training reaction' in lines[0]:
+                    comment = lines[1]+' '+lines[0]
```
mliu49: Do we need to do this, or could we just get rid of the line with 'from training reaction' since it doesn't seem like we use it?

mjohnson541: It gets looked for in retrieveOriginalEntry, which expects it to be on the same line as the primary comment.

rmgpy/tools/uncertaintyTest.py
```diff
@@ -0,0 +1,86 @@
...
+        uncertainty.compileAllSources()
+        uncertainty.assignParameterUncertainties()
+        
+        thermoUnc = uncertainty.thermoInputUncertainties
+        kineticUnc = uncertainty.kineticInputUncertainties
+        
+        logging.error(thermoUnc)
+        logging.error(kineticUnc)
+        
+        self.assertEquals(thermoUnc,[1.5, 1.5, 2.0, 1.8999999999999999, 3.1000000000000001, 1.5, 1.8999999999999999, 2.0, 2.0, 1.8999999999999999, 2.2000000000000002, 1.8999999999999999, 2.0, 1.5])
```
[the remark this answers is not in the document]
mjohnson541: Good idea!

rmgpy/tools/uncertaintyTest.py
```diff
@@ -0,0 +1,86 @@
...
+            
+        uncertainty.extractSourcesFromModel()
+        uncertainty.compileAllSources()
+        uncertainty.assignParameterUncertainties()
+        
+        thermoUnc = uncertainty.thermoInputUncertainties
+        kineticUnc = uncertainty.kineticInputUncertainties
+        
+        logging.error(thermoUnc)
+        logging.error(kineticUnc)
```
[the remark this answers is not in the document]
mjohnson541: I thought it might be useful to someone trying to debug if the test fails, but I don't have a strong opinion about it.  I'll remove it.

rmgpy/tools/uncertainty.py
```diff
@@ -341,37 +326,66 @@ def loadModel(self, chemkinPath, dictionaryPath, transportPath=None):
                                                               dictionaryPath=dictionaryPath,
                                                               transportPath=transportPath)
 
+    def retrieveSaturatedSpeciesFromList(self,species):
+        """
+        Given a radical `species`, this function retrieves the saturated species objects from a list of species objects.
```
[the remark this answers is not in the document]
mjohnson541: added

rmgpy/tools/uncertainty.py
```diff
@@ -341,37 +326,66 @@ def loadModel(self, chemkinPath, dictionaryPath, transportPath=None):
...
+        saturatedStruct.saturate()
+        for otherSpecies in self.speciesList:
+            if otherSpecies.isIsomorphic(saturatedStruct):
+                return otherSpecies, False
+        
+        #couldn't find saturated species in the model, try libraries
+        newSpc = Species(molecule=[saturatedStruct])
+        thermo = self.database.thermo.getThermoDataFromLibraries(newSpc)
+        
+        if not thermo is None:
```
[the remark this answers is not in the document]
mjohnson541: done

rmgpy/data/kinetics/family.py
mliu49: I don't think retrieveOriginalEntry uses it. If you look at line 2150 below, it splits the comment and takes the fourth token, which is the original template which was matched, which is the only argument given to retrieveOriginalEntry.

mjohnson541: Oh, you're right it grabs that information from the comment on the rule not the original kinetics comment even though it's present in both.

mjohnson541: This should be fixed now

rmgpy/data/kinetics/family.py
```diff
@@ -2162,7 +2166,9 @@ def assignWeightsToEntries(entryNestedList, weightedEntries, N = 1):
                                     re.sub(r"(?<=]) \+ (?!Average)", ",'",      # + sign between average and non-average
                                     re.sub(r"(?<!]) \+ (?!Average)", "','",     # + sign between non-averages
                                     comment)))))))
-
+            
+                evalCommentString = evalCommentString.replace("),",")',")
```
mliu49: Do you have an example of the type of comment that this deals with? Would it be possible to update the regex above instead of adding this line?

mjohnson541: I can't remember where I got the original failures, but this 'made up' comment replicates the parsing problem.  I just added parentheses around Csrad everywhere it showed up.  I think I missed one, but it shouldn't matter in the parsing.  

"""Average of [Average of [O2b;O_(Csrad)] + Average of [O_atom_triplet;O_(Csrad) + CH2_triplet;O_(Csrad)] + Average of [H_rad;O_(Csrad) + Average of [Ct_rad/Ct;O_(Csrad) from training reaction 0] + Average of [O_pri_rad;O_(Csrad) + Average of [O_rad/NonDeC;O_(Csrad) + O_rad/NonDeO;O_(Csrad)]] + Average of [Cd_pri_rad;O_(Csrad)] + Average of [CO_pri_rad;O_(Csrad)] + Average of [C_methyl;O_(Csrad) + Average of [C_rad/H2/Cs;O_Csrad + C_rad/H2/Cd;O_(Csrad) + C_rad/H2/O;O_(Csrad)] + Average of [C_rad/H/NonDeC;O_(Csrad)] + Average of [Average of [C_rad/Cs3;O_(Csrad)]]]]]
Estimated using template [Y_rad_birad_trirad_quadrad;O_(Csrad)] for rate rule [CH_quartet;O_(Csrad)]
Euclidian distance = 2.0
family: Disproportionation
"""

After processing it becomes:  
"[['O2b;O_(Csrad)'],['O_atom_triplet;O_(Csrad)','CH2_triplet;O_(Csrad)'],['H_rad;O_(Csrad),['Ct_rad/Ct;O_(Csrad) from training reaction 0'],['O_pri_rad;O_(Csrad),['O_rad/NonDeC;O_(Csrad)','O_rad/NonDeO;O_(Csrad)']],['Cd_pri_rad;O_(Csrad)'],['CO_pri_rad;O_(Csrad)'],['C_methyl;O_(Csrad),['C_rad/H2/Cs;O_Csrad','C_rad/H2/Cd;O_(Csrad)','C_rad/H2/O;O_(Csrad)'],['C_rad/H/NonDeC;O_(Csrad)'],[['C_rad/Cs3;O_(Csrad)']]]]]"

This can't be evaluated because where there should be "),'" there are "),".  It should be possible to adapt the regex, but I think it's a lot harder to do and harder to read than my fix.

mliu49:
I added some stuff to the unit test you wrote, and I fixed the regex to incorporate the bug you discovered. Let me know your thoughts.

A test that checks Cantera output files is failing for an unknown reason, which I'm working on figuring out right now. I can't reproduce the failure locally.

mjohnson541:
Your changes look good to me.

mliu49:
I rebased and removed the debugging commits. If @amarkpayne could take a final look (since I also made some changes), I think this is ready to be merged once the tests pass.

mjohnson541:
@mliu49 did you want to merge this in?

mliu49:
Sure, I was hoping for @amarkpayne to do so, but I can do it now.

mjohnson541:
@amarkpayne doesn't have merge permissions

mliu49:
Ah, sorry, I didn't realize that.

PR #1170, opened 2017-08-23
Stop calculate degeneracy errors

goldmanm:
This PR will prevent calculate degeneracy from throwing an error and will just warn the user that the degeneracy is incorrect. This is due to issues in aramatic vs. kekulized structures using different rates.

These commits were pulled from `only_core_degeneracy`, but moved here to speed up merging of bug fixes.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1577,13 +1577,14 @@ class KineticsFamily(Database):
             
         # log issues
         if len(reactions) != 1:
-            for reactant in reaction.reactants:
-                logging.error("Reactant: {0!r}".format(reactant))
-            for product in reaction.products:
-                logging.error("Product: {0!r}".format(product))
-            raise KineticsError(('Unable to calculate degeneracy for reaction {0} '
+            logging.warning(('Unable to calculate degeneracy for reaction {0} '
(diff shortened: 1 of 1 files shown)


mliu49:
1. Should there be a test for the first issue (same reactant instances)? It seems like a significant issue, so I'm surprised that it hasn't caused more problems.
2. I don't really like turning that error into a warning. Also, the issue that @amarkpayne posted about is only an issue on the only_core_degeneracy branch. His example works without any problems on master.

goldmanm:
1. Good idea to write a test. I will do that. This only occurs when calculateDegeneracy is called after a reaction has been added to the model, which is only occuring in 'only_core_degeneracy' branch.
2. I am not a huge fan of hiding errors with warnings. it does seem sad to have a long job killed due to something (like a few reactions not finding degeneracies properly), which probably won't make a difference in the final model. I guess the idea was that since we can tolerate some incorrect degeneracies, we shouldn't be stopping a job for it. Though I am kind of ambivalent on getting this part pulled in since I don't want to hide when RMG calculates degeneracy improperly.

mliu49:
The first issue has been addressed by #1276. For the second issue, I think it's better to keep it as an error for now.

PR #1171, opened 2017-08-23
make bounds for change bonds more flexible

goldmanm:
Numerical drift was causing errors when adding bonds due to lack
of flexibility in the change bond methods in the Bond class. 

This PR added flexibility to the thousandths place.

This was pulled from `only_core_degeneracy`, but moved here to speed up merging of bug fixes.

### DIFF
--- a/rmgpy/molecule/molecule.py
+++ b/rmgpy/molecule/molecule.py
@@ -601,7 +601,7 @@ class Bond(Edge):
         Update the bond as a result of applying a CHANGE_BOND action to
         increase the order by one.
         """
-        if self.order <=2: 
+        if self.order <=2.0001:
             self.order += 1
         else:
             raise gr.ActionError('Unable to increment Bond due to CHANGE_BOND action: '+\
@@ -612,7 +612,7 @@ class Bond(Edge):
         Update the bond as a result of applying a CHANGE_BOND action to
         decrease the order by one.
         """
-        if self.order >=1: 
+        if self.order >=0.9999:
             self.order -= 1
         else:
             raise gr.ActionError('Unable to decrease Bond due to CHANGE_BOND action: '+\
@@ -625,7 +625,7 @@ class Bond(Edge):
         in bond order, and can be any real number.
         """
         self.order += order
-        if self.order < 0 or self.order >3:
+        if self.order < -0.0001 or self.order >3.0001:
             raise gr.ActionError('Unable to update Bond due to CHANGE_BOND action: Invalid resulting order "{0}".'.format(self.order))
 
     def applyAction(self, action):
--- a/rmgpy/molecule/molecule.py
+++ b/rmgpy/molecule/molecule.py
@@ -565,7 +565,7 @@ class Bond(Edge):
         NOTE: we can replace the absolute value relation with math.isclose when
         we swtich to python 3.5+
         """
-        return abs(self.order - otherOrder) <= 1e-6
+        return abs(self.order - otherOrder) <= 1e-4
 
         
     def isSingle(self):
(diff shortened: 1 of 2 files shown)


rmgpy/molecule/moleculeTest.py
```diff
@@ -478,7 +478,7 @@ def testApplyActionDecrementBond(self):
             try:
                 bond.applyAction(action)
             except ActionError:
-                self.assertTrue(order0 < 1)
```
mjohnson541: So on one of my other branches I ran into a unit test issue with this check, which I fixed by changing 1 to 1.0001, but I guess if this is passing unit tests this is fine.

mliu49:
Just to confirm, we want to keep the `is[Single/Double/...]` methods within 1e-6 and only have these methods be within 1e-4?

goldmanm:
@mliu49, thanks for bringing this up. If anything, we might want to change the 10^-6  for the isDouble/Single/etc to 10^-4. I was hesitant to use 10^-6 since the error appeared nearby 2e-7. 

I am happy to add that.

mliu49:
I think 1e-4 tolerance should be sufficient.

goldmanm:
Reasons regarding the changing of symmetry number to account for enantiomers:

Separately, each enantiomer should have the thermo value that RMG currently gives it. RMG, however, lumps all the enantiomers together. Since there are now 2 different forms of the species that RMG represents as one, the entropy should be increased by Rln(2), since otherwise RMG would get the equilibrium of the following reaction wrong (where s and r are the enantiomeric forms and sr is the lumped form):

in reality:

A <-> B(s)
A <-> B(r)

in RMG:

A <-> B(sr)

For there to be the same concentration of s and r forms in RMG as in the separated representation (at equilibrium), the entropy of B(sr) needs to be Rln2 higher than that of B(s) or B(r).

RMG can accomplish this by modifying the symmetry number because accounting for symmetry and enantiomer relationship separately can be accomplished by halving the symmetry number

$Rln(2) - Rln(\sigma) = -Rln(\sigma/2)$

This explination is already partially discussed in [RMG documentation](http://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/thermo.html#symmetry-and-chirality) but RMG currently does not account for it, added by Nick.

This implementation makes the assumption that all enantiomeric centers have additive effects, which is made in GENYSIS [source](http://onlinelibrary.wiley.com/doi/10.1002/jcc.23788/full)

# examples from RMG-tests

## one chiral center

The molecule from RMG-tests, [CH2]C(C)C=C, has 1 chiral center. Labeling the centes looks like [CH2]C*(C)C=C. This molecule should have an increase of (ln2), and indeed RMG tests shows that the entropy is increased by 1.37 ~ R(ln2). [calculation](https://duckduckgo.com/?q=1.987+*ln(2)&t=canonical&ia=calculator)
[CH2]C(C)C=C

## two chiral centers

The molecule from RMG-tests, [CH2]CC([CH2])C(C)C=C, has two chiral centers. Labeling the centes looks like [CH2]CC*([CH2])C*(C)C=C. This molecule should have an increase of 2*R(ln2), and indeed RMG tests shows that the entropy increased by 2.76kcal/mol ~ 2*R(ln2). [calculation](https://duckduckgo.com/?q=2+*+1.987+*ln(2)&t=canonical&ia=calculator)

I can write an example like this for RMG unittests

# molecule on a symmetric ring.

Another case of interest is a chiral center on a symmetric ring. I will write a unittest for this too.


Bear in mind that RMG's symmetry (and non-existent) enantiomer corrections are ad-hoc heuristics and no one is going get perfect values in all cases since it isn't possible to go from a 2D graph to 3D representation infaliably even with force-field like methods.

mliu49:
Comment in wrong thread?

rmgpy/molecule/molecule.py
```diff
@@ -578,43 +578,43 @@ def isOrder(self, otherOrder):
...
-        return abs(self.order-3) <= 1e-6
+        return self.isOrder(3)
 
     def isBenzene(self):
         """
         Return ``True`` if the bond represents a benzene bond or ``False`` if
         not.
         """
-        return abs(self.order-1.5) <= 1e-6
+        return self.isOrder(4)
```
mliu49: This should be `1.5`.

goldmanm: where did that come from? fixing now.

goldmanm:
@mliu49, rebased this for merging.

PR #1081, opened 2017-07-11
Added info of reaction type to PDep network files

alongd:
Previously no info of the reaction type (library/family) was given in network.py files generated by Cantherm during an RMG run. This small PR adds such a comment to each reaction() function in PDep network files.

PR #1174, opened 2017-08-24
Fixed info of reaction type in PDep network files

alongd:
Relocated the library/family comment from Reaction to kinetics to avoid ending up with two `comment` instances in PDep network files.

Thanks @jimchu10 for [pointing this issue up](https://github.com/ReactionMechanismGenerator/RMG-Py/commit/1ae3a89b24122f7cdedf0bbfd6dd568e1da5ea9e#commitcomment-23796076).

### DIFF
--- a/rmgpy/cantherm/pdep.py
+++ b/rmgpy/cantherm/pdep.py
@@ -629,13 +629,11 @@ class PressureDependenceJob(object):
                 f.write('    products = [{0}],\n'.format(', '.join([repr(str(spec)) for spec in rxn.products])))
                 f.write('    transitionState = {0!r},\n'.format(rxn.transitionState.label))
                 if rxn.kinetics is not None:
+                    if isinstance(rxn, LibraryReaction) and 'Reaction library:' not in rxn.kinetics.comment:
+                        rxn.kinetics.comment += 'Reaction library: {0!r}'.format(rxn.library)
                     f.write('    kinetics = {0!r},\n'.format(rxn.kinetics))
                 if ts.tunneling is not None:
                     f.write('    tunneling = {0!r},\n'.format(ts.tunneling.__class__.__name__))
-                if isinstance(rxn,LibraryReaction):
-                    f.write('    comment = "Library reaction: {0!r}",\n'.format(rxn.library))
-                else:
-                    f.write('    comment = "Template reaction: {0!r}",\n'.format(rxn.family))
                 f.write(')\n\n')
             
             # Write network


rmgpy/cantherm/pdep.py
```diff
@@ -629,13 +629,15 @@ def saveInputFile(self, path):
                 f.write('    products = [{0}],\n'.format(', '.join([repr(str(spec)) for spec in rxn.products])))
                 f.write('    transitionState = {0!r},\n'.format(rxn.transitionState.label))
                 if rxn.kinetics is not None:
+                    if isinstance(rxn, LibraryReaction):
+                        if 'Reaction library:' not in rxn.kinetics.comment:
+                            rxn.kinetics.comment += 'Reaction library: {0!r}'.format(rxn.library)
+                    else:
+                        if 'Reaction family:' not in rxn.kinetics.comment:
```
mjohnson541: So with proper seed mechanisms I actually added a kinetics comment that notes the family the kinetics object came from.  So I think it's redundant to add the reaction family.  (I didn't however add the library so the library comment still needs added)

alongd: Will it also output the family if seed generation is turned off?
If we both use the same syntax, we can keep the code here since it first checks whether the comment exists or not

goldmanm: I am unsure the status of the relationship between proper seed mechanisms and all our other outputs. Will this break the seed mechanisms? 

Is this redundant in chemkin output or other output formats or just the seed mechanisms?

mjohnson541: It's added to the kinetics comment when it is made so it is entirely independent of seed mechanism generation (I just needed something to parse the family out of).  This means it will be redundant in all cases.  

Since this is appended to the kinetics comment when the input network file is saved the comment would only ever be parsed if the network file was loaded and the comment parsed.  I'm not aware of anywhere that happens.  If I am correct and that doesn't happen anywhere I wouldn't expect this to break anything.

mjohnson541: There's also no point in tagging it the same way we do for seed mechanisms because the network file format is different. So this should be fine.

mjohnson541: The comment I added is just 'family:', and it is always present if it was made from a family so your catch both won't work and is unnecessary.

alongd:
But what if seed generation is turned off? The user won't get any comment re the family.
Perhaps we should both use the same format (which we should do anyway) and catch if the comment already exists?

mjohnson541:
The addition of the comment is independent of seed mechanism generation.  It will be appended to every single kinetics object generated using the families.

mjohnson541:
If run ethane-oxidation right now, you will see that it is redundant in all cases.

alongd:
OK. Would you like to add the library comment to your code, so both are close, or do we prefer to keep it here?

mliu49:
The kinetics family comment is added when estimating the rate for a reaction, so it's unrelated to the library comment. I think it's fine to just add the library comment here.

Honestly, I don't like that we save the kinetics family to the comment at all, since the reaction has an attribute for the family. I would be really nice to consolidate our metadata reading and writing methods.

mjohnson541:
I'm not seeing family as an attribute the reaction.py file, did you mean TemplateReaction?

mliu49:
Yes, it's in the `TemplateReaction` class, but all family reactions are created as `TemplateReaction` objects.

alongd:
@mjohnson541 , done. Ready to merge.

alongd:
@goldmanm or @mjohnson541 , could you approve and merge?

alongd:
tag #1081

PR #1178, opened 2017-08-24
Fix LiquidReactor sensitivity analysis

### LINKED ISSUE #993: Sensitivity analysis doesn't work for liquid reactor
After a reaction model is generated, are there any methods to analyze the model to see which reaction step causes the discrepancy between the model and experiments?
For example, there may be a four times difference in kinetics between the simulation using the generated model and experiments. If we can figure out which step is the rate-limiting step, that step can be used to modify the model if the rates or thermodynamics is not accurate enough.

cgrambow:
This PR enables sensitivity analysis for the LiquidReactor by adding a Jacobian method, which was previously unavailable. The Jacobian method is copied from SimpleReactor and amended to ensure constant volume.

Fixes #993.


mjohnson541:
@KEHANG can we get Colin added?

KEHANG:
Colin should be able to create branch under `ReactionMechanismGenerator` now.

cgrambow:
Which `sensitivityTest.py` are you referring to? Do you mean `testSensitivity.py` in `tools`?

mjohnson541:
Oh sorry! I forgot that one uses the other convention! Yes that one!

cgrambow:
I'm pretty much done. Fixed `jacobian` for `liquidReactor` and fixed the corresponding unit test. The unit test was using a simple forward difference to estimate and compare the Jacobian, which was unstable for two of the test reactions. Instead, I derived the Jacobians for those two by hand and am using that to compare. I also noticed that simulations using `rmgpy/tools/sensitivity.py` were not working for `liquidReactor` with constant species, so I enabled that.

The only thing that I would still like to do, is to enable diffusion-limited adjustment of rate constants in `sensitivity.py`, which I tried doing by adding
```python
rmg.loadDatabase()
Species.solventData = rmg.database.solvation.getSolventData(rmg.solvent)
Species.solventName = rmg.solvent
Species.solventStructure = rmg.database.solvation.getSolventStructure(rmg.solvent)
diffusionLimiter.enable(Species.solventData, rmg.database.solvation)
```
in `rmgpy/tools/sensitivity.py` under my added check for the liquid reactor, but this is giving me
```python
TypeError: 'NoneType' object is not subscriptable
Exception TypeError: "'NoneType' object is not subscriptable" in 'rmgpy.reaction.Reaction.getRateCoefficient' ignored
```
Evidently it's having issues indexing the reactions, but I'm not sure what's happening. Ideas?

mjohnson541:
Can you post the full traceback?

cgrambow:
No traceback, that's all it prints. It prints it one time per reaction in the model, I think.

mjohnson541:
Ok, I think something isn't being initialized properly.  Can you push those changes so I can see the context?

cgrambow:
Done

mjohnson541:
Ok, I have a commit that fixes this now, but I can't push to your repository.  You should be able to add ReactionMechanismGenerator as a remote and push your local branch to it.  You'll need to make a new pull request and reference this one to that one.  I know of course you couldn't do this originally, but in general you should create pull requests from branches on ReactionMechanismGenerator rather than your own repository.

mjohnson541:
moved to #1181

Issue #1177, opened 2017-08-24
Are enantiomers really accounted for?


goldmanm:
The [RMG documentation](http://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/thermo.html#symmetry-and-chirality) claims that enantiomers are accounted for in thermo correction. I cannot find it implemented anywhere in the code.

Testing the thermo output also does not seem to account for enantiomers. Take two molecules: propyl aldehyde, which has no enantiomers, and propyl aldehyde with one hydrogen on the center carbon replaced with deuterium, which has a chiral center. The entropy of the second molecule should be Rln2 higher, according to the [documentation](http://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/thermo.html#symmetry-and-chirality), but the group additivity values are identical. 

since github can't parse spaces in URLS: paste the following into your browser to see the thermo comparison

propyl aldehyde - http://rmg.mit.edu/database/thermo/molecule/1 C u0 p0 c0 {2,S} {5,S} {6,S} {7,S}%0A2 C u0 p0 c0 {1,S} {3,S} {8,S} {9,S}%0A3 C u0 p0 c0 {2,S} {4,D} {10,S}%0A4 O u0 p2 c0 {3,D}%0A5 H u0 p0 c0 {1,S}%0A6 H u0 p0 c0 {1,S}%0A7 H u0 p0 c0 {1,S}%0A8 H u0 p0 c0 {2,S}%0A9 H u0 p0 c0 {2,S}%0A10 H u0 p0 c0 {3,S}%0A
propyl aldehyde with a deuterium - http://rmg.mit.edu/database/thermo/molecule/1 C u0 p0 c0 {2,S} {5,S} {6,S} {7,S}%0A2 C u0 p0 c0 {1,S} {3,S} {8,S} {9,S}%0A3 C u0 p0 c0 {2,S} {4,D} {10,S}%0A4 O u0 p2 c0 {3,D}%0A5 H u0 p0 c0 {1,S}%0A6 H u0 p0 c0 {1,S}%0A7 H u0 p0 c0 {1,S}%0A8 H u0 p0 c0 i2 {2,S}%0A9 H u0 p0 c0 {2,S}%0A10 H u0 p0 c0 {3,S}%0A

PR #1179, opened 2017-08-25
Account for enantiomers in entropy

goldmanm:
This PR is meant to solve issue #1177. It implements:

1. turning the symmetry number into a float (we don't compare or change symmetries, so this should be easier than bonds or degeneracy)
1. halving the symmetry number when a chiral center is found. This occurs in the calculateAtomSymmetryNumber method since finding a chiral center (an atom with four different groups), already occurs there. We half the symmetry number because a chiral center increases entropy by $R\ln(2)$, which is the same effect as halving the symmetry number (the symmetry number effects entropy by $-R\ln(\sigma)$).
1. adding a unittest to test for a change in atom symmetry number with a chiral center (we didn't have a chiral center in symmetry unittests before this one)
1. updating documentation to describe that the symmetry attribute contains chiral center modification.

Since unittests were written, all the review need to do is understand the theory written above and check for any differences in RMG-tests

### DIFF
--- a/rmgpy/molecule/symmetryTest.py
+++ b/rmgpy/molecule/symmetryTest.py
@@ -76,6 +76,76 @@ class TestMoleculeSymmetry(unittest.TestCase):
                 symmetryNumber *= calculateAtomSymmetryNumber(molecule, atom)
         self.assertEqual(symmetryNumber, 9)
     
+    def testAtomSymmetryNumberEthanewithDeuteriumTritium(self):
+        """
+        Test the Molecule.calculateAtomSymmetryNumber() on CC(D)(T)
+
+        This is meant to test whether chirality is accounted for, which
+        should half the symmetry term.
+
+        The total number is 1.5 because the methyl group contributes *3 and
+        the chiral center contributes *0.5
+        """
+        molecule = Molecule().fromAdjacencyList(
+"""
+1 C u0 p0 c0 {2,S} {3,S} {4,S} {5,S}
+2 C u0 p0 c0 {1,S} {6,S} {7,S} {8,S}
+3 D u0 p0 c0 {1,S}
+4 T u0 p0 c0 {1,S}
+5 H u0 p0 c0 {1,S}
+6 H u0 p0 c0 {2,S}
+7 H u0 p0 c0 {2,S}
+8 H u0 p0 c0 {2,S}
+
+""")
+        symmetryNumber = 1
+        for atom in molecule.atoms:
+            if not molecule.isAtomInCycle(atom):
+                symmetryNumber *= calculateAtomSymmetryNumber(molecule, atom)
+        self.assertAlmostEqual(symmetryNumber, 1.5)
+
+    def testAtomSymmetryNumberWithTwoChiralCenters(self):
+        """
+        Test the Molecule.calculateAtomSymmetryNumber() on [CH2]CC([CH2])C(C)C=C
+
+        This is meant to test whether chirality is accounted for, which
+        should half the symmetry term.
+
+        The molecule has one methyl group (*3), two CH2dot groups (*2 each), and
+        two chiral centers (*0.5), leading to a total atom symmetry number of 3
+        """
+        molecule = Molecule().fromAdjacencyList(
+"""
+multiplicity 3
+1  C u1 p0 c0 {2,S} {3,S} {4,S}
+2  H u0 p0 c0 {1,S}
+3  H u0 p0 c0 {1,S}
+4  C u0 p0 c0 {1,S} {5,S} {13,S} {14,S}
+5  C u0 p0 c0 {4,S} {6,S} {9,S} {15,S}
+6  C u1 p0 c0 {5,S} {7,S} {8,S}
+7  H u0 p0 c0 {6,S}
+8  H u0 p0 c0 {6,S}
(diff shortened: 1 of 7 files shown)


rmgpy/molecule/symmetryTest.py
```diff
@@ -76,6 +76,31 @@ def testAtomSymmetryNumberEthane(self):
                 symmetryNumber *= calculateAtomSymmetryNumber(molecule, atom)
         self.assertEqual(symmetryNumber, 9)
     
+    def testAtomSymmetryNumberEthanewithDeuteriumTritium(self):
```
KEHANG: more complicated molecules (based on RMG's imagination) might need to be tested. Can we justify the entropy changes for the molecules from eg1 in RMG-tests (https://travis-ci.org/ReactionMechanismGenerator/RMG-tests/builds/268177870?utm_source=github_status&utm_medium=notification)? They seem to differ by `R*ln2`. But I'm not fully clear about why the chiral center will half the symmetry number on top of previous symmetry calculation. Do you foresee any potential issues with kinetic estimation for a reaction with one side having chiral center?

KEHANG:
Review comments on things to be done before merging:

- [ ] fix the requested changes

- [ ] git rebase

- [ ] need evidence from other examples in RMG-tests to support this PR.

goldmanm:
# Reasons regarding the changing of symmetry number to account for enantiomers:

Separately, each enantiomer should have the thermo value that RMG currently gives it. RMG, however, lumps all the enantiomers together. Since there are now 2 different forms of the species that RMG represents as one, the entropy should be increased by Rln(2), since otherwise RMG would get the equilibrium of the following reaction wrong (where s and r are the enantiomeric forms and sr is the lumped form):

in reality:

A <-> B(s)
A <-> B(r)

in RMG:

A <-> B(sr)

For there to be the same concentration of s and r forms in RMG as in the separated representation (at equilibrium), the entropy of B(sr) needs to be Rln2 higher than that of B(s) or B(r).

RMG can accomplish this by modifying the symmetry number because accounting for symmetry and enantiomer relationship separately can be accomplished by halving the symmetry number

$Rln(2) - Rln(\sigma) = -Rln(\sigma/2)$

This explination is already partially discussed in [RMG documentation](http://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/thermo.html#symmetry-and-chirality) but RMG currently does not account for it, added by Nick.

This implementation makes the assumption that all enantiomeric centers have additive effects, which is made in GENYSIS [source](http://onlinelibrary.wiley.com/doi/10.1002/jcc.23788/full)

# examples from RMG-tests

## one chiral center

The molecule from RMG-tests, [CH2]C(C)C=C, has 1 chiral center. Labeling the centes looks like [CH2]C*(C)C=C. This molecule should have an increase of (ln2), and indeed RMG tests shows that the entropy is increased by 1.37 ~ R(ln2). [calculation](https://duckduckgo.com/?q=1.987+*ln(2)&t=canonical&ia=calculator)

## two chiral centers

The molecule from RMG-tests, [CH2]CC([CH2])C(C)C=C, has two chiral centers. Labeling the centes looks like [CH2]CC*([CH2])C*(C)C=C. This molecule should have an increase of 2*R(ln2), and indeed RMG tests shows that the entropy increased by 2.76kcal/mol ~ 2*R(ln2). [calculation](https://duckduckgo.com/?q=2+*+1.987+*ln(2)&t=canonical&ia=calculator)

I can write an example like this for RMG unittests

# molecule on a symmetric ring.

Another case of interest is a chiral center on a symmetric ring. I will write a unittest for this too.


Bear in mind that RMG's symmetry (and non-existent) enantiomer corrections are ad-hoc heuristics and no one is going get perfect values in all cases since it isn't possible to go from a 2D graph to 3D representation infaliably even with force-field like methods. 

# kinetics with enantiomers

You asked a good question. RMG does not take enantiomers into account when doing kinetics. It doesn't take symmetry into account either. In both these cases, some effect is captured by reaction path degeneracy, and some is probably not. Fixing that is a very different beast and might be a later pull request.

goldmanm:
@kehang, I added a unittest for the case of multiple chiral centers in a molecule. There are already tests (`testTotalSymmetryNumber12Dimethylbenzene` & `testTotalSymmetryNumber14Dimethylbenzene`) ensuring that RMG does not account for symmetry number on rings, so I didn't duplicate those. 

Is there anything I missed? Let me know when I should rebase for you to merge.

KEHANG:
BTW, one typo I mentioned in `thermo.rst` has not been fixed yet. Otherwise, it's good and please go ahead and run RMG-tests, and provide a table similar to the one in https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1187

goldmanm:
memory benchmark|memory new|time benchmark|time new|example
---|---|---|---|---
466.57|472.91|00:00:03:09|00:00:03:10|eg1
399.64|399.70|00:00:12:11|00:00:12:19|eg3
355.73|357.62|00:00:01:44|00:00:01:45|eg5
346.75|346.96|00:00:00:52|00:00:00:52|eg6
337.49|337.58|00:00:00:35|00:00:00:35|eg7
352.40|351.99|00:00:01:50|00:00:01:50|NC

goldmanm:
appears very similar. rebasing to master (for you to merge)

goldmanm:
@kehang, could you merge the pull request now that you accepted it?

PR #1039, opened 2017-06-03
Enhancing heteroatoms in RMG-Py

alongd:
This PR contains new sulfur AtomTypes (+ tests)
and allows database-test to make sample nitrogen atoms

This should be merged together with the corresponding [RMG-database PR](https://github.com/ReactionMechanismGenerator/RMG-database/pull/194).

PR #1180, opened 2017-08-25
Added a raise DatabaseError to thermo.py

amarkpayne:
to avoid infinite looping incase a thermo pointer is corrupt

### DIFF
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -1738,13 +1738,16 @@ class ThermoDatabase(object):
         if node is None:
             raise DatabaseError('Unable to determine thermo parameters for {0}: no data for node {1} or any of its ancestors.'.format(molecule, node0) )
 
-        data = node.data; comment = node.label
-        while isinstance(data, basestring) and data is not None:
-            for entry in database.entries.values():
+        data = node.data
+        comment = node.label
+        while isinstance(data, basestring):
+            for entry in database.entries.itervalues():
                 if entry.label == data:
                     data = entry.data
                     comment = entry.label
                     break
+            else:
+                raise DatabaseError("Node {0} points to a non-existant group called {1} in database: {2}".format(node.label, data, database.label))
         data.comment = '{0}({1})'.format(database.label, comment)
 
         # This code prints the hierarchy of the found node; useful for debugging
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -1740,7 +1740,14 @@ class ThermoDatabase(object):
 
         data = node.data
         comment = node.label
+        loop_count = 0
         while isinstance(data, basestring):
+            loop_count += 1
+            if loop_count > 100:
+                raise DatabaseError("Maximum iterations reached while following thermo group data pointers. A circular"
+                                    " reference may exist. Last node was {0} pointing to group called {1} in "
+                                    "database {2}".format(node.label, data, database.label))
+
             for entry in database.entries.itervalues():
                 if entry.label == data:
                     data = entry.data


amarkpayne:
I was looking at a long job that I was running on the RMG server, and I noticed that the model generation had not advanced over the course of 15 days. I know that @cgrambow has also had issues with RMG jobs becoming hung-up. I have no idea what is causing these two cases, but it made me remember that @alongd had experienced an issue like this, and had made the commit above to remedy the situation.

This commit was originally on #1039, but I believe that merging this commit is urgent enough to be its own pull request. This might not be the cause of the other hang-ups, but I think it is best to fix this possibility either way.

cgrambow:
I agree, merging the commit probably can't hurt, but I'm not sure it's the cause of the current hang-ups. My jobs became hung-up after about two days on the RMG server, but they have been running without issue on Pharos for four days now. It may be an issue of SLURM losing communication with jobs.

rmgpy/data/thermo.py
```diff
@@ -1745,6 +1745,7 @@ def __addGroupThermoData(self, thermoData, database, molecule, atom):
                     data = entry.data
                     comment = entry.label
                     break
+            else: raise DatabaseError("Node {0} points to a non-existant group called {1} in database: {2}".format(node.label, data, database.label))
```
mjohnson541: To my understanding the while loop above should only execute once, so shouldn't it be an if statement instead?

mjohnson541: Also there are two ways for the while loop above it to terminate:  
1) node = None 
or
2 ) node.data has something that isn't None
in case 1 it will error when it sets data = node.data

so to reach this while loop we have to start with data is not None

since the while loop only executes once we don't need to check that data is not None in the if statement

amarkpayne:
Updated with suggestions from @mjohnson541 to make the code more logical. Thanks for pointing out that the 'while' statement was unnecessary since the code has failed if the loop is run more than once.

mliu49:
I think @mjohnson541 interpreted the code incorrectly. The while loop ensures that we continue searching until we reach a node that has actual data. Here's a pseudocode interpretation:
```
while data is a string and not None
    search through all the entries 
        if we find an entry that has a label which matches data
            set data to the new entry's data
            stop searching
    (implicit: if the data is still a string, restart the while loop; if the data is actual thermo data or None, exit the while loop)
```
If none of the entries match the label that it is searching for, the while loop will never exit. It is also possible that there is a circular reference in the thermo entries, which would also prevent the while loop from exiting. This seems to rely heavily on the assumption that the database is properly written...

mjohnson541:
Oh, sorry I wasn't aware the database had references that way.

amarkpayne:
@mliu49 is the 'and not None' necessary in the while statement? If data is a string, can it be None?

mliu49:
I think this looks good. I agree that `and not None` is not necessary.

I'm wondering if it's necessary to put an iteration limit on the while loop to catch cases with circular references, or if we trust our ability to prevent such cases from being added to the database.

mjohnson541:
Given how difficult it can be to track some of these issues I certainly think that would be worthwhile.

amarkpayne:
Loop counter added as requested.

mliu49:
Could you change the error message to "Maximum iterations reached while following thermo group data pointers. A circular reference may exist. Last node was ..."

amarkpayne:
Error message updated.

amarkpayne:
All I did is rebase, should this actually require another review? Or did I mess something up?

amarkpayne:
Nevermind, Github won't let authors approve their own pull request. Makes sense

mliu49:
The settings now require an approved review before merging, but if you update the branch, prior approvals get removed. This can be annoying if you only rebased, but is intended to prevent approvals from carrying over if new changes are made.

I'll approve and merge this once Travis finishes.

PR #1181, opened 2017-08-28
Fix LiquidReactor sensitivity analysis PR2

cgrambow:
Same PR as #1178, but on `ReactionMechanismGenerator` repo.


cgrambow:
Still giving me the error...

mjohnson541:
I just fixed the pickling error are you referring to that or the same error?  The current branch passes all unit tests for me locally right now.  Did you make?

mjohnson541:
So the liquid sensitivity test is taking 45 seconds on my computer and the simple reactor sensitivity is only taking 1.16 seconds.  Can we use a smaller test case?

cgrambow:
I recompiled and no it's working.

I don't think the unit test is taking long because of the size of the model (which is hardly larger than the simple test), it's taking so long because it takes more than half a minute for RMG to set up `diffusionLimiter`. We could add a flag somewhere telling it to only turn on `diffusionLimiter` in `sensitivity.py` if explicitly requested.

mjohnson541:
Oh, I'm not really concerned about an extra 30 sec in an actual real run, just in the unit tests.  We already load diffusionLimiter in diffusionLimitedTest.py so in the future may be a good idea to combine these diffusion related tests into one test, but I don't think that belongs in this pull.

cgrambow:
Right, but because we now incorporated diffusion limiter into rmgpy/tools/sensitivity.py and the unit test uses that file, it will take an additional 30 sec for the unit test regardless of test model size if we don't turn off diffusionLimiter for the test. Are you suggesting we keep it that way for this pull request or change it somehow?

On Aug 28, 2017 1:04 PM, Matt Johnson <notifications@github.com> wrote:

Oh, I'm not really concerned about an extra 30 sec in an actual real run, just in the unit tests. We already load diffusionLimiter in diffusionLimitedTest.py so in the future may be a good idea to combine these diffusion related tests into one test, but I don't think that belongs in this pull.

mjohnson541:
Ok, I've turned it off.

rmgpy/solver/liquidTest.py
```diff
@@ -53,36 +53,56 @@ class LiquidReactorCheck(unittest.TestCase):
 
     def setUp(self):
         """
-        Here we choose a kinetic model consisting of the hydrogen abstraction reaction
-        CH4 + C2H5 <=> CH3 + C2H6.
-
+        Here we choose a kinetic model consisting of the hydrogen abstraction
```
mjohnson541: This is ok, but in general don't make these sorts of changes because while it's better to have code more compliant with PEP8 they're difficult to review (I have to go through and guarantee that the code will run the same as before), usually a waste of your time and can in other cases crowd out important changes in a commit log.  It's very good to keep your changes and the close vicinity around them PEP8 compliant, but don't go overboard with it.

mliu49: To add on, updates to code style are always appreciated, but if you plan on making a lot of style changes, it is helpful to put them in a separate commit to make reviewing easier (i.e. commit with style changes only, then commit with functional changes).

cgrambow: Ok, thanks for the advice!

rmgpy/solver/liquidTest.py
```diff
@@ -144,10 +141,9 @@ def testComputeFlux(self):
 
         # Convert the solution vectors to numpy arrays
         t = numpy.array(t, numpy.float64)
-        y = numpy.array(y, numpy.float64)
+        # y = numpy.array(y, numpy.float64)
```
mjohnson541: can you remove this commented out line?

rmgpy/tools/loader.py
```diff
@@ -180,4 +189,3 @@ def loadRMGJavaJob(inputFile, chemkinFile=None, speciesDict=None, useChemkinName
         species.molecule[0].draw(str(path))
     
     return rmg
-
```
mjohnson541: Can you remove the white space changes here?  These tend to clog our commit log and make it more difficult to determine what changes were made in a given commit.

rmgpy/tools/sensitivity.py
```diff
@@ -73,8 +74,6 @@ def plotSensitivity(outputDirectory, reactionSystemIndex, sensitiveSpeciesList,
         ReactionSensitivityPlot(csvFile=csvFile, numReactions=number).barplot(reactionPlotFile)
         ThermoSensitivityPlot(csvFile=csvFile, numSpecies=number).barplot(thermoPlotFile)
 
-
```
mjohnson541: also here

cgrambow: You mean make another commit and add the blank lines back in? Wouldn't that double clog the commit log?

rmgpy/tools/loader.py
```diff
@@ -36,10 +36,8 @@
 import os.path
```
mjohnson541: Can you add a constant species to the liquid sensitivity test to improve our coverage?

cgrambow:
Updated history. Should be good to go.

rmgpy/tools/sensitivity.py
```diff
@@ -77,7 +76,7 @@ def plotSensitivity(outputDirectory, reactionSystemIndex, sensitiveSpeciesList,
 
 
 
-def simulate(rmg):
+def simulate(rmg, diffusionLimited=True):
```
mjohnson541: Was this bit added in by accident?  I think this stuff belongs with the last commit.

rmgpy/tools/loader.py
```diff
@@ -36,8 +36,8 @@
...
```
mjohnson541: Can you be a bit more detailed in the message for this commit?

rmgpy/tools/sensitivity.py
cgrambow: Indeed it was. Not sure how that got in there.

cgrambow:
Pushed again

mjohnson541:
Ok, I need you to update your current master and then rebase this branch onto master, I believe it should be a clean rebase with no merge conflicts.

cgrambow:
Done. I also added a flag in `scripts/sensitivity.py` to enable turning off of diffusion-limited rates.

mjohnson541:
Ok, I've run some of my own tests and this looks good.  Anything else before I merge this?

cgrambow:
Looks good to me.

PR #1186, opened 2017-08-29
Updated BAC for CBS-QB3 according to Paraskevas2013

aelong:
I updated CBS-QB3 BAC corrections according to the work of Paraskevas et al.
http://onlinelibrary.wiley.com/doi/10.1002/chem.201301381/full
This update was recommended by Professor Sabbe at Ghent University.

### DIFF
--- a/rmgpy/cantherm/statmech.py
+++ b/rmgpy/cantherm/statmech.py
@@ -699,16 +699,13 @@ def applyEnergyCorrections(E0, modelChemistry, atoms, bonds):
             'O-H':  0.07, 'C-O': 0.25, 'C=O': -0.03, 'O-O': 0.26, 'C-N': -0.20,
             'C=N': -0.30, 'C#N': -1.33, 'N-O': 1.01, 'N_O': -0.03, 'N=O': -0.26,
             'N-H':  0.06, 'N-N': -0.23, 'N=N': -0.37, 'N#N': -0.64,}
-    
-    # BAC corrections from Table IX in http://dx.doi.org/10.1063/1.477794 for CBS-Q method
-    # H-Cl correction from CBS-QB3 enthalpy difference with Gurvich 1989, HF298=-92.31 kJ
-    # 'N=O', 'N-N', 'N=N', 'N-O' corrections taken from Table 2 in R. Ashcraft, S. Raman, W.H. Green J. Phys. Chem. B, Vol. 111, No. 41, 2007; doi: 10.1021/jp073539t
-    elif modelChemistry == 'CBS-QB3':
-        bondEnergies = { 'C-H': -0.11, 'C-C': -0.3, 'C=C': -0.08, 'C#C': -0.64,
-            'O-H': 0.02, 'C-O': 0.33, 'C=O': 0.55, 'N#N': -2.0, 'O=O': -0.2,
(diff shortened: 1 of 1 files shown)


rmgpy/cantherm/statmech.py
```diff
@@ -704,8 +704,15 @@ def applyEnergyCorrections(E0, modelChemistry, atoms, bonds):
     # H-Cl correction from CBS-QB3 enthalpy difference with Gurvich 1989, HF298=-92.31 kJ
     # 'N=O', 'N-N', 'N=N', 'N-O' corrections taken from Table 2 in R. Ashcraft, S. Raman, W.H. Green J. Phys. Chem. B, Vol. 111, No. 41, 2007; doi: 10.1021/jp073539t
     elif modelChemistry == 'CBS-QB3':
-        bondEnergies = { 'C-H': -0.11, 'C-C': -0.3, 'C=C': -0.08, 'C#C': -0.64,
-            'O-H': 0.02, 'C-O': 0.33, 'C=O': 0.55, 'N#N': -2.0, 'O=O': -0.2,
+        # bondEnergies = { 'C-H': -0.11, 'C-C': -0.3, 'C=C': -0.08, 'C#C': -0.64,
```
amarkpayne: Is there a reason to keep the old bond corrections in the comments? The values will still be here in the git history if anyone needs them. If not then just get rid of the commented out old code, but otherwise this looks good.

aelong: I left them because I didn't update every value so by leaving the old its easy to see the few that changed, and also by leaving them its easiest to switch between either set if for example someone is curious as to the impact for their system. I don't have strong feelings one way or the other, just thought leaving them was slightly more beneficial for the stated reasons.

mjohnson541:
@KEHANG can we get Alan added?

rmgpy/cantherm/statmech.py
mjohnson541: If we're sure this is an improvement we should remove the commented ones, if not we shouldn't make changes.  If there are compelling reasons to use one or the other for different cases we should implement an option that's controllable through the input files.  I understand why you find it useful to leave them in, but having commented out lines of code is largely considered unprofessional in major coding projects.

KEHANG: Leaving such comments is usually not recommended in RMG master code. I think we update the numbers in Cantherm when we've done tests and are sure the new number are better than the old ones. So in this case if Paraschevas has done the tests and we convince ourselves, decide to trust it, we should eliminate old numbers to avoid confusion. Otherwise, imagine we have multiple places with such binary options for user to choose from, it's gonna be troublesome for them and they might not have clear criterion to select which option to use. I guess the key question is how confident are we about the new numbers?

aelong: I don't think there's any reason to have the user choose between the two in the input file. As far as the new ones being better, the Paraskevas correlation is 15 years more recent, actually for CBS-QB3, and I believe uses a larger set of molecules to fit the corrections. The updated values give me an improved H298 for cyclopentadieneyl as well.

alongd:
@aelong , could you rebase and force push?

goldmanm:
closed since #1272 is a duplicate (and is already merged)

PR #1187, opened 2017-08-30
Add carbene species constraint

mliu49:
Adds two new species constraint options:
- `maxSingletCarbenes`: limits the number of carbon atoms with lone pairs in a molecule
- `maxCarbeneRadicals`: limits the number of unpaired electrons a molecule with a carbene can have

This eliminates the need for the global forbidden groups added in ReactionMechanismGenerator/RMG-database@2e18d62346778526ba25ca3ae10543ac926fc581, and addresses the speed concerns raised in #1079. Since the intention of the forbidden groups was to prevent molecules with 2 carbenes or 1 carbene and 1 radical from being formed, using species constraints provides a more robust filter than the forbidden groups.

__Note:__ I set the default for `maxSingletCarbenes` to 1 and the default for `maxCarbeneRadicals` to 0, which makes them the only species constraints with a default value. The reasoning is that this accomplishes the same thing as having the forbidden groups without any extra input from the user. Any thoughts on whether or not this is a good idea would be appreciated.

The following table compares results of eg1 for different RMG-database versions on current master of RMG-Py (f3550f4a9ee9b5a7b66b5a86dbec822bfa06e9e5), with the last column being on this branch of RMG-Py.

 -- | Before Aromatics 1 PR | After Aromatics 1 PR | Revert forbidden groups | Add species constraints
----- | ----- | ----- | ----- | -----
Total time (s) | 586 | 1386 | 765 | 705
isMoleculeForbidden (s) | 61 | 704 | 69 | 67
Core species | 32 | 37 | 37 | 37
Core rxns | 164 | 206 | 206 | 206
Edge species | 2055 | 2449 | 2646 | 2325
Edge rxns | 6138 | 7487 | 8049 | 7216

Conclusions which can be drawn from these results:
- The aromatics PR resulted in an increase in the number of core species and reactions
- The added forbidden groups contributed over 10 min to total runtime
- The forbidden groups successfully forbid 197 edge species and 562 edge reactions
- Using species constraints removed 321 edge species and 833 edge reactions
- The extra species and reactions removed were molecules which were too large to be caught by the forbidden groups
- The time difference between reverting forbidden groups and adding species constraints is probably not statistically significant, but could simply be a result of having a smaller edge

### DIFF
--- a/examples/rmg/commented/input.py
+++ b/examples/rmg/commented/input.py
@@ -1,42 +1,42 @@
 # Data sources
 database(
-	#overrides RMG thermo calculation of RMG with these values.
-	#libraries found at http://rmg.mit.edu/database/thermo/libraries/
-	#if species exist in multiple libraries, the earlier libraries overwrite the 
-	#previous values
-    thermoLibraries = ['BurkeH2O2','primaryThermoLibrary','DFT_QCI_thermo','CBS_QB3_1dHR'],
-	#overrides RMG kinetics estimation if needed in the core of RMG. 
-	#list of libraries found at http://rmg.mit.edu/database/kinetics/libraries/
-	#libraries can be input as either a string or tuple of form ('library_name',True/False) 
-     #where a `True` indicates that all unused reactions will be automatically added 
-	#to the chemkin file at the end of the simulation. Placing just string values
-     #defaults the tuple to `False`. The string input is sufficient in almost
-     #all situations
-    reactionLibraries = [('C3', False)],
-	#seed mechanisms are reactionLibraries that are forced into the initial mechanism 
-	#in addition to species listed in this input file.  
-	#This is helpful for reducing run time for species you know will appear in 
-	#the mechanism.  
-    seedMechanisms = ['BurkeH2O2inN2','ERC-FoundationFuelv0.9'],
-	#this is normally not changed in general RMG runs.  Usually used for testing with 
-	#outside kinetics databases
-    kineticsDepositories = 'default', 
-	#lists specific families used to generate the model. 'default' uses a list of 
-	#families from RMG-Database/input/families/recommended.py
-	#a visual list of families is available in PDF form at RMG-database/families
-    kineticsFamilies = 'default',
-	#specifies how RMG calculates rates.  currently, the only option is 'rate rules'
-    kineticsEstimator = 'rate rules',
+    # overrides RMG thermo calculation of RMG with these values.
+    # libraries found at http://rmg.mit.edu/database/thermo/libraries/
+    # if species exist in multiple libraries, the earlier libraries overwrite the
+    # previous values
+    thermoLibraries=['BurkeH2O2', 'primaryThermoLibrary', 'DFT_QCI_thermo', 'CBS_QB3_1dHR'],
+    # overrides RMG kinetics estimation if needed in the core of RMG.
+    # list of libraries found at http://rmg.mit.edu/database/kinetics/libraries/
+    # libraries can be input as either a string or tuple of form ('library_name',True/False)
+    # where a `True` indicates that all unused reactions will be automatically added
+    # to the chemkin file at the end of the simulation. Placing just string values
+    # defaults the tuple to `False`. The string input is sufficient in almost
+    # all situations
+    reactionLibraries=[('C3', False)],
+    # seed mechanisms are reactionLibraries that are forced into the initial mechanism
+    # in addition to species listed in this input file.
+    # This is helpful for reducing run time for species you know will appear in
+    # the mechanism.
+    seedMechanisms=['BurkeH2O2inN2', 'ERC-FoundationFuelv0.9'],
+    # this is normally not changed in general RMG runs.  Usually used for testing with
+    # outside kinetics databases
+    kineticsDepositories='default',
+    # lists specific families used to generate the model. 'default' uses a list of
+    # families from RMG-Database/input/families/recommended.py
+    # a visual list of families is available in PDF form at RMG-database/families
+    kineticsFamilies='default',
+    # specifies how RMG calculates rates.  currently, the only option is 'rate rules'
+    kineticsEstimator='rate rules',
 )
 
 # List of species
-#list initial and expected species below to automatically put them into the core mechanism.  
-#'structure' can utilize method of SMILES("put_SMILES_here"), 
-#adjacencyList("""put_adj_list_here"""), or InChI("put_InChI_here")
-#for molecular oxygen, use the smiles string [O][O] so the triplet form is used
+# list initial and expected species below to automatically put them into the core mechanism.
+# 'structure' can utilize method of SMILES("put_SMILES_here"),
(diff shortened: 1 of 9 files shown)


mliu49:
Unit tests for `constraints.py` to be added.

mliu49:
ReactionMechanismGenerator/RMG-database@2e18d62346778526ba25ca3ae10543ac926fc581 should be reverted along with this pull request.

KEHANG:
Side note (not necessarily need to be done in this PR): `isMoleculeForbidden()` is currently done by looping over all the forbidden groups one by one, which obviously is slow when having large core/edge and lots of forbidden structures. Parallelism would be suitable in this case. But if we could come up with a tree structure to organize forbidden groups would be another option.

rmgpy/molecule/moleculeTest.py
```diff
@@ -1091,6 +1091,23 @@ def testRadicalCH2CH2CH2(self):
...
+        self.assertEqual(mol.getRadicalCount(), 0)
+        self.assertEqual(mol.getSingletCarbeneCount(), 1)
+
+    def testTripletCarbene(self):
+        """Test radical and carbene count on triplet carbene."""
+        mol = Molecule().fromAdjacencyList('1 C u2 p0', saturateH=True)
+        self.assertEqual(mol.getRadicalCount(), 2)
+        self.assertEqual(mol.getSingletCarbeneCount(), 0)
+
+    def testSingletCarbon(self):
```
KEHANG: the test examples need a bit more complicated. Ideally test example should come from ReactionMechanismGenerator/RMG-database@2e18d62346778526ba25ca3ae10543ac926fc581 since you are trying to replace it. Please pick some representatives from there.

examples/rmg/commented/input.py
```diff
@@ -223,6 +223,11 @@
     # maximumHeavyAtoms=20,
     # maximum radicals on a molecule
     maximumRadicalElectrons=1,
+    # maximum number of singlet carbenes (lone pair on a carbon atom) in a molecule
+    maximumSingletCarbenes=1,
+    # maximum number of radicals on a molecule with a singlet carbene
+    # should be lower than maximumRadicalElectrons in order to have an effect
```
KEHANG: change "lower" to "lower or equal"

KEHANG:
Review comments on things to be done before merging:

- [ ] fix requested changes

- [ ] git rebase and git squash certain related commits

- [ ] pull in ReactionMechanismGenerator/RMG-database@2e18d62346778526ba25ca3ae10543ac926fc581

- [ ] check and paste RMG-tests results on speed up to the conversation

rmgpy/molecule/moleculeTest.py
mliu49: What do you mean by more complicated? A molecule with 2 carbenes?

examples/rmg/commented/input.py
mliu49: If it's equal, then it still won't have any effect, since it will enforce the same limitation as `maximumRadicalElectrons`, which returns first.

mliu49:
A pull request has been made for reverting the forbidden group additions to RMG-database. The [RMG-tests build](https://travis-ci.org/ReactionMechanismGenerator/RMG-tests/builds/271761695) for that branch is up, but still fails to complete the second example, presumably because the benchmark version (with the forbidden groups) is taking to long to finish. The only job to finish (eg1) took about half as long after removing the forbidden groups, as expected. It also had many additional species and reactions, which this PR should remove.

rmgpy/molecule/moleculeTest.py
KEHANG: You are trying to replace the 18 forbidden groups with this PR. So I think the sample molecules from the 18 groups should be tested and more representative.

examples/rmg/commented/input.py
KEHANG: what I'm trying to say is that it's confusing to have comment written as "lower than" but you have example with "equal"

rmgpy/molecule/moleculeTest.py
mliu49: I don't think that's necessary. First of all, these tests are just for `getSingletCarbeneCount()`, which counts the number of carbene carbons in a molecule. I don't think the test molecules here need to be more complicated to test that RMG can count properly.

I also wrote tests for the actual species constraint module, to make sure that it catches what we expect it to catch. There I tested sample molecules with carbenes/radicals on adjacent carbons. Since the 18 forbidden groups were just trying to ban carbenes/radicals on the same molecule, I think this should be sufficient. Adding more carbons won't change the effect of the test.

examples/rmg/commented/input.py
mliu49: I'll change the value to 0 then.

KEHANG:
The code looks good now. Can you run a local RMG-tests with testing version being `carbene_constraint -rmgpy` and `revert_forbidden -db`, and provide a table of speed-up comparison?

mliu49:
Here are the RMG-tests results.

 -- | Benchmark Time | Tested Time | Benchmark Mem | Tested Mem
----- | ----- | ----- | ----- | -----
eg1 | 3:08 | 1:36 | 449.96 | 468.78
eg3 | 12:13 | 1:51 | 399.68 | 399.90
eg5 | 1:44 | 1:04 | 357.26 | 351.57
eg6 | 0:52 | 0:48 | 344.66 | 345.05

There were fewer speices and reactions for eg1 in the testing version, mostly due to additional carbenes which were caught by the species constraints.

The test suite stopped at eg7 due to [CH] from GRI-Mech3.0 being forbidden by the species constraints. Because of the way the `maxCarbeneRadicals` constraint is implemented, a single carbon atom with a radical electron and a lone pair would be caught by the default settings. I don't think we want this, so the constraint probably needs to be reformulated.

```
Benchmark vs. Testing Version Summary
=========================================
benchmark version of RMG
c5868701bb437c8e4c6a3cb0c520eb33e15ed03f
Fri Sep 1 23:22:39 2017 -0400

testing version of RMG
bb2e2c9b843da3f1520873812426709fe300e0e2
Tue Sep 5 11:28:36 2017 -0400

benchmark version of RMG-database
05619c48fe836ddd97651add611546447008e9fe
Fri Sep 1 09:38:24 2017 -0400

testing version of RMG-database
8b97642d5d26b212c5e4728f2d230856f955551d
Mon Sep 4 12:19:15 2017 -0400
=========================================
```

mliu49:
After talking with @zjburas, I think we should just add the `allowed=['reaction libraries]` parameter to the problematic examples. We still want to forbid RMG from generating these species since they will probably have very bad thermo estimates.

KEHANG:
Thanks for the update on RMG-tests. Will fix them soon. In terms of next step for this PR, please git rebase and squash accordingly.

KEHANG:
Looks good to me now. I'm going to merge https://github.com/ReactionMechanismGenerator/RMG-database/pull/211 and this one soon.

PR #1188, opened 2017-08-30
Weights

mjohnson541:
This adds the option to add weights for dynamics criterion calculations.  Three different weights are allowed:  
1)  radical change weights:  You set some tolerance t in the input file and for a reaction that changes the number of radicals by dr the weight W is calculated W = exp(t * abs(dr)).  This weight is multiplied by the dynamics number of that reaction during calculations increasing the weight of reactions that change the number of radicals.  

2) Species Specific weights:  You set a dictionary in the input file mapping species labels to a weight.  This weight is multiplied by the contribution of that species in any dynamics number calculation i.e. W_spc,i*abs(Ln(PI_Ac_spc,i))+ W_spc,i+1*abs(Ln(PI_Ac_spc,i+1)) + ...

3) Rate ratio weights:  set useFluxRatioInWeights to True in the input file.  This causes it to use the rate ratio of a given species at a given time as a species weight.  This weight behaves the same and is multiplied by any weight from 2.  

based from #1082


mjohnson541:
If weights are implemented they won't be implemented this way.

PR #1189, opened 2017-09-01
clean up surface management

mjohnson541:
This pull request cleans up surface management by creating three new functions in model.py for surface management:
addNewSurfaceObjects
adjustSurface
clearSurfaceAdjustments

these are now called in main.py and surface management operations have been moved to these functions.

### DIFF
--- a/rmgpy/rmg/model.py
+++ b/rmgpy/rmg/model.py
@@ -45,7 +45,7 @@ from rmgpy.constraints import failsSpeciesConstraints
 from rmgpy.quantity import Quantity
 import rmgpy.species
 from rmgpy.thermo.thermoengine import submit
-
+from rmgpy.reaction import Reaction
 from rmgpy.exceptions import ForbiddenStructureException
 from rmgpy.data.kinetics.depository import DepositoryReaction
 from rmgpy.data.kinetics.family import KineticsFamily, TemplateReaction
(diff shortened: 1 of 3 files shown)


mjohnson541:
Ok, added a unittest for addNewSurfaceObjects.

rmgpy/rmg/model.py
```diff
@@ -210,9 +210,11 @@ def __init__(self, core=None, edge=None):
             self.edge = ReactionModel()
         else:
             self.edge = edge
-        
-        self.surfaceSpecies = []
-        self.surfaceReactions = []
+        if surface is None:
```
goldmanm: I am not super knowledgeable about surface, but should we be creating a surface object at all if the user turns the surface off? This might be good to note for a overhaul of model.py

mjohnson541: Turning the surface "off" right now is unrelated to the presence of the surface object.  Surface calculations are turned "on" and "off" based on whether or not a surface related tolerance was set not based on the presence of a surface object.  If the surface is "off" the surface object will just always be empty.  We also don't pass model objects directly to reactionSystem objects so without a larger overhaul we still have to rely on the tolerances to tell if it should be "off" or "on".  I suppose if you're expressing concern about instantiating an object we don't use I can leave it as none and add if-statements to manage it, but I don't think it would improve speed or readability much.

mjohnson541:
Ok, I just rebased.

mjohnson541:
@goldmanm can you reapprove and merge this?  The only change since you approved it last was a rebase.

PR #1190, opened 2017-09-01
only one Species class

goldmanm:
We previously had two species classes, one inheriting from the other, which was a source of confusion. This PR merges both classes into species.py and moves the class attributes of the other Species class to other areas of the code.

### DIFF
--- a/rmgpy/thermo/thermoengine.py
+++ b/rmgpy/thermo/thermoengine.py
@@ -37,7 +37,7 @@ from rmgpy.statmech import Conformer
 from rmgpy.thermo import Wilhoit, NASA, ThermoData
 import rmgpy.data.rmg
 
-def processThermoData(spc, thermo0, thermoClass=NASA):
+def processThermoData(spc, thermo0, thermoClass=NASA, solventName = ''):
     """
     Converts via Wilhoit into required `thermoClass` and sets `E0`.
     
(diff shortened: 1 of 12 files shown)


rmgpy/thermo/thermoengine.py
```diff
@@ -36,8 +36,9 @@
 from rmgpy.statmech import Conformer
 from rmgpy.thermo import Wilhoit, NASA, ThermoData
 import rmgpy.data.rmg
+from rmgpy.exceptions import DatabaseError
```
mliu49: This import is unused.

rmgpy/thermo/thermoengine.py
```diff
@@ -176,4 +183,8 @@ def submit(spc):
     the result.
 
     """
-    spc.thermo = submit_(evaluator, spc)
+    if solventName:
+        evaluator_wrapper = lambda x: evaluator(x,solventName)
+        spc.thermo = submit_(evaluator_wrapper, spc)
```
mliu49: Can you not feed `solventName` directly to submit?

goldmanm:
@mliu49, Thanks for the three great suggestions. Your two code improvements worked seamlessly and made the code more readable (pushed here)

I am also not super familiar with using liquid reactors. I tried to refactor everything so that the functionality is the same. When running `minimal_liq` test for branches `master` and `one_species`, The same species, thermo, kinetics, and reactions were found for both cases at a model size of 30 species (which is when I stopped the simulation). It is run locally, and I can show you the results.

Issue #903, opened 2017-02-16
RMG quantity objects return unexpected results when compared


goldmanm:
RMG quantity objects (at least ScalarQuantity specifically) don't compare values to each other or have valid addition/subtraction operations. A simple example:

```
from rmgpy.quantity import ScalarQuantity
ScalarQuantity(1.,'s') > ScalarQuantity(0.5,'s')
```

Intuitively, we would expect the comparison to always be True, since 1 second is more than 0.5 seconds. However, RMG currently returns either True or False for this depending on other attributes of the object (like memory location, etc.).

goldmanm:
For me, this issue leads to less robust programming. For example, I currently compare `quantity.value > 1`, which may cause an issue if later the the quantity classes were modified to store time values in milliseconds instead of seconds (where 0.5 seconds would then return true, since it is stored as 500).

KEHANG:
I guess using `quantity.value_si > 1` for comparison is recomended.

rwest:
The workaround for current implementation is to always compare the .value_si attributes, but I think a modification to the class so that just comparing the objects does this for you (and checks the dimensions are the same!) would reduce chances of people making mistakes. It's not unreasonable to expect "a > b" to work! (So I'm not sure this should have been closed yet)

Issue #1135, opened 2017-08-07
get coverage to measure cythonized code


goldmanm:
Currently, coverage is not measured for cythonized code. Now that we use codecov, it may be beneficial to measure cythonized code. We just need to add a few lines to .travis.yml. The downside is that it may significantly slowdown the testing. 

More information on this measuring coverage of cythonized code: #785 .

PR #1191, opened 2017-09-01
Add Basic Operations to Scalar Quantities

amarkpayne:
Addresses some of the issues in #903

### DIFF
--- a/rmgpy/quantity.py
+++ b/rmgpy/quantity.py
@@ -38,6 +38,8 @@ import quantities as pq
 
 import rmgpy.constants as constants
 from rmgpy.exceptions import QuantityError
+import re
+from logging import warn
 
 ################################################################################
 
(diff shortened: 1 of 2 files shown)


amarkpayne:
This PR is now ready for review. In particular if anyone who is familiar with Cython could look over this PR and recommend improvements that would be very much appreciated. 

This PR adds the following features, as discussed at the Code Organization Hackathon:

- Ability to add and subtract scalar quantities. These quantities must have the same dimensions or else an error is thrown. For A + B or A - B, the returned scalar quantity has units of A since it came first.

- Ability to multiply and divide scalar quantities together. In this operation, units are compounded together, and basic simplification is performed (simplifies m**3/m**2 to m, but won't simplify N*m or kcal/J).

- Uncertainties are propogated appropriately for addition, subtraction, division, and multiplication.

- Scalar Quantities can be compared using <,>, ==, !=, =<, =>

- Added method 'has_same_dimensions_as' to check if two scalar quantities have the same dimensions.


Final note: The unit tests added for each of these methods is not reflected in codecov, since we currently are not tracking cythonized code (see #1135).

mjohnson541:
Sorry, I'm a bit backlogged with the aromatic pull requests, but I'll try to get to this one in the next two weeks.

mjohnson541:
My understanding of the new contributor guidelines is that all RMG-Py PR's must not reduce code coverage.  Since this PR reduces code coverage it needs additional unittests.

amarkpayne:
The unit tests added in this PR don't get included in the coverage report since it all deals with cythonized code. Sorry, I should have been more clear.

I checked by inspection to make sure that any code I added was unit tested, but if I need to write another unit test for some code I missed just let me know.

goldmanm:
This looks like an awesome PR. Just a few things we discussed:

1. Fix the multiplicative uncertainty handling
2. Have some way to handle when incompatibilities are created in uncertainty (possibly warn user and make the value `np.nan`)

PR #1193, opened 2017-09-06
estimate highly unsaturated bicyclic correction via heuristic method

KEHANG:
Bicyclic decomposition method works well when our database has comprehensive set of bicyclics and monocyclics. However, this assumption is not exactly true; database has a comprehensice set of saturated bicyclics while only a few of unsaturated ones. As one can imagine, data coverage is quite sparse there since one single bicyclic frame can have multiple ways of placing an unsaturated bond in the rings. It also seems not practical to calculate those unsaturated ones.

We had this idea of estimating unsaturated bicyclic correction from saturated bicyclics and monocyclics (see an example below)

bicyclic_estimation_formula

This PR adds implemented this idea, which serves as an additional layer to bicyclic decomposition method. I tested the algorithm on 195 unsaturated bicyclics (the dataset is crafted by @zjburas and @adeeljamal, called `zb_bicyclics_195_table` hosted in `Thermo Central Database`). The results below shows little impact on saturated cyclics (the first 6 tables) but great improvements on unsaturated cyclics (the last table), which is exactly what we want it to behave.

 thermo estimation error (kcal/mol) | master-rmgpy & master-db | bicyclic_formula-rmgpy & Aromatics_updates_2-db 
----- | ----- | -----
small_cyclic_table | 3.39 | 3.42 
large_linear_polycyclic_table | 9.10 | 9.05 
large_fused_polycyclic_table | 9.56 | 9.54
small_O_only_polycyclic_table | 6.93 | 6.95
large_linear_O_only_polycyclic_table | 10.31 | 10.31
large_fused_O_only_polycyclic_table | 5.33 | 5.30
zb_bicyclics_195_table | 62.89 | 9.59

This PR should be merged with the database PR (https://github.com/ReactionMechanismGenerator/RMG-database/pull/206) made by @zjburas.

### DIFF
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -1604,15 +1604,24 @@ class ThermoDatabase(object):
         be applied.
         """
         # look up polycylic tree directly
-        matched_group_thermodata, _, isPartialMatch = self.__addRingCorrectionThermoDataFromTree(None, self.groups['polycyclic'], molecule, polyring)
+        matched_group_thermodata, matched_group, isPartialMatch = self.__addRingCorrectionThermoDataFromTree(None, self.groups['polycyclic'], molecule, polyring)
         
         # if partial match (non-H atoms number same between 
         # polycylic ring in molecule and match group)
         # otherwise, apply heuristic algorithm
         if not isPartialMatch:
-            thermoData = addThermoData(thermoData, matched_group_thermodata, groupAdditivity=True, verbose=True)
-            # By setting verbose=True, we turn on the comments of polycyclic correction to pass the unittest.
-            # Typically this comment is very short and also very helpful to check if the ring correction is calculated correctly.
+            if isBicyclic(polyring) and matched_group.label in self.groups['polycyclic'].genericNodes:
+                # apply secondary decompostion formula
+                # to get a estimated_group_thermodata
+                estimated_bicyclic_thermodata = self.getBicyclicCorrectionThermoDataFromHeuristic(polyring)
+                if not estimated_bicyclic_thermodata:
+                    estimated_bicyclic_thermodata = matched_group_thermodata
+                thermoData = addThermoData(thermoData, estimated_bicyclic_thermodata, groupAdditivity=True, verbose=True)
+            else:
+                # keep matched_group_thermodata as is
+                thermoData = addThermoData(thermoData, matched_group_thermodata, groupAdditivity=True, verbose=True)
+                # By setting verbose=True, we turn on the comments of polycyclic correction to pass the unittest.
+                # Typically this comment is very short and also very helpful to check if the ring correction is calculated correctly.
         else:
             self.__addPolyRingCorrectionThermoDataFromHeuristic(thermoData, polyring)
             
@@ -1633,8 +1642,19 @@ class ThermoDatabase(object):
         
         # loop over 2-ring cores
         for bicyclic in bicyclicsMergedFromRingPair:
-            self.__addRingCorrectionThermoDataFromTree(thermoData, 
+            matched_group_thermodata, matched_group, _ = self.__addRingCorrectionThermoDataFromTree(None, 
                 self.groups['polycyclic'], bicyclic, bicyclic.atoms)
+
+            if matched_group.label in self.groups['polycyclic'].genericNodes:
+                # apply secondary decompostion formula
+                # to get a estimated_group_thermodata
+                estimated_bicyclic_thermodata = self.getBicyclicCorrectionThermoDataFromHeuristic(bicyclic.atoms)
+                if not estimated_bicyclic_thermodata:
+                    estimated_bicyclic_thermodata = matched_group_thermodata 
+                thermoData = addThermoData(thermoData, estimated_bicyclic_thermodata, groupAdditivity=True)
+            else:
+                # keep matched_group_thermodata as is
+                thermoData = addThermoData(thermoData, matched_group_thermodata, groupAdditivity=True)
 
         # loop over 1-ring 
         for singleRingTuple, occurance in ringOccurancesDict.iteritems():
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -496,6 +496,31 @@ def splitBicyclicIntoSingleRings(bicyclic_submol):
     return [convertRingToSubMolecule(SSSR[0])[0], 
                 convertRingToSubMolecule(SSSR[1])[0]]
 
+def saturateRingBonds(ring_submol):
+    """
+    Given a ring submolelcule (`Molecule`), makes a deep copy and converts non-single bonds 
+    into single bonds, returns a new saturated submolecule (`Molecule`)
+    """
+    atomsMapping = {}
+    for atom in ring_submol.atoms:
+        if atom not in atomsMapping:
+            atomsMapping[atom] = atom.copy()
+
+    mol0 = Molecule(atoms=atomsMapping.values())
+
+    alreadySaturated = True
+    for atom in ring_submol.atoms:
+        for bondedAtom, bond in atom.edges.iteritems():
+            if bondedAtom in ring_submol.atoms:
+                if bond.order > 1.0: alreadySaturated = False
+                if not mol0.hasBond(atomsMapping[atom],atomsMapping[bondedAtom]):
+                    mol0.addBond(Bond(atomsMapping[atom],atomsMapping[bondedAtom],order=1.0))
+    
+    mol0.updateAtomTypes()
+    mol0.updateMultiplicity()
+    mol0.updateConnectivityValues()
+    return mol0, alreadySaturated
+
 ################################################################################
 
 class ThermoDepository(Database):
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -485,6 +485,17 @@ def bicyclicDecompositionForPolyring(polyring):
         bicyclicsMergedFromRingPair.append(mergedRing)
 
     return bicyclicsMergedFromRingPair, ringOccurancesDict
+
+def splitBicyclicIntoSingleRings(bicyclic_submol):
+    """
+    Splits a given bicyclic submolecule into two individual single 
+    ring submolecules (a list of `Molecule`s ).
+    """
+    SSSR = bicyclic_submol.getDeterministicSmallestSetOfSmallestRings()
+
+    return [convertRingToSubMolecule(SSSR[0])[0], 
+                convertRingToSubMolecule(SSSR[1])[0]]
+
 ################################################################################
 
 class ThermoDepository(Database):
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -1652,6 +1652,68 @@ class ThermoDatabase(object):
                 # Typically this comment is very short and also very helpful to check if the ring correction is calculated correctly.
 
 
+    def getBicyclicCorrectionThermoDataFromHeuristic(self, bicyclic):
+
+        # saturate if the bicyclic has unsaturated bonds
+        # otherwise return None
+        bicyclic_submol = convertRingToSubMolecule(bicyclic)[0]
+        saturated_bicyclic_submol, alreadySaturated = saturateRingBonds(bicyclic_submol)
+
+        if alreadySaturated:
+            return None
+        # split bicyclic into two single ring submols
+        single_ring_submols = splitBicyclicIntoSingleRings(bicyclic_submol)
+
+        # split saturated bicyclic into two single ring submols
+        saturated_single_ring_submols = splitBicyclicIntoSingleRings(saturated_bicyclic_submol)
+
+        # apply formula: 
+        # bicyclic correction ~= saturated bicyclic correction - 
+        # saturated single ring corrections + single ring corrections
+        saturated_bicyclic_thermoData = self.__addRingCorrectionThermoDataFromTree(None, 
+                self.groups['polycyclic'], saturated_bicyclic_submol, saturated_bicyclic_submol.atoms)[0]
+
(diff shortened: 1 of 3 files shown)


rmgpy/data/thermo.py
```diff
@@ -1660,6 +1660,68 @@ def __addPolyRingCorrectionThermoDataFromHeuristic(self, thermoData, polyring):
...
+        # saturated single ring corrections + single ring corrections
+        saturated_bicyclic_thermoData = self.__addRingCorrectionThermoDataFromTree(None, 
+                self.groups['polycyclic'], saturated_bicyclic_submol, saturated_bicyclic_submol.atoms)[0]
+
+        estimated_bicyclic_thermodata = saturated_bicyclic_thermoData
+        
+        for submol in saturated_single_ring_submols:
+            
+            if not isAromaticRing(submol):
+                aromaticBonds = findAromaticBondsFromSubMolecule(submol)
```
mliu49: Why do you try to find aromatic bonds if the submol is not an aromatic ring? Also, isn't the ring saturated already?

rmgpy/data/thermoTest.py
```diff
@@ -716,6 +716,52 @@ def testAddPolyRingCorrectionThermoDataFromHeuristicUsingAlkaneTricyclic(self):
...
+
+        self.assertIn('Cyclopentane', ringGroupLabels)
+        self.assertIn('Cyclopropane', ringGroupLabels)
+        self.assertIn('Cyclopentene', ringGroupLabels)
+        self.assertIn('Cyclopropene', ringGroupLabels)
+        self.assertIn('s2_3_5_ane', polycyclicGroupLabels)
+
+    def testGetBicyclicCorrectionThermoDataFromHeuristic2(self):
+        """
+        Test bicyclic correction estimated properly from heuristic formula
```
mliu49: Could you differentiate the docstrings of these two tests? For example, you could add the SMILES of the species you're testing.

rmgpy/data/thermoTest.py
```diff
@@ -899,6 +899,26 @@ def testIsAromaticRing(self):
...
+        """
+        smiles = 'C1=CCC2C1=C2'
+        mol = Molecule().fromSMILES(smiles)
+        polyring = mol.getDisparateRings()[1][0]
+
+        self.assertTrue(isBicyclic(polyring))
+
+    def testIsBicyclic2(self):
+        """
+        Test isBicyclic identifies bicyclic correctly
```
mliu49: Could you also differentiate the docstrings for these two tests?

rmgpy/data/thermo.py
```diff
@@ -1612,15 +1612,27 @@ def __addPolycyclicCorrectionThermoData(self, thermoData, molecule, polyring):
...
+        matched_group_thermodata, matched_group, isPartialMatch = self.__addRingCorrectionThermoDataFromTree(None, self.groups['polycyclic'], molecule, polyring)
         
         # if partial match (non-H atoms number same between 
         # polycylic ring in molecule and match group)
         # otherwise, apply heuristic algorithm
         if not isPartialMatch:
-            thermoData = addThermoData(thermoData, matched_group_thermodata, groupAdditivity=True, verbose=True)
-            # By setting verbose=True, we turn on the comments of polycyclic correction to pass the unittest.
-            # Typically this comment is very short and also very helpful to check if the ring correction is calculated correctly.
+            if isBicyclic(polyring):
```
mliu49: Could these if statements be simplified?
```
if isBicyclic(polyring) and matched_group.label in self.groups['polycyclic'].genericNodes:
    estimated_bicyclic_thermodata = self.getBicyclicCorrectionThermoDataFromHeuristic(polyring)
    if not estimated_bicyclic_thermodata:
        estimated_bicyclic_thermodata = matched_group_thermodata
    thermoData = addThermoData(thermoData, estimated_bicyclic_thermodata, groupAdditivity=True)
else:
    thermoData = addThermoData(thermoData, matched_group_thermodata, groupAdditivity=True, verbose=True)
```

rmgpy/data/thermoTest.py
```diff
@@ -1184,6 +1184,68 @@ def testSaturateRingBonds1(self):
...
+        expected_saturated_ring_submol.updateConnectivityValues()
+
+        self.assertTrue(alreadySaturated)
+        self.assertEqual(saturated_ring_submol.multiplicity, \
+                            expected_saturated_ring_submol.multiplicity)
+        self.assertTrue(saturated_ring_submol.isIsomorphic(expected_saturated_ring_submol))
+
+    def testSaturateRingBonds3(self):
+        """
+        Test unsaturated bonds can be saturated properly
```
mliu49: Could you update these docstrings too?

rmgpy/data/thermoTest.py
```diff
@@ -1091,6 +1091,33 @@ def testSplitBicyclicIntoSingleRings(self):
...
+        """
+        smiles = 'C1=CCC2=C1C2'
+        mol = Molecule().fromSMILES(smiles)
+        ring_submol = convertRingToSubMolecule(mol.getDisparateRings()[1][0])[0]
+
+        saturated_ring_submol, alreadySaturated = saturateRingBonds(ring_submol)
+
+        expected_saturated_ring_submol = Molecule().fromSMILES('C1CCC2C1C2')
+        # remove hydrogen
```
mliu49: Would `Molecule.deleteHydrogens()` not work here?

rmgpy/data/thermo.py
mliu49: I saw the other commit where you don't saturate aromatic bonds anymore. However, you set the bond order of aromatic bonds to 1 here anyway, so what is the reason for not saturating them to begin with?

rmgpy/data/thermo.py
```diff
@@ -1719,10 +1719,17 @@ def getBicyclicCorrectionThermoDataFromHeuristic(self, bicyclic):
...
+            Tdata = ([300,400,500,600,800,1000,1500],"K"),
+            Cpdata = ([0.0,0.0,0.0,0.0,0.0,0.0,0.0],"J/(mol*K)"),
+            H298 = (0.0,"kJ/mol"),
+            S298 = (0.0,"J/(mol*K)"),
+        )
         saturated_bicyclic_thermoData = self.__addRingCorrectionThermoDataFromTree(None, 
                 self.groups['polycyclic'], saturated_bicyclic_submol, saturated_bicyclic_submol.atoms)[0]
 
-        estimated_bicyclic_thermodata = saturated_bicyclic_thermoData
+        estimated_bicyclic_thermodata = addThermoData(estimated_bicyclic_thermodata,\
```
mliu49: How does this compare to a deep copy? Would it accomplish the same thing? Would a deep copy be slower?

rmgpy/molecule/molecule.py
```diff
@@ -1858,7 +1858,9 @@ def getDeterministicSmallestSetOfSmallestRings(self):
                 # Keep the smallest of the cycles found above
                 cycleCandidate_tups = []
                 for cycle0 in cycles:
-                    tup = (cycle0, len(cycle0), -sum([originConnDict[v] for v in cycle0]), -sum([v.element.number for v in cycle0]))
+                    tup = (cycle0, len(cycle0), -sum([originConnDict[v] for v in cycle0]), \
+                            -sum([v.element.number for v in cycle0]),\
```
mliu49: The backslashes are also unnecessary here.

rmgpy/data/thermoTest.py
```diff
@@ -1172,10 +1172,13 @@ def testSplitBicyclicIntoSingleRings(self):
...
                                 key=lambda submol: len(submol.atoms))
 
+        single_ring_submol_a.updateAtomTypes()
+        single_ring_submol_b.updateAtomTypes()
+
         expected_submol_a = Molecule().fromSMILES('C1=CC1')
         atomsToRemove = []
         for atom in expected_submol_a.atoms:
-            if len(atom.bonds) == 1: 
+            if atom.isHydrogen():
```
mliu49: Would `Molecule.deleteHydrogens()` work here as well?

rmgpy/data/thermo.py
```diff
@@ -1720,14 +1720,23 @@ def getBicyclicCorrectionThermoDataFromHeuristic(self, bicyclic):
...
+            Tdata = ([300,400,500,600,800,1000,1500],"K"),
+            Cpdata = ([0.0,0.0,0.0,0.0,0.0,0.0,0.0],"J/(mol*K)"),
+            H298 = (0.0,"kJ/mol"),
+            S298 = (0.0,"J/(mol*K)")
+        )
         
         saturated_bicyclic_thermoData = self.__addRingCorrectionThermoDataFromTree(None, 
                 self.groups['polycyclic'], saturated_bicyclic_submol, saturated_bicyclic_submol.atoms)[0]
+        
+        estimated_bicyclic_thermodata = addThermoData(estimated_bicyclic_thermodata,
```
mliu49: You fixed this once in e2cefaefba671d890b73414252f515c60b4fb928, then you unfixed it in 73c3e28e3fed40cacbf9136724e7997aec5cb502, then you fix it again here. I think you could probably squash the first two commits, and only fix it once here.

rmgpy/data/thermo.py
```diff
@@ -502,6 +502,33 @@ def splitBicyclicIntoSingleRings(bicyclic_submol):
     return [convertRingToSubMolecule(SSSR[0])[0], 
                 convertRingToSubMolecule(SSSR[1])[0]]
 
+def saturateRingBonds(ring_submol):
```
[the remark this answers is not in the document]
KEHANG: `Molecule.toSingleBonds()` is not customized enough for this purpose, e.g., here we don't want to saturate benzene bond

rmgpy/data/thermo.py
KEHANG: Yeah, thanks for pointing it out.

KEHANG:
@mliu49 thanks for your review comments, new commits from https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1193/commits/12d163e763a7ba863d2a2df1fffa98efa09a2e7a to the end are the ones to address the issues you mentioned one by one. If anything looks good to you, will squash them properly afterwards.

rmgpy/data/thermoTest.py
```diff
@@ -295,6 +295,20 @@ def testSpeciesThermoGenerationLibrary(self):
...
 
+    def testThermoEstimationNotAffectDatabase(self):
+
+        poly_root = self.database.groups['polycyclic'].entries['PolycyclicRing']
+        previous_enthalpy = poly_root.data.getEnthalpy(298)/4184.0
+        smiles = 'C1C2CC1C=CC=C2'
+        spec = Species().fromSMILES(smiles)
+        spec.generateResonanceIsomers()
+
+        thermo_gav = self.database.getThermoDataFromGroups(spec)
```
mliu49: Should you add an assertion that the generated thermo data actually hit the `PolycyclicRing` group? Or is that always guaranteed?

mliu49:
Thanks for making the changes! I think you can rebase and squash when you're ready.

rmgpy/data/thermoTest.py
KEHANG: will add that.

rmgpy/data/thermo.py
```diff
@@ -485,6 +500,47 @@ def bicyclicDecompositionForPolyring(polyring):
...
+
+    return [convertRingToSubMolecule(SSSR[0])[0], 
+                convertRingToSubMolecule(SSSR[1])[0]]
+
+def saturateRingBonds(ring_submol):
+    """
+    Given a ring submolelcule (`Molecule`), makes a deep copy and converts non-single bonds 
+    into single bonds, returns a new saturated submolecule (`Molecule`)
+    """
+    from rmgpy.molecule.molecule import Molecule, Bond
```
mliu49: These have already been imported at the top of this file.

rmgpy/data/thermo.py
```diff
@@ -1616,6 +1692,80 @@ def __addPolyRingCorrectionThermoDataFromHeuristic(self, thermoData, polyring):
...
+
+        for submol in saturated_single_ring_submols:
+            
+            if not isAromaticRing(submol):
+                aromaticBonds = findAromaticBondsFromSubMolecule(submol)
+                for aromaticBond in aromaticBonds:
+                    aromaticBond.setOrderNum(1)
+                
+                submol.update()
+                single_ring_thermoData = self.__addRingCorrectionThermoDataFromTree(None, \
```
mliu49: Could you remove this backslash and the ones in the following lines?

KEHANG:
@mliu49 I think it's ready for a final check

rmgpy/data/thermo.py
```diff
@@ -502,6 +502,33 @@ def splitBicyclicIntoSingleRings(bicyclic_submol):
...
+    for atom in ring_submol.atoms:
+        for bondedAtom, bond in atom.edges.iteritems():
+            if bondedAtom in ring_submol.atoms:
+                if bond.order > 1.0: alreadySaturated = False
+                if not mol0.hasBond(atomsMapping[atom],atomsMapping[bondedAtom]):
+                    mol0.addBond(Bond(atomsMapping[atom],atomsMapping[bondedAtom],order=1.0))
+    
+    mol0.updateAtomTypes()
+    mol0.updateMultiplicity()
+    mol0.updateConnectivityValues()
```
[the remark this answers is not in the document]
KEHANG: @mliu49 yeah, `updateCharge()` is not suitable for submolecule. which consist a substructure of the original molecule, so e.g., the valence 4 of carbon is not alway satisfied.

rmgpy/data/thermo.py
KEHANG: thanks, they are removed now

rmgpy/data/thermo.py
KEHANG: they are removed now

PR #1200, opened 2017-09-12
Seed mechanism improvements

mjohnson541:
This improves two things with regards to seed mechanisms:  
1) Fixes family name parsing for some of the more complicated kinetics comments
2) Automatically loads the libraries Library Reactions are originally from without them needing to be a seed mechanism or a reaction library.  

It also removes some of the more verbose logging messages.


rmgpy/data/kinetics/library.py
```diff
@@ -143,7 +143,6 @@ def getLibraryReactions(self):
             if entry._longDesc and 'Originally from reaction library: ' in entry._longDesc:
                 lib = entry._longDesc.replace('Originally from reaction library: ','')
                 lib = lib.replace('\n','')
-                logging.info(lib)
```
amarkpayne: Should these be logging.debug statements instead? (Lines 146, 155, 162)

mjohnson541: For stuff that occurs so early in simulation, (makes it easy to test) I'm comfortable requiring them to add their own debugging statements.  These statements were not particularly helpful when I was debugging the parsing issue and after thinking about it I've decided that they clutter up the log file too much to be worth keeping.

rmgpy/data/kinetics/library.py
```diff
@@ -150,7 +150,8 @@ def getLibraryReactions(self):
                  )
             elif entry._longDesc and 'rate rule' in entry._longDesc: #template reaction
                 c = entry._longDesc.split('\n')
-                familyname = c[-1].replace('family: ','')
+                cnew = [i for i in c if 'family: ' in i]
```
goldmanm: It is not intuitively obvious from the variable name that 'cnew' contains only strings of family names. Maybe we could rename it to 'family_comments' or something like that.

rmgpy/rmg/model.py
```diff
@@ -1415,7 +1416,10 @@ def addSeedMechanismToCore(self, seedMechanism, react=False):
 
         if react: raise NotImplementedError("react=True doesn't work yet")
         database = rmgpy.data.rmg.database
-
+
```
goldmanm: slight whitespace issues here.

rmgpy/data/kinetics/library.py
```diff
@@ -123,9 +123,9 @@ class KineticsLibrary(Database):
     A class for working with an RMG kinetics library.
     """
 
-    def __init__(self, label='', name='', solvent=None, shortDesc='', longDesc='', duplicatesChecked=True):
```
goldmanm: Great name improvement!

rmgpy/data/kinetics/library.py
```diff
@@ -141,7 +141,7 @@ def getLibraryReactions(self):
         rxns = []
         for entry in self.entries.values():                
             if entry._longDesc and 'Originally from reaction library: ' in entry._longDesc:
-                lib = entry._longDesc.replace('Originally from reaction library: ','')
+                lib = entry._longDesc.split('\n')[0].replace('Originally from reaction library: ','')
```
goldmanm: Does this have the same reliability concerns as the [first commit](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1200/commits/d81e5d6dd7772b3b688a73977a2ef9b35d51e75f)

rmgpy/data/kinetics/library.py
mjohnson541: fixed

rmgpy/rmg/model.py
mjohnson541: fixed

rmgpy/data/kinetics/library.py
mjohnson541: I've changed it now so that it extracts the library name from the first line that has 'Originally from reaction library:  ' in it

mjohnson541:
Ok, I think the best way to test this is to start a job that involves libraries stop it and try to use the resulting seed mechanism in a run without including the libraries used to generate it.  That checks the second problem this fixes.  I think the fix for the first problem (the addition of the 'family:  ' comment) is well covered by testGetLibraryReactions in libraryTest.py.

goldmanm:
I am wondering if it would be possible to make a small unittest for this, so that if it breaks with later changes to RMG, people will be aware and keep it functioning. My conception of a unittest for it.

1. load a test database (which is fast to load) without any libraries (and possible without families either)
1. Create a seed mechanism (with only one reaction) from either `ethane-oxidation` or `GRI-Mech3.0` which are in the testing database and add it to a `KineticsPrimaryDatabase`
2. initialize a CERM model
3. call `addSeedMechanismToCore`
4. check and see that the library was loaded. 

Hopefully, this will make it so this fix lasts, and I think this should avoid any file creation complications. What are your thoughts?

mjohnson541:
The rmg run in mainTest.py now loads a library as a seed mechanism, so when the generated seed is tested it also tests if that library is loaded automatically.

mjohnson541:
Ok, this should be ready for final review, I think it would be prudent to keep the parsing commits in this PR unless your stuff is ready right now, it should prevent changes from kinetics comments from crashing the seed mechanism stuff until your stuff is finished.

PR #1205, opened 2017-09-19
More badges in README and other improvements

mliu49:
Small update to the README:
- Consolidate badges at the top of the document for visibility
- Add more badges!
- Minor changes to content

Notes:
- I removed the section on the current version (replaced with badge) so that we don't need to update the file with each new version.
- I removed the section on continuous integration, since I didn't feel like it was particularly important explicitly state that we have it.


KEHANG:
Looks good to me

PR #1207, opened 2017-09-25
print out reverse rates in cantherm kinetics jobs

mjohnson541:
Causes the equilibrium constant and reverse rate coefficients to be printed out along with the forward rate coefficients in a Cantherm kinetics job.

### DIFF
--- a/rmgpy/cantherm/kinetics.py
+++ b/rmgpy/cantherm/kinetics.py
@@ -170,6 +170,8 @@ class KineticsJob:
         
         ks = []
         k0s = []
+        k0revs = []
+        krevs = []
         
         logging.info('Saving kinetics for {0}...'.format(reaction))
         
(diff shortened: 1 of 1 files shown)


goldmanm:
what files exactly are being modified? Is it just adding to the comments or is a machine-readable Arrhenius form output?

I'm not implying changing it. Just want clarification.

mjohnson541:
It was just adding comments-putting the Keq(T) and krev(T) values in a table, I'm not sure how useful machine readable Arrhenius would be, but now that you mention it, I think at least providing an Arrhenius fit of the data for kf and krev would be helpful.  I'll add that.

mjohnson541:
I just added a commit to print comments with Arrhenius fits for reverse rate coefficients with and without tunneling.

goldmanm:
Looks great @mjohnson541. Thanks for the reminder about reviewing.

PR #1214, opened 2017-10-12
Extending the lone pair <=> radical resonance

alongd:
Previously the `generateLonePairRadicalResonanceStructures` was limited to certain structures, since it was written mainly with N2O in mind:

image <=> image

This PR extends it to consider RO[O], RO[NH], RN[O], RN[NH] as well, e.g.:
image <=> image
image <=> image

Resonance tests were also added.

Since RDkit cannot process hypervalance oxygen (e.g. `–[O+]=`), some ROO inchiParsingTests had to be removed . Any comments/suggestions on these deletions are of course welcome.

### DIFF
--- a/rmgpy/molecule/generator.py
+++ b/rmgpy/molecule/generator.py
@@ -492,7 +492,7 @@ def find_lowest_u_layer(mol, u_layer, equivalent_atoms):
 def generate_minimum_resonance_isomer(mol):
     """
     Select the resonance isomer that is isomorphic to the parameter isomer, with the lowest unpaired
-    electrons descriptor.
+    electrons descriptor, unless this unnecessarily forms a charged molecule.
 
     First, we generate all isomorphic resonance isomers.
     Next, we return the candidate with the lowest unpaired electrons metric.
@@ -501,24 +501,29 @@ def generate_minimum_resonance_isomer(mol):
     """
 
     cython.declare(
+        atom=Atom,
         candidates=list,
         sel=Molecule,
         cand=Molecule,
         metric_sel=list,
+        charge_sel=int,
+        charge_cand=int,
         metric_cand=list,
         )
 
-
     candidates = resonance.generateIsomorphicResonanceStructures(mol)
-    
-    sel = candidates[0]
-    metric_sel = get_unpaired_electrons(sel)
-    for cand in candidates[1:]:
-       metric_cand = get_unpaired_electrons(cand)
-       if metric_cand < metric_sel:
-            sel = cand
-            metric_sel = metric_cand
 
+    sel = mol
+    metric_sel = get_unpaired_electrons(sel)
+    charge_sel = sum([abs(atom.charge) for atom in sel.vertices])
+    for cand in candidates:
+        metric_cand = get_unpaired_electrons(cand)
+        if metric_cand < metric_sel:
+            charge_cand = sum([abs(atom.charge) for atom in cand.vertices])
+            if charge_cand <= charge_sel:
+                sel = cand
+                metric_sel = metric_cand
+                charge_sel = charge_cand
     return sel
 
 
--- a/rmgpy/molecule/pathfinder.py
+++ b/rmgpy/molecule/pathfinder.py
@@ -263,28 +263,24 @@ def findAllDelocalizationPaths(atom1):
 def findAllDelocalizationPathsLonePairRadical(atom1):
     """
     Find all the delocalization paths of lone electron pairs next to the radical center indicated
-    by `atom1`. Used to generate resonance isomers.
+    by `atom1`. Used to generate resonance isomers for species such as RN[O], RO[NH], RO[O], RN[NH].
+    (currently only N and O are considered, will soon include S as well)
     """
     cython.declare(paths=list)
     cython.declare(atom2=Atom, bond12=Bond)
-    
-    # No paths if atom1 is not a radical
-    if atom1.radicalElectrons <= 0:
-        return []
-    
-    # In a first step we only consider nitrogen and oxygen atoms as possible radical centers
-    if not ((atom1.lonePairs == 0 and atom1.isNitrogen()) or(atom1.lonePairs == 2 and atom1.isOxygen())):
-        return []
-    
-    # Find all delocalization paths
+
     paths = []
-    for atom2, bond12 in atom1.edges.items():
-        # Only single bonds are considered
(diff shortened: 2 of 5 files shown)


rmgpy/molecule/pathfinder.py
```diff
@@ -263,28 +263,24 @@ def findAllDelocalizationPaths(atom1):
 def findAllDelocalizationPathsLonePairRadical(atom1):
     """
     Find all the delocalization paths of lone electron pairs next to the radical center indicated
-    by `atom1`. Used to generate resonance isomers.
+    by `atom1`. Used to generate resonance isomers for species such as RN[O], RO[NH], ROO, RN[NH].
```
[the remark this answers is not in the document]
alongd: right. corrected.

alongd:
Some InChI issues were fixed to allow the new resonance transformations in this PR.
- Main modification: Certain resonance isomers won't get an AugmentedInChI in species.py (using a try: except block)
- parser.check() now uses the `aug_inchi`. Fixes #1218 
- generate_minimum_resonance_isomer() now also considers charge
- other minor related fixes

@mjohnson541, could you take a look?

mliu49:
Is it necessary for the augmented InChI fixes to be part of this pull request? It seems somewhat unrelated, and I'm currently working on a major overhaul of the generator/parser modules which they would fit better with.

alongd:
An alternative is to merge this w/o three inchiparsingTests.
When generating the resonance structures in this PR, these tests fail since RDkit cannot handle hyper-valance atoms. I'm OK with merging without the InChI commits here, commenting out the parsing tests. The tests can then be brought back along with your overhaul.

mliu49:
Ahh sorry, I forgot about that. I think it is reasonable to mark those tests as work_in_progress for this pull request then.

rmgpy/molecule/generator.py
```diff
@@ -501,24 +501,33 @@ def generate_minimum_resonance_isomer(mol):
...
-       metric_cand = get_unpaired_electrons(cand)
-       if metric_cand < metric_sel:
-            sel = cand
-            metric_sel = metric_cand
 
+    sel = mol
+    metric_sel = get_unpaired_electrons(sel)
+    charge_sel = 0
+    for atom in sel.vertices:
+        charge_sel += abs(atom.charge)
```
mliu49: This and the instance below could be replaced by a list comprehension:
```
charge_sel = sum([abs(atom.charge) for atom in sel.vertices])
```

alongd:
As discussed with @mliu49 , and suggested by @KEHANG in #1218, I updated the check() function to use the `aug_inchi` but also check the `mol.multiplicity`.
Blank line deletion commits were removed to keep rebase conflicts to a minimum in the future parsing overhaul.
@mjohnson541, this should now be ready for review.

rmgpy/molecule/generator.py
```diff
@@ -501,24 +501,33 @@ def generate_minimum_resonance_isomer(mol):
...
+    for cand in candidates:
+        metric_cand = get_unpaired_electrons(cand)
+        if metric_cand < metric_sel:
+            charge_cand = 0
+            for atom in cand.vertices:
+                charge_cand += abs(atom.charge)
+            if charge_cand < charge_sel:
```
mliu49: Perhaps this should be `<=` so that the selected structure will still be updated if the indices are reduced but the charge remains the same.

rmgpy/molecule/resonanceTest.py
```diff
@@ -40,14 +40,20 @@ def testAllylShift(self):
...
+    def testLonePairRadical(self):
+        """Test resonance structure generation for CC=N[O], ROO, RN[NH] radicals
 
-        Simple case for lone pair <=> radical resonance"""
+        Selected cases for lone pair <=> radical resonance"""
         molList = generateResonanceStructures(Molecule(SMILES="CC=N[O]"))
         self.assertEqual(len(molList), 3)
         self.assertTrue(any([any([atom.charge != 0 for atom in mol.vertices]) for mol in molList]))
 
+        molList = generateResonanceStructures(Molecule(SMILES="CO[O]"))
```
mliu49: Could you split these into separate test methods?

rmgpy/species.py
```diff
@@ -519,19 +520,22 @@ def copy(self, deep=False):
...
         self.generateResonanceIsomers()
         for mol in self.molecule:
-            cand = mol.toAugmentedInChI()
-            candidates.append(cand)
-
-        candidates.sort()
-        return candidates[0] 
+            try:
+                cand = [mol.toAugmentedInChI(),mol]
+                candidates.append(cand)
```
mliu49: The `candidates.append(cand)` statement would ideally go into an `else` clause after the `except` clause.

Also, it would be best if you could identify the exception that's raised for the failure that you expect and specify it in the `except` statement.

rmgpy/species.py
```diff
@@ -519,19 +520,22 @@ def copy(self, deep=False):
...
+            except:
+                pass  # not all resonance isomers can be parsed into InChI (e.g. if containing a hypervalance atom)
+        candidates = sorted(candidates, key=itemgetter(0))
+        for cand in candidates:
+            if not any(atom.charge != 0 for atom in cand[1].vertices):
+                return cand[0]
```
mliu49: Just to confirm, this preferentially returns InChI's of charged resonance forms?

alongd: This should preferentially return InChI's of *uncharged* forms.
The `any()` is TRUE if one or more atoms are charged, and the `not` reverses it.

alongd:
Thanks for the comments, @mliu49 !
All requested changes were squashed here.

rmgpy/species.py
mliu49: Thanks, I missed the not...

Would this be equivalent to `all(atom.charge == 0 for atom in cand[1].vertices)`?

alongd: Yes, thanks. I'll change it.

alongd:
This PR should be final now.
Let me know if there are any other concerns. Otherwise, it's ready to merge.

@mjohnson541 , could you approve your review if all looks good to you too?

alongd:
Thanks! yep, ready to merge.

alongd:
@mliu49, @mjohnson541, this is now rebased, Travis checks are clean.
Could one of you re-approve?
Also ready to merge

PR #1217, opened 2017-10-18
Do not consider template differences from resonance structures as multiple TS

mliu49:
The multiple TS pull request (#1055) created duplicate reactions any time isomorphic reactions had different reaction templates. That led to cases where there could be 6 or more duplicates of a single reaction due to different resonance structures. This pull request reverses that change for reactions which have different reaction templates due to different resonance structures.

The new algorithm takes advantage of the isIdentical method to distinguish between cases where templates are different due to resonance structures vs. different transition states. This was previously used to distinguish between reactions which contribute to degeneracy vs. ones which don't. The logic is similar for multiple transition states, in that if the exact same atoms are participating in the reaction in the same way, then we can assume that it's just one transition state.

As a side effect of the change, the old issue with non-deterministic kinetics is raised again, since we choose to keep the first reaction we encounter instead of all of them. However, the true issue behind that is non-deterministic resonance structure order. Reaction generation and template matching is reasonably deterministic, but the order in which reactions are generated depends on the order of resonance structures.

In most cases, the order is determined by sorting by H298, which is fairly deterministic, although there are definitely cases where it is inaccurate. Overall, it seems reasonable to trade excessive duplicates for a chance of non-deterministic-ness.

### DIFF
--- a/rmgpy/data/kinetics/common.py
+++ b/rmgpy/data/kinetics/common.py
@@ -267,7 +267,7 @@ def ensure_independent_atom_ids(input_species, resonance=True):
             species.generate_resonance_structures(keepIsomorphic=True)
 
 
-def find_degenerate_reactions(rxnList, same_reactants=None, kinetics_database=None, kinetics_family=None):
+def find_degenerate_reactions(rxnList, same_reactants=None, template=None, kinetics_database=None, kinetics_family=None):
     """
     given a list of Reaction object with Molecule objects, this method
     removes degenerate reactions and increments the degeneracy of the
@@ -278,11 +278,24 @@ def find_degenerate_reactions(rxnList, same_reactants=None, kinetics_database=No
     This algorithm used to exist in family.__generateReactions, but was moved
     here because it didn't have any family dependence.
     """
+    # If a specific reaction template is requested, filter by that template
+    if template is not None:
+        selected_rxns = []
+        template = frozenset(template)
+        for rxn in rxnList:
+            if template == frozenset(rxn.template):
+                selected_rxns.append(rxn)
+        if not selected_rxns:
+            # Only log a warning here. If a non-empty output is expected, then the caller should raise an exception
+            logging.warning('No reactions matched the specified template, {0}'.format(template))
+            return []
+    else:
+        selected_rxns = rxnList
 
     # We want to sort all the reactions into sublists composed of isomorphic reactions
     # with degenerate transition states
     rxnSorted = []
-    for rxn0 in rxnList:
+    for rxn0 in selected_rxns:
         # find resonance structures for rxn0
         rxn0.ensure_species()
         if len(rxnSorted) == 0:
--- a/rmgpy/data/kinetics/common.py
+++ b/rmgpy/data/kinetics/common.py
@@ -294,36 +294,39 @@ def find_degenerate_reactions(rxnList, same_reactants=None, kinetics_database=No
                 # Try to determine if the current rxn0 is identical or isomorphic to any reactions in the sublist
                 isomorphic = False
                 identical = False
-                sameTemplate = False
+                sameTemplate = True
                 for rxn in rxnList1:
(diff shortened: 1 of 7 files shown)


mliu49:
Seems to be failing eg1 on RMG-tests. I will investigate.

mliu49:
The problem is an interesting case where the forward reaction has 1 transition state, but the reverse reaction has 2 transition states to an isomorphic product.

image

The reaction was initially found via methylcyclopropyl, which generated two duplicate reactions with different templates. When calculating the proper degeneracy in the forward direction, each of the duplicates only generated a single reaction, which does not necessarily have the same template as before, causing `calculateDegeneracy` to mistakenly find no reactions.

Prior to the multiple TS PR, one reaction would have been generated in both the forward and reverse directions. After the multiple TS PR, two reactions were generated in both directions. This PR caused the problematic case of one reaction being generated in the forward direction and two reactions being generated in the reverse direction.

I think the root cause is because `Birad_recombination` has two separate sub-trees for the radical sites, even though they are effectively interchangeable in terms of labeling. The result is that one possible template is `['R3_SS', 'Y_rad_out', 'Cpri_rad_out_2H']` and the other template is `['R3_SS', 'C_rad_out_2H', 'Ypri_rad_out']`. The *groups* are actually identical, but they come from two different trees (`Y_rad_out` and `Ypri_rad_out`). `R_Recombination` gets around this by having a single tree, then doing some additional processing which is hard-coded in `applyRecipe`.

I think the optimal solution might be to change `Birad_recombination` to use a single radical tree as well, which would require additional hard-coding in RMG-Py. There might be another way to solve it via the degeneracy algorithm, but I don't think it would be very straightforward.

mliu49:
My last two comments can be effectively ignored, as I've changed my approach to this problem.

Now, instead of removing duplicate reactions at the `generate_reactions` stage, they are removed at the `checkForExistingReaction` stage. This is accomplished through the use of the duplicate attribute. Previously, both methods were assigning duplicate flags, meaning that the ones assigned by `generate_reactions` were essentially meaningless. Now, they are only assigned by `generate_reactions`, and `checkForExistingReaction` will use them to determine whether or not to add a reaction to the model.

I think this is now ready to be reviewed and merged. This should eliminate issues with duplicate reactions, which should also have an appreciable speedup effect.

mliu49:
RMG-tests results:

---|---|---|---|---
eg1|00:00:03:33|00:00:03:20|469.24|465.87
eg3|00:00:07:37|00:00:07:23|418.91|416.54
eg5|00:00:02:18|00:00:02:18|373.74|373.69
eg6|00:00:01:59|00:00:01:58|373.90|375.81
eg7|00:00:01:26|00:00:01:31|365.50|365.24
NC|00:00:02:36|00:00:02:17|378.70|380.82
solvent_hexane|00:00:01:46|00:00:01:49|364.81|362.57
MCH|00:00:05:03|00:00:04:55|732.25|733.36

[results.log](https://github.com/ReactionMechanismGenerator/RMG-Py/files/1749722/results.log)

mliu49:
As we discussed, I tried coming up with a deterministic way to sort resonance structures when the existing sorting methods in thermo estimation are not sufficient. Unfortunately, my attempts were not very successful. You can check out my attempt (including a failing unit test) on the `sorting_int` branch.

Essentially, I tried to derive an integer based on a mix of atom properties and sorting indices. One difficulty I had was getting the number to be different for different resonance structures. The other difficulty was with getting a deterministic value, since I found that the sorting indices are not deterministic.

If you have other ideas for the sorting metric, I would be happy to try them out. Otherwise, I think it would be best to go ahead and merge this PR for now, and continue developing that separately.

goldmanm:
That's unfortunate. I'm writing an issue about it for future knowledge (so your effort is not for not). Anyways, we should merge this. Do you mind rebasing the request, and I'll approve it?

mliu49:
Thanks, it's been rebased.

PR #1219, opened 2017-10-19
Resonance fixes and method renaming

mliu49:
Fixes a couple bugs in the resonance module:
- Molecules incorrectly identified as non-aromatic were not being properly handled
- Aromatic resonance structures not properly generated in some cases
- Incorrectly placed parenthesis

I also decided to bite the bullet and completely rename `generateResonanceIsomers` to `generateResonanceStructures`. I previously renamed the method in the resonance module (7cadcbd4f2349c2b653ca1a9cba7eafba9fb5f47), but did not rename the molecule or species methods due to their prevalent use. Since having both names was somewhat confusing, this pr standardizes to `generateResonanceStructures`. Additional pull requests to update usage in RMG-database and RMG-website to follow soon.


rmgpy/molecule/resonance.py
```diff
@@ -461,9 +461,18 @@ def generateAromaticResonanceStructures(mol, features=None):
     # Then determine which ones are aromatic
     aromaticBonds = molecule.getAromaticRings(rings)[1]
 
+    # Check if the bonds in at least one ring are alternating single/double
+    # If not, then we need to shift electrons around to obtain the true aromatic form
+    valid = False
+    for bonds in aromaticBonds:
+        orders = [bond.order for bond in bonds]
```
goldmanm: I am slightly confused how this works. I thought `bond.order` was a float, so a list of floats should be returned (and `getOrderStr()` should be used to get the string form. Maybe I am missing something.

rmgpy/species.py
```diff
@@ -162,17 +162,17 @@ def setMolecularWeight(self, value):
         self._molecularWeight = quantity.Mass(value)
     molecularWeight = property(getMolecularWeight, setMolecularWeight, """The molecular weight of the species. (Note: value_si is in kg/molecule not kg/mole)""")
 
-    def generateResonanceIsomers(self, keepIsomorphic=True):
+    def generate_resonance_structures(self, keepIsomorphic=True):
         """
```
goldmanm: this is beautiful! now I won't have to look up to determine whether structures or isomers is used!

rmgpy/molecule/resonance.py
mliu49: That's true... I have no idea what I was doing, or why this appears to work.

mliu49: Thanks for pointing this out. Turns out that this wouldn't have worked anyway, since the bonds are not guaranteed to be in the correct order. I couldn't figure out a robust solution, so I decided to go with the assumption that it is always invalid instead.

mliu49:
I added another commit to generate aromatic resonance structures when comparing products in `__generateReactions()`. I had added this on one of my degeneracy branches at some point, but it got lost at some point during the transition to @goldmanm's degeneracy branch.

PR #1222, opened 2017-10-20
make model enlargement occur after each reactor

mjohnson541:
After each reactor all of the processing steps except pruning will occur before the next reactor is run.  This severely reduces the number of simulations needed to be run for runs with multiple reactors because it prevents different reactors from picking up the same species.  

Defining a simulation efficiency metric:  Species Per Simulation Time (SPST) that is the number of species added per simulation.  We can see this does very poorly the more reactors you add.  

screen shot 2017-10-20 at 3 14 25 pm

For a system where only one species is taken at a time this PR forces SPST = 1, (provided no simulations finish) no matter how many reactors are used.  Accounting for finishing simulations it will decrease from 1 over the mechanism generation process as the model becomes more complete.

### DIFF
--- a/rmgpy/rmg/model.py
+++ b/rmgpy/rmg/model.py
@@ -1167,34 +1167,35 @@ class CoreEdgeReactionModel:
         """
 
         ineligibleSpecies = []     # A list of the species which are not eligible for pruning, for any reason
-
-        numEdgeSpecies = len(self.edge.species)
+        prunableSpecies = reactionSystems[0].prunableSpecies
+        prunableNetworks = reactionSystems[0].prunableNetworks
+        
+        numPrunableSpecies = len(prunableSpecies)
         iteration = self.iterationNum
         # All edge species that have not existed for more than two enlarge
         # iterations are ineligible for pruning
-        for spec in self.edge.species:
+        for spec in prunableSpecies:
             if iteration - spec.creationIteration <= minSpeciesExistIterationsForPrune:
                 ineligibleSpecies.append(spec)
 
         # Get the maximum species rates (and network leak rates)
         # across all reaction systems
-        maxEdgeSpeciesRateRatios = numpy.zeros((numEdgeSpecies), numpy.float64)
+        maxEdgeSpeciesRateRatios = numpy.zeros((numPrunableSpecies), numpy.float64)
         for reactionSystem in reactionSystems:
-            for i in range(numEdgeSpecies):
+            for i in range(numPrunableSpecies):
                 rateRatio = reactionSystem.maxEdgeSpeciesRateRatios[i]
                 if maxEdgeSpeciesRateRatios[i] < rateRatio:
                     maxEdgeSpeciesRateRatios[i] = rateRatio
 
-            for i in range(len(self.networkList)):
-                network = self.networkList[i]
+            for i,network in enumerate(prunableNetworks):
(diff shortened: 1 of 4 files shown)


mjohnson541:
The trick with this however is that the edge species have changed between when you start pruning and when you finish pruning.  So pruning has been reorganized to only track and manage the species that were in the edge before a given reactor cycle started.

rmgpy/solver/base.pyx
```diff
@@ -224,21 +227,40 @@ cdef class ReactionSystem(DASx):
...
     def initialize_solver(self):
         DASx.initialize(self, self.t0, self.y0, self.dydt0, self.senpar, self.atol_array, self.rtol_array)
     
+    def set_prunable_indices(self,edgeSpecies,pdepNetworks):
+        cdef object spc
+        cdef list temp
+        temp = []
+        for i,spc in enumerate(self.prunableSpecies):
+            if spc in edgeSpecies:
+                temp.append(edgeSpecies.index(spc))
```
mliu49: Idea: Could this be replaced with a try/except statement so we don't need to search the list for the species twice?
```
try:
    temp.append(edgeSpecies.index(spc))
except ValueError:
    self.maxEdgeSpeciesRateRatios[i] = numpy.inf

Comment (mjohnson541) [reply rmgpy/solver/base.pyx]: Ok, this should be changed.

Comment (mjohnson541) [reply rmgpy/solver/base.pyx]: (This is fixed)

**Comment by mjohnson541** [comment]
I've been using this branch pretty extensively for speed up for quite a while now and I've had no problems at all with it.  Can we get this merged?

File: rmgpy/solver/base.pyx [hunk 2]
```diff
@@ -224,21 +227,40 @@ cdef class ReactionSystem(DASx):
...
         DASx.initialize(self, self.t0, self.y0, self.dydt0, self.senpar, self.atol_array, self.rtol_array)
     
+    def set_prunable_indices(self,edgeSpecies,pdepNetworks):
+        cdef object spc
+        cdef list temp
+        temp = []
+        for i,spc in enumerate(self.prunableSpecies):
+            try:
+                temp.append(edgeSpecies.index(spc))
+            except:
```
mliu49: Could you add an exception class here? I think a ValueError is raised if an item is not in the list.

mliu49:
Sorry, I forgot about this. I can merge it once you address the comment I just made.

mjohnson541:
Ok that should be fixed.

rmgpy/rmg/model.py
```diff
@@ -1167,34 +1167,36 @@ def prune(self, reactionSystems, toleranceKeepInEdge, maximumEdgeSpecies, minSpe
...
-            for i in range(numEdgeSpecies):
+            for i in range(numPrunableSpecies):
                 rateRatio = reactionSystem.maxEdgeSpeciesRateRatios[i]
                 if maxEdgeSpeciesRateRatios[i] < rateRatio:
                     maxEdgeSpeciesRateRatios[i] = rateRatio
 
-            for i in range(len(self.networkList)):
-                network = self.networkList[i]
+            for i in range(len(prunableNetworks)):
+                network = prunableNetworks[i]
```
mliu49: I realize this wasn't really part of this PR, but could you replace this with `enumerate`? Sorry to keep adding more comments. I'll definitely merge after this.

mjohnson541:
Ok, are we good now?

Issue #1221, opened 2017-10-20
Duplicate species in a model?


mjohnson541:
Conducting simulation of reaction system 1...
initializing surface ...
surface initialization complete
At time 7.4247e-09 s, PDepNetwork #5 at 0.0117306639562 exceeded the minimum rate for simulation interruption of 0.01
Reached max number of objects...preparing to terminate
At time 7.4247e-09 s, PDepNetwork #5 at 0.0117306639562 exceeded the minimum rate for exploring of 0.01
terminating simulation due to interrupt...
Making seed mechanism...
Traceback (most recent call last):
  File "/Users/mattjohnson/RMGCODE/RMG-Py/rmg.py", line 173, in <module>
    main()
  File "/Users/mattjohnson/RMGCODE/RMG-Py/rmg.py", line 167, in main
    rmg.execute(**kwargs)
  File "/Users/mattjohnson/RMGCODE/RMG-Py/rmgpy/rmg/main.py", line 637, in execute
    self.makeSeedMech()
  File "/Users/mattjohnson/RMGCODE/RMG-Py/rmgpy/rmg/main.py", line 841, in makeSeedMech
    shortDesc = species.thermo.comment
  File "/Users/mattjohnson/RMGCODE/RMG-Py/rmgpy/data/thermo.py", line 604, in loadEntry
    raise DatabaseError('Adjacency list and multiplicity of {0} matches that of existing molecule {1} in thermo library {2}.  Please correct your library.'.format(label, entry.label, self.name))
rmgpy.exceptions.DatabaseError: Adjacency list and multiplicity of O2-2 matches that of existing molecule O2 in thermo library Syngas.  Please correct your library.  

I got this error running a syngas example on master.  I haven't yet been able to duplicate this in one of the standard examples yet, but I haven't tried all of them.  I've confirmed in the dictionary file that the O2's (O2(3) and O2(15)) are identical.  I also got a curious adjacency list for OO:  

OO
1 O u0 p1 c+1 {2,S} {3,S} {4,S}
2 O u0 p3 c-1 {1,S}
3 H u0 p0 c0 {1,S}
4 H u0 p0 c0 {1,S}

mjohnson541:
Also 
HO2(6)
multiplicity 2
1 H u0 p0 c0 {2,S}
2 O u1 p1 c+1 {1,S} {3,S}
3 O u0 p3 c-1 {2,S}

mjohnson541:
[...]

PR #1223, opened 2017-10-20
Revise findAllDelocalizationPathsLonePairRadical to allow only NO2

alongd:
Since the reactivity of structures such as `R[O+.][O-]` (formed from `RO[O.]`) isn't yet implemented in RMG, the previous changes to `findAllDelocalizationPathsLonePairRadical` are reverted here to account only for NO2-like structures.

### DIFF
--- a/rmgpy/molecule/pathfinder.py
+++ b/rmgpy/molecule/pathfinder.py
@@ -263,24 +263,20 @@ def findAllDelocalizationPaths(atom1):
 def findAllDelocalizationPathsLonePairRadical(atom1):
     """
     Find all the delocalization paths of lone electron pairs next to the radical center indicated
-    by `atom1`. Used to generate resonance isomers for species such as RN[O], RO[NH], RO[O], RN[NH].
-    (currently only N and O are considered, will soon include S as well)
+    by `atom1`. Used to generate resonance isomers in adjacent N and O as in NO2.
     """
     cython.declare(paths=list)
     cython.declare(atom2=Atom, bond12=Bond)
 
     paths = []
-    # In a first step we only consider nitrogen and oxygen atoms as possible radical centers
-    if ((atom1.isNitrogen() and atom1.lonePairs in [0,1])
-            or (atom1.isOxygen() and atom1.lonePairs in [1,2]))\
-            and atom1.radicalElectrons > 0:
+    if atom1.isNitrogen() and atom1.radicalElectrons >= 1 and atom1.lonePairs == 0:
         for atom2, bond12 in atom1.edges.items():
-            if bond12.isSingle():  # Only single bonds are considered
-                # Neighboring atom must posses a lone electron pair to loose it
-                if ((atom2.isNitrogen() and atom2.lonePairs in [1,2])
-                        or (atom2.isOxygen() and atom2.lonePairs in [2,3]))\
-                        and atom2.radicalElectrons == 0:
-                    paths.append([atom1, atom2])
+            if atom2.isOxygen() and atom2.radicalElectrons == 0 and atom2.lonePairs == 3 and bond12.isSingle():
+                paths.append([atom1, atom2])
+    elif atom1.isOxygen() and atom1.radicalElectrons >= 1 and atom1.lonePairs == 2:
+        for atom2, bond12 in atom1.edges.items():
+            if atom2.isNitrogen() and atom2.radicalElectrons == 0 and atom2.lonePairs == 1 and bond12.isSingle():
+                paths.append([atom1, atom2])
     return paths
 
 def findAllDelocalizationPathsN5dd_N5ts(atom1):
--- a/rmgpy/molecule/resonanceTest.py
+++ b/rmgpy/molecule/resonanceTest.py
@@ -48,22 +48,6 @@ class ResonanceTest(unittest.TestCase):
         self.assertEqual(len(molList), 3)
         self.assertTrue(any([any([atom.charge != 0 for atom in mol.vertices]) for mol in molList]))
 
-    def ROO(self):
-        """Test resonance structure generation for RO[O]
-
-        Selected case for lone pair <=> radical resonance"""
-
-        molList = generateResonanceStructures(Molecule(SMILES="CO[O]"))
-        self.assertEqual(len(molList), 2)
-
-    def RNN(self):
-        """Test resonance structure generation for RN[N]
-
-        Selected case for lone pair <=> radical resonance"""
-
-        molList = generateResonanceStructures(Molecule(SMILES="CN[NH]"))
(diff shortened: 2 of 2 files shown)


alongd:
This solves #1221 and ready for review. @mliu49 , could you take a look?

rmgpy/molecule/pathfinder.py
```diff
@@ -263,24 +263,20 @@ def findAllDelocalizationPaths(atom1):
...
     cython.declare(atom2=Atom, bond12=Bond)
 
     paths = []
-    # In a first step we only consider nitrogen and oxygen atoms as possible radical centers
-    if ((atom1.isNitrogen() and atom1.lonePairs in [0,1])
-            or (atom1.isOxygen() and atom1.lonePairs in [1,2]))\
-            and atom1.radicalElectrons > 0:
+    if atom1.isNitrogen() and atom1.radicalElectrons == 1 and atom1.lonePairs == 0:
         for atom2, bond12 in atom1.edges.items():
-            if bond12.isSingle():  # Only single bonds are considered
```
mliu49: Do you not need if it's a single bond anymore?

mliu49:
@mjohnson541, could you run the test case you had to confirm that this fixes the issues?

rmgpy/molecule/pathfinder.py
alongd: good point, it was unintentionally omitted. Now added back.

mliu49:
The method is now slightly more restrictive than before. Previously, atom 1 had to have at least 1 radical, now it can only have exactly 1. Also, (N with 0 LP connected to N with 1 LP) or (O with 2 LP connected to O with 3 LP) were previously acceptable paths, but are now not acceptable.

Running RMG-tests locally shows some differences compared to 673d9dc6e4a1b706c8ff0318fd8934162e9a6f7a, before #1214. Run times are slightly faster, presumably due to fewer resonance structures (which is good). However, there were model discrepancies in the MCH example.

Action item: Decide whether the more restrictive algorithm is desirable.

---|---|---|---|---
eg1|00:00:02:17|00:00:02:08|435.99|441.10
eg3|00:00:05:03|00:00:04:35|396.78|397.55
eg7|00:00:00:42|00:00:00:36|336.80|339.08
NC|00:00:01:51|00:00:01:45|342.57|342.59
solvent_hexane|00:00:00:54|00:00:00:54|332.36|330.06
MCH|00:00:03:33|00:00:03:18|691.32|682.37

```
core for MCH:
Test model has 49 species.
Original model has 50 species.
Test model has 166 reactions.
Original model has 171 reactions.
The original model has 1 species that the tested model does not have.
spc: ipropyl(119)
The original model has 5 reactions that the tested model does not have.
rxn: H(3) + C3H6(29) <=> allyl(107) + H2(6)             origin: H_Abstraction
rxn: npropyl(86) <=> ipropyl(119)               origin: intra_H_migration
rxn: ipropyl(119) + C2H3(25) <=> C2H4(27) + C3H6(29)            origin: Disproportionation
rxn: H(3) + C3H6(29) <=> ipropyl(119)           origin: R_Addition_MultipleBond
rxn: H(3) + C2H5(31) <=> H2(6) + C2H4(27)               origin: Disproportionation
edge for MCH:
Test model has 376 species.
Original model has 376 species.
Test model has 1140 reactions.
Original model has 1146 reactions.
The original model has 6 reactions that the tested model does not have.
rxn: H(3) + C3H6(29) <=> allyl(107) + H2(6)             origin: H_Abstraction
rxn: H(3) + C3H6(29) <=> ipropyl(119)           origin: R_Addition_MultipleBond
rxn: H(3) + C2H5(31) <=> H2(6) + C2H4(27)               origin: Disproportionation
rxn: H(3) + C3H6(29) <=> propen2yl(153) + H2(6)         origin: H_Abstraction
rxn: propen1yl(132) + H2(6) <=> H(3) + C3H6(29)         origin: H_Abstraction
rxn: C3H6(T)(68) + H(3) <=> ipropyl(119)                origin: R_Recombination
```

mliu49:
As a side note, current master (with the expanded lone pair <=> radical resonance) times out on RMG-tests, and eg3 was unfinished after running for over 20 minutes.

alongd:
Thanks for looking into it, @mliu49 .

Your side note is indeed concerning.

I ran the MCH test locally for (1) current master (10f1d26), (2) the 673d9dc benchmark, (3) master + the current PR, and (4) checking out the 673d9dc benchmark and manually changing `findAllDelocalizationPathsLonePairRadical()` in pathfinder.py to the version in this PR (i.e., not considering other unrelated changes introduced into master since the 673d9dc benchmark).

#|run|time|memory|core species|edge species
---|---|---|---|---|---
1|master|2:16|777.78|59|476
2|673d9dc benchmark|1:47|686.65|49|327
3|current PR rebased on master|1:49|696.91|50|326
4|current PR manually based on benchmark|1:48|685.80|49|327

My observations are:
- The current PR alone did not introduce any significant change to the MCH test vs. the benchmark
- The current master deviates significantly from the benchmark w.r.t. MCH
- The current PR rebased on master gives closer results to the benchmark than master

Indeed, the current PR is more restrictive than the original code. The original code considered any atom1/atom2 combination from the table below:

atom1|atom2
---|---
`N u1 p0`|`N u0 p1`
`O u1 p2`|`O u0 p3`

This is understood in light of the resonance it was tailored for:
image
In other words, it was meant for the following two combinations:
- `O u1 p2` adjacent to `N u0 p1` (LHS), or
- `N u1 p0` adjacent to `O u0 p3` (RHS).

I did not want to allow the remaining two combinations in this PR:
- `O u1 p2` adjacent to `O u0 p3`: This structure shouldn't exist in RMG (monorad charged O2, `[::O.][O:::-]`)
- `N u1 p0` adjacent to `N u0 p1`: This structure could be possible, I will consider it in the heteroatom resonance reexamination PR currently in the oven (`[NH3.][:NH2]` <-> `[:NH3-][NH2+.]`)

In any case, these restrictions should not affect our tests. I thought, based on our recent past experience, that this resonance transformation should be rewritten in a more restrictive way, allowing only for the desired transformations. Obviously, I did a poor job communicate this correctly when I made this PR. I still think that the current restrictive form is better, but I'm also OK with restoring the original code for this path (we shouldn't be forming the monorad charged O2 anyway).

alongd:
Right, now the path is restricted to just one radical on atom1, and I think I should allow at least one, as you correctly observed, @mliu49 .
This will allow birad transformations like: `C[:N.][::O.]`<->`C[N..+][:::O-]`. Weird, but principally should be allowed. I'll correct accordingly.

mjohnson541:
I've tested the current version of this on my syngas example and I no longer encounter #1221.

Issue #1220, opened 2017-10-20
PDep jobs fail due to Species 'collisionModel' AttributeError


alongd:
Running the ethane-oxidation example on current master (an example with PDep) gives the following trace:

```
After model enlargement:
    The model core has 6 species and 0 reactions
    The model edge has 0 species and 0 reactions


Species with SMILES of [H] named H based on thermo library name
Species with SMILES of [CH2] named CH2(S) based on thermo library name
Species with SMILES of C named CH4 based on thermo library name
Generating kinetics for new reactions...
For reaction CH2(S)(6) + CH4(7) <=> ethane(1) Ea raised from -1.5 to 0 kJ/mol.
Updating 1 modified unimolecular reaction networks (out of 1)...
Updating PDepNetwork #1
Warning: For CC, more characteristic frequencies were generated than vibrational modes allowed. Removed 1 groups (6 frequencies) to compensate.
Warning: Maximum number of iterations reached when fitting spectral data for CC.
Warning: Statmech node 'C_R1' and all its parents have data = None
Warning: For C[CH2], more characteristic frequencies were generated than vibrational modes allowed. Removed 1 groups (2 frequencies) to compensate.
Warning: Statmech node 'C_R0' and all its parents have data = None
Warning: Statmech node 'C_R0' and all its parents have data = None
Traceback (most recent call last):
  File "/home/alongd/Code/RMG-Py//rmg.py", line 173, in <module>
    main()
  File "/home/alongd/Code/RMG-Py//rmg.py", line 167, in main
    rmg.execute(**kwargs)
  File "/home/alongd/Code/RMG-Py/rmgpy/rmg/main.py", line 567, in execute
    bimolecularReact=self.bimolecularReact)
  File "/home/alongd/Code/RMG-Py/rmgpy/rmg/model.py", line 683, in enlarge
    self.updateUnimolecularReactionNetworks()
  File "/home/alongd/Code/RMG-Py/rmgpy/rmg/model.py", line 1642, in updateUnimolecularReactionNetworks
    network.update(self, self.pressureDependence)
  File "/home/alongd/Code/RMG-Py/rmgpy/rmg/pdep.py", line 545, in update
    spec.collisionModel = SingleExponentialDown(alpha0=(4.86,'kcal/mol'))
AttributeError: 'rmgpy.species.Species' object has no attribute 'collisionModel'
```

mjohnson541:
I suspect this is a result of pulling in #1190.

[...]

PR #1224, opened 2017-10-20
remove collisionModel Attribute definition

mjohnson541:
The attribute is unused and its definition is the cause of #1220.  It started causing #1220 because now that both Species classes have been combined into one cythonized Species calss it can no longer add new attributes to the object as it did when it was added to the uncythonized Species class.

### DIFF
--- a/rmgpy/rmg/pdep.py
+++ b/rmgpy/rmg/pdep.py
@@ -540,9 +540,8 @@ class PDepNetwork(rmgpy.pdep.network.Network):
         bathGas = [spec for spec in reactionModel.core.species if not spec.reactive]
         self.bathGas = {}
         for spec in bathGas:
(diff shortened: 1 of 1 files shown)


goldmanm:
I added a small change to the commented out piece of code to remove reference to `alpha0`

mliu49:
Now that RMG-tests should be working again, I will merge this once it finishes.

mliu49:
Well, I guess I spoke too soon. This will need to wait until #1227 is fixed.

mliu49:
You lost Mark's fixup for the comment. I'm going to force push again.

mjohnson541:
Oh, sorry, I only saw the first few words.

mliu49:
That's fine. It'll need to be squashed before merging though.

mjohnson541:
@mliu49 can you merge this?

PR #1228, opened 2017-10-24
Testing to enable status updating by RMG-tests

mliu49:
Testing GitHub API. Please ignore.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -63,6 +63,9 @@ git clone ${REPO} ${TARGET_DIR}
 # go inside the newly created folder:
 cd $TARGET_DIR
 
+# checkout test branch
+git checkout --track origin/redesign
+
 # create a new branch in RMG-tests with the name equal to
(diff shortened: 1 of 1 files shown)


mliu49:
I figured I should update the status of this task. Using the GitHub API is pretty straightforward, and the result is great, as can be seen below in the status box. However, right now there is no way to send the completion status report since the RMG-tests builds always time out (which is why the status shows pending).

I've contacted Travis CI support about this, and they said that it's a known issue that `after_success` and `after_failure` job steps cannot be executed if the job times out. Therefore, the only way to have it successfully update the status upon completion is to not have the build time out.

Based on that, we have a few different options if we want automatic status updates from RMG-tests
1. Significantly pare down RMG-tests such that it can finish within 50 minutes
2. Develop automated testing on our own servers without run time limitations
3. Switch to the paid Travis plan (most unlikely)

mliu49:
Status updates are working and are now live on RMG-tests.

Issue #1227, opened 2017-10-24
RMG-tests still errors on eg1


mliu49:
```
Traceback (most recent call last):
 File "/home/travis/build/ReactionMechanismGenerator/RMG-tests/code/benchmark/RMG-Py/scripts/checkModels.py", line 288, in <module>
 main()
 File "/home/travis/build/ReactionMechanismGenerator/RMG-tests/code/benchmark/RMG-Py/scripts/checkModels.py", line 80, in main
 check(name, benchChemkin, benchSpeciesDict, testChemkin, testSpeciesDict)
 File "/home/travis/build/ReactionMechanismGenerator/RMG-tests/code/benchmark/RMG-Py/scripts/checkModels.py", line 92, in check
 execute(benchChemkin, benchSpeciesDict, benchThermo, testChemkin, testSpeciesDict, testThermo, **kwargs)
 File "/home/travis/build/ReactionMechanismGenerator/RMG-tests/code/benchmark/RMG-Py/rmgpy/tools/diff_models.py", line 313, in execute
 model1.species, model1.reactions = loadChemkinFile(chemkin1, speciesDict1, thermoPath = thermo1)
 File "/home/travis/build/ReactionMechanismGenerator/RMG-tests/code/benchmark/RMG-Py/rmgpy/chemkin.py", line 884, in loadChemkinFile
 reactionList = readReactionsBlock(f, speciesDict, readComments = readComments)
 File "/home/travis/build/ReactionMechanismGenerator/RMG-tests/code/benchmark/RMG-Py/rmgpy/chemkin.py", line 1307, in readReactionsBlock
 raise e
rmgpy.exceptions.ChemkinError: Unexpected species identifier C_methyl;C_rad/H2/Cs encountered in flux pairs for reaction C3H7(15) + CH3(2) <=> CCCC(11).
```

Correctly splitting the kinetics comment at `family:` in 10f1d267a11bbee726cec4950f7c3ffe9a4f1ed7 did not fully address the issue.

mliu49:
Example of problematic reaction comment:
```
! Reaction index: Chemkin #174; RMG #1055
! Template reaction: Disproportionation
! Flux pairs: O2(2), HO2(10); C4H5(257), C4H4(332); 
! O2b;Cds/H2_d_Crad from training reaction 2
! Exact match found for rate rule [O2b;Cds/H2_d_Crad]
! Euclidian distance = 0
[...]

PR #1229, opened 2017-10-25
change kinetics commenting of training reactions

goldmanm:
Previously kinetics comments from training reactions started
with the rate rule. Due to the way loadChemkinFile parses the
comments by identifying starts of lines, this caused parsing errors.
This commit reworded that line of the comments to start with
identifiable words. It also updated the uncertainty module parsing
to work with this change as well as updated the unittests.

### DIFF
--- a/rmgpy/chemkinTest.py
+++ b/rmgpy/chemkinTest.py
@@ -390,7 +390,7 @@ This reaction matched rate rule [C_methyl;C_methyl]
 Reaction index: Chemkin #2; RMG #4
 Template reaction: R_Recombination
 Flux pairs: [CH3], CC; [CH3], CC; 
-From training reaction 21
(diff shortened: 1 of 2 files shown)


goldmanm:
This pull request can be tested locally:

1. run 'make eg1'
2. stop the run after a minute or two
2. test that the proper string was created with `grep training chem_edge_annotated.inp`
3. in `testing/eg1/chemkin`, import chem_edge_annotated.inp and species_edge_dictionary.txt into loadChemkinFile.

goldmanm:
This PR fixes #1227

mjohnson541:
So this doesn't appear to actually solve #1227 according to the current RMG-tests run.

mliu49:
Talk about timing...

I think the RMG-tests is failing due to the benchmark failing. It's a really annoying side effect. We can take a look at the build for master to confirm.

Issue #1192, opened 2017-09-05
Website kinetics search broken


mliu49:
Recent changes to how reaction are generated in RMG-Py has broken the kinetics search functionality of the website. The live website is still using an old version of RMG-Py, so it doesn't interfere with website usage at the moment. However, the website cannot be updated until these issues are addressed.

For the kinetics search function, the website uses `KineticsDatabase.generateReactionsFromFamilies()` which calls `KineticsFamily.generateReactions()`. Previously, that was sufficient to get properly processed reactions. However, the post-processing has now been moved to the `rmgpy.rmg.react` module.

Additionally, group additivity for kinetics estimation now works again (#1061), so the website will actually give group additivity estimates. However, I have no idea whether or not those estimates are good or not.

Finally, the website doesn't handle multiple transition states at all, so changes will need to be made there as well.

goldmanm:
I can refactor `KineticsDatabase.generateReactionsFromFamilies()` to use `react.reactSpecies()` instead of `KineticsFamily.generateReactions()`, so that no extra work is needed on the website for that issue....if that would be helpful. 

Maybe I'll manage to write a unittest for `generateReactionsFromFamilies` while I am at it :)

mliu49:
Thanks, I don't think you need to work on it right away though. There are some broader organizational issues that I'm not quite sure about and would like to discuss with you at some point after I sort out my thoughts.

mliu49:
The website has been updated following #1231, so this issue is now fixed.

PR #1231, opened 2017-10-30
Refactor reaction generation

mliu49:
This pull request does significant restructuring of the methods related to reaction generation, motivated by #1192. These changes are intended to reduce code duplication, improve organization, and reduce cyclic import risks. For those interested, a basic [function call map](https://github.com/ReactionMechanismGenerator/RMG-Py/files/1428453/Reaction.Generation.pdf) was made to assist with the restructuring.

Essentially, the primary reaction generation method is now `KineticsDatabase.generate_reactions_from_families` instead of `rmgpy.rmg.react.reactSpecies`. Supporting methods were moved to `rmgpy.data.kinetics.common` in cases where usage was shared with `KineticsFamily` methods. The `react` module is again concerned only with parallelization of reaction generation, with `reactSpecies` calling `generate_reactions_from_families`. The `kineticsTest` module was also significantly refactored to use the updated methods.

Technically, there should be no effects on normal model generation, so anything that does come up will need to be investigated. This will affect how scripts and ipython notebooks handle reaction generation though. The process is now more streamlined and intuitive if only generating reactions:
1. Load KineticsDatabase
2. Call KineticsDatabase.generate_reactions

Of course, more steps are needed if also generating kinetics.


mliu49:
New commits remove the need for `filter_reactions` by allowing `generate_reactions_from_families` to pass products all the way to `__generateReactions`, which is able to filter reactions by products. Also `Reaction.matchSpecies` was updated to also accept and check products so that `generate_reactions_fromLibraries` also does require `filter_reactions` anymore.

Since `filter_reactions` is no longer used anywhere, I added a deprecation warning. I thought that it might be good to keep around for a while longer, but I would be fine with completely removing it if other people think that it would be better.

mliu49:
Added new commits:
- New clear_labels argument allows retention of atom labels, useful for training reactions script
- The resonance argument has been extended to products, by passing the prod_resonance argument down to `__generateReactions`. This gives slightly more intuitive behavior, in that turning resonance off will give only the reactions that match the specified product resonance structures.

Also rebased onto master.

rmgpy/reaction.py
```diff
@@ -360,17 +360,29 @@ def hasTemplate(self, reactants, products):
...
         The reactants should be :class:`Molecule` objects.
+
+        If ``products`` are provided, then they must match as well.
         """
         # Check forward direction
         if _isomorphicSpeciesList(self.reactants, reactants):
-            return True
+            if products is None:
+                return True
+            elif _isomorphicSpeciesList(self.products, products):
```
KEHANG: can you merge `if` and `elif` since they both produce True.

rmgpy/data/kinetics/family.py
```diff
@@ -2029,7 +2029,7 @@ def getLabeledReactantsAndProducts(self, reactants, products):
         If RMG cannot find appropriate labels, (None, None) will be returned.
         """
```
KEHANG: the docstring should be updated since now the input reactants and products are `Species` type?

rmgpy/data/kinetics/family.py
```diff
@@ -2029,7 +2029,7 @@ def getLabeledReactantsAndProducts(self, reactants, products):
...
         template = self.forwardTemplate
-        reactants0 = [reactant.copy(deep=True) for reactant in reactants]
+        reactants0 = [reactant.molecule[0].copy(deep=True) for reactant in reactants]
         if len(reactants0) == 1:
```
KEHANG: Why not catch that before? maybe should add unittests?

rmgpy/data/kinetics/database.py
```diff
@@ -408,7 +408,7 @@ def generate_reactions_from_library(self, library, reactants, products=None):
 
         return reaction_list
 
-    def generate_reactions_from_families(self, reactants, products=None, only_families=None, resonance=True):
+    def generate_reactions_from_families(self, reactants, products=None, only_families=None, resonance=True, clear_labels=True):
```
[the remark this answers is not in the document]
mliu49: I'm actually planning on removing the commits related to the clear_labels argument, since it didn't work as expected. In the convert library to training ipynb, we generate reactions from families then regenerate the atom labels, so I thought it would be easier to not clear the labels to begin with so we wouldn't need to regenerate them. It turned out to be much more complicated though.

rmgpy/data/kinetics/family.py
mliu49: I think this method is only used in the convert library to training ipynb. With the changes to generateReactionsFromFamilies, this method no longer worked properly. I can add unit tests.

rmgpy/data/kinetics/database.py
```diff
@@ -400,37 +404,83 @@ def generateReactionsFromLibrary(self, reactants, products, library):
...
+                reactants[1] = reactants[1].copy(deep=True)
+                same_reactants = True
+            elif reactants[0].isIsomorphic(reactants[1]):
+                same_reactants = True
+
+        # Convert to Species objects if necessary
+        reactants = ensure_species(reactants)
+
+        # Label reactant atoms for proper degeneracy calculation
+        label_list_of_species(reactants, resonance=resonance)
```
KEHANG: rename this method?

rmgpy/data/kinetics/database.py
```diff
@@ -412,10 +412,20 @@ def generate_reactions_from_families(self, reactants, products=None, only_famili
...
         Generate all reactions between the provided list or tuple of one or two
         `reactants`, which can be either :class:`Molecule` objects or :class:`Species`
-        objects. This method applies all of the loaded reaction families.
+        objects. This method can apply all kinetics families or a selected subset.
 
-        If `only_families` is a list of strings, only families with those labels
-        are used.
+        Args:
+            reactants:      Molecules or Species to react
+            products:       Desired product structures (optional)
```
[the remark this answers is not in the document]
mliu49: Updated

rmgpy/data/kinetics/database.py
```diff
@@ -412,10 +412,20 @@ def generate_reactions_from_families(self, reactants, products=None, only_famili
...
+            only_families:  List of family labels to generate reactions from (optional)
+                            Default is to generate reactions from all families
+            resonance:      Flag to generate resonance structures for reactants and products (optional)
+                            Default is True, resonance structures will be generated
+            clear_labels:   Flag to clear atom labels from reactants after reaction generation (optional)
+                            Default is True, labels will be cleared
+
+        Returns:
+            List of reactions with the specified reactants and products.
```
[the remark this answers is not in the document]
mliu49: Updated

rmgpy/reaction.py
```diff
@@ -360,35 +360,19 @@ def hasTemplate(self, reactants, products):
             (all([spec in self.products for spec in reactants]) and
             all([spec in self.reactants for spec in products])))
 
-    def matchesMolecules(self, reactants):
+    def matchesSpecies(self, reactants):
         """
         Return ``True`` if the given ``reactants`` represent the total set of
         reactants or products for the current ``reaction``, or ``False`` if not.
         The reactants should be :class:`Molecule` objects.
```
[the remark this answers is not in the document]
mliu49: Fixed

rmgpy/rmg/react.py
```diff
@@ -68,32 +68,10 @@ def reactSpecies(speciesTuple):
     given one species tuple, will find the reactions and remove degeneracy
     from them.
     """
```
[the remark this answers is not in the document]
mliu49: Updated

rmgpy/data/kinetics/database.py
mliu49: Changed to `ensure_independent_atom_ids`

mliu49:
I addressed most of your comments. I'm planning to make some additional changes after you review the fixup commits that I've added:

- [ ] Rebase and squash fixup commits
- [ ] Remove commits related to the clear_labels attribute. It did not work as I was expecting, and @goldmanm was working on a similar idea that is likely better implemented.
- [ ] Move commits related to fixing the ipython notebook for training reactions to a different branch for further work. The commits I made were quick fixes, and need a bit more thought and unit tests. They are also not very related to the remainder of this PR.

@KEHANG, please let me know when you're ready for me to rebase this PR.

KEHANG:
@mliu49 the fixup commits look good. Please go ahead squash and rebase.

rmgpy/data/kinetics/database.py
```diff
@@ -400,37 +404,83 @@ def generateReactionsFromLibrary(self, reactants, products, library):
...
+                reactants[1] = reactants[1].copy(deep=True)
+                same_reactants = True
+            elif reactants[0].isIsomorphic(reactants[1]):
+                same_reactants = True
+
+        # Convert to Species objects if necessary
+        reactants = ensure_species(reactants)
+
+        # Label reactant atoms for proper degeneracy calculation
+        ensure_independent_atom_ids(reactants, resonance=resonance)
```
KEHANG: if `resonance =True`, this method only generate resonance isomers when ids are not independent, possible non-deterministic behavior?

rmgpy/data/kinetics/family.py
```diff
@@ -1605,7 +1606,8 @@ def __generateReactions(self, reactants, products=None, forward=True, prod_reson
                             Default is True, labels will be cleared
 
         Returns:
-            List of all reactions with the specified reactants and products within this family.
+            List of all reactions containing Species objects with the
```
KEHANG: I think `family.generateReactions` and `__generateReactions` returns reactions containing `Molecule` objects instead of `Species` objects. `generate_reactions_from_families` indeed returns  reactions containing `Species` objects because it calls `find_degenerate_reactions` which calls `ensure_species_in_reaction`. Can you double-check that? also if that's true, is it good to have this Molecule-Species conversion happen so implicitly?

KEHANG:
@mliu49 I checked again and added a couple more comments.

rmgpy/data/kinetics/database.py
mliu49: You're right, I updated it to also generate resonance structures if the IDs are already independent.

rmgpy/data/kinetics/family.py
mliu49: Thanks for catching that! They do return reactions with `Molecule` objects. I don't think there's much we can do about the conversion though. Let me know if you have any ideas.

mliu49:
I've rebased and cleaned up this branch. I think it is ready to be merged, pending any further comments.

KEHANG:
Good! Before I merge it, can you summarize the behaviour of this branch when running RMG-tests?

mliu49:
From RMG-tests:

---|---|---|---|---
eg1|00:00:01:44|00:00:01:43|435.01|439.63
eg3|00:00:03:59|00:00:04:10|396.23|395.57
eg5|00:00:01:15|00:00:01:14|355.48|356.08
eg6|00:00:00:47|00:00:00:42|341.17|340.43
eg7|00:00:00:26|00:00:00:25|337.60|337.94
NC|00:00:01:07|00:00:01:07|342.25|340.77
MCH|00:00:02:37|00:00:02:37|661.82|661.57

PR #1241, opened 2017-11-16
Fix ch3no2 example

mjohnson541:
Adds a simulator block to the ch3no2 example.  This is necessary after the Accelerator Tools pull request that enables model concatenation


alongd:
OK.
Could you just add the `atol=`, `rtol=` for readability?

mjohnson541:
Ok, that should be added now.

PR #1243, opened 2017-11-17
Fix Uncertainty Sensitivity Analysis

mjohnson541:
adapt uncertainty.sensitivityAnalysis to new call syntax.

### DIFF
--- a/rmgpy/tools/uncertainty.py
+++ b/rmgpy/tools/uncertainty.py
@@ -585,7 +585,7 @@ class Uncertainty:
         from rmgpy.quantity import Quantity
         from rmgpy.tools.sensitivity import plotSensitivity
         from rmgpy.rmg.listener import SimulationProfileWriter, SimulationProfilePlotter
-
+        from rmgpy.rmg.settings import ModelSettings, SimulatorSettings
         T = Quantity(T)
         P = Quantity(P)
         termination=[TerminationTime(Quantity(terminationTime))]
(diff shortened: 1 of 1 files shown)


mjohnson541:
@mliu49 can you test this fixes your problem?

rmgpy/tools/uncertainty.py
```diff
@@ -605,14 +605,22 @@ def sensitivityAnalysis(self, initialMoleFractions, sensitiveSpecies, T, P, term
         reactionSystem.attach(SimulationProfilePlotter(
             self.outputDirectory, reactionSystemIndex, self.speciesList))
         
+        simulatorSettings = SimulatorSettings() #defaults
+        
+        modelSettings = ModelSettings() #defaults
+        modelSettings.fluxToleranceMoveToCore = 0.1
```
mliu49: This was previously 1, does it have any affect since we're running sensitivity?

mjohnson541: If that matters I think we have much bigger problems than whether its set to 0.1 or 1.0.

mjohnson541:
Ok, I think this is ready then.

rmgpy/tools/uncertainty.py
mliu49: I guess I'm wondering if the reactor still does all the flux calculations and identifies species to be moved to the core, even though we aren't actually moving anything. A more direct question might be whether sensitivity analysis would be any faster with a tolerance of 1 vs 0.1?

mjohnson541: The edge is empty it doesn't have anything to check 0.1 or 1.0 against.

mliu49: Thanks, I forgot about that.

Issue #1239, opened 2017-11-13
CanTherm CH3OH+HCO example crashes with a TypeError


alongd:
The `reactions/CH3OH+HCO` CanTherm example crashes with the following trace:
```
Saving kinetics for ch3oh + hco <=> ch3oh + hco...
Traceback (most recent call last):
  File "/home/alongd/Code/RMG-Py//cantherm.py", line 61, in <module>
    cantherm.execute()
  File "/home/alongd/Code/RMG-Py/rmgpy/cantherm/main.py", line 269, in execute
    job.execute(outputFile=outputFile, plot=self.plot)
  File "/home/alongd/Code/RMG-Py/rmgpy/cantherm/kinetics.py", line 127, in execute
    self.save(outputFile)
  File "/home/alongd/Code/RMG-Py/rmgpy/cantherm/kinetics.py", line 227, in save
    kinetics0rev = Arrhenius().fitToData(Tlist, numpy.array(k0revs), kunits=self.krunits)
TypeError: Argument 'Tlist' has incorrect type (expected numpy.ndarray, got list)
```

Looks like the default values assignment [here](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/cantherm/kinetics.py#L152)
```
if Tlist is None:
    Tlist = 1000.0/numpy.arange(0.4, 3.35, 0.05)
```
generate an unexpected var type.

(This example doesn't crash when a `Tlist` is specified in the kinetics block)

alongd:
solved via #1244

PR #1244, opened 2017-11-23
Enable RCCSD(T)-F12 energies in Cantherm

cgrambow:
There's no reason not to have functionality for reading RCCSD(T)-F12 energies from Molpro output files, so this PR enables that functionality.

Also fixes a Cantherm error related to the data format of the Tlist when no Tlist is specified explicitly in the Cantherm input file.

### DIFF
--- a/rmgpy/cantherm/molepro.py
+++ b/rmgpy/cantherm/molepro.py
@@ -66,14 +66,17 @@ class MoleProLog:
         E0=None
         if f12a:
             while line!='':
-                #first one is for radicals second is for non radicals
-                if 'RHF-UCCSD(T)-F12a energy' in line or 'CCSD(T)-F12a total energy  ' in line:
+                if ('RHF-UCCSD(T)-F12a energy' in line
+                    or 'RHF-RCCSD(T)-F12a energy' in line
+                    or 'CCSD(T)-F12a total energy  ' in line):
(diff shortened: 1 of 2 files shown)


alongd:
Should fix #1239

rmgpy/cantherm/molepro.py
```diff
@@ -66,14 +66,17 @@ def loadEnergy(self,frequencyScaleFactor=1.):
         E0=None
         if f12a:
             while line!='':
-                #first one is for radicals second is for non radicals
-                if 'RHF-UCCSD(T)-F12a energy' in line or 'CCSD(T)-F12a total energy  ' in line:
+                if ('RHF-UCCSD(T)-F12a energy' in line
+                    or 'RHF-RCCSD(T)-F12a energy' in line
+                    or 'CCSD(T)-F12a total energy  ' in line):
```
alongd: Was the blank space left here on purpose (`'CCSD(T)-F12a total energy  '`)?

cgrambow: I didn't change that part, but I am fairly certain that the space is not necessary.

alongd: I see. Let's not change it then, unless we can confirm it's unnecessary.

PR #1084, opened 2017-07-12
Allow RMG to handle a specific species as a third body collider

alongd:
Previously RMG could not handle reactions of the sort `A + B (+C) = D (+C)`. Moreover, reactions like `H2 + Ar <=> H + H + Ar` were only handled correctly if the kinetics is of Arrhenius type (like in the BurkeH2O2 libraries), but if the kinetics is in any PDep form, RMG would consider it as `H2 + Ar (+M) <=> H + H + Ar (+M)`, wrongly accounting for an additional third body.

**Usage**
Simple add a specific third body collider to your library reaction as you would intuitively (i.e., no need for a SpecificCollider keyword).
For example (and this rxn is the motivation for this PR):
```
entry(
    index = 226,
    label = "SO2 + O (+N2) <=> SO3 (+N2)",
    degeneracy = 1,
    kinetics = Troe(
        arrheniusHigh = Arrhenius(A=(3.7e+11, 'cm^6/(mol^2*s)'), n=0, Ea=(1689, 'cal/mol'), T0=(1, 'K')),
        arrheniusLow = Arrhenius(A=(2.9e+27, 'cm^9/(mol^3*s)'), n=-3.58, Ea=(5206, 'cal/mol'), T0=(1, 'K')),
        alpha=0.43, T3=(371, 'K'), T1=(7442, 'K'), efficiencies={}),
    shortDesc = u"""[Marshall2005],[Marshall2006]""",
)
```

Another example w/o parenthesis, which RMG should also handle correctly now, is:
```
entry(
    index = 18,
    label = "HNCN + N2 <=> H + NCN +N2",
    degeneracy = 1,
    kinetics = ThirdBody(
        arrheniusLow = Arrhenius(A=(1.79e+28, 'cm^3/(mol*s)'), n=-3.44, Ea=(64502, 'cal/mol'), T0 = (1, 'K'), Tmin=(2000, 'K'), Tmax=(4000, 'K'))),
    shortDesc = u"""[Lin2000a]""",
)
```

[...]

PR #1246, opened 2017-12-07
Add specificCollider to areIdenticalSpeciesReferences()

alongd:
We previously added a specificCollider attribute to the Reaction Class (#1084).
Now RMG recognized these reactions, however it won't add them to the edge if a similar reaction with no specificCollider exists (or vice versa), claiming "This library reaction was not new".
This small fix allows RMG to recognize a reaction as non-identical based on the specificCollider attribute, not only its reactants and products.
This fix was confirmed to work on an unmerged branch and to include `O(6) + SO2(1) (+N2) <=> SO3(4) (+N2)` in addition to `O(6) + SO2(1) <=> SO3(4)`.

### DIFF
--- a/rmgpy/rmg/model.py
+++ b/rmgpy/rmg/model.py
@@ -468,7 +468,8 @@ class CoreEdgeReactionModel:
         # Determine the proper species objects for all reactants and products
         reactants = [self.makeNewSpecies(reactant)[0] for reactant in forward.reactants]
         products  = [self.makeNewSpecies(product)[0]  for product  in forward.products ]
-        if forward.specificCollider is not None: forward.specificCollider = self.makeNewSpecies(forward.specificCollider)[0]
+        if forward.specificCollider is not None:
+            forward.specificCollider = self.makeNewSpecies(forward.specificCollider)[0]
 
         if forward.pairs is not None:
             for pairIndex in range(len(forward.pairs)):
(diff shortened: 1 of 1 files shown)


rmgpy/rmg/model.py
```diff
@@ -1931,6 +1932,6 @@ def areIdenticalSpeciesReferences(rxn1, rxn2):
     are identical, in either direction.
     """
 
-    return any([rxn1.reactants == rxn2.reactants and rxn1.products == rxn2.products, \
-            rxn1.reactants == rxn2.products and rxn1.products == rxn2.reactants
-            ])
+    return all([any([rxn1.reactants == rxn2.reactants and rxn1.products == rxn2.products,
```
mjohnson541: given there aren't many things being compared I think it would be clearer to just avoid one-lining it and do away with  the 'any' and 'all' in favor of 'and's and 'or's

alongd: Done

alongd:
Had some tab\spaces issues, now solved and ready for review\merging.

mjohnson541:
Sorry, if I'm being picky, I've added a commit with what I had in mind.  If what I did makes sense to you can you squash it?

alongd:
It does. Thanks!

mjohnson541:
Can you squash it into yours?

alongd:
Squashed

PR #1247, opened 2017-12-07
Minor kinetics fixes

goldmanm:
This pull request adds a few changes to kinetics: 

* Replaced general exceptions to more specific ones in `family.py` for better error catching
* Improved the robustness of a method by also checking for reverse reactions
* The template is now output when RMG matches a training reaction (since knowing the template is necessary to relabel reactants if the depository isn't loaded)
* Make the default kinetics estimator 'rate rules' instead of group additivity, which I don't think people use much. This allows for the line to be omitted from input files.
* Fix tab error.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1877,7 +1877,10 @@ class KineticsFamily(Database):
                 kineticsList.append([deepcopy(entry.data), entry, entry.item.isIsomorphic(reaction, eitherDirection=False)])
         for kinetics, entry, isForward in kineticsList:
             if kinetics is not None:
-                kinetics.comment += "Matched reaction {0} {1} in {2}".format(entry.index, entry.label, depository.label)
+                kinetics.comment += "Matched reaction {0} {1} in {2}\nThis reaction matched rate rule {3}".format(entry.index, 
+                                                      entry.label, 
+                                                      depository.label,
+                                                      '[{0}]'.format(';'.join([g.label for g in template])))
         return kineticsList
     
     def __selectBestKinetics(self, kineticsList):
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1881,6 +1881,7 @@ class KineticsFamily(Database):
                                                       entry.label, 
                                                       depository.label,
                                                       '[{0}]'.format(';'.join([g.label for g in template])))
+                kinetics.comment += "\nfamily: {}".format(self.label)
         return kineticsList
     
     def __selectBestKinetics(self, kineticsList):
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -53,7 +53,7 @@ from .rules import KineticsRules
 from rmgpy.exceptions import InvalidActionError, ReactionPairsError, KineticsError,\
                              UndeterminableKineticsError, ForbiddenStructureException,\
                              KekulizationError, ActionError, DatabaseError
-
+import itertools
 ################################################################################
 
 class TemplateReaction(Reaction):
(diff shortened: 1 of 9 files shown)


rmgpy/data/kinetics/family.py
```diff
@@ -1877,7 +1877,10 @@ def getKineticsFromDepository(self, depository, reaction, template, degeneracy):
                 kineticsList.append([deepcopy(entry.data), entry, entry.item.isIsomorphic(reaction, eitherDirection=False)])
         for kinetics, entry, isForward in kineticsList:
             if kinetics is not None:
-                kinetics.comment += "Matched reaction {0} {1} in {2}".format(entry.index, entry.label, depository.label)
+                kinetics.comment += "Matched reaction {0} {1} in {2}\nThis reaction matched rate rule {3}".format(entry.index,
```
mliu49: Why is this needed? When we match a training reaction exactly, we technically aren't matching a rate rule.

Also, assuming that this is added, does it break uncertainty?

rmgpy/data/kinetics/family.py
```diff
@@ -2064,30 +2064,35 @@ def getLabeledReactantsAndProducts(self, reactants, products):
         elif len(reactants0) == 2:
             moleculeA = reactants0[0]
             moleculeB = reactants0[1]
+            # get mappings in forward direction
             mappingsA = self.__matchReactantToTemplate(moleculeA, template.reactants[0])
             mappingsB = self.__matchReactantToTemplate(moleculeB, template.reactants[1])
+            mappingPairs = list(itertools.product(mappingsA,mappingsB))
+            # get mappings in the reverse direction
```
mliu49: This method does not check both orderings because it is called by `addAtomLabelsForReaction`, which does check both orderings.

I would suggest seeing if `addAtomLabelsForReaction` is appropriate for your use case, or also modifying that method to account for your changes here.

goldmanm: Thanks for brining this up. I wasn't aware that `addAtomLabelsForReaction` accounted for reverse order. The current issue with using `addAtomLabelsForReaction` for my application is that it is written to require reactants and products as molecule objects and outputs a reaction object with species objects (though without resonance structures).  In my application, I start with a reaction with species objects containing possibly resonance structures, and I need to obtain labeled molecule objects for the reactants. It seems like I would have to some finagling of recreating a bunch of reaction objects to input into `addAtomLabelsForReaction` and then extract out the molecules from their respective Species objects. Though I am fine making this correction (or any others) if that would be preferable.

I guess my thoughts on the code currently:

1. Nothing within `getLabeledReactantsAndProducts` seems to imply that the intended outcome is to throw an error if the order of reactants is wrong. If we want it to throw an error in these cases, we should probably make it a private method so users like me don't go trying to use it unintentionally....though making it private anyways may be a good idea.
2. `addAtomLabelsForReaction` seems to do multiple things in one go, which is not good usage of object oriented design (and this is why it doesn't really work well for my application). The conversion to Species objects seems outside the scope of the method and removing this extraneous part could make the method more applicable for other uses.
3. I couldn't find any usage for `addAtomLabelsForReaction` in any of the RMG-Py or RMG-website repositories (the same is true for `getLabeledReactantsAndProducts`). Neither method is tested. I think another alternative to deal with  `addAtomLabelsForReaction` is to deprecate it for removal in RMG 2.2, and only support `getLabeledReactantsAndProducts`.
4. Having only one public method, either `getLabeledReactantsAndProducts` or a more robust and single-tasked version of `addAtomLabelsForReaction` seems like a code improvement to me.

Let me know what you think the best path forward is, and I can implement it.

rmgpy/data/kinetics/family.py
goldmanm: That's a really good comment (and made me think back a bit). Here's the logic in why I think it is necessary: When loading a reaction from a chemkin file, I think a desired output would be similar information to when the reaction was saved in a chemkin file. Loading reactions from chemkin files does not currently require a loaded database. If only the depository index was saved (as it is before this PR), it would not be possible to repopulate the TemplateReaction.template attribute for reaction objects. In my isotope work, I need the template attribute to be populated so I can ensure that I don't include different transition states when generating reactions with isotopes.

rmgpy/data/kinetics/family.py
mliu49: These two methods were written for `convertKineticsLibraryToTrainingReactions.ipynb` in RMG-database. I was looking into them for some fixes to the ipynb, and found that `addAtomLabelsForReaction` broke with reactions with Species objects , as you mentioned. I had made a temporary hack fix by having the method take the first molecule of each of the inputted species, which would then be converted back to a species at the end. I agree that both methods need to be reworked to accept and return reactions with Species objects.

I've written a basic unit test for `addAtomLabelsForReaction`, located [here](https://github.com/mliu49/RMG-Py/commit/e7cbcba9a991c8dad1c22edcb6516247941d24cc), which you can use or expand on. I think you have a lot of flexibility in however you want to remake these two methods, since they're only used in that one notebook. It would be great if you could condense it into a single method. I assumed that the reason it was broken into two was to reduce code duplication.

rmgpy/data/kinetics/family.py
goldmanm: For the potential to break uncertainty, I did a visual inspection of uncertainty parsing in `family.py` when first making this commit. To double check, I just added the extra line in kinetics which is tested for uncertainty by `TestKineticsParsing` in `kineticsTest.py`, which still passes with the added line.

rmgpy/data/kinetics/family.py
goldmanm: Thanks for the advice. I think it really helped improve the code. I moved over `ensure_species` to make it a function of Reaction objects. I then used that method to make `addAtomLabelsForReaction` much more flexible. It can accept either species- or molecule-based reaction objects and then can output either species- or molecule-based reaction objects. I left the checking-order thing that sparked the discussion in `getLabeledReactantsAndProducts` since it seems more like the intended behavior of the method. 

In the limited case `addAtomLabelsForReaction` was originally used for(input molecule-based reaction object and get out species-based reaction object), the only change to the method is that resonance structures are now output, whereas they were not before. 

The only think I can think of that is still left to be decided is whether we should make `getLabeledReactantsAndProducts` private. I am ambivalent either way (now that `addAtomLabelsForReaction` functions more robustly)

rmgpy/data/kinetics/family.py
```diff
@@ -2130,7 +2152,7 @@ def getTrainingDepository(self):
             if depository.label.endswith('training'):
                 return depository
         else:
-            raise Exception('Could not find training depository in family {0}.'.format(self.label))
+            raise AttributeError('Could not find training depository in family {0}.'.format(self.label))
```
mliu49: Would a `DatabaseError` be more appropriate? I feel like `AttributeError` has a fairly specific scope.

rmgpy/data/kinetics/family.py
```diff
@@ -2059,68 +2063,86 @@ def getLabeledReactantsAndProducts(self, reactants, products):
...
+                # must have gotten the wrong pair
+                pass
+        if labeled_reactants is None or labeled_products is None:
+            raise ActionError("Could not find labeled reactants for reaction {} from family {}.".format(reaction,self.label))
+
+        # place the molecules in reaction object
+        reaction.products = labeled_products
+        reaction.reactants = labeled_reactants
+
+        if output_with_species:
```
mliu49: Is there a use case where output with molecule is preferred?

I think a resonance argument to toggle resonance structure generation might be more useful.

mliu49:
Would you be ok with adding [this commit](https://github.com/mliu49/RMG-Py/commit/c98f423eb223e5cf4b257784c4b557579ce12e75) to this PR as well? It loads database species dictionaries as ordered dictionaries, which helps maintain the order of species when loading and saving.

rmgpy/data/kinetics/family.py
goldmanm: Awesome observation! Fortunately, this comment is used in other parts of the code, so "'This reaction matched rate rule'" is already on the list of line breaks in the master branch.

rmgpy/data/kinetics/family.py
goldmanm: In my isotope work, I use the output molecule objects for some molecular graph operations. 

I also like the idea of only outputting species objects. One potential issue I realized with this method would be the loss of any species-level data (like thermo, symmetry number, etc) when using this method. 

I propose that I make the following changes to the method:

1. only output species objects with the option to create resonance structures
2. store the updated & labeled graphs in the current species objects to prevent loss of any species-level attributes.

How does this sound?

rmgpy/data/kinetics/family.py
goldmanm: Sounds great! I'll make this change. :)

rmgpy/data/kinetics/family.py
mliu49: Ah, I see.

rmgpy/data/kinetics/family.py
mliu49: I think that sounds good. Will using the original species objects be straightforward to implement?

goldmanm:
Thanks again for the suggestions. Definitely helped me improve the code. Just to let you know about the changes:

1. rebased the more appropriate error message into the original commit
2. added a new commit changing the output of addLabelsToReaction to what we discussed
3. added the small commit you recommended.
4. rebased to current master.

Let me know if you have any other thoughts on this.

goldmanm:
@mliu49, is it possible to merge this? The only failing tests are the thermo central database, since those were down at the time of running (and possibly still are down). 

I'm restarting the build just in case those were put back up.

PR #1248, opened 2017-12-07
Minor thermo isotope tweaks

goldmanm:
This PR improves thermo generation and isotope representation (put into one PR since each is small leading to hassle with re-basing before each merge). 

Thermo fixes:

* Adds a method that modifies the entropy of a NASA object without converting it to wilholt, which leads to some extra conversion error.
* Only submit thermo calculations if thermo is not already provided (for example if it is loaded from a file)

Isotope fixes:

* Color the element of enriched carbon when drawing isotopes
* Allow isotopes to be properly pickled when they are elements in molecules

etc:

* Increase the sig-figs output to the chemkin file. 3 decimal places was too much error when comparing isotope effects.

### DIFF
--- a/rmgpy/molecule/draw.py
+++ b/rmgpy/molecule/draw.py
@@ -864,8 +864,11 @@ class MoleculeDrawer:
         self.symbols = symbols = [atom.symbol for atom in atoms]
         for i in range(len(symbols)):
             # Don't label carbon atoms, unless there are only one or two heavy atoms
+            # or they are isotopically labeled
             if symbols[i] == 'C' and len(symbols) > 2:
-                if len(atoms[i].bonds) > 1 or (atoms[i].radicalElectrons == 0 and atoms[i].charge == 0):
+                if (len(atoms[i].bonds) > 1 or \
+                        (atoms[i].radicalElectrons == 0 and atoms[i].charge == 0))\
+                        and atoms[i].element.isotope== -1:
                     symbols[i] = ''
         # Do label atoms that have only double bonds to one or more labeled atoms
         changed = True
@@ -1165,7 +1168,8 @@ class MoleculeDrawer:
             boundingRect = [x1, y1, x2, y2]
     
             # Set color for text
-            if   heavyAtom == 'C':  cr.set_source_rgba(0.0, 0.0, 0.0, 1.0)
+            if   atom.element.isotope != -1: cr.set_source_rgba(0.0, 0.5, 0.0, 1.0)
+            elif heavyAtom == 'C':  cr.set_source_rgba(0.0, 0.0, 0.0, 1.0)
             elif heavyAtom == 'N':  cr.set_source_rgba(0.0, 0.0, 1.0, 1.0)
             elif heavyAtom == 'O':  cr.set_source_rgba(1.0, 0.0, 0.0, 1.0)
             elif heavyAtom == 'F':  cr.set_source_rgba(0.5, 0.75, 1.0, 1.0)
--- a/rmgpy/molecule/molecule.py
+++ b/rmgpy/molecule/molecule.py
@@ -135,7 +135,11 @@ class Atom(Vertex):
             'atomType': self.atomType.label if self.atomType else None,
             'lonePairs': self.lonePairs,
         }
-        return (Atom, (self.element.symbol, self.radicalElectrons, self.charge, self.label), d)
+        if self.element.isotope == -1:
+            element2pickle = self.element.symbol
+        else:
+            element2pickle = self.element
+        return (Atom, (element2pickle, self.radicalElectrons, self.charge, self.label), d)
 
     def __setstate__(self, d):
         """
--- a/rmgpy/chemkin.py
+++ b/rmgpy/chemkin.py
@@ -1622,7 +1622,7 @@ def writeKineticsEntry(reaction, speciesList, verbose = True, javaLibrary = Fals
     string += '{0!s:<51} '.format(reaction_string)
 
     if isinstance(kinetics, _kinetics.Arrhenius):
-        string += '{0:<9.3e} {1:<9.3f} {2:<9.3f}'.format(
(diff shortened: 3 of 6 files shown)


rmgpy/molecule/draw.py
```diff
@@ -864,8 +864,11 @@ def __generateAtomLabels(self):
         self.symbols = symbols = [atom.symbol for atom in atoms]
         for i in range(len(symbols)):
             # Don't label carbon atoms, unless there are only one or two heavy atoms
+            # or they are isotopically labeled
             if symbols[i] == 'C' and len(symbols) > 2:
-                if len(atoms[i].bonds) > 1 or (atoms[i].radicalElectrons == 0 and atoms[i].charge == 0):
+                if (len(atoms[i].bonds) > 1 or \
+                        (atoms[i].radicalElectrons == 0 and atoms[i].charge == 0))\
+                        and atoms[i].element.isotope== -1:
```
mjohnson541: So this implies that it will draw the letter of any isotopic element along with coloring it green?

rmgpy/thermo/convertTest.py
```diff
@@ -102,7 +102,7 @@ def test_convert_Wilhoit_to_NASA(self):
             self.assertAlmostEqual(Cp_nasa, Cp_wilhoit, delta=1e0)
             H_wilhoit = wilhoit.getEnthalpy(T)
             H_nasa = nasa.getEnthalpy(T)
-            self.assertAlmostEqual(H_nasa, H_wilhoit, delta=2e1)
```
mjohnson541: How does adding a pair of independent functions affect the accuracy of converting Wilhoit to NASA?

rmgpy/molecule/molecule.py
```diff
@@ -135,7 +135,11 @@ def __reduce__(self):
             'atomType': self.atomType.label if self.atomType else None,
             'lonePairs': self.lonePairs,
         }
-        return (Atom, (self.element.symbol, self.radicalElectrons, self.charge, self.label), d)
+        if self.element.isotope == -1:
```
mjohnson541: Why do we not need additional changes in processing when we unpickle the isotopic element?

rmgpy/thermo/convertTest.py
goldmanm: If we use the typical way to modify entropy of a NASA object, it is converted to Wilhoit and then reconverted back into NASA. There is some error in the fitting procedure from Wilhoit to NASA (and vise versa). This error is small compared to most things in RMG, but for isotope studies, it is significant to not allow proper enrichments to be derived. This method is meant to modify the NASA object without converting to Wilhoit (thereby reducing the error). To answer your question, this method does not affect the accuracy of converting between Wilhoit and NASA, it does around the whole issue by keeping objects as NASA.

rmgpy/molecule/molecule.py
goldmanm: When we unpickle the pickled file, it performs the following operations:

```
# create object of name Argument 0 using args from Argument 1
new_object = [Argument 0]([Argument 1])
# call custom_made method to set other arguments
new_object.__setstate__([Argument 2])
```

For the atom class could look something like 

```
new_object = Atom('C',0,0,'')
new_object.__setstate__(d)
```

Since the __init__ method for Atom already can properly accept either a string or an Element object, no changes in unpickling are necessary.

rmgpy/thermo/convertTest.py
mjohnson541: I understand what they are intended to do, but where do these functions interface with the conversion method?  Since you had to define them in the pxd file I assume the methods did not exist before you made them?  and they aren't called except by each other anywhere in the changes here?  Which logically implies to me that they aren't called by existing RMG functions/objects?  Are you missing a commit where you modify the toNASA function?

rmgpy/molecule/molecule.py
mjohnson541: Ok

rmgpy/thermo/convertTest.py
goldmanm: Good questions. These methods are never intended to interface with the conversion methods (like toNASA or toWilhoit).

Currently they are not called anywhere in the RMG code, since they are not necessary in a typical RMG generation (to my knowledge). I use these methods in the isotope module, which there will be a pull request for soon. I could have just written separate methods in the isotope module to perform this action, but I thought that would be better object oriented design to have a NASA method to modify these values since it doesn't necessarily involve anything related to isotopes. (plus, someone else might find modifying NASA polynomials without introducing extra error useful too). 

If there something I should rewrite in the PR that could make this clearer, let me know.

rmgpy/molecule/draw.py
goldmanm: Exactly! So for a long linear hydrocarbon, where typically no letters are created, if a carbon is enriched, the algorithm will print the 'CH2' and color the 'CH2' green, so the user can notice that the enrichment is in that part of the molecule. If the letter is typically drawn anyways, then the only change is coloring it green.

mjohnson541: Ok

rmgpy/thermo/convertTest.py
mjohnson541: Ok, that makes sense, but if this doesn't interface with toNASA why does this require changes in test_convert_Wilhoit_to_NASA?

goldmanm: That is a very good question, sorry I misunderstood the previous two attempts. 

That question baffled me when I first wrote this commit, hence the comment at the end of the commit. My best guess is that the scipy.optimize function behaved differently when the cythonized object of NASA was modified. The nonlinear fitting within the `toNASA` function could have found a different minima. 

For some reason, after updating to most recent master, and trying out the lower values, the lower values worked (which is rebased into the current branch). I think there might be a potential for finding different minima in the test method.

goldmanm:
@mjohnson541, let me know if anything else would be helpful for this PR.

mjohnson541:
Can you rebase it?

PR #1257, opened 2018-01-10
Modify loading of thermo groups

mliu49:
The main functional change is that loading of generic nodes has been moved from `ThermoDatabase.load()` to `ThermoDatabase.loadGroups()`. The reason is that RMG-website uses `loadGroups()` and not `load()`, so the generic nodes were not being populated. Because `load()` calls `loadGroups()`, there is no effect on normal operation.

As a bonus commit, I also simplified the code in `loadGroups()` to reduce code duplication.

### DIFF
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -796,9 +796,6 @@ class ThermoDatabase(object):
         self.loadLibraries(os.path.join(path, 'libraries'), libraries)
         self.loadGroups(os.path.join(path, 'groups'))
 
-        self.recordRingGenericNodes()
-        self.recordPolycylicGenericNodes()
-        
     def loadDepository(self, path):
         """
         Load the thermo database from the given `path` on disk, where `path`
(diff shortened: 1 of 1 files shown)


rmgpy/data/thermo.py
```diff
@@ -859,6 +856,9 @@ def loadGroups(self, path):
         self.groups['longDistanceInteraction_cyclic']   =   ThermoGroups(label='longDistanceInteraction_cyclic').load(os.path.join(path, 'longDistanceInteraction_cyclic.py'  ), self.local_context, self.global_context)
         self.groups['longDistanceInteraction_noncyclic']   =   ThermoGroups(label='longDistanceInteraction_noncyclic').load(os.path.join(path, 'longDistanceInteraction_noncyclic.py'  ), self.local_context, self.global_context)
 
+        self.recordRingGenericNodes()
+        self.recordPolycylicGenericNodes()
```
mjohnson541: I know they weren't part of the PR, but I noticed these two functions don't have docstrings, would it be possible for you to add ones?

mliu49:
I've added docstrings for those two methods. Thanks for reviewing!

PR #1259, opened 2018-01-17
Group extensions

mjohnson541:
This PR adds functions for generating a set of more specific groups and their respective complements (when possible) for a given group structure.  Allowed extensions are:  
1) atom specification
2) internal bond creation
3) external bond creation (form a bond to a new unspecified atom)
4) bond order specification
5) unpaired electron number specification

This is a step on the way to automatic generation of isomorphic trees.


mjohnson541:
merged with automaticTreeGeneration

PR #1261, opened 2018-01-24
fix pruning message

mjohnson541:
Retrieve the correct Gibbs number for printing in thermoFilterDown.  Can test this by running minimal_thermofilter and noting that the printed Gibbs numbers are distinct and sequential when it filters down to maximumEdgeSpecies.

### DIFF
--- a/rmgpy/rmg/model.py
+++ b/rmgpy/rmg/model.py
@@ -1116,19 +1116,20 @@ class CoreEdgeReactionModel:
             ind = 0
             removeSpcs = []
             
-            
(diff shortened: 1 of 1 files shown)


mjohnson541:
Ok, I've rebased it now.

goldmanm:
@mjohnson541, I just rebased it. The errors are only in thermoCentralDB, which is down now. Do you have merge access?

mjohnson541:
No, I can't override a Travis failure.  I think @mliu49 can?

PR #443, opened 2015-08-27
New handling Molecule parsing from SMILES, InChI, etc.

connie:
This pull request is now on the official repository, for easy edits by multiple users.  Please refer to the previous pull request https://github.com/ReactionMechanismGenerator/RMG-Py/pull/435 for more details.

=====
# Summary:
This PR bundles three major, related improvements to the `rmgpy.molecule` package:

1. A refactoring of the identifier generation and parsing functionality of the molecule module
2. A new way of sorting atoms
3. A first attempt in creating an canonical (augmented) inchi for species with complex electronic structures.

## Identifier generation and parsing

A subpackage of the molecule package is created: `rmgpy.molecule.parser`. This package bundles modules related to the generation and parsing of molecule identifiers such as SMILES and InChI. This stuff was originally to be found in the `rmgpy.molecule.molecule` module. With different backends (rdkit, openbabel) and different identifiers (SMILES, SMARTS, InChI, InChI key, augmented inchi) this became quite large, hence the idea of moving it to its own sub-package.

## Sorting atoms

Prior to this PR, vertices were sorted based on a vertex connectivity value that calculated the connectivity of a vertex until the 3rd degree. This is rather limited, oftentimes leading to multiple equivalent sorting possibilities of atoms.

The new sorting algorithm implements Morgan's idea, by recursively updating the connectivity value of the vertices, until the number of distinct connectivity values is maximized.

Next, `molecule.sortAtoms()` uses not only the connectivity value of the vertex (`sortVertices`), but also its atom attributes like charge, radicalElectrons, lonePairs.

I believe this leads to a unique sorting order of atoms in contrast to the previous implementation.

## Canonical (augmented) inchi

Prior to this PR, the augmented inchi was an inchi with the total multiplicity of the molecule appended at the end.

E.g.:
the triplet biradical .C=C. was represented by:
`InChI=1/C2H2/c1-2/h1-2H/mult3`

Because backends are only able to convert the InChI to a graph, this lead to wrong graphs.

E.g.:`InChI=1/C2H2/c1-2/h1-2H/mult3` would be converted into ethyne (C#C) rather than the
biradical with the identical atom connectivity.

This PR tries to heuristically correct for those mistakes, e.g. by converting the triple bond into
a double bond, with neighboring radical atoms.

In addition, in some cases the multiplicity info is not enough to unambiguously reconstruct the 
graph from the augmented InChI, leading to errors downstream. Therefore, a new (optional) layer is created that indicate the atoms that bear unpaired electrons.

[...]

PR #1262, opened 2018-01-24
Refactor translation of molecule identifiers and conversion of molecule formats

mliu49:
This pull request performs major refactoring of modules related to reading and writing molecule identifiers.

The original modules involved are:
- `rmgpy.molecule.parser`
- `rmgpy.molecule.generator`
- `rmgpy.molecule.inchi`

The new modules after refactoring are:
- `rmgpy.molecule.translator`
- `rmgpy.molecule.converter`
- `rmgpy.molecule.inchi`

The motivation behind the refactoring was to improve the organization of these modules for better flexibility and extensibility and reduced likelihood of cyclic imports. The logic behind the new modules are to group translation between Molecule objects and various identifier types in one module and conversion between RMG, RDKit, and OpenBabel molecules in a separate module.
- Inside `rmgpy.molecule.translator`, individual functions are retained for various identifier types, but they are all wrappers for the core `_read` and `_write` functions, which in turn call the requested backends via `_rdkit_translator` and `_openbabel_translator`.
- The `rmgpy.molecule.converter` module only contains methods for converting to and from RDKit and OpenBabel molecules, which are imported by a couple of different modules.
- Methods related to augmented InChI were previously dispersed among the three modules listed, but are now all located in `rmgpy.molecule.inchi`.

Most methods were also renamed following lowercase_with_underscore convention, and many methods were marked as "private" methods with a leading underscore. More commonly used methods were not renamed to avoid confusion (e.g. `fromSMILES`).

Ultimately, this PR contains no functional changes, and only improves code organization.


rmgpy/molecule/molecule.py
```diff
@@ -44,20 +44,17 @@
...
-    pass
 from .graph import Vertex, Edge, Graph, getVertexConnectivityValue
 import rmgpy.molecule.group as gr
 from .atomtype import AtomType, atomTypes, getAtomType, AtomTypeError
 import rmgpy.constants as constants
-import rmgpy.molecule.parser as parser
-import rmgpy.molecule.generator as generator
+import rmgpy.molecule.element as elements
+import rmgpy.molecule.converter as converter
+import rmgpy.molecule.translator as translator
```
goldmanm: since converter and translator import Atom and Bond, would moving these import statements within the Molecule class avoid circular imports, or is this change just a placebo?

rmgpy/molecule/inchi.py
```diff
@@ -761,40 +761,6 @@ def fixCharge(mol, u_indices):
...
-                dummy.remove(test_index)
-                mobile_h_atom_couple.remove(test_index)
-                new_partner = mobile_h_atom_couple[0]
-                central = dummy[0]
-                return mol.atoms[central - 1], mol.atoms[original_atom - 1], mol.atoms[new_partner - 1]
-
-    raise Exception('We should always have found the mobile-H system. All mobile H couples: {}, test indices: {}'
-                    .format(all_mobile_h_atoms_couples, test_indices))
-
-
```
goldmanm: Just making sure you don't see any benefit to deprecate them instead. If  on second thought you don't think they are useful to anyone, might ask well remove them.

rmgpy/molecule/converter.py
```diff
@@ -229,6 +239,8 @@ def fromOBMol(mol, obmol):
     # cython.declare(i=cython.int)
     # cython.declare(radicalElectrons=cython.int, charge=cython.int, lonePairs=cython.int)
     # cython.declare(atom=Atom, atom1=Atom, atom2=Atom, bond=Bond)
+    if not OB_INSTALLED:
+        raise DependencyError('OpenBabel is not installed. Please install or use RDKit.')
```
goldmanm: beautifully clear!

rmgpy/molecule/translator.py
```diff
@@ -134,72 +134,76 @@
...
-def toInChI(mol, backend='try-all'):
+def toInChI(mol, backend='try-all', aug_level=0):
     """
-    Convert a molecular structure to an InChI string. Uses
-    `RDKit <http://rdkit.org/>`_ to perform the conversion.
-    Perceives aromaticity.
+    Convert a molecular structure to an InChI string.
+    For aug_level=0, generates the canonical InChI.
+    For aug_level=1, appends the molecule multiplicity.
+    For aug_level=2, appends positions of unpaired and paired electrons.
```
goldmanm: `aug_level` of 2 seems to also append multiplicity. 

Can you think of a usage where someone wants the aug_level 2 but not have multiplicity? Maybe it doesn't exist, but if it does, it might be useful to have aug_layer be a list or tuple.

goldmanm: Though maybe this is outside the scope of the PR also

rmgpy/molecule/translatorTest.py
```diff
@@ -32,6 +32,7 @@
 This module contains unit test for the translator module.
 """
 
+import mock
```
goldmanm: this import seems not needed

rmgpy/molecule/molecule.py
mliu49: You're right. This doesn't actually eliminate these circular imports. These changes were originally motivated by me running into circular import issues when trying to add new functionality, but I've forgotten the exact situation.

rmgpy/molecule/inchi.py
mliu49: I note in the commit message that the first function was moved to a new location (and the original version wasn't removed), and the second function is not used at all, though I think it was probably replaced by other methods as well.

rmgpy/molecule/translator.py
mliu49: Do you mean based on the docstring or in the code? I'm pretty sure that `aug_level=2` doesn't include multiplicity. I chose to leave `aug_level=2` as the augmented InChI as Nick had implemented. `aug_level=1` is essentially the old augmented InChI before Nick revamped it.

I think Nick's idea was that specifying unpaired electrons would be sufficient to calculate multiplicity, which was the case before we started worrying about excited states and such. I could add the multiplicity to `aug_level=2`, which I don't think would hurt anything, since it's only being used for labeling stuff right now.

mliu49:
I went through and fixed all of the straightforward Codacy errors. Unfortunately, that led me down a rabbit hole of digging through history. There is a unit test for RDKit molecule conversion which had an unused variable flagged by Codacy. Turned out that the test wasn't working properly, and it was actually failing once I fixed it.

Upon investigation, the issue was due to improper atom sorting, leading to incorrect atom mapping being returned. However, this was an issue that was thoroughly addressed in #443 and #456 via a nice atom sorting method. Interestingly, the improved `sortAtoms()` method did not exist on master! It took me a while to find out that it was removed in #478, presumably by mistake, with no code review.

Thus, I restored the improved `sortAtoms()` method. The main effect is a change in atom ordering, which should be inconsequential in most applications.

I also added a fix to RDKitMol conversion to address disappearing lone pairs, caused by the fact that RDKit has implicit lone pairs. Turns out the fix was as simple as having RMG calculate lone pairs before updating the molecule.

mliu49:
I rebased this branch and added a few more items. I think I am finally satisfied with the robustness and stability of this branch.

New changes:
- When using OpenBabel for SMILES generation, use 'can' for canonical SMILES
- Basic output checking when generating identifiers
- Do not sanitize when generating SMILES with RDKit (restores current functionality)
- Add two unit tests for aromatic SMILES and InChI

mliu49:
RMG-tests doesn't show anything concerning. There were two differences which I think are due to non-deterministic behavior since they've also been showing up when testing other branches.

---|---|---|---|---
eg1|00:00:03:20|00:00:03:16|467.17|469.73
eg3|00:00:07:22|00:00:07:22|416.56|419.56
eg5|00:00:03:39|00:00:03:36|375.49|377.52
eg6|00:00:03:04|00:00:03:06|372.20|372.92
eg7|00:00:01:22|00:00:01:22|362.56|363.08
NC|00:00:02:23|00:00:02:26|373.27|376.17
solvent_hexane|00:00:01:42|00:00:01:43|361.84|362.40
MCH|00:00:05:29|00:00:05:36|795.70|808.12

Could this one be due to different symmetry numbers? The groups are all identical, and the only difference is in the entropy.
```
Non-identical thermo!
original:       [CH]1CCC1CC1CC1
tested: [CH]1CCC1CC1CC1
Hf(300K)  |S(300K)   |Cp(300K)  |Cp(400K)  |Cp(500K)  |Cp(600K)  |Cp(800K)  |Cp(1000K) |Cp(1500K) 
     65.21|     97.54|     32.87|     44.26|     54.81|     63.99|     77.53|     85.89|    100.40
     65.21|     98.92|     32.87|     44.26|     54.81|     63.99|     77.53|     85.89|    100.40
thermo: Thermo group additivity estimation: group(Cs-CsCsCsH) + group(Cs-CsCsCsH) + group(Cs-CsCsHH) + group(Cs-CsCsHH) + group(Cs-CsCsHH) + group(Cs-CsCsHH)
thermo: Thermo group additivity estimation: group(Cs-CsCsCsH) + group(Cs-CsCsCsH) + group(Cs-CsCsHH) + group(Cs-CsCsHH) + group(Cs-CsCsHH) + group(Cs-CsCsHH)
```

The MCH example also has an extra species, which also seems random.
```
The tested model has 1 species that the original model does not have.
spc: ipropyl(119)
The tested model has 5 reactions that the original model does not have.
rxn: H(3) + C3H6(29) <=> H2(6) + allyl(106)             origin: H_Abstraction
rxn: npropyl(83) <=> ipropyl(119)               origin: intra_H_migration
rxn: ipropyl(119) + C2H3(25) <=> C3H6(29) + C2H4(27)            origin: Disproportionation
rxn: H(3) + C3H6(29) <=> ipropyl(119)           origin: R_Addition_MultipleBond
rxn: H(3) + C2H5(31) <=> H2(6) + C2H4(27)               origin: Disproportionation
```

PR #1263, opened 2018-01-25
Generate reaction pairs error converted to debug warning

mjohnson541:
This allows failures in reaction pair generation in family to fall back to Reaction.generatePairs.  This makes it easier to write mult-molecular to mult-molecular families.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1830,7 +1830,8 @@ class KineticsFamily(Database):
             error = True
             
         if error:
-            raise ReactionPairsError('Unable to determine reaction pairs for {0!s} reaction {1!s}.'.format(self.label, reaction))
+            logging.debug('Preset mapping missing for determining reaction pairs for family {0!s}, falling back to Reaction.generatePairs'.format(self.label))
+            return []
         else:
             return pairs
         


rmgpy/data/kinetics/family.py
```diff
@@ -1830,7 +1830,8 @@ def getReactionPairs(self, reaction):
             error = True
             
         if error:
-            raise ReactionPairsError('Unable to determine reaction pairs for {0!s} reaction {1!s}.'.format(self.label, reaction))
+            logging.warn('Preset mapping missing for determining reaction pairs for family {0!s}, falling back to Reaction.generatePairs'.format(self.label))
```
mliu49: I think the command is `logging.warning`. Could you add a unit test to confirm that the warning is logged?

See [here](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/data/kinetics/familyTest.py#L606) for an example of using mock to test logging.

cgrambow: `warn` and `warning` both work. In fact, there is no difference in Python 2.7 ([see here](https://stackoverflow.com/questions/15539937/whats-the-difference-between-logging-warn-and-logging-warning-in-python-or-are)), but it's probably better to use `warning` for posterity. Based on some jobs I've run, I can confirm that the warning works, but it does tend to occupy a large chunk of the log file; something to think about...

mjohnson541: Shouldn't we just add a general test that the logging.warning function works instead of testing individual cases?

mliu49: That's good to know! I wonder if debug would perhaps be more appropriate then?

mliu49: I think testing in this case would be to confirm that the warning is logged under the correct circumstances. We can probably assume that the method itself works properly. It's not a big deal though.

PR #1265, opened 2018-01-26
Automatic tree generation part 1

mjohnson541:
This PR adds the functionality to automatically generate rate trees from a top node a set of training reactions based on optimization of an objective function.


mjohnson541:
This pull request lays the testable groundwork for automatic tree optimization.  The regularization algorithms are not included in this (so the trees generated are as general as possible).  There's plenty of room for speed improvements, but most of the ones I foresee will complicate the code.  I thought it best to start with this PR, which has all of the important algorithms and is fully testable.

rmgpy/molecule/group.py
```diff
@@ -1274,7 +1521,27 @@ def isIdentical(self, other):
...
+        initialMap = dict()
+        for atom in self.atoms:
+            if atom.label and atom.label != '':
+                L = [a for a in other.atoms if a.label == atom.label]
+                assert len(L) == 1, 'more than one label for {0} in {1}'.format(atom.label,other)
+                initialMap[atom] = L[0]
+        
+        #isSubgraphIsomorphic ignores the atoms in the initial map so we have to do it seperately without
+        #the map to confirm that the atoms in the initial map are subgraph-isomorphic
+        return self.isSubgraphIsomorphic(other,initialMap=initialMap) and self.isSubgraphIsomorphic(other) #necessary
```
mliu49: I'm not sure if this will always hold true. You're better off confirming that the initial map is valid using `isMappingValid` then running the subgraph isomorphism check. I also think that you could simply add automatic generation of the initial map as an option to the existing `isSubgraphIsomorphic` method instead of creating a new method.

Same applies for the method in `Molecule`.

rmgpy/molecule/molecule.py
```diff
@@ -45,6 +45,12 @@
 import itertools
 from copy import deepcopy
 
+import element as elements
+try:
+    import openbabel
+except:
+    pass
```
mliu49: Did you add these, or are they a remnant of a rebase?

rmgpy/data/kinetics/family.py
```diff
@@ -2431,3 +2432,21 @@ def getTopLevelGroups(self, root):
             allGroups = all([isinstance(entry.item, Group) for entry in groupList])
 
         return groupList
+
+def informationGain(ks1,ks2):
+    return len(ks1)*np.std(ks1)+len(ks2)*np.std(ks2)
```
mliu49: Could you add a docstring?

rmgpy/data/kinetics/family.py
```diff
@@ -2461,6 +2461,63 @@ def extendNode(self,parent,thermoDatabase=None,obj=None,T=1000.0):
...
+        more than one training reaction associated with it a set of potential more specific extensions 
+        are generated and the extension that optimizing the objective function combination is chosen 
+        and the iteration starts over at the beginning
+        
+        additionally the tree structure is simplified on the fly by removing groups that have no kinetics data associated
+        if their parent has no kinetics data associated and they either have only one child or
+        have two children one of which has no kinetics data and no children
+        (its parent becomes the parent of its only relevant child node)
+        """
+        boo = True
```
mliu49: Use a more descriptive variable name or add a comment?

rmgpy/molecule/group.py
mjohnson541: I made a lot of changes related to subgraphIsomorphism checking in the not yet a PR part 2 of automatic tree generation.  Since we don't intend to use part 1 without part 2 (I broke up the PRs to make it easier to review) is it alright if I make this change in part 2?

mjohnson541:
So I don't really think of these as "automatic trees" vs "regular trees" so much as automatically generated sections of a tree.  This framework is designed to operate on a tree that already exists.  So you would delete everything in the tree you want the tree to regenerate (leaving the top node and logic nodes at the top) and then run automatic tree generation on that "trimmed" tree.  This means both that we don't have to restructure all of the trees and we don't need to regenerate the trees all at the same time.  

There's no real reason to merge this without part 2, which will make all aspects of tree generation functional.  Later parts should be independent of parts 1 and 2.  Part 2 involves a lot of modification of the code in this PR and a little bit of new code.

rmgpy/molecule/group.py
mjohnson541: this should be fixed now

rmgpy/molecule/molecule.py
mjohnson541: removed

rmgpy/data/kinetics/family.py
mjohnson541: added

rmgpy/data/kinetics/family.py
mjohnson541: comment added

alongd:
How do we want to deal with resonance structures in training reactions?
Specifically, if we train for, e.g., `CH2CHO + H = CH2CHOH`
(`CH2CHO` has the following resonance structures: [70% `[CH2]C=O` <--> 30% `C=C[O]`])
and the training hits a theoretical `H;Orad_C` node, then we're actually training this node with a much slower reaction than equivalent ones w/o resonance.
Could we make the automatically generated nodes be specific enough to represent resonance structures as different that species w/o resonance?

mjohnson541:
The algorithm I developed is a machine learning algorithm, so if resonance is important to rate estimation we should expect it to develop nodes specific enough to resolve it.  However, it would probably improve the generated trees if we added resonance attributes similar to the ring attribute that let it examine resonance effects without having to add lots of extra atoms first.

mjohnson541:
commits merged with #1363

mliu49:
Merged via #1363.

PR #1266, opened 2018-01-29
Allow a kinetic family to be irreversible

alongd:
This PR makes it possible to define a family with a `reversible = False` attribute (and sets it to True by default). As a result, all reactions generated by that family will be irreversible (`=>`).
This is meant to be used for families that involve a photon emission such as triplet/singlet conversions. Used correctly, this does not defy thermodynamics.

Since we don't keep track of photons in our system (yet? :smiley:), we cannot and should not try to calculate the reverse rate. The solver/Chemkin should also treat these families as irreversible.

A unit test was added (along with a demo family that is part of the [twin -database PR](https://github.com/ReactionMechanismGenerator/RMG-database/pull/239)).

For the reviewer: checkout the [ twin -database branch](https://github.com/ReactionMechanismGenerator/RMG-database/pull/239) as well, and run a job with species that match the new families (e.g., `O=O` or adjacent birad alkanes). The relevant reactions in the Chemkin output should be unidirectional.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -371,6 +371,7 @@ class KineticsFamily(Database):
     Attribute           Type                            Description
     =================== =============================== ========================
     `reverse`           ``string``                      The name of the reverse reaction family
+    `reversible`        `Boolean`                       Is family reversible? (True by default)
     `forwardTemplate`   :class:`Reaction`               The forward reaction template
     `forwardRecipe`     :class:`ReactionRecipe`         The steps to take when applying the forward reaction to a set of reactants
     `reverseTemplate`   :class:`Reaction`               The reverse reaction template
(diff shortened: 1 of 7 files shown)


alongd:
@goldmanm and @mjohnson541 , I asked both of you to review this, but really just one is enough (unless you both find this interesting). @mjohnson541 also reviews the twin -db, so I though I should divide the workload, but perhaps it's better to review both as one feature? Let me know what you think.

alongd:
I do get the expected output on these branches with only `O=O` in the input:
```
! Reaction index: Chemkin #1; RMG #1
! Template reaction: Singlet_Val6_to_triplet
! Flux pairs: O2(S)(1), O2(2); 
! Matched reaction 1 O2(S) => O2(T) in Singlet_Val6_to_triplet/training
! This reaction matched rate rule [O2]
! family: Singlet_Val6_to_triplet
O2(S)(1)=>O2(2)                                     4.500000e+10 0.000     0.397    
```

Could you make sure you have `maximumRadicalElectrons` set at 2 and not 1?

mjohnson541:
XD, yes I always forget the commented example has that set to 1 unlike almost everything else.  False alarm.

PR #1270, opened 2018-02-06
Add ring attribute to atoms

mliu49:
This PR enables specification of ring membership for Atoms and GroupAtoms. The level of implementation is currently very basic, but can easily be extended in the future.

Essentially, Group adjacency lists now support a ring property, denoted by `r0` or `r1`, indicating whether or not an atom is part of a ring, e.g.
```
1 R!H u0 p0 c0 r0 {2,S}
2 R!H u0 p0 c0 r1 {1,S}
```
This allows greater specificity when designing groups and can help with ring perception issues when estimating rates.

Internally, this property is stored in a `props` dictionary attribute of Atom and GroupAtom with the `inRing` key. Adjacency list support was not added for Molecules since this is a derived property and not essential to the representation.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1300,6 +1300,8 @@ class KineticsFamily(Database):
         for struct in productStructures:
             if isinstance(struct, Molecule):
                 struct.update()
+            else:
+                struct.resetRingMembership()
 
         # Return the product structures
         return productStructures
--- a/rmgpy/molecule/group.py
+++ b/rmgpy/molecule/group.py
@@ -373,13 +373,18 @@ class GroupAtom(Vertex):
                     if charge1 == charge2: break
                 else:
                     return False
+        # Other properties must have an equivalent in other (and vice versa)
+        # Absence of the 'inRing' prop indicates a wildcard
+        if 'inRing' in self.props and 'inRing' in group.props:
+            if self.props['inRing'] != group.props['inRing']:
+                return False
         # Otherwise the two atom groups are equivalent
         return True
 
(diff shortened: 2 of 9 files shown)


rmgpy/data/kinetics/family.py
```diff
@@ -1300,6 +1300,8 @@ def applyRecipe(self, reactantStructures, forward=True, unique=True):
         for struct in productStructures:
             if isinstance(struct, Molecule):
```
mjohnson541: Why does this not need done for molecule objects?

mliu49: If you're referring to resetting ring membership, the `Molecule.update` method will also update the ring attributes.

rmgpy/molecule/group.py
```diff
@@ -373,13 +373,18 @@ def equivalent(self, other):
...
+            if self.props['inRing'] != group.props['inRing']:
+                return False
         # Otherwise the two atom groups are equivalent
         return True
 
     def isSpecificCaseOf(self, other):
         """
-        Returns ``True`` if `other` is the same as `self` or is a more
-        specific case of `self`. Returns ``False`` if some of `self` is not
+        Returns ``True`` if `self` is the same as `other` or is a more
```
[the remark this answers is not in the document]
mliu49: Ok, I'll do that.

rmgpy/data/kinetics/family.py
mjohnson541: Ok

mliu49:
I've updated that commit message and rebased this branch. I think it can be merged at your discretion.

PR #1271, opened 2018-02-08
New Flux Diagram Features

cgrambow:
This PR makes some improvements and adds new features to RMG's flux diagram capabilities. The solver algorithm is slightly updated to reflect the algorithm used in `rmgpy/solver/base.pyx`. This is done for consistency because slight differences in the resulting simulation profiles were noticed.

The larger part of this PR deals with the treatment of "central species". Previously, specifying a central species only resulted in a flux diagram containing that species and all species directly connected to it through single reactions. Now, multiple central species can be specified, each of which will appear on the flux diagram whether or not they are connected to each other. It is also possible to specify a graph radius, which will draw the connections up to each central species through a series of reactions. Lastly, one can superimpose all central species onto the entire automatically generated flux diagram to ensure that they appear in it. This is especially useful for systems in which species are produced via low-flux channels but still accumulate to significant concentrations. Identifying high concentration species that would not appear on the flux diagram and specifying them as central species (preferably with a radius larger than 1) ensures that they now appear on the diagram.

Edit: I added another option to save/read the simulation to/from a file. Since I am dealing with many large models, which take a long time to simulate, this prevents me from having to simulate the model over and over again.

### DIFF
--- a/rmgpy/tools/fluxdiagram.py
+++ b/rmgpy/tools/fluxdiagram.py
@@ -324,16 +324,20 @@ def simulate(reactionModel, reactionSystem, settings=None):
     edgeReactionRates = []
 
     nextTime = initialTime
+    stepTime = initialTime
     terminated = False
+
     while not terminated:
         # Integrate forward in time to the next time point
-        reactionSystem.advance(nextTime)
-        
-        time.append(reactionSystem.t)
-        coreSpeciesConcentrations.append(reactionSystem.coreSpeciesConcentrations)
-        coreReactionRates.append(reactionSystem.coreReactionRates)
-        edgeReactionRates.append(reactionSystem.edgeReactionRates)
-        
+        reactionSystem.step(stepTime)
+
+        if reactionSystem.t >= 0.9999 * nextTime:
+            nextTime *= timeStep
+            time.append(reactionSystem.t)
+            coreSpeciesConcentrations.append(reactionSystem.coreSpeciesConcentrations)
+            coreReactionRates.append(reactionSystem.coreReactionRates)
+            edgeReactionRates.append(reactionSystem.edgeReactionRates)
+
         # Finish simulation if any of the termination criteria are satisfied
         for term in reactionSystem.termination:
             if isinstance(term, TerminationTime):
@@ -347,14 +351,14 @@ def simulate(reactionModel, reactionSystem, settings=None):
                     break
 
         # Increment destination step time if necessary
-        if reactionSystem.t >= 0.9999 * nextTime:
-            nextTime *= timeStep
+        if reactionSystem.t >= 0.9999 * stepTime:
+            stepTime *= 10.0
 
     time = numpy.array(time)
     coreSpeciesConcentrations = numpy.array(coreSpeciesConcentrations)
     coreReactionRates = numpy.array(coreReactionRates)
     edgeReactionRates = numpy.array(edgeReactionRates)
-    
+
     return time, coreSpeciesConcentrations, coreReactionRates, edgeReactionRates
 
 ################################################################################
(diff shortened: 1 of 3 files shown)


cgrambow:
Another issue (not necessarily relevant for this PR) is that the flux diagram function adds new nodes to the diagram based on maximum fluxes between species (regardless of which times these occur at), which is not necessarily a good criterion for selecting nodes because it could fail to select nodes which accumulate to high concentrations over time and would definitely be of interest to the user. If anyone has any good suggestions for how to improve this behavior, I could implement something like that in a future PR.

cgrambow:
I now modified the method so that when specifying `superimpose`, the algorithm ensures that all `centralSpecies` connect to the main flux graph. This is done by sorting the reactions based on maximum flux and only selecting the top `centralReactionCount` reactions at each recursive step. Using a value of 1 or 2 for `centralReactionCount` typically results in acceptable diagrams, but I should note that the number of connections still increases exponentially for `centralReactionCount` > 1, which could be an issue for systems where the central species nodes are not close to the main graph.

In any case, I think this is ready for review and I'll be happy to take suggestions on how to incorporate the desired behavior in a better way.

cgrambow:
I rearranged/squashed code to hopefully make this easier to review.

rmgpy/tools/fluxdiagram.py
```diff
@@ -60,6 +60,7 @@
 maximumNodePenWidth = 10.0      # The thickness of the border around a node at maximum concentration
 maximumEdgePenWidth = 10.0      # The thickness of the edge at maximum species rate
 radius = 1                      # The graph radius to plot around a central species
+centralReactionCount = None     # The maximum number of reactions to draw from each central species (None draws all)
```
mjohnson541: Is this the total number of reactions from every species? or just the central species and the remaining species that connect to the central species only get one reaction?

rmgpy/tools/fluxdiagram.py
```diff
@@ -308,7 +327,9 @@ def addAdjacentNodes(targetNodeIndex, nodes, edges, speciesList, reactionList, m
                                          edges,
                                          speciesList,
                                          reactionList,
+                                         maxReactionRates,
                                          maxSpeciesRates,
+                                         reactionCount=reactionCount,
```
mjohnson541: so it's applied to every node?

rmgpy/tools/fluxdiagram.py
```diff
@@ -166,6 +166,24 @@ def generateFluxDiagram(reactionModel, times, concentrations, reactionRates, out
...
+                    # Recursively add nodes until they connect with main graph
+                    addAdjacentNodes(centralSpeciesIndex,
+                                     nodes,
+                                     edges,
+                                     speciesList,
+                                     reactionList,
+                                     maxReactionRates,
+                                     maxSpeciesRates,
+                                     reactionCount=centralReactionCount,
+                                     rad=-1,
```
mjohnson541: You should both comment and note in the commit that this rad=-1 is a signaling value and doesn't represent a real radius value, otherwise this kind of coding can be confusing.

rmgpy/tools/fluxdiagram.py
```diff
@@ -321,7 +321,6 @@ def simulate(reactionModel, reactionSystem, settings=None):
     time = []
     coreSpeciesConcentrations = []
     coreReactionRates = []
-    edgeReactionRates = []
```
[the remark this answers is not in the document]
cgrambow: Well the current code only loads a single Chemkin file, so `reactionModel.edge.species` and `reactionModel.edge.reactions` will be empty. I removed these lines because they weren't actually being used in the rest of the code. I guess I could add functionality to load an edge Chemkin file as well, but maybe that's better left for a future PR.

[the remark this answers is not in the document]
mjohnson541: Sorry, that was an old comment from before you reformatted the PR.

rmgpy/tools/fluxdiagram.py
mjohnson541: Can you change the comment to reflect that?

mjohnson541:
If you can fix that comment and rebase I think I'm ready to merge this.

cgrambow:
Added comment, squashed, and rebased.

PR #1276, opened 2018-02-08
Update calculateDegeneracy to deal with identical reactants

goldmanm:
calculateDegeneracy did not assess whether two reactants were
actually pointing to the same reference, as did the newly written
database.generate_reactions_from_families method does. This led
calculateDegeneracy not to be able to track atom IDs (since the
molecules had the same values) and get degeneracy incorrect.

This commit adds code which is used in database.generate_reactions_from_families
to calculateDegeneracy, so it can give accurate degeneracy values.

### DIFF
--- a/rmgpy/data/kinetics/kineticsTest.py
+++ b/rmgpy/data/kinetics/kineticsTest.py
@@ -854,12 +854,38 @@ class TestKinetics(unittest.TestCase):
         self.assertEqual(len(reaction_list), 1)
         self.assertEqual(reaction_list[0].degeneracy, 2)
 
-        reaction_list = self.database.kinetics.generate_reactions_from_families(reactants, products, only_families=['H_Abstraction'], resonance=False)
 
+
+    def test_generate_reactions_from_families_product_resonance2(self):
+        """Test that we can specify the no product resonance structure when generating reactions"""
+        reactants = [
+            Molecule().fromSMILES('CCC=C'),
+            Molecule().fromSMILES('[H]'),
+        ]
+        products = [
+            Molecule().fromSMILES('CC=C[CH2]'),
+            Molecule().fromSMILES('[H][H]'),
+        ]
+
+        reaction_list = self.database.kinetics.generate_reactions_from_families(reactants, products, only_families=['H_Abstraction'], resonance=False)
         self.assertEqual(len(reaction_list), 0)
+
+        self.assertTrue(isinstance(products[0],Species))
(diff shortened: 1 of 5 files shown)


goldmanm:
@mliu49 , This entire PR uses code copied from what you wrote recently. Let me know your thought and if you think this warrants a test method.

rmgpy/data/kinetics/kineticsTest.py
```diff
@@ -688,6 +688,18 @@ def test_ensure_independent_atom_ids(self):
...
+        """
+        Ensure ensure_independent_atom_ids does not generate resonance
+        """
+        s1 = Species().fromSMILES('CCC')
+        s2 = Species().fromSMILES('C=C[CH]C')
+        self.assertEqual(s2.molecule[0].atoms[0].id, -1)
+
+        ensure_independent_atom_ids([s1, s2],resonance=False)
+        # checks resonance structures
+        self.assertEqual(len(s2.molecule),1)
```
mliu49: Perhaps add another assertion that the atom id is no longer -1?

goldmanm:
@mliu49, I added the ID check to the test method.

mliu49:
I think you were running into some odd behavior as a result of changing ensure_species to modify in place. Is that still something that we need to worry about?

goldmanm:
The refactoring of ensure_species led to some issue with getting independent atom IDs, but I believe that should be resolved now. This was about a month ago, so my memory is a little bit hazy. We should definitely check travis RMG-tests before merging, just in case. 

I also found out that assignAtomLabelsToReaction also faced the same issue with identical references that calculateDegeneracy had, so I added one more commit to this PR to fix that method as well (with multiple unittests :)

goldmanm:
I checked [RMG tests](https://travis-ci.org/ReactionMechanismGenerator/RMG-tests/builds/355688759). It takes ~50 minutes and only runs one example (clocking 6 minutes of total runtime) before timing out. I will try to run it on the rmg server to see if it errors.

goldmanm:
I ran on the RMG servers. Results are the same for model generation time. For some reason, only the NC example was run. One species had different thermo 
[C]1=NO1, which was due to whether RMG used the group additivity value of ring(Cyclopropene) or ring(oxirene), a 30 kcal difference. Neither of these groups is likely giving accurate values for the thermo, but the change either shows some indeterminate nature of RMG or that we are losing resonance structures in the simulation with these changes.

PR #1283, opened 2018-02-12
Check vital dependencies after make

mliu49:
Thanks to @alongd for the idea for this PR!

This PR refines the dependency checking done at the conclusion of the `make` target. Previously, we would check that `symmetry` and `rdkit` were installed. Now, a script has been added which checks more dependencies and is run at the end of the `make` target. If any of the checked dependencies are missing, the user is asked to install them via anaconda.

The `QM` target has been replaced by the new `check` target since the only thing it was doing was checking the dependencies required for QM (it previously did other stuff that was removed following the transition to anaconda). That functionality is still retained in `check_dependencies.py`.

The idea is that only dependencies that are vital to running RMG jobs are checked here. If there are any others that I missed, please comment.

### DIFF
--- a/check_dependencies.py
+++ b/check_dependencies.py
@@ -1,0 +1,96 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""
+This script checks for the major dependencies that RMG-Py requires.
+"""
+
+import re
+import subprocess
+
+missing = False
+
+print '\nChecking vital dependencies...\n'
+print '{0:<15}{1:<15}{2}'.format('Package', 'Version', 'Location')
+
+# Check for symmetry
+try:
+    result = subprocess.check_output('symmetry -h', stderr=subprocess.STDOUT, shell=True)
+except subprocess.CalledProcessError:
+    print '{0:<30}{1}'.format('symmetry', 'Not found. Please install in order to use QM.')
+    missing = True
+else:
+    match = re.search(r'\$Revision: (\S*) \$', result)
+    version = match.group(1)
+    location = subprocess.check_output('which symmetry', shell=True)
+
+    print '{0:<15}{1:<15}{2}'.format('symmetry', version, location.strip())
+
+# Check for RDKit
+try:
+    import rdkit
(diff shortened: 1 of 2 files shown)


check_dependencies.py
```diff
@@ -0,0 +1,92 @@
...
+"""
+This script checks for the major dependencies that RMG-Py requires.
+"""
+
+import re
+import subprocess
+
+missing = False
+
+print 'Checking vital dependencies...\n'
```
alongd: Perhaps add `\n` before `Checking...` so the output is separated from the other make build

check_dependencies.py
```diff
@@ -0,0 +1,92 @@
...
+    missing = True
+else:
+    location = py_rdl.__file__
+    print '{0:<30}{1}'.format('pyrdl', location)
+
+if missing:
+    print """
+There are missing dependencies as listed above. Please install them before proceeding.
+
+Using Anaconda, these dependencies can be installed from the RMG channel as follows:
```
alongd: the `rmg_env` should also be activated before installing the dependencies (at least when I checked it). We could assume it already is (sine they did `make`) but wouldn't hurt to note that.

check_dependencies.py
```diff
@@ -0,0 +1,92 @@
...
+
+    conda install -c rmg <package name>
+"""
+else:
+    print '\nEverything was found!'
```
alongd: Perhaps we shouldn't finish with an exclamation mark, it looks like a warning or an error

Add an empty line at the end of the file?

check_dependencies.py
mliu49: Done

check_dependencies.py
mliu49: Done

check_dependencies.py
mliu49: Done :)

mliu49:
I can rebase after you review the fixup.

mliu49:
This is ready to be merged.

PR #1284, opened 2018-02-13
Remove duplicate simulation script

cgrambow:
Changes made:
- Removed `sensitivity.py` script because it was a duplicate of `simulate.py`
- Added missing scripts in `setup.py`
- Renamed occurrences of "sensitivity" to "simulate" (or similar)
- Ensure that simulation profiles are generated when running `simulate` in `tools/simulate.py`

The duplicate script has been bothering me for a while, so I decided to change it. Should only take a few minutes to review.


rmgpy/tools/simulate.py
```diff
@@ -139,7 +138,7 @@ def simulate(rmg, diffusionLimited=True):
         if reactionSystem.sensitiveSpecies:
             plotSensitivity(rmg.outputDirectory, index, reactionSystem.sensitiveSpecies)
 
-def runSensitivity(inputFile, chemkinFile, dictFile, diffusionLimited=True):
+def runSimulation(inputFile, chemkinFile, dictFile, diffusionLimited=True):
```
mliu49: If you're renaming this anyway, how about switching to lowercase with underscores? (And the other methods as well.)

setup.py
```diff
@@ -185,9 +185,10 @@ def getCanthermExtensionModules():
             if os.path.splitext(source)[1] == '.pyx':
                 ext_modules.append(module)
 
-scripts=['cantherm.py', 'rmg.py', 'scripts/diffModels.py', 'scripts/generateFluxDiagram.py',
-         'scripts/generateReactions.py', 'scripts/mergeModels.py','scripts/sensitivity.py', 'scripts/thermoEstimator.py',
-         'testing/databaseTest.py']
+scripts=['cantherm.py', 'rmg.py', 'scripts/checkModels.py', 'scripts/convertFAME.py', 'scripts/diffModels.py',
+         'scripts/generateChemkinHTML.py', 'scripts/generateFluxDiagram.py', 'scripts/generateReactions.py',
+         'scripts/machineWriteDatabase.py', 'scripts/mergeModels.py', 'scripts/simulate.py',
+         'scripts/standardizeModelSpeciesNames.py', 'scripts/thermoEstimator.py', 'testing/databaseTest.py']
```
mliu49: Could you separate these into one item per line to help make future commits cleaner?

rmgpy/tools/simulate.py
```diff
@@ -93,12 +93,11 @@ def simulate(rmg, diffusionLimited=True):
...
-        if rmg.saveSimulationProfiles:
-            reactionSystem.attach(SimulationProfileWriter(
-                rmg.outputDirectory, index, rmg.reactionModel.core.species))   
-            reactionSystem.attach(SimulationProfilePlotter(
-                    rmg.outputDirectory, index, rmg.reactionModel.core.species))
+
+        reactionSystem.attach(SimulationProfileWriter(
+            rmg.outputDirectory, index, rmg.reactionModel.core.species))
+        reactionSystem.attach(SimulationProfilePlotter(
+            rmg.outputDirectory, index, rmg.reactionModel.core.species))
```
mliu49: What was the reason for removing the `if` statement?

scripts/simulate.py
```diff
@@ -49,27 +49,26 @@ def parse_arguments():
...
+    runSimulation(inputFile, chemkinFile, dictFile, dflag)
 
 ################################################################################
 
-
 if __name__ == '__main__':
     main()
 
     
-
```
mliu49: Remove extra blank lines and end file with newline?

rmgpy/tools/simulate.py
mliu49: Sorry, I just read your commit message. What if my purpose is running sensitivity and I don't want simulation profiles?

rmgpy/tools/simulate.py
cgrambow: Is it more important to have consistent style within RMG or to comply with PEP8? I would think the former, but I presume all of you have had discussions about this already, so I'm happy to make the change.

amarkpayne: I think we made the decision to implement future code using PEP8, and any changes to old code along the way as we can.

rmgpy/tools/simulate.py
cgrambow: Sure, that's possible, but it doesn't really take any extra time to write the profiles and I figure it's better to have it by default rather than having to rerun a very long simulation just to get the profiles if one forgot to turn on `saveSimulationProfiles`.

mliu49: Ok, that's fine.

rmgpy/tools/simulate.py
cgrambow: Should I also rename `simulateTest.py` to `simulate_test.py` or something similar?

mliu49: Hmm, I don't think we considered that previously... The [style guide](https://github.com/ReactionMechanismGenerator/RMG-Py/wiki/Programming-Style-Guide) says to use `simulateTest.py` because that's how we've been doing it, but it doesn't make much sense.

I think you can keep it as `simulateTest.py` for now, but we might consider renaming all our test files in the future.

cgrambow:
I made the requested changes.

PR #1288, opened 2018-02-15
fix uncertainty parsing issue

mjohnson541:
Comments of the form:
From training reaction 8 for rate rule C_rad/H2/Cs;HCO
Exact match found for rate rule [C_rad/H2/Cs;HCO]
Euclidian distance = 0
family: CO_Disproportionation

result in C_rad/H2/Cs;HCO being set in rateRuleString instead of [C_rad/H2/Cs;HCO], which threw off the parser
this commit differentiates between the two possible rateRuleString formats

### DIFF
--- a/rmgpy/chemkinTest.py
+++ b/rmgpy/chemkinTest.py
@@ -393,7 +393,7 @@ This reaction matched rate rule [C_methyl;C_methyl]
 Reaction index: Chemkin #2; RMG #4
 Template reaction: R_Recombination
 Flux pairs: [CH3], CC; [CH3], CC; 
-From training reaction 21 for rate rule [C_rad/H2/Cs;C_methyl]
+From training reaction 21 used for C_rad/H2/Cs;C_methyl
 Exact match found for rate rule [C_rad/H2/Cs;C_methyl]
 Euclidian distance = 0
 """]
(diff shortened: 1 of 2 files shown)


mliu49:
It seems that the original intention would be to only match one of those lines (the second one?).

1. Would these two lines always reference the same rate rule?
2. Could we not just make the first regex more specific to only match the originally intended line?

mjohnson541:
Yes.  The lines will always reference the same rate, because the training reaction kinetics has to be at the same location the training reaction itself hits.  I'm not sure why we include it.  

Probably, but it's a bit trickier to do since word before 'for rate rule' varies between comments.  If we try to do that the simplest solution would probably be to identify the line by checking if the character before ' for rate rule' is a number or not.  Could also do a fancy regex, which would be more robust, but that makes the code much harder to read.  Doing so might save us if someone decides we need to add a third line that says 'for rate rule' in it to the kinetics comment, but it would certainly complicate the code. Would feel cleaner though.

mliu49:
I suggest changing [this line](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/data/kinetics/family.py#L1070) and line 1118 after that to not use `for rate rule`. I think it's probably better overall to avoid having multiple potential matches. Perhaps something like `From training reaction {1} applied to {2}` would be ok.

You should also update chemkinTest [here](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/chemkinTest.py#L396). Because the test case had brackets, this issue was not caught.

I think something that we really need to do next, is to have the comment parsing unit tests generate the comments instead of having them hard-coded.

mjohnson541:
So I've added the original fix, I've changed the 'from training reaction' comment and I've adjusted the test comment to be in the same format as the 'new 'from training reaction' comment.

mjohnson541:
I agree, we do have some training reactions in the Disprop test database, If we choose the right compounds to hit them with we should be able to generate a decent set of kinetics comments.

Issue #1136, opened 2017-08-08
Improvements to Object Oriented nature of RMG


goldmanm:
There are some cases where RMG does not intuitively implement object oriented design. This issue is meant to be a place where multiple people can list attributes or methods that should be added, moved or deleted to make RMG more intuitive.

Possibly we could have a hackathon on a topic like this. Post some improvements below. We can also discuss why things are how they are in this issue too.

goldmanm:
- create a multiply attribute to KineticsData (and possibly add, divide, subtract, etc) and replace instances where multiplication occurs with the simple `x * y` syntax.
- create attributes to add, subtract, multiply and divide ScalarQuantity objects.
- merge the two Species classes into one class
- move `convertToSpeciesObjects` method from rmgpy.rmg.react to rmgpy.reaction (this is my fault)
- breakup untestable methods (like `enlarge` in rmgpy.rmg.model)
- remove degeneracy attribute from Reaction and place in TemplateReaction (as long as degeneracy for Libraries is not used)

mjohnson541:
I think execute in main and simulate in base need to be simplified using subfunctions, but I'd prefer to build these off the AcceleratorTools branches, since I've changed them so much.

goldmanm:
I also think we might want to better modularize some parts of RMG. For example, if we want to have multiple ways to calculate which species should be added to the core, we might want to have separate classes that implement the different methods, and have an instance of one of the classes as an attribute in CoreEdgeReactionModule. If we have more than just core and edge, maybe a different class than core edge reaction module to represent the differnt methods of storing reactions. If some of this requires different information returned by solvers (or other classes), we might want to have separate inherited classes for those.

Another integration that is pretty finicky might be separating the pdep specifics from core edge reaction module, though I am less sure how this would be done. Also, the structure of estimation from families using rate rules, training reations and group additivity is not structured ideally since there is no inheritance of various estimation methods and rate rule specifics are hardcoded into family.py, database building etc. 

Ideally, I think RMG should be designed to not require changes to the core functionality, but have various inherited classes that allow for flexibility and robustness. We do this already really well with KineticsData, ThermoData, but other specific options are not implemented with independent and separate modules.

goldmanm:
The Database objects could have structural improvements: 

* KineticsDatabase does not inherit from Database but has many similar methods
[...]

PR #1295, opened 2018-02-17
Speed up loading Chemkin files

cgrambow:
Running `loadChemkinFile` is very slow for large models (of which I have many). After some profiling, I identified the duplicate reaction processing section as the main culprit. I separated that part of the code as a new method and cythonized it with static types and cimports from other Cython modules. For a model with 18,000 reactions, the timings of the method are:
- Previous: ~35 s
- Cythonized: ~12 s

A large model I am working with has 143,000 reactions and takes ~17 min. This is significantly faster than the time I had to wait before (probably around 1 hour), but it's still a long time to wait just to load a model. The main issue is that the duplicate reaction processing method is O(n<sup>2</sup>) in the number of reactions and waiting around for 10 billion loop iterations takes a while... Does anyone have ideas how to modify this PR to achieve further speed-up, perhaps by modifying the algorithm itself?


mjohnson541:
I think you might be able to get away with skipping duplicate checking in most cases especially if it's RMG generated.  Adding a checkDuplicates flag might solve your problem.  I know from working on automatic seed mechanism generation its the primary bottleneck in reading and writing reaction information.

mjohnson541:
These comparison operations (is spc in spcList, is rxn in rxnList) can actually be done in log(N) rather than N.  If each item in the list can be mapped to ordered identifiers and the list is reordered this way a binary search algorithm can be used for which membership can be determined in log(N) rather than O(N) time.

goldmanm:
based on Matt's point. I was working on hash functions for species and reaction objects so we can use dictionary of species and reactions instead of lists reducing the efficiency to O(1).  

I should look back into what I was able to accomplish. Some of the questions deal with what attributes should go into a hash function.

goldmanm:
Also, @mjohnson541, is there a method for loading seed mechanisms that might be faster than loadChemkinFile?

Maybe having a better format for reading/writing RMG species and reactions could improve speed significantly, as discussed in #1136.

cgrambow:
Although the main cost here doesn't come from checking membership in a list, it's because you have to compare all the other reactions to see if any of them match and then you have to do that for each reaction.

I guess I'll add a flag to turn off duplicate checking then.

mjohnson541:
The principle is the same, you have some object and a list of objects and you're attempting to check if some attribute of that object is the same as that of any of the objects in the list of objects.  In one case that attribute is something that defines the whole object in this case its just the reactants and products.

cgrambow:
True, that makes sense. Based on your previous comment, are you saying that RMG-generated mechanisms should not contain duplicate reactions? Because I have several RMG mechanisms containing duplicates. Wouldn't I have to process those?

mjohnson541:
Processing the duplicates means finding all the ones that are the same and clustering them into one multi-arrhenius rate.  In theory the duplicates should already be labeled in the chemkin file if they come from an RMG generated mechanism so you don't need to do duplicate checking to identify if a given reaction has duplicates.

cgrambow:
I added a flag for checking duplicates, which I also added to the simulate and flux diagram scripts. If desired, I can modify the checking duplicates function at a later date to try to incorporate some of the algorithmic changes we have been discussing here. Otherwise, just adding the flag and having the speed-up from Cython suits my purposes fairly well right now.

rmgpy/tools/loader.py
```diff
@@ -39,19 +39,23 @@
...
 from rmgpy.solver.base import TerminationConversion
 
-def loadRMGJob(inputFile, chemkinFile=None, speciesDict=None, generateImages=True, useJava=False, useChemkinNames=False):
+def loadRMGJob(inputFile, chemkinFile=None, speciesDict=None, generateImages=True, useJava=False,
+               useChemkinNames=False, checkDuplicates=True):
 
     if useJava:
         # The argument is an RMG-Java input file
-        rmg = loadRMGJavaJob(inputFile, chemkinFile, speciesDict, generateImages, useChemkinNames=useChemkinNames)
+        rmg = loadRMGJavaJob(inputFile, chemkinFile, speciesDict, generateImages,
```
amarkpayne: Before I begin the rest of my review, Codacy is flagging this line. One thing I noticed is that `generateImages` is being passed here, but this variable is not defined in the function definition in line 129. I think this argument needs to be removed, unless I am missing something here.

cgrambow: You're absolutely right, not sure how that got through previous reviews. Instead of removing the argument, I just fixed the function. I also went ahead and rebased on master.

rmgpy/tools/fluxdiagram.py
```diff
@@ -442,7 +442,7 @@ def loadChemkinOutput(outputFile, reactionModel):
 ################################################################################
 
 def createFluxDiagram(inputFile, chemkinFile, speciesDict, savePath=None, speciesPath=None, java=False, settings=None,
-                      chemkinOutput='', centralSpecies=None, diffusionLimited=True):
+                      chemkinOutput='', centralSpecies=None, diffusionLimited=True, checkDuplicates=True):
```
amarkpayne: Does this tool only work for RMG-Py mechanisms given that it needs an RMG  input file? If so I think the default option here should be `checkDuplicates=False` (as well as for scripts/generateFluxDiagram.py).

@mliu49 will this break the website in any way?

If this works for models other than just RMG ones then we should update the documentation for Generating Flux Diagrams to instruct RMG users to set this flag for speed

rmgpy/tools/loader.py
```diff
@@ -39,19 +39,23 @@
 from rmgpy.solver.liquid import LiquidReactor
 from rmgpy.solver.base import TerminationConversion
 
-def loadRMGJob(inputFile, chemkinFile=None, speciesDict=None, generateImages=True, useJava=False, useChemkinNames=False):
+def loadRMGJob(inputFile, chemkinFile=None, speciesDict=None, generateImages=True, useJava=False,
```
amarkpayne: Same comment here, should the default option be False? There are others here in this commit as well.

My opinion is that if the function can only be used on RMG generated mechanisms, then by default we should not check for duplicates. If there is even the slightest chance though that a non-RMG mechanism could be used, then I think we should check for duplicates (to avoid risk of error) and update the documentation in this case to tell RMG mechanism users to set this flag to not check for duplicates.

amarkpayne: You can also update line 58 of rmgpy/solver/baseTest.py so that it doesn't check duplicates. I think you can do the same for line 81 of rmgpy/reduction/input.py but I don't know if non-RMG mechanisms can be used.

amarkpayne:
Sorry, I forgot to look at codecov. Can you add a commit with a unit test for `_processDuplicateReactions` ? I don't think this will change codecov because we don't currently look at coverage for cython code, but at least we can then explain that we haven't added anyhting that isn't unit tested.

scripts/generateFluxDiagram.py
```diff
@@ -71,18 +73,19 @@ def parse_arguments():
     chemkinOutput = os.path.abspath(args.chemkinOutput) if args.chemkinOutput is not None else ''
     useJava = args.java
     dflag = args.dlim
+    checkDuplicates = args.checkDuplicates
```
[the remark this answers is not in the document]
cgrambow: I set `action` to `store_false` when the `--no-dupcheck` flag is used, so I think it should be correct the way it is right now.

rmgpy/tools/fluxdiagram.py
cgrambow: In theory, one could use a non-RMG Chemkin file and create the corresponding species dictionary and input file and pass it to the function, but it is unlikely that anyone would do that.

rmgpy/tools/loader.py
cgrambow: See above comment

cgrambow:
Does is make sense to add a separate unit test? Any unit test that makes use of `loadChemkinFile` would also be testing `_process_duplicate_reactions`, at least with the current default argument of `True` for `checkDuplicates` and the unit tests in `chemkinTest` do that.

scripts/generateFluxDiagram.py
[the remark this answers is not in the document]
amarkpayne: You are right, sorry about that. Part of my confusion was that `store_false` has `True` as its default setting

amarkpayne:
You are right of course that we do test `loadChemkinFile` in chemkinTest.py, which does call `_process_duplicate_reactions`. In this way, we know that this function isn't completely broken. I am looking through this test though to see if any of the tests make sure that duplicate reactions are handled properly (i.e. check that `_process_duplicate_reactions` actually removes duplicates when it should). As long as there is something there that does this I am fine, though I could see how this might be overkill. Let me know what you think

rmgpy/tools/fluxdiagram.py
amarkpayne: Gotcha. In this case it might be best to leave the defaults like you have them. The best thing to do would be to update the documentation to make RMG users aware of this flag, but we don't currently do that for the other options, and it would show up if the user looked through the options or executed the help flag. Therefore, I leave this decision to you.

cgrambow:
I addressed all your comments. I noticed that `generateChemkinHTML` already has a `foreign` flag for non-RMG-generated mechanisms, so I decided to use that syntax instead while keeping the default set to True for `checkDuplicates`, except in the `simulate` and `generateFluxDiagram` scripts, which do not check duplicates by default. I also updated the documentation for those two scripts.

I have not yet implemented a unit test for duplicate checking, but I can probably do so at a later time.

cgrambow:
I added a unit test for `_process_duplicate_reactions`. Should be good to go.

documentation/source/users/rmg/modules/generateFluxDiagram.rst
```diff
@@ -7,12 +7,32 @@ Generating Flux Diagrams
...
+To use this method, you just need a Chemkin input file and an RMG species dictionary.
+The syntax is as follows::
 
-	python generateFluxDiagram.py input.py
-	
-where ``input.py`` is the input file for the completed RMG run.  The program will use the automatically
-generated file structure to find the other necessary files to create the movie.
+    python generateFluxDiagram.py [-h] [--java] [--no-dlim] [-s SPECIES] [-f]
+                                  [-n N] [-e N] [-c TOL] [-r TOL] [-t S]
+                                  INPUT CHEMKIN DICTIONARY [CHEMKIN_OUTPUT]
```
amarkpayne: Should the positional arguments be listed before the optional ones?

rmgpy/chemkinTest.py
```diff
@@ -310,6 +312,73 @@ def testReadSpecificCollider(self):
 
         self.assertEqual(reaction.specificCollider.label, 'N2(5)')
 
+    def test_process_duplicate_reactions(self):
```
amarkpayne: This is excellent and very thorough!

documentation/source/users/rmg/modules/generateFluxDiagram.rst
cgrambow: The way I have it right now, is the same way it would be displayed when showing usage information from the command line, so I think I'll keep it this way.

Issue #741, opened 2016-08-08
Reaction appears in both Arrhenius and Chebyshev form in the same mechanism


zjburas:
I noticed recently that when I run an RMG job with P-dep (input and output files attached), the two reactions shown below appear twice in both an Arrhenius and P-dep Chebyshev form. Both reactions are estimated by RMG (i.e., neither are from a library/seed mechanisms) and they are not labeled as duplicates. I can still run the mechanism in Chemkin and visualize it on the website, but sometimes if I try to merge this mechanism with another one I get an error due to the different rate expression forms. If I simply delete on of the forms, then the merge can happen. 

Is there a reason that the same reaction should appear in both Arrhenius and Chebyshev form in the same mechanism? If so, the merge tool seems unaware that this is acceptable.

**Reaction 1:**
! Template reaction: H_Abstraction
! Flux pairs: O(11), OH(12); CH2O2(239), CHO2(57); 
! Exact match found for rate rule (O/H/OneDeC;O_atom_triplet)
O(11)+CH2O2(239)=CHO2(57)+OH(12)                    1.700e+08 1.500     9.559    

! PDep reaction: PDepNetwork # 754
! Flux pairs: CH2O2(239), CHO2(57); O(11), OH(12); 
O(11)+CH2O2(239)(+M)=CHO2(57)+OH(12)(+M)            1.000e+00 0.000     0.000  
    TCHEB/ 300.000   3200.000 /
    PCHEB/ 0.010     98.692   /
    CHEB/ 6 4/
    CHEB/ 2.258e+00    -6.953e-04   -4.838e-04   -2.685e-04  /
    CHEB/ 9.159e+00    2.178e-04    1.516e-04    8.409e-05   /
    CHEB/ 2.900e-01    1.990e-04    1.385e-04    7.681e-05   /
    CHEB/ 6.626e-02    8.873e-05    6.174e-05    3.425e-05   /
    CHEB/ 2.171e-02    3.737e-05    2.600e-05    1.443e-05   /
    CHEB/ 8.895e-03    1.768e-05    1.231e-05    6.834e-06   /

**Reaction 2:**
! Template reaction: Disproportionation
! Flux pairs: O2(2), HO2(15); C3H7(40), C3H6(164); 
! Exact match found for rate rule (O2b;Cmethyl_Csrad)
! Multiplied by reaction path degeneracy 24
[...]

Issue #566, opened 2016-02-05
Unmarked Duplicate Reactions From 2 Pdep Networks


connie:
Pdep reaction in chebyshev not properly marked as duplicates when they are coming from different pdep networks..

```
# Data sources
database(
    thermoLibraries = ['primaryThermoLibrary'],
    reactionLibraries = [],
    seedMechanisms = [],
    kineticsDepositories = ['training'], 
    kineticsFamilies = 'default',
    kineticsEstimator = 'rate rules',
)

# Constraints on generated species
generatedSpeciesConstraints(
    maximumCarbonAtoms = 7,
    maximumRadicalElectrons = 1,
)

# List of species
species(
    label='n-heptane',
    structure=SMILES("CCCCCCC"),
)

species(
    label='Ar',
    reactive=False,
    structure=SMILES("[Ar]"),
)

simpleReactor(
    temperature=(1000,'K'),
    pressure=(400,'Pa'),
    initialMoleFractions={
        "n-heptane": 0.02,
        "Ar": 0.98,
    },
    terminationConversion={
        'n-heptane': 0.99,
    },
    terminationTime=(1e6,'s'),
)

simpleReactor(
    temperature=(1500,'K'),
    pressure=(400,'Pa'),
    initialMoleFractions={
        "n-heptane": 0.02,
        "Ar": 0.98,
    },
    terminationConversion={
        'n-heptane': 0.99,
    },
    terminationTime=(1e6,'s'),
)

simpleReactor(
    temperature=(2000,'K'),
    pressure=(400,'Pa'),
    initialMoleFractions={
        "n-heptane": 0.02,
        "Ar": 0.98,
    },
    terminationConversion={
        'n-heptane': 0.99,
    },
    terminationTime=(1e6,'s'),
)

simulator(
    atol=1e-16,
    rtol=1e-8,
)

model(
    toleranceMoveToCore=0.05,
    toleranceInterruptSimulation=0.05
)

pressureDependence(
    method='modified strong collision',
    maximumGrainSize=(0.5,'kcal/mol'),
    minimumNumberOfGrains=250,
    temperatures=(300,3000,'K',8),
    pressures=(0.001,100,'bar',5),
    interpolation=('Chebyshev', 6, 4),
)


```

rmgpy.chemkin.ChemkinError: Encountered unmarked duplicate reaction C4H5(156) + H(7) <=> CH3(10) + C3H3(100).

[...]

Issue #394, opened 2015-06-11
Incorrectly declared and undeclared duplicate reactions when pdep is on


connie:
Running into a lot of problems with undetected duplicates, or falsely labeled duplicates in the chemkin file..  Will try to post a short example later (so far the jobs I see them on are pretty long.)

PR #1296, opened 2018-02-19
Don't add a PDep reaction to the model if it exists as an unflagged LibraryReaction

alongd:
Addresses #741, #394 

This still does not address Josh's comment from #566, and in fact will not allow a second network reaction passing through a different isomer to be included in the model. Let's discuss how we'd like to handle that.

(tag @jimchu10, who expressed interest in this fix)

### DIFF
--- a/rmgpy/data/kinetics/library.py
+++ b/rmgpy/data/kinetics/library.py
@@ -73,6 +73,7 @@ class LibraryReaction(Reaction):
                  degeneracy=1,
                  pairs=None,
                  library=None,
+                 has_pdep_route=False,
                  entry=None
                  ):
         Reaction.__init__(self,
@@ -85,7 +86,8 @@ class LibraryReaction(Reaction):
                           transitionState=transitionState,
                           duplicate=duplicate,
                           degeneracy=degeneracy,
-                          pairs=pairs
+                          pairs=pairs,
+                          has_pdep_route=has_pdep_route
                           )
         self.library = library
         self.family = library
@@ -106,6 +108,7 @@ class LibraryReaction(Reaction):
                                   self.degeneracy,
                                   self.pairs,
                                   self.library,
+                                  self.has_pdep_route,
                                   self.entry
                                   ))
 
@@ -139,13 +142,13 @@ class KineticsLibrary(Database):
         and returns at list of all of these LibraryReaction and TemplateReaction objects
         """
         rxns = []
-        for entry in self.entries.values():                
+        for entry in self.entries.values():
             if entry._longDesc and 'Originally from reaction library: ' in entry._longDesc:
                 lib = [line for line in entry._longDesc.split('\n') if 'Originally from reaction library: ' in line]
                 lib = lib[0].replace('Originally from reaction library: ','')
                 lib = lib.replace('\n','')
-                rxn = LibraryReaction(reactants=entry.item.reactants[:], products=entry.item.products[:],\
-                 library=lib, specificCollider=entry.item.specificCollider, kinetics=entry.data, duplicate=entry.item.duplicate,\
+                rxn = LibraryReaction(reactants=entry.item.reactants[:], products=entry.item.products[:],
+                 library=lib, specificCollider=entry.item.specificCollider, kinetics=entry.data, duplicate=entry.item.duplicate,
                  reversible=entry.item.reversible
                  )
             elif entry._longDesc and 'rate rule' in entry._longDesc: #template reaction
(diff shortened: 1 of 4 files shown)


alongd:
I changed this fix so it only checks whether a PDep reaction already exists as a LibraryReaction. This does not conflict with cases where the same reaction is generated via a different networks (which should still be addressed, see #566).

One thing to consider is that sometimes the same reaction can have both PDep and non-PDep pathways. If a library reaction represents only the non-PDep pathway, then the network reaction should be kept. This is still not the case in the present PR.
(examples of this situation are HOSO2+O2<=>SO3+HO2, and SH+SH<=>H2S+S -- where both an H-abstraction channel and a PDep channel exist.)

The above could be solved by tagging such reactions in our database.

alongd:
@mliu49, this PR is now ready for review.
I didn't find an appropriate place for a unit test, let me know if you think of something.

mliu49:
Sorry, I meant #741.

alongd:
@jimchu10, Could you try running your model after cherry-picking the commits from this PR? We'd like to know how many instances of PDep reactions by RMG were _not_ added to your model due to this fix (can be traced by searching for `matched an existing core reaction` in the log file)
Thanks!

jimchu10:
Sure. I will cherry-pick this commit and run the RMG P-dep job. The job might take more than a day to converge, and I will update the results once I get the final mechanism.

jimchu10:
[out.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/1844889/out.txt)

The log files is attached. The message "matched an existing core reaction" appear 8 time in the file. 

Although this job crashed due some other issues after 37 hours, I can still see the chemistry found is different from the output mechanism from the same RMG-Py and RMG-database without the commits in this issue. 

For example, benzyl didn't show up in previous fun, but it did appear in this run. The main reason should be that previous duplicate P-dep and library reactions made the rates over-estimated, and thus affect the later iterations of mechanism generation.

alongd:
Thanks for sharing your log file, @jimchu10, I'm glad to hear this fix improved the reactivity of your system (assuming that the appearance of benzyl was desired...)! I also experience an improvement in the predictions of my sulfur systems using this PR.

I went through the reactions in the attached log file to make sure they already include the PDep route:
- `HCO+O=CO2+H` taken from FFCM-1. Originally from [Baulch 2005](https://aip.scitation.org/doi/pdf/10.1063/1.1748524), it is based on experimental data so it considers all pathways at the relevant conditions.
- `CH3+HO2=CH3O+OH` taken from FFCM-1. Rate constant expression uses the 1000K result (95%) from flow reactor experiments and modelling of Scire et al.(2001) and the temperature dependence from the theoretical study by [Zhu & Lin(2001)](https://pubs.acs.org/doi/pdf/10.1021/jp010698i), which is a PDep calculation done at G2M.
- `CH2OH+O2=CH2O+HO2` from FFCM-1. Originally from [Baulch 2005](https://aip.scitation.org/doi/pdf/10.1063/1.1748524), based on experimental data.
- `C2H5+O=CH2O+CH3` from FFCM-1. Originally from [Baulch 2005](https://aip.scitation.org/doi/pdf/10.1063/1.1748524), based on experimental data.
- `C2H5+O2=C2H4+HO2` from FFCM-1. Originally from [Baulch 2005](https://aip.scitation.org/doi/pdf/10.1063/1.1748524), based on experimental data.
- `CH2OH+O2=CH2O+HO2` from FFCM-1. Originally from [Baulch 2005](https://aip.scitation.org/doi/pdf/10.1063/1.1748524), based on experimental data.
- `CH3CO+O2=CH2CO+HO2` from FFCM-1. Estimated equal to the CH2CHO+O2 rate constant, which is taken from [Baulch 2005](https://aip.scitation.org/doi/pdf/10.1063/1.1748524), based on experimental data.
- `HCO+OH=CO+H2O` from FFCM-1. Originally from [Baulch 2005](https://aip.scitation.org/doi/pdf/10.1063/1.1748524), based on experimental data.

In conclusion, although the entries are only taken from FFCM and are not representative to all RMG's libraries, all reactions that were kept from a library and were not supplemented with an RMG network reaction do take into account the PDep route, which justifies our approach in this PR. I can also confirm that this is the case in my sulfur system.

In my case RMG captured 7 instances where a network reaction was not added (out of ~600 rxns), and here there are 8 instances out of an unfinished model of ~4300 rxns.

@mliu49, I'm rebasing this, let me know if you think additional discussion is needed.

mliu49:
Just ran RMG-tests. Everything looked good except for eg1. The odd thing is that eg1 doesn't even have pdep on. Do you have any ideas for what could have caused the difference?

[eg1.log](https://github.com/ReactionMechanismGenerator/RMG-Py/files/1857464/eg1.log)

jimchu10:
Thank you Alon. I totally agree with you that this PR match the approach we would like to do for P-dep RMG job. I will try more jobs which would make RMG jobs actually converge, and then compare the detailed modeling results of the mechanisms with/without this PR.

PR #1298, opened 2018-02-21
Allow tri-rads in findAllDelocalizationPaths

alongd:
We currently find the ally resonance transformation `[C..]=C[CH2.] --> [C...]C=C`, but we don't find the transformation in the reverse direction, creating a hysteresis (and duplicate species).
(dots above were added to explicitly denote radicals in the SMILES)

As a result, the following are currently different species if they are generated in the following order:
image
image

No test was added now since this will also be implemented in a future lone pair resonance PR with appropriate testing for this case as well.


alongd:
Thanks @mliu49 .
This fix looks trivial, no more comments on my end.
Just rebased, ready to merge in.

Issue #1300, opened 2018-02-27
ValueError when generating a reaction network


HiroumiTani:
A job with Pdep crashes when generating a reaction network with the following message. I haven't detect from which commit this issue happens, but I am sure that rmg.py at the commit of e32da33dd628c9386ffeffdf2e439a316c60387c didn't have this problem. Please let me know if you have an idea on this issue. Thank you.

```
Generating kinetics for new reactions...
Updating 1 modified unimolecular reaction networks (out of 1)...
Updating PDepNetwork #1
Warning: Statmech node 'N_R0' and all its parents have data = None
Warning: Statmech node 'N_R0' and all its parents have data = None
Warning: Maximum number of iterations reached when fitting spectral data for NN.
For reaction hydrazine(1) <=> H(8) + N2H3(58), Ea raised from 320.8 to 338.9 kJ/mol to match endothermicity of reaction.
========================================================================
1 network information
---------------------
Isomers:
    hydrazine(1)                                          86.2423 kJ/mol
Reactant channels:
Product channels:
    NH2(46) + NH2(46)                                     352.223 kJ/mol
    H(8) + N2H3(58)                                       425.149 kJ/mol
Path reactions:
    hydrazine(1) <=> NH2(46) + NH2(46)                     364.75 kJ/mol
    hydrazine(1) <=> H(8) + N2H3(58)                      425.149 kJ/mol
========================================================================

Calculating densities of states for 1 network...
Using 513 grains from 86.24 to 1157.35 kJ/mol in steps of 2.09 kJ/mol to compute densities of states
Traceback (most recent call last):
  File "/home/hiroumi/rmg/RMG-Py/rmg.py", line 173, in <module>
    main()
  File "/home/hiroumi/rmg/RMG-Py/rmg.py", line 167, in main
    rmg.execute(**kwargs)
  File "/home/hiroumi/rmg/RMG-Py/rmgpy/rmg/main.py", line 567, in execute
    bimolecularReact=self.bimolecularReact)
  File "/home/hiroumi/rmg/RMG-Py/rmgpy/rmg/model.py", line 686, in enlarge
    self.updateUnimolecularReactionNetworks()
  File "/home/hiroumi/rmg/RMG-Py/rmgpy/rmg/model.py", line 1659, in updateUnimolecularReactionNetworks
    network.update(self, self.pressureDependence)
[...]

PR #1273, opened 2018-02-08
Major documentation update

mliu49:
This PR includes many documentation updates and fixes:

1. Updates theme to match RMG website
2. Fix a number of issues with docstrings and documentation pages that caused warnings during Sphinx processing
3. Change spelling of MolePro to Molpro everywhere (modules, classes, docstrings, comments)
4. Add a few new pages to the API documentation to reflect changes to RMG

Notes:
- The design was intentionally kept slightly different from the RMG website. One reason is that because the theme and general design is based on a Sphinx theme, it's not straightforward to mirror RMG website's style. Another reason is that slightly differentiating the designs might help people realize that they're separate websites.
- The number of warnings given by Sphinx has been reduced to 16, mainly duplicate labels or citations. I don't plan on following up on those in this PR unless someone else wants to do so.
- I'm not sure why MolePro was spelled that way, but based on their website, it should be Molpro. Since I added a API page on this module, I figured it would be best if we spelled it correctly.

Conflicting PR's:
- This conflicts with #1146 which modifies the MolePro module.
- This might conflict with #1183 which adds to the documentation.

It would be nice if this could be merged as soon as possible in order to update the documentation, but I wouldn't mind fast-tracking the conflicting PR's and then rebasing this on top.

PR #1303, opened 2018-02-28
fix network.__repr__

mjohnson541:
Fixes issue #1300.  Improves handling of arrays in network.__repr__ method and fixes eg5.

### DIFF
--- a/rmgpy/pdep/network.py
+++ b/rmgpy/pdep/network.py
@@ -129,15 +129,15 @@ class Network:
         if self.netReactions: string += 'netReactions="{0!r}", '.format(self.netReactions)
         if self.T != 0.0: string += 'T="{0}", '.format(self.T)
         if self.P != 0.0: string += 'P="{0}", '.format(self.P)
-        if self.Elist: string += 'Elist="{0}", '.format(self.Elist)
-        if self.Jlist: string += 'Jlist="{0}", '.format(self.Jlist)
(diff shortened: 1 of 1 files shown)


goldmanm:
i think the method also needs modifications for other list variables, shown in #1304. 

Sorry for doubling your work. I thought Alon's comment in #1300 was asking me to look into this.

mjohnson541:
Actually it doesn't, it's only a problem if the quantity is a numpy array.  Lists are fine.

mjohnson541:
That's ok, I figured it out and I figured I'd just put it into a PR.

alongd:
Yes, @goldmanm , I did asked you to look at it. Thanks for both of you for your fixes.
Let's decide on a version to merge.

mliu49:
@mjohnson541 If you could change this to `is not None`, then I think we'll merge this one.

alongd:
#1304 has some more fixes under `__repr__`. @mjohnson541, would you like to adopt them?
Let's agree on the changes and get this merged in so PDep is back to life on master

mjohnson541:
It should be changed to is not None.  @alongd  the other fixes in #1304 are unnecessary only the array variables need to have this change.

mliu49:
Thanks! In terms of merging, I think #1273 should be merged first once it passes. I will rebase and merge this so we don't need to wait for Travis a second time.

PR #1304, opened 2018-02-28
Fix network __repr__ method

goldmanm:
The __repr__ method throws an error when comparing the truth value
of a list. This commit adds more specific comparison to prevent an
error in the __repr__ method.

This fixes issue #1300.

I ran `make eg5` with no errors on this

### DIFF
--- a/rmgpy/pdep/network.py
+++ b/rmgpy/pdep/network.py
@@ -121,23 +121,23 @@ class Network:
     def __repr__(self):
         string = 'Network('
         if self.label != '': string += 'label="{0}", '.format(self.label)
-        if self.isomers: string += 'isomers="{0!r}", '.format(self.isomers)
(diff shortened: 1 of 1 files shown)


mjohnson541:
This mostly duplicates #1303.

alongd:
Thanks!
Let's decide which we'd like to keep, this or https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1303

mliu49:
Both of these look fine. I'm not sure if the changes to the other attributes are necessary if they are python lists. It seems that the issue was with ambiguous truth values of numpy arrays?

I do prefer the `is not` syntax though.

goldmanm:
I don't think my `is not None` for the first part will actually work (though it won't cause an error), since the __init__ message replaces them automatically with lists. I think it might be better to go with #1303 or replace the conditional statements checking for empty lists.

mliu49:
Addressed by #1303.

PR #1206, opened 2017-09-20
Charged atom types

nyee:
This must be pulled in simultaneously with the RMG-database PR also named chargedAT. **Be sure to remove the first commit redirecting the travis build before merging**

This PR adds the charge attribute to atomtypes with the goal to increase the diversity of nitrogen and sulfur groups we can represent. This also had some effect on the atomtypes for carbon monoxide. Additionally, we have renamed carbon and oxygen atomtypes to conform to the format of nitrogen and sulfur. 

This was a major functional change, so many coding edits (besides refactoring atom type names in code and test database) were also included. Some of the major ones are listed below:

1. Molecule objects need correct lone pairs and charge to get correct atomtypes now. The Molecule.update() function and many tests were changed accordingly
2. Because of the previous point, sub-molecules used in the polycyclic thermo heuristic now need to be saturated before descending the tree for corrections. Otherwise, the charge/lone pairs get assumed incorrectly.
3. Now that we define lone pairs as part of the atom types, we no longer need to convert lone pairs to bi-radicals in solvation

PR #888, opened 2017-02-09
RMG-Py/nitrogen's PR

alongd:
This PR includes the new nitrogen atomTypes and their corresponding recipe actions.
The AtomTypeError was elaborated with single bonds and number of lone pairs.
AtomType tests were added.

Delocalization paths of allyl to a bi-radical center were allowed. This passed travis, but we should still be on the watch of any unexpected behavior.

The temporary branch change commit in .travis.yml should be deleted before merging into master

Issue #717, opened 2016-07-18
Chlorine in RMG


Ilyasem:
I am very new to Python and chemical modeling in general.
I am trying to model a flash photolysis of chlorine in the presence of methanol.

Here is the error I am stuck on:

Generating thermodynamics for new species...
Traceback (most recent call last):
  File "rmg.py", line 165, in <module>
    rmg.execute(inputFile, output_dir, *_kwargs)
  File "C:\Users\ilya_\Anaconda2\envs\rmg_env\lib\site-packages\rmgpy\rmg\main.py", line 541, in execute
    bimolecularReact=self.bimolecularReact)
  File "C:\Users\ilya_\Anaconda2\envs\rmg_env\lib\site-packages\rmgpy\rmg\model.py", line 764, in enlarge
    spec.generateThermoData(database, quantumMechanics=self.quantumMechanics)
  File "C:\Users\ilya_\Anaconda2\envs\rmg_env\lib\site-packages\rmgpy\rmg\model.py", line 100, in generateThermoData
    thermo0 = database.thermo.getThermoData(self, trainingSet=None, quantumMechanics=quantumMechanics)
  File "C:\Users\ilya_\Anaconda2\envs\rmg_env\lib\site-packages\rmgpy\data\thermo.py", line 724, in getThermoData
    tdata = self.estimateThermoViaGroupAdditivity(molecule)
  File "C:\Users\ilya_\Anaconda2\envs\rmg_env\lib\site-packages\rmgpy\data\thermo.py", line 999, in estimateThermoViaGroupAdditivity
    thermoData = self.estimateRadicalThermoViaHBI(molecule, self.computeGroupAdditivityThermo)
  File "C:\Users\ilya_\Anaconda2\envs\rmg_env\lib\site-packages\rmgpy\data\thermo.py", line 936, in estimateRadicalThermoViaHBI
    thermoData_sat = stableThermoEstimator(saturatedStruct)
  File "C:\Users\ilya_\Anaconda2\envs\rmg_env\lib\site-packages\rmgpy\data\thermo.py", line 1042, in computeGroupAdditivityThermo
    self.__addGroupThermoData(thermoData, self.groups['group'], molecule, {'_':atom})
  File "C:\Users\ilya_\Anaconda2\envs\rmg_env\lib\site-packages\rmgpy\data\thermo.py", line 1126, in __addGroupThermoData
    while node.data is None and node is not None:
[...]

PR #1310, opened 2018-03-07
Added Chlorine `Cl1s` atomType

alongd:
This PR allows Chlorine to be reactive in RMG. Only valance 1 is considered, and no hypervalance atomTypes were defined.
A respective atomType test was added, and the documentation was changed accordingly.
I also made a small change to the isomorphismTest which otherwise insists that Cl should be charged.

This PR will soon be accompanied by a sister -db PR with relevant kinetics and thermo data by @zjburas. They should be merged together, but this branch can principally be reviewed.


mliu49:
@rwest mentioned that he was working on chlorine in #717. Are there any updates on the status of that?

alongd:
@rwest , regarding @mliu49's comment above, are there any chlorine additions you'd like to add to either this PR or https://github.com/ReactionMechanismGenerator/RMG-database/pull/257? Or shall we continue with these as they are?

rwest:
Sorry, I've not been keeping on top of this or https://github.com/ReactionMechanismGenerator/RMG-database/pull/257. Here's a brief recap. 

Our work was done by @faribas  and can be found on her `halogens` branches. If she still receives github notifications, she can correct what I get wrong...

* RMG-Py: https://github.com/ReactionMechanismGenerator/RMG-Py/compare/master...faribas:halogens
* Database: https://github.com/ReactionMechanismGenerator/RMG-database/compare/master...faribas:halogens

Because this forked three years ago (March 2015) it's probably hard to merge them directly, but there may be value in copying across what she did. 

**Thermochemistry:**
as well as adding some thermo libraries and Benson's group additivity values for chlorinated hydrocarbons, she found she had to calculate her own hydrogen bond increment (HBI) values in order to estimate radicals containing chlorines. Next-nearest-neighbour effects were required to get decent estimates. She derived these from her own quantum calculations.

**Reactions:**
She allowed Cl to react as the radical for the two families _H abstraction_ and _radical recombination_, and created new reaction families for _H-Cl or Cl-Cl insertion into a multiple bond_ and _Cl abstraction_.  Training rates were from the literature.

More details are in chapter 3 of her PhD thesis at http://hdl.handle.net/2047/D20213055 

Also worth mentioning is that I have just received a grant to work on adding halogens to RMG (https://www.nsf.gov/awardsearch/showAward?AWD_ID=1751720), so would love to start a conversation about the best way to do this! It will be @skrsna working on that project, starting soon. Perhaps we can all meet.

alongd:
Congrats on the grant! I love that this includes training undergrads in this kind of problem solving.

Other stakeholders here are @zjburas (currently in CA), and @sarahkha from the GreenGroup implementing iodine.

I'll look at @faribas's -Py repo, @zjburas, could you take a look at the -db repo?

zjburas:
Sure @alongd , I can take a look at the -db repo. 

Congratulations @rwest on the grant. My immediate interest in halogen chemistry is pretty narrow: I just want to use RMG as a tool to account for chlorine side-chemistry during chlorine initiated oxidation experiments from 300-1000 K. For that purpose, what @faribas already did is probably sufficient. Hopefully we can talk sometime.

documentation/source/reference/molecule/atomtype.rst
```diff
@@ -116,3 +116,8 @@ Atom type       Description
 ``S6tt``        sulfur atom with no lone pairs (valance 6) and two triple bonds
 ``S6tdc``       charged sulfur atom with no lone pairs (valance 6), one to two triple bonds, up to two double bonds, and up to four single bonds
 =============== ==============================================================================================================================================================
```
amarkpayne: You need to delete line 118 for the table to format properly. Otherwise, it tries to end the table early

documentation/source/users/rmg/database/introduction.rst
```diff
@@ -189,6 +189,8 @@ table below shows all atoms types in RMG.
 +----------+-------------------+------------------------------------------------------------------------------------------------------------------+
 |S6tdc     |Sulfur             |No lone pairs, one to two triple bonds, up to two double bonds, up to four single bonds, charged -1/-1            |
 +----------+-------------------+------------------------------------------------------------------------------------------------------------------+
+|Cl1s      |Chlorine           |Three lone pairs, zero to one single bonds                                                                        |
```
amarkpayne: Should you add an entry for Cl as well?

rmgpy/cantherm/statmech.py
```diff
@@ -705,10 +705,10 @@ def applyEnergyCorrections(E0, modelChemistry, atoms, bonds):
             'N-H':  0.06, 'N-N': -0.23, 'N=N': -0.37, 'N#N': -0.64,}
     elif modelChemistry == 'CBS-QB3': 
         bondEnergies = { 
-            'C-C': -0.495, 'C-H': -0.045, 'C=C': -0.825, 'C-O': 0.378, 'C=O': 0.743, 'O-H': -0.423, #Table2: Paraskevas, PD (2013). Chemistry-A European J., DOI: 10.1002/chem.201301381
-            'C#C': -0.64,  'C#N': -0.89,  'C-S': 0.43,   'O=S': -0.78, 'S-H': 0.0,   'C-N': -0.13, # Table IX: Petersson GA (1998) J. of Chemical Physics, DOI: 10.1063/1.477794 
-            'N-H': -0.42,  'N=O': 1.11,   'N-N': -1.87,  'N=N': -1.58, 'N-O': 0.35, #Table 2: Ashcraft R (2007) J. Phys. Chem. B; DOI: 10.1021/jp073539t
-            'N#N': -2.0,   'O=O': -0.2,   'H-H': 1.1, # Unknown source
+            'C-C': -0.495,'C-H': -0.045,'C=C': -0.825,'C-O': 0.378,'C=O': 0.743,'O-H': -0.423,  #Table2: Paraskevas, PD (2013). Chemistry-A European J., DOI: 10.1002/chem.201301381
```
amarkpayne: It looks like your editor changed some of the spacing here

rmgpy/molecule/group.py
```diff
@@ -1117,18 +1117,25 @@ def updateFingerprint(self):
...
         
         self.carbonCount   = 0
         self.nitrogenCount = 0
         self.oxygenCount   = 0
         self.sulfurCount   = 0
+        self.chlorineCount = 0
+        self.siliconCount = 0
         self.radicalCount  = 0
         for atom in self.vertices:
             if len(atom.atomType) == 1:
```
amarkpayne: This is outside the scope of your commit, but what should happen if len(atom.atomType) != 1?

rmgpy/molecule/group.py
```diff
@@ -1137,17 +1144,24 @@ def updateFingerprint(self):
...
-                    self.oxygenCount += 1
-                elif isSulfur and not isCarbon and not isNitrogen and not isOxygen:
-                    self.sulfurCount += 1
-            if len(atom.radicalElectrons) == 1:
-                radical = atom.radicalElectrons[0]
-                self.radicalCount += radical
+                isChlorine = atomType.equivalent(chlorine)
+                isSilicon  = atomType.equivalent(silicon)
+                sum_is_atom = isCarbon + isNitrogen + isOxygen + isSulfur + isChlorine + isSilicon
+                if sum_is_atom == 1:
```
amarkpayne: What happens if sum_is_atom != 1? I can't imagine this happening, but maybe I am missing a subtlety here. I would think that this would either always evaluate to `True` (and thus it is not needed) or we need some sort of error handling if it does occur.

rmgpy/molecule/molecule.pxd
```diff
@@ -61,7 +61,15 @@ cdef class Atom(Vertex):
 
     cpdef bint isOxygen(self)
 
+    cpdef bint isFluorine(self)
+
+    cpdef bint isSilicon(self)
```
amarkpayne: Make a note in the commit that `isSilicon` was added to cython file as well, good catch!

rmgpy/molecule/molecule.py
```diff
@@ -307,6 +307,13 @@ def isOxygen(self):
         """
         return self.element.number == 8
 
+    def isFluorine(self):
+        """
+        Return ``True`` if the atom represents an oxygen atom or ``False`` if
```
amarkpayne: Change doc string element name to "fluorine"

rmgpy/molecule/molecule.py
```diff
@@ -321,6 +328,20 @@ def isSulfur(self):
         """
         return self.element.number == 16
 
+    def isChlorine(self):
+        """
+        Return ``True`` if the atom represents an sulfur atom or ``False`` if
```
amarkpayne: Same here, but for "chlorine"

rmgpy/molecule/molecule.py
```diff
@@ -321,6 +328,20 @@ def isSulfur(self):
...
+        not.
+        """
+        return self.element.number == 17
+
+    def isIodine(self):
+        """
+        Return ``True`` if the atom represents an sulfur atom or ``False`` if
```
amarkpayne: Same here, but for "Iodine"

rmgpy/molecule/group.py
```diff
@@ -1117,18 +1117,25 @@ def updateFingerprint(self):
...
-        cython.declare(isCarbon=cython.bint, isNitrogen=cython.bint, isOxygen=cython.bint, isSulfur=cython.bint, radical=cython.int)
-        
+        cython.declare(carbon=AtomType, nitrogen=AtomType, oxygen=AtomType, sulfur=AtomType, chlorine=AtomType, silicon=AtomType)
+        cython.declare(isCarbon=cython.bint, isNitrogen=cython.bint, isOxygen=cython.bint, isSulfur=cython.bint,
+                       isChlorine=cython.bint, isSilicon=cython.bint, radical=cython.int)
+
         carbon   = atomTypes['C']
         nitrogen = atomTypes['N']
         oxygen   = atomTypes['O']
+        # flourine = atomTypes['F']
```
amarkpayne: I think we should either go ahead and add fluorine and iodine atom types in this PR, or remove this commented line until we are ready.

rmgpy/cantherm/statmech.py
alongd: That was actually me, trying to make the lines a bit shorter (now with the 'C-Cl', 'C-F' additions, the second line was too long). It still functions good and it's readable, let me know if you think otherwise.

rmgpy/molecule/group.py
alongd: My understanding is that since it is a Group and not a Molecule, an "atom" can be defined as `1 [C,N,O] ux px cx {2,S}`, in which case we can't fingerprint it and we correctly skip this condition. I think this also means we're missing out on cases like `1 [Cs,Cd] ux px cx {2,S}` where we could fingerprint, but don't (for that atom in the Group). However, this should be working smoothly for Molecules.

rmgpy/cantherm/statmech.py
amarkpayne: Gotcha, this is fine by me then, just making sure

rmgpy/molecule/group.py
amarkpayne: That makes sense. I wonder if there is a way to catch cases like `1 [Cs,Cd] ux px cx {2,S}` and if that would be useful. If so this is something for another PR though.

rmgpy/molecule/group.py
alongd: I think this is related to the previous comment: an "atom" in a Group could be defined as `1 [C,N,O] ux px cx {2,S}`. In this case, atomType = [C,N,O], and `.equivalent` will return `True` for more than one isAtom. In this case, `sum_is_atom` will be 3.
I agree that it seems unnecessary (now that I rethink it) since it's already under the `if len(atom.atomType) == 1:` condition discussed above. We could either keep it to keep the same functionality as before in case I'm also missing a subtlety, or remove. I'll check.

alongd:
Thanks for the timely review, @amarkpayne! I addressed all comments but one in fixup commits. I'll double check the implications of removing the `if sum_is_atom == 1:` condition.

alongd:
I just saw your comment. I pushed an optional solution, if it won't do the trick I'll squash w/o it

alongd:
We believe this PR is a comprehensive example for adding new reactive elements into RMG.
See also https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1206 for sulfur and https://github.com/ReactionMechanismGenerator/RMG-Py/pull/888 for nitrogen

PR #1314, opened 2018-03-12
Catmerge3

mazeau:
Merging catmerge 2 into rmg master using a new branch called catmerge3


rwest:
I think this is jumping the gun, and at least initially you intended to send the PR to cfgoldsmith:cat rather than ReactionMechanismGenerator:master. 
Although this is not a bad time to start a conversation about how to keep these things in sync and how and when to merge...

PR #1319, opened 2018-03-15
Minor changes to forbidden structure loading and checking

mliu49:
This fixes the capability to specify a molecule as a forbidden structure and adds the capability to specify a species as a forbidden structure.

The benefit of these over groups is that they can use normal isomorphism, which allows comparison of molecular formulas. In the future, they could also be changed to an InChI comparison to completely eliminate the need for isomorphism.

### DIFF
--- a/rmgpy/data/base.py
+++ b/rmgpy/data/base.py
@@ -1267,17 +1267,27 @@ class ForbiddenStructures(Database):
         contains forbidden functionality, or ``False`` if not. Labeled atoms
         on the forbidden structures and the molecule are honored.
         """
+        from rmgpy.species import Species
+
         for entry in self.entries.values():
-            entryLabeledAtoms = entry.item.getLabeledAtoms()
-            moleculeLabeledAtoms = molecule.getLabeledAtoms()
-            initialMap = {}
-            for label in entryLabeledAtoms:
-                # all group labels must be present in the molecule
-                if label not in moleculeLabeledAtoms: break  
-                initialMap[moleculeLabeledAtoms[label]] = entryLabeledAtoms[label]
-            else:
-                if molecule.isMappingValid(entry.item, initialMap) and molecule.isSubgraphIsomorphic(entry.item, initialMap):
+            if isinstance(entry.item, Molecule) or isinstance(entry.item, Species):
+                # Perform an isomorphism check
+                if entry.item.isIsomorphic(molecule):
                     return True
+            elif isinstance(entry.item, Group):
+                # We need to do subgraph isomorphism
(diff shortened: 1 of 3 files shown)


rmgpy/data/base.py
```diff
@@ -1267,17 +1267,27 @@ def isMoleculeForbidden(self, molecule):
...
+                initialMap = {}
+                for label in entryLabeledAtoms:
+                    # all group labels must be present in the molecule
+                    if label not in moleculeLabeledAtoms: break
+                    initialMap[moleculeLabeledAtoms[label]] = entryLabeledAtoms[label]
+                else:
+                    if molecule.isMappingValid(entry.item, initialMap) and molecule.isSubgraphIsomorphic(entry.item, initialMap):
+                        return True
+            else:
+                raise NotImplementedError
```
alongd: Could you add an error message?

rmgpy/data/base.py
```diff
@@ -1267,17 +1267,27 @@ def isMoleculeForbidden(self, molecule):
...
-            entryLabeledAtoms = entry.item.getLabeledAtoms()
-            moleculeLabeledAtoms = molecule.getLabeledAtoms()
-            initialMap = {}
-            for label in entryLabeledAtoms:
-                # all group labels must be present in the molecule
-                if label not in moleculeLabeledAtoms: break  
-                initialMap[moleculeLabeledAtoms[label]] = entryLabeledAtoms[label]
-            else:
-                if molecule.isMappingValid(entry.item, initialMap) and molecule.isSubgraphIsomorphic(entry.item, initialMap):
+            if isinstance(entry.item, Molecule) or isinstance(entry.item, Species):
```
alongd: Perhaps we could first check equality between `len(molecule.vertices)` before entering the other conditions?
I understand we check fingerprints inside Molecule.isIsomorphic(), but it might still save us a bit, since most comparisons will have a different number of atoms than entry.item

rmgpy/data/base.py
mliu49: Done

rmgpy/data/base.py
mliu49: I think I'll leave it like this for now. I did some time trials, and it didn't seem like adding that check gave that much speedup. I do plan on revisiting this in the near future though, since I'm considering replacing this with an InChI comparison.

mliu49:
Thanks for reviewing! I updated the forbidden structures part of the documentation. They do still go in the same file. The only difference is changing the keyword from `group` to `molecule` or `species`.

Issue #1315, opened 2018-03-13
Inadequate model chemistry designations


cgrambow:
According to the documentation, the `DFT_G03_b3lyp` model chemistry is a "B3LYP calculation with a moderately large basis set", but this designation is completely useless unless we also know what the exact basis set is (similarly, `DFT_ks_b3lyp` and `DFT_uks_b3lyp` are also inadequate designations). For example, using a 6-31G(2df,p) basis set (which I would describe as moderately large) gives an atomic energy for carbon that differs from `DFT_G03_b3lyp` by almost 7 kcal/mol, which means that heats of formation for any medium sized molecule will likely be off by 50 kcal/mol or more (rates would be less affected).

Most users are probably not aware of this issue, so I propose either finding out what the basis sets are and adding them to the designation, or to remove the atomic energies and no longer allow those model chemistries because they more likely than not will give erroneous results. Thoughts?

Going beyond this, it might also make sense to add an option to Cantherm to allow the user to supply their own atomic energies, which would greatly increase its flexibility by not being restricted to the hard-coded model chemistries.

Also see #742

alongd:
closed via #1324

Issue #742, opened 2016-08-09
CanTherm model chemistries


bslakman:
I'm a bit confused about what is meant by some of the model chemistries in CanTherm. I want to do calculations using G3//B3LYP (i.e. Gaussian-3 theory on B3LYP geometries). Is the appropriate model chemistry to use `DFT_G03_b3lyp`, `G3`, or neither? I used to think the former referred to calculations done with the _program_ Gaussian 03, but this seems incorrect now as none of the others refer to the program used.
However if it is the former, I'm not sure that we should be adding in the spin orbit corrections, as we don't for `G3` itself.
Edit: see https://github.com/ReactionMechanismGenerator/RMG-Py/issues/263

nyee:
Maybe the documentation under model chemistry helps? Does G3 imply a specific basis set like CBS-QB3 does? If so, I would guess you want to use the G3 corrections.

https://github.com/ReactionMechanismGenerator/RMG-Py/blob/15880913bbfab08e2e29b5199b4ae58ea0301857/documentation/source/users/cantherm/input.rst

alongd:
closed via #1324

PR #1324, opened 2018-03-20
Remove atoms parameter in Cantherm and allow user-defined atomic energies

cgrambow:
When parsing quantum chemistry output in a Cantherm calculation, there is no reason for the user to specify the types and numbers of atoms, because the information can be obtained from the output files. In fact, mistyping atoms will lead to incorrect thermochemistry and rates, which is avoided in this PR.

This PR also increases the flexibility of Cantherm calculations by allowing the user to supply a dictionary of atomic energies, which means that we are no longer restricted by the hard-coded model chemistries in `rmgpy/cantherm/statmech.py`.

Lastly, another issue in #1315 and #742 is addressed by removing model chemistries that do not specify which basis set was used for the atomic energy calculations. This is a relatively drastic measure, but it should ensure that users do not accidentally end up with completely wrong results.

### DIFF
--- a/documentation/source/users/cantherm/input.rst
+++ b/documentation/source/users/cantherm/input.rst
@@ -155,7 +155,7 @@ Allowed bond types for the ``bonds`` parameter are, e.g., ``'C-H'``, ``'C-C'``, 
 
 ``'O=S=O'`` is also allowed.
 
-The order of elements in the bond correction label is important and generally follows the order specified under "allowed atom symbols" above, i.e., ``'C=N'`` is correct while ``'N=C'`` is incorrect. Use ``-``/``=``/``#`` to denote a single/double/triple bond, respectively. For example, for formaldehyde we would write::
+The order of elements in the bond correction label is not important. Use ``-``/``=``/``#`` to denote a single/double/triple bond, respectively. For example, for formaldehyde we would write::
 
     bonds = {'C=O': 1, 'C-H': 2}
 
--- a/documentation/source/users/cantherm/input.rst
+++ b/documentation/source/users/cantherm/input.rst
@@ -67,9 +67,6 @@ Model Chemistry                                  AEC   BC   SOC
 ``'B-CCSD(T)-F12/cc-pVnZ-F12'``, *n = D,T,Q*      v         v
 ``'B-CCSD(T)-F12/cc-pCVnZ-F12'``, *n = D,T,Q*     v         v
 ``'B-CCSD(T)-F12/aug-cc-pVnZ-F12'``, *n = D,T,Q*  v         v
-``'DFT_G03_b3lyp'``                               v    v    v
-``'DFT_ks_b3lyp'``                                          v
-``'DFT_uks_b3lyp'``                                         v
 ``'G03_PBEPBE_6-311++g_d_p'``                     v         v
 ``'MP2_rmp2_pVnZ'``, *n = D,T,Q*                  v         v
 ``'FCI/cc-pVnZ'``, *n = D,T,Q*                    v         v
--- a/documentation/source/users/cantherm/input.rst
+++ b/documentation/source/users/cantherm/input.rst
@@ -39,7 +39,7 @@ which accepts a string describing the model chemistry.
 CanTherm uses this information to adjust the computed energies to the usual gas-phase reference
 states by applying atom, bond and spin-orbit coupling energy corrections. This is particularly
 important for ``thermo()`` calculations (see below). Note that the user must specify under the
-``species()`` function the type and number of atoms and bonds for CanTherm to apply these corrections.
+``species()`` function the type and number of bonds for CanTherm to apply these corrections.
 The example below specifies CBS-QB3 as the model chemistry::
 
     modelChemistry("CBS-QB3")
@@ -114,7 +114,6 @@ The species input file accepts the following parameters:
 ======================= =========================== ====================================
 Parameter               Required?                   Description
 ======================= =========================== ====================================
-``atoms``               yes                         Type and number of atoms in the species
 ``bonds``               optional                    Type and number of bonds in the species
 ``linear``              yes                         ``True`` if the molecule is linear, ``False`` if not
 ``externalSymmetry``    yes                         The external symmetry number for rotation
@@ -128,12 +127,11 @@ Parameter               Required?                   Description
 ``rotors``              optional                    A list of :class:`HinderedRotor()` and/or :class:`FreeRotor()` objects describing the hindered/free rotors
 ======================= =========================== ====================================
 
-The ``atom`` and ``bond`` parameters are used to apply atomization energy corrections (AEC), bond corrections (BC), and spin orbit corrections (SOC) for a given ``modelChemistry()`` (see `Model Chemistry`_).
+The types and number of atoms in the species are automatically inferred from the quantum chemistry output and are used
+to apply atomization energy corrections (AEC) and spin orbit corrections (SOC) for a given ``modelChemistry()``
+(see `Model Chemistry`_).
 
-Allowed atom symbols for the ``atoms`` parameter are
-``'C'``, ``'N'``, ``'O'``, ``'S'``, ``'P'``, and ``'H'``. For example, for formaldehyde we would write::
-
-    atoms = {'C': 1, 'O': 1, H': 2}
+The ``bond`` parameter is used to apply bond corrections (BC) for a given ``modelChemistry()``.
 
 Allowed bond types for the ``bonds`` parameter are, e.g., ``'C-H'``, ``'C-C'``, ``'C=C'``, ``'N-O'``, ``'C=S'``, ``'O=O'``, ``'C#N'``...
 
@@ -189,11 +187,6 @@ the ``species()`` function in the input file should look like the following exam
     species('C2H6', 'C2H6.py')
 
 and the species input file (``C2H6.py`` in the example above) should look like the following::
-
-    atoms = {
-    'C': 2,
-    'H': 6,
-    }
 
     bonds = {
         'C-C': 1,
@@ -299,11 +292,6 @@ puts a lower bound on the impact of that rotor on the species's overall partitio
 in between these two extremes.
 
 To summarize, the species input file with hindered/free rotors should look like the following example (different options for specifying the same ``rotors`` entry are commented out)::
-
-    atoms = {
-        'C': 2,
-        'H': 6,
-    }
 
     bonds = {
         'C-C': 1,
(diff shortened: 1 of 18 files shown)


documentation/source/users/cantherm/input.rst
```diff
@@ -128,12 +127,11 @@ Parameter               Required?                   Description
 ``rotors``              optional                    A list of :class:`HinderedRotor()` and/or :class:`FreeRotor()` objects describing the hindered/free rotors
 ======================= =========================== ====================================
 
-The ``atom`` and ``bond`` parameters are used to apply atomization energy corrections (AEC), bond corrections (BC), and spin orbit corrections (SOC) for a given ``modelChemistry()`` (see `Model Chemistry`_).
+The types and number of atoms in the species are automatically inferred from the quantum chemistry output and are used
+to apply atomization energy corrections (AEC) and spin orbit corrections (SOC) for a given ``modelChemistry()``
+(see `Model Chemistry`_).
 
-Allowed atom symbols for the ``atoms`` parameter are
```
alongd: We may still want to leave some description in the documentation of the atoms we commonly have AEC for

rmgpy/cantherm/statmech.py
```diff
@@ -318,6 +311,16 @@ def load(self):
         logging.debug('    Reading optimized geometry...')
         coordinates, number, mass = geomLog.loadGeometry()
 
+        # Infer atoms from geometry
+        atoms = {}
+        for atom_num in number:
+            symbol = atom_num_dict[atom_num]
```
alongd: We should capture exception when `atom_num` is not in `atom_num_dic`, and output an error message that this atom is currently not supported in Cantherm

rmgpy/cantherm/statmech.py
```diff
@@ -318,6 +311,16 @@ def load(self):
...
+            atoms[symbol] = atoms.get(symbol, 0) + 1
```
alongd: excellent implementation!

alongd:
In the spirit of this PR, could you also add a tiny fix for annoying cases when the bond correction is entered in reverse, e.g., `O=C` instead of `C=O`?
I believe [this line](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/cantherm/statmech.py#L721) should be changed to also check if `symbol[::-1]` is in `bondEnergies`.
Thanks!

rmgpy/cantherm/statmech.py
cgrambow: Done. Since we're now allowing the user to input atomic energies, I also added more enthalpies of formation of atoms so that we now support more elements.

cgrambow:
Should be good to go!

documentation/source/users/cantherm/input.rst
```diff
@@ -13,12 +13,13 @@ Each section is made up of one or more function calls, where parameters are
 specified as text strings, numbers, or objects. Text strings must be wrapped in
 either single or double quotes.
 
-The following is a list of all the functions of a CanTherm input file for thermodynamics and high-pressure limit kinetics computations:
+The following is a list of all the elements of a CanTherm input file for thermodynamics and high-pressure limit kinetics computations:
```
alongd: I'm uncertain regarding this terminology change: "elements" sound like chemical elements. By the way, the table below still says `Function`, as well as other references in the documentation below. Since it's a python file, it's OK to call them functions. But we could consider "components" as well.

documentation/source/users/cantherm/input.rst
```diff
@@ -33,16 +34,27 @@ Function                    Description
...
 states by applying atom, bond and spin-orbit coupling energy corrections. This is particularly
 important for ``thermo()`` calculations (see below). Note that the user must specify under the
-``species()`` function the type and number of atoms and bonds for CanTherm to apply these corrections.
+``species()`` function the type and number of bonds for CanTherm to apply these corrections.
 The example below specifies CBS-QB3 as the model chemistry::
 
-    modelChemistry("CBS-QB3")
+    modelChemistry = "CBS-QB3"
+
+Alternatively, the atomic energies at the ``modelChemistry`` level of theory can be directly
```
alongd: We should tell the users that no bond corrections will be applied in this case, and perhaps allow them to add their own?

documentation/source/users/cantherm/input.rst
```diff
@@ -131,18 +139,17 @@ Parameter               Required?                   Description
 ``rotors``              optional                    A list of :class:`HinderedRotor()` and/or :class:`FreeRotor()` objects describing the hindered/free rotors
 ======================= =========================== ====================================
 
-The ``atom`` and ``bond`` parameters are used to apply atomization energy corrections (AEC), bond corrections (BC), and spin orbit corrections (SOC) for a given ``modelChemistry()`` (see `Model Chemistry`_).
-
-Allowed atom symbols for the ``atoms`` parameter are
-``'C'``, ``'N'``, ``'O'``, ``'S'``, ``'P'``, and ``'H'``. For example, for formaldehyde we would write::
```
alongd: We may still want to leave some description in the documentation of the atoms we commonly have AEC for

rmgpy/cantherm/statmech.py
```diff
@@ -318,6 +314,19 @@ def load(self):
         logging.debug('    Reading optimized geometry...')
         coordinates, number, mass = geomLog.loadGeometry()
 
+        # Infer atoms from geometry
+        atoms = {}
+        for atom_num in number:
+            try:
+                symbol = atom_num_dict[atom_num]
+            except KeyError:
+                raise Exception('Atom number {} is not yet supported.'.format(atom_num))
```
alongd: The error message needs a tiny refinement: "atom number" could also be inferred as the number of the atom in the xyz / Z matrix. Could you print the element instead?

documentation/source/users/cantherm/input.rst
cgrambow: The reason I changed it is because many of the options aren't functions, they're assignment statements. That was probably changed in the past, but the documentation never got updated. I thought I went through and checked for all occurrences of "function", but evidently I wasn't thorough enough. I agree that "elements" could be misleading. I'll give it a different name.

documentation/source/users/cantherm/input.rst
cgrambow: Bond corrections would still be applied if `modelChemistry` matches one of the options in `applyEnergyCorrections`, although of course it is unlikely that someone would specify `atomEnergies` themselves if they are already available within Cantherm. I could add an option to add bond corrections, but deriving them is typically associated with a lot more work than calculating atomic energies, so I doubt that option would be used. Ideally, if new bond corrections become available, they should be added to Cantherm. What do you think?

documentation/source/users/cantherm/input.rst
cgrambow: Ah, I wasn't sure what you meant by this earlier, because the information about which AECs are available is still in the table, which I didn't delete. You mean we should have a description about which atomic energies are available for the implemented model chemistries, correct? I'll add that.

rmgpy/cantherm/statmech.py
cgrambow: Yup, I'll change it

documentation/source/users/cantherm/input.rst
alongd: I think it's enough to mention in the documentation something like:
If a `modelChemistry` other than the ones specified in the table is given, then the user should add the respective `atomEnergies` to get meaningful results, and BAC would not be applied.

documentation/source/users/cantherm/input.rst
alongd: Yes, something like what elements Canthem supports (from the atomHf dictionary you elaborated). But in second though, each modelChemistry supports slightly different elements... Perhaps we could list the common ones?
And more importantly, what do you think of outputting an Error is we don't have one of the elements within the `atomEnergies` dictionary for the selected modelChemistry (or within the user inputted atomEnergies values)?

cgrambow: I don't think we should output that error because if the user is only interested in kinetics and doesn't care about the thermo, then it doesn't matter if atomic energies are available or not because they will cancel out between the products and reactants. Cantherm already prints a warning if that is the case, so at least the user would be aware of the issue.

cgrambow:
Updated documentation, let me know what you think.

documentation/source/users/cantherm/input.rst
```diff
@@ -33,61 +34,75 @@ Function                    Description
...
+        'O': -74.987979,
+        'S': -397.658253,
+    }
+
+The table below shows which model chemistries have atomization energy corrections (AEC), bond
+corrections (BC), and spin orbit corrections (SOC). It also lists which atoms types are available
+for a given model chemistry.
+
+================================================ ===== ==== ==== ==================
+Model Chemistry                                  AEC   BC   SOC  Atom Types
```
alongd: I love this addition!
Could you just change "Atom Types" to something like "Supported elements"? "Atom Types" means something else in RMG.

cgrambow: Sure thing! Too much RMG lingo...

cgrambow: Done

cgrambow:
One thing I was wondering about: Cantherm prints a warning if an atom type/element is unknown, which will cause it to not subtract the atomic reference energy for that atom. In that case, I feel like the returned energy will not make any physical sense, i.e., it will not be useful for thermo, but should still be relevant for kinetics. If that happens, should we just print a warning that the energy corrections will not be applied and that the user should not trust the thermo values?

Cantherm also prints the equilibrium constant and reverse rates in comments in the output file. Those would not be affected, correct?

alongd:
Is there a reason for a user to only use partial atom energy corrections?
I think Cantherm should crush if AEC is used, but is partial.
So perhaps we need a flag for `useAtomEnergyCorrections` (like `useBondCorrections`), that defaults to True and allows Cantherm to be consciously used with no corrections if it is False?

cgrambow:
I made the changes. `useAtomCorrections` is `True` by default and Cantherm raises an error if elements are not found in the `atomEnergies` dictionary. I also added this to the documentation.

cgrambow:
Rebased

cgrambow:
I added the atomic energies for B3LYP/6-311+G(3df,2p). While doing so, I noticed that most of the model chemistry atomic energies in `rmgpy/cantherm/statmech.py` don't actually match the outputs of single-atom quantum calculations, so they probably were fitted in some way in the past, but that information has since been lost. We should try to find out where the values come from and start adding some documentation for posterity, but that can be done in future pull requests.

I also changed the code such that model chemistries are now case-insensitive. This means that future developers should make sure to add atomic energies and bond correction model chemistries in lower-case in `statmech.py`.

PR #1325, opened 2018-03-21
Corrected known SMILES in generator.py

alongd:
- Added known SMILES for: H2S, Cl2, HCl, I2, HI, S, N, CH2,HS, NH2, NH, Cl, I.
- Corrected atom order in formula to be alphabetical (but allowing C to precede H), e.g., `SO` -> `OS`, so this is in accordance with Molecule.getFormula()
- Removed formula which could have several meanings: C2H4 (could be a carbene), SO2, SO3, H2SO4, HSO3

This is a relatively small PR, would anyone like to review?

### DIFF
--- a/rmgpy/molecule/generator.py
+++ b/rmgpy/molecule/generator.py
@@ -66,15 +66,16 @@ _known_smiles_molecules = {
                  'CH4O': 'CO',
                  'CO2': 'O=C=O',
                  'CO': '[C-]#[O+]',
-                 'C2H4': 'C=C',
                  'O2': 'O=O',
                  'C': '[C]',  # for this to be in the "molecule" list it must be singlet with 2 lone pairs
-                 'SO2': 'O=S=O',
-                 'SO3': 'O=S(=O)=O',
-                 'H2SO4': 'OS(=O)(=O)O',
+                 'H2S': 'S',
                  'N2O': 'N#[N+][O-]',
                  'NH3': 'N',
                  'O3': '[O-][O+]=O',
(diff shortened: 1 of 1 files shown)


rmgpy/molecule/generator.py
```diff
@@ -66,35 +66,40 @@
...
                  'C': '[C]',  # for this to be in the "molecule" list it must be singlet with 2 lone pairs
-                 'SO2': 'O=S=O',
-                 'SO3': 'O=S(=O)=O',
-                 'H2SO4': 'OS(=O)(=O)O',
+                 'H2S': 'S',
                  'N2O': 'N#[N+][O-]',
                  'NH3': 'N',
                  'O3': '[O-][O+]=O',
+                 'Cl2': '[Cl][Cl]',
+                 'ClH': 'Cl',
```
goldmanm: Do you think it is worthwhile to add 'HCl' in addition to, or in place of, 'ClH'?

I'm not familiar with writing hydrochloric acid as ClH

goldmanm: Otherwise looks great to me.

alongd: Following the usage of this dictionary, it is being called on [lines 282-287](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/molecule/generator.py#L284) using Molecule.getFormula(). Since the formula is in alphabetical order (except for cases where C is present, then it precedes H), the dictionary is only called with the `ClH` key. For this reason we also have `HO` instead of `OH`

Issue #409, opened 2015-07-08
Ordering of species in reaction is nondeterministic


connie:
When a reaction string is printed in Chemkin in different jobs, the ordering in which they are listed gets jumbled.  For instance in 2 jobs, I can get:

```
CH3(17)+CH3(17)=H(2)+C2H5(27)                       7.845e+12 0.100     10.600   
```

and 

```
CH3(17)+CH3(17)=C2H5(27)+H(2)                       7.845e+12 0.100     10.600  
```

Which are identical except for the ordering.  This is correct but makes it difficult for the user when one has to search multiple permutations of a single reaction to identify it.  I think we should try to make str(reaction) print the species in order of index.

rwest:
#388 may be relevant here

mliu49:
Species sorting now uses comparison methods (implemented in 0c0c7ed13c5eac36a2b4d073b91156e448e5c3d7), which use the new `sorting_key` property (implemented in c90c407ad93eb32de7cc5a6c0cd2d8192a4b5145). This will provide deterministic sorting, although the current implementation depends on the species label and index, which can change depending on thermo libraries and the order that species are generated.

In cases like this, where the indices are the same, this will guarantee deterministic results.

Issue #388, opened 2015-05-22
rxn.reactants.sort() without providing sorting key


KEHANG:
In `checkForExistingReaction(self, rxn)`, `rxn.reactant.sort()` is firstly implemented to created unique order of multiple reactants. But this sort() doesn't provide any sorting key. After some testing, I guess `sort()` method if no sorting key provided, is sorting based on object's memory address (it's only a hypothesis, anyone who is familiar with it please let us know.), which makes the order of reactants in a same reaction non-deterministic if you run the same RMG job multiple times (I've already observed this phonomenon). 

But the reason why currently RMG still can produce same final mechanism is that we only allow one species instance to exist in memory for one species. 

The potential problem of this `sort()` implementation is when unpickling restart files, the memory address of loaded reactants will change so the order of reactants in reactions should change accoordingly, if not, some existing reactions cannot be recognized further leading to multiple DUPLICATE reactions in `rmg.reactionModel.reactionDict`.

I suggestion this `sort()` should be provided by some appropriate sorting key to make it more deterministic. I've tried species.index since every reactive species will have a unique index at this time point (when implementing `checkForExistingReaction(self, rxn)`), which can now solve the non-deterministic problem. But more thoughts about sorting key are welcome.

nickvandewiele:
The example you describe in the third paragraph would be an excellent unit test. 

Would the following example sometimes fail because the memory addresses are allocated in a random way?

```python
import cPickle
from rmgpy.species import Species
from rmgpy.reaction import Reaction
 
a = Species(...)
b = Species(...)
 
r = Reaction()
r.reactants = [a,b]
 
r.reactants.sort()
 
r_pickle = cPickle.loads(cPickle.dumps(r))
r_pickle.reactants.sort()
 
for spc, spc_pickle in zip(r.reactants, r_pickle.reactants):
    assert.assertTrue(spc.isIsomorphic(spc_pickle)
```

KEHANG:
The example you provided will always return True. The non-deterministic feature isn't because the order of memory allocation (unpickling) for pickled objects is random. The unpickling order is actually fixed but would be complicated enough when you are unpickling a large hierachical structure like `self.reactionModel` and the unpickling order is not same as the order you create those objects previously. The following example (it will fail) can demonstrate what I just said:
```python
import cPickle
from rmgpy.species import Species
from rmgpy.reaction import Reaction

a = Species(index=1, label='CO')
[...]

PR #1329, opened 2018-03-22
Chemical identity comparison for Molecules and Species

mliu49:
Improvements:
- Molecules can be instantiated with SMILES or InChI
- New read-only properties for Molecules and Species (fingerprint, InChI, SMILES, multiplicity)
- New `is_same` method checks "chemical identity" via InChI and multiplicity (better name suggestions welcome)
- Use `is_same` in `checkForExistingSpecies()` instead of isomorphism. Testing has shown that this is much faster and also more accurate.

-----

# April 2019 Revision

I have finally refactored and updated this branch. There are a number of changes from the prior version, but I decided not to make a new PR.

I decided to abandon the idea of creating a new `is_same` method, since it seemed to add unnecessary complexity, especially since people are already familiar with using `isIsomorphic` and `isIdentical`.

## Major changes
- Introduce `strict` argument to almost all isomorphism related methods. The default is `strict=True` which is the same as current behavior. Specifying `strict=False` ignores electrons when comparing atoms and bonds, meaning bond orders, radicals, and lone pairs are not considered at all. The result is a resonance structure independent comparison.
- Use `strict=False` comparisons for:
    - Checking if species exist
    - Reaction degeneracy determination
    - Product checking during reaction generation
- This allows us to do these tasks without extra resonance structure generation. While isomorphism with `strict=False` is slower than strict isomorphism (due to more potential matches), it is more than compensated for by eliminating resonance structure generation.
- This also eliminates the need for keeping non-representative resonance structures in the molecule list for Species. Most of that code was removed from `checkForExistingSpecies`, though non-reactive Molecules can still be created.

## Quality of life changes
- Update Molecule and Species properties (fingerprint, InChI, SMILES, multiplicity)
- Both Molecule and Species objects can be instantiating by providing InChI or SMILES argument instead of using `fromInChI` or `fromSMILES`
- Sort species in reactions by index, which allows more consistent reaction strings in log files and Chemkin files. This solves #388 and #409. No idea why this wasn't done earlier.

## Bugfixes
- In `ensure_independent_atom_ids`, make sure we select a reactive molecule when generating resonance structures
- In `ensure_independent_atom_ids`, pass `keep_isomorphic=True` to `ensure_species` for accurate degeneracy determination


rmgpy/molecule/translator.py
```diff
@@ -457,6 +457,21 @@ def _check_output(mol, identifier):
...
+                if isinstance(key, tuple):
+                    # Add to the corresponding symbol
+                    mol_elementcount[key[0]] += count
+                    # Indicate that we found an isotope
+                    found = True
+                    # Remove the entry from the dictionary
+                    del mol_elementcount[key]
+            # If we didn't find anything, that would mean the result is wrong
+            conditions.append(found)
+
```
mjohnson541: So the check below (inchi_elementcount == mol_elementcount) works because util.retrieveElementCount(identifier) doesn't account for isotopes and the above loop moves isotopes to there corresponding element?

rmgpy/data/kinetics/family.py
```diff
@@ -1841,17 +1841,13 @@ def generate_products_and_reactions(order):
             
                 products0 = reaction.products[:] if forward else reaction.reactants[:]
 
-                # For aromatics, generate aromatic resonance structures to accurately identify isomorphic species
-                if prod_resonance:
```
mjohnson541: Why can we get rid of this stuff?

mliu49:
This PR is out of date now. The commits related to InChI generation for isotopes has been merged via #1453. The species comparison changes will be refactored and a new PR will be opened once that's ready.

mliu49:
Please see the updated initial comment for summary and explanation of the new changes.

Performance comparison with eg1:
- master: 430 seconds
- species_identity: 370 seconds

This is a 14% reduction in runtime. I expect more significant differences for models with larger species and/or more resonance structures.

The model core is identical on these two branches. The edge on this branch has 7 fewer species, most likely due to species being identified as duplicates using the non-strict comparison. I think this is a positive change though, since this is a more robust method of identifying whether species are truly the same (as compared to generating resonance structures).

rmgpy/reaction.py
```diff
@@ -1321,28 +1326,33 @@ def get_mean_sigma_and_epsilon(self, reverse=False):
         mean_epsilons = reduce((lambda x, y: x * y), epsilons) ** (1 / len(epsilons))
         return mean_sigmas, mean_epsilons
 
-def isomorphic_species_lists(list1, list2, check_identical=False, only_check_label=False, generateInitialMap=False):
+
+def same_species_lists(list1, list2, check_identical=False, only_check_label=False, generate_initial_map=False, strict=True):
```
mjohnson541: If we're renaming this already can we either generalize it so it works for N species or rename it in a less general way?

rmgpy/species.py
```diff
@@ -154,6 +155,14 @@ def __reduce__(self):
         """
         return (Species, (self.index, self.label, self.thermo, self.conformer, self.molecule, self.transportData, self.molecularWeight, self.energyTransferModel, self.reactive, self.props))
 
+    @property
+    def InChI(self):
+        """InChI string representation of this species. Read-only."""
+        if self._inchi is None:
+            if self.molecule:
+                self._inchi = self.molecule[0].InChI
```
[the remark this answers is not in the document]
mliu49: `Molecule.InChI` is a property which will set the InChI if it's not already there.

rmgpy/reaction.py
mliu49: You mean since it can only compare list of up to 3 species? Do you any name suggestions?

rmgpy/species.py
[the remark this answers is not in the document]
mjohnson541: Ok

rmgpy/reaction.py
mjohnson541: isomorphic_species_trios?

mjohnson541: isomorphic_species_lists_le3, isomorphic_species_lists_lt4?

mjohnson541: Do you think it's not worth worrying about the confusion that much?

PR #1331, opened 2018-03-26
Sampling reactors

mjohnson541:
The pull request enables reactors to sample over a range of conditions rather than a single condition.  Conditions are chosen by maximizing distance from prior run conditions (weighted by how recent they were, the conversion/time reached and how many objects they returned).  These variable condition reactors terminate after a defined number of iterations (nSimsTerm).  This should allow us to avoid missing chemistry that occurs between two fixed reactors and improve our efficiency with simulations.


alongd:
Could you add some documentation on this feature?

mjohnson541:
Documentation should be added now.

rmgpy/rmg/main.py
```diff
@@ -1512,7 +1547,119 @@ def initializeLog(verbose, log_file_name):
...
+                self.tmax = term.time.value_si
+        
+        self.normalize = hasattr(reactionSystem,'initialMoleFractions')
+        self.conditionList = []
+        self.scaledConditionList = []
+        self.ts = []
+        self.convs = []
+        self.Ns = []
+        
+    def addtConvN(self,t,conv,N):
```
alongd: Please change new function names to lowercase

rmgpy/rmg/main.py
```diff
@@ -552,10 +568,16 @@ def execute(self, **kwargs):
...
+                    conditions = self.RMG_Memories[index].getCond()
                 )
                 self.updateReactionThresholdAndReactFlags(
                     rxnSysUnimolecularThreshold=reactionSystem.unimolecularThreshold, 
                     rxnSysBimolecularThreshold=reactionSystem.bimolecularThreshold)
+        else:
+            for index, reactionSystem in enumerate(self.reactionSystems):
+                self.RMG_Memories.append(RMG_Memory(reactionSystem))
+                self.RMG_Memories[index].generateCond()
+                logging.info('conditions choosen for reactor {0} were {1}'.format(index,self.RMG_Memories[index].getCond()))
```
alongd: What's the difference between this an `logConditions()`?

rmgpy/rmg/main.py
```diff
@@ -764,15 +798,15 @@ def execute(self, **kwargs):
         # Run sensitivity analysis post-model generation if sensitivity analysis is on
         for index, reactionSystem in enumerate(self.reactionSystems):
             
-            if reactionSystem.sensitiveSpecies:
+            if reactionSystem.sensitiveSpecies and reactionSystem.sensConditions:
```
alongd: I think that `sensConditions` is `None` if ranges aren't specified. Does this condition allow "normal" SA to be run?

rmgpy/rmg/main.py
```diff
@@ -782,8 +816,9 @@ def execute(self, **kwargs):
                     pdepNetworks = self.reactionModel.networkList,
                     sensitivity = True,
                     sensWorksheet = sensWorksheet,
-                    modelSettings = self.modelSettingsList[-1],
+                    modelSettings = ModelSettings(toleranceMoveToCore=1e8,toleranceInterruptSimulation=1e8),
```
alongd: Could you explain this change?

rmgpy/rmg/mainTest.py
```diff
@@ -151,7 +152,19 @@ def testRMGSeedWorks(self):
...
+        """
+        test that RMG Memory objects function properly
+        """
+        for rxnsys in self.rmg.reactionSystems:
+            Rmem = RMG_Memory(rxnsys)
+            Rmem.generateCond()
+            Rmem.getCond()
+            Rmem.addtConvN(1.0,.2,2)
+            Rmem.generateCond()
+            Rmem.getCond()
```
alongd: Perhaps add relevant assertions and helpful error messages?

rmgpy/solver/base.pyx
```diff
@@ -200,7 +200,21 @@ cdef class ReactionSystem(DASx):
         if surfaceReactions is None:
             surfaceReactions = []
             
-
+        if conditions:
+            isConc = hasattr(self,'initialConcentrations')
+            keys = conditions.keys()
+            if 'T' in keys and hasattr(self,'T'):
```
alongd: Could `T` be in `keys` w/o having a self.T attribute?

rmgpy/solver/base.pyx
```diff
@@ -200,7 +200,21 @@ cdef class ReactionSystem(DASx):
...
+                self.T = Quantity(conditions['T'],'K')
+            if 'P' in keys and hasattr(self,'P'):
+                self.P = Quantity(conditions['P'],'Pa')
+            for k in keys:
+                if isConc:
+                    if k in self.initialConcentrations.keys():
+                        self.initialConcentrations[k] = Quantity(conditions[k],'mol/m^3')
```
alongd: I'm not too familiar with LiqReactor, but couldn't the conditions be specified in different units? If so, we're missing a conversion

rmgpy/rmg/main.py
```diff
@@ -537,12 +540,25 @@ def execute(self, **kwargs):
...
+            if x.Trange:
+                Tlist.append(x.Trange[1].value_si)
+            elif x.T:
+                Tlist.append(x.T.value_si)
+                
+        self.Tmax = max(Tlist)
+        
+        self.nSimsTerms = [0 for i in xrange(len(self.reactionSystems))]
         
+        self.RMG_Memories = []
```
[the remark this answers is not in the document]
mjohnson541: Done

rmgpy/rmg/main.py
mjohnson541: fixed

rmgpy/rmg/main.py
mjohnson541: done

rmgpy/rmg/main.py
```diff
@@ -1512,7 +1547,119 @@ def initializeLog(verbose, log_file_name):
...
+        find the next condition to run at by solving an optimization problem
+        this optimization problem maximizes distance from prior conditions weighted more if they are more recent,
+        minimizes conversion and completion time and maximizes number of objects added
+        the resulting condition is added to the end of conditionList
+        """
+        if self.conditionList == []:
+            self.conditionList.append({key:value[0] for key,value in self.Ranges.iteritems()})
+            self.scaledConditionList.append({key:0.0 for key,value in self.Ranges.iteritems()})
+        elif len(self.conditionList[0]) == 0:
+            return
```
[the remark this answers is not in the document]
mjohnson541: I only added the return because a line there is required and it provided more information than pass.  I'll convert it to a pass.

rmgpy/rmg/main.py
```diff
@@ -537,12 +540,25 @@ def execute(self, **kwargs):
...
+        Tlist = []
+        for x in self.reactionSystems:
+            if x.Trange:
+                Tlist.append(x.Trange[1].value_si)
+            elif x.T:
+                Tlist.append(x.T.value_si)
+                
+        self.Tmax = max(Tlist)
+        
+        self.nSimsTerms = [0 for i in xrange(len(self.reactionSystems))]
```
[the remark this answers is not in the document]
mjohnson541: Since self.reactionSystems isn't a numpy array object that doesn't clean the syntax up too much so I'd prefer to leave it as is.

rmgpy/rmg/input.py
```diff
@@ -133,36 +133,43 @@ def adjacencyList(string):
...
+                  sensitivityThreshold=1e-3,
+                  sensitivityTemperature=None,
+                  sensitivityPressure=None,
+                  sensitivityMoleFractions=None,
                   ):
     logging.debug('Found SimpleReactor reaction system')
     
     for value in initialMoleFractions.values():
-        if value < 0:
+        if value != list and value < 0:
```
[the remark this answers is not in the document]
mjohnson541: I'm really embarrassed about these ones...

rmgpy/rmg/input.py
```diff
@@ -133,36 +133,43 @@ def adjacencyList(string):
...
             raise InputError('Initial mole fractions cannot be negative.')
+        elif value == list:
```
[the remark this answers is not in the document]
mjohnson541: I'm really embarrassed about these ones...

rmgpy/rmg/input.py
```diff
@@ -133,36 +133,43 @@ def adjacencyList(string):
 def simpleReactor(temperature,
                   pressure,
                   initialMoleFractions,
+                  nSimsTerm=4,
```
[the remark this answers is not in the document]
mjohnson541: I chose 4 originally because its a reasonable number of reactors for spanning a 1-D temperature interval, but perhaps it should be increased.

rmgpy/rmg/input.py
```diff
@@ -133,36 +133,43 @@ def adjacencyList(string):
...
-        if value < 0:
+        if value != list and value < 0:
             raise InputError('Initial mole fractions cannot be negative.')
+        elif value == list:
+            if len(value) > 2:
+                raise InputError('mole ranges can only have one or two entries')
+            elif type(value[0]) < 0:
+                raise InputError('Initial mole fractions cannot be negative.')
+            elif value[1] < value[0]:
+                raise InputError('Initial mole fraction range out of order')
```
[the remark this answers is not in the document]
mjohnson541: done

rmgpy/rmg/input.py
```diff
@@ -133,36 +133,43 @@ def adjacencyList(string):
...
     
     for value in initialMoleFractions.values():
-        if value < 0:
+        if value != list and value < 0:
             raise InputError('Initial mole fractions cannot be negative.')
+        elif value == list:
+            if len(value) > 2:
+                raise InputError('mole ranges can only have one or two entries')
+            elif type(value[0]) < 0:
+                raise InputError('Initial mole fractions cannot be negative.')
```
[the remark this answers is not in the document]
mjohnson541: Must have not been a good day for me...

rmgpy/rmg/main.py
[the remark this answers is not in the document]
alongd: Perhaps we could do without the `elif` here?

rmgpy/rmg/main.py
mjohnson541: It should do so now since the non-ranged values will default as sensConditions

rmgpy/rmg/input.py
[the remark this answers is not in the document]
mjohnson541: I've set it to 6 for now

rmgpy/rmg/mainTest.py
mjohnson541: I couldn't really think of anything I felt the need to specifically check in the object other than that the basic operations could be done.  It is deterministic so I could hard code the test (check whether its generating the same condition), but I feel like unless I'm checking something I have reason to worry about that just forces someone to change the hard coded numbers anytime they change something in the weighing or optimization algorithm.

mjohnson541:
I made some significant changes to the documentation so if you could look that over I'd appreciate it.

mliu49:
I would like to see some more test results before this gets merged, including but not limited to:
1) RMG-tests
2) Comparison of output/runtime between multiple reactors vs. sampling reactor over a T range (and potentially P and composition ranges as well)

To be honest, the mistakes which Alon pointed out were a bit concerning. It would be best if you could improve the test coverage of new code which you added.

examples/rmg/SR_test/input.py
```diff
@@ -0,0 +1,59 @@
...
+    label='O2',
+    reactive=True,
+    structure=SMILES('[O][O]')
+)
+
+# Reaction systems
+simpleReactor(
+    temperature=[(1000,'K'),(1500,'K')],
+    pressure=[(1.0,'bar'),(10.0,'bar')],
+    nSimsTerm=12,
```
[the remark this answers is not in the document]
alongd: Could you direct me to the comment that was added?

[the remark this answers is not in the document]
mjohnson541: Sorry, in my tests it turns out 12 is actually a really good number for this one so I don't think a comment is necessary.  I've also added a rule of thumb for choosing nSimsTerm to the documentation.

mjohnson541:
@mliu49 I've greatly improved coverage from the original PR, however it doesn't show up because except for testing the new RMG_Memory objects it makes the most sense to test the rest of this system using functional tests.  I believe under the functional tests nearly all lines of code I've added are covered now.  

I'm planning to present some results related to the output in RMG meeting this week.  What question are you trying to answer with runtime information?

mliu49:
I understand that it may be difficult to write unit tests for these changes, and I wasn't referring explicitly to the coverage percentage. I mainly meant that the fact that you didn't catch those errors during testing suggested that you weren't running the appropriate tests, whether unit or functional.

I mentioned runtime comparison since you said in your initial comment that this should provide improved efficiency. That makes sense, but I was hoping you could quantify it, even if just in one example case.

mjohnson541:
I apologize for the state of the original PR, but I'm pretty confidant the tests I've added since are sufficient to thoroughly test this feature.  

Ok, I expect I can demonstrate that.

mjohnson541:
So after discovering a space sampling issue with my original method for choosing conditions I discussed the sampling problem with Kevin Silmore from the Swan lab and he proposed a different sampling algorithm I've now adopted in this PR.  In this algorithm we evaluate the objective function at a grid of condition points, normalize these values and then choose a point where each point has probability of its normalized objective function value of being chosen.  After a grid point is chosen a random step (of maximum length half the distance to the next grid point) is taken from the grid point to choose a final sample point.

mjohnson541:
The deterministic-ness issue was actually unrelated to random number generation and actually related to non-deterministic solver output.  By removing termination conversion and time from the objective function the algorithm is now as deterministic as regular RMG.

mjohnson541:
Ok, I believe I've fixed everything that needed fixed in this.

documentation/source/users/rmg/input.rst
```diff
@@ -257,6 +257,61 @@ sensitivities for dln(C_i)/dln(k_j) > sensitivityThreshold  or dlnC_i/d(G_j) > s
...
+		sensitivityThreshold=0.001,
+		balanceSpecies = "N2",
+		)
+
+Note that increasing ``nSimsTerm`` improves convergence over the entire range, but convergence is only guaranteed at the 
+last ``nSimsTerm`` reaction conditions. Theoretically if ``nSimsTerm`` is set high enough the RMG model converges over the 
+entire interval.  Even convergence for smaller ``nSimsTerm`` (values near the number of individual reactors you would 
+use to span the system) is usually superior to the convergence achieved (over the whole interval) using individual
+reactors.  If there is a particular reaction condition you expect to converge more slowly than the rest of the range 
+and are worried might be missed in theory there is virually no cost to using a single condition reactor at that
```
alongd: please add a comma `worried might be missed, in theory there is`

documentation/source/users/rmg/input.rst
```diff
@@ -257,6 +257,61 @@ sensitivities for dln(C_i)/dln(k_j) > sensitivityThreshold  or dlnC_i/d(G_j) > s
...
+		sensitivityMoleFractions = {"ethane":0.1,"O2":0.9},
+		sensitivity=["ethane","O2"],
+		sensitivityThreshold=0.001,
+		balanceSpecies = "N2",
+		)
+
+Note that increasing ``nSimsTerm`` improves convergence over the entire range, but convergence is only guaranteed at the 
+last ``nSimsTerm`` reaction conditions. Theoretically if ``nSimsTerm`` is set high enough the RMG model converges over the 
+entire interval.  Even convergence for smaller ``nSimsTerm`` (values near the number of individual reactors you would 
+use to span the system) is usually superior to the convergence achieved (over the whole interval) using individual
```
alongd: I only understood this in a second read, please emphasize: `(values near the number of individual reactors you would 
have used to span the system if not implementing range-based reactors)`

documentation/source/users/rmg/input.rst
```diff
@@ -257,6 +257,61 @@ sensitivities for dln(C_i)/dln(k_j) > sensitivityThreshold  or dlnC_i/d(G_j) > s
...
+and are worried might be missed in theory there is virually no cost to using a single condition reactor at that 
+condition along with the range based reactor.  Interval convergence on most of the interval can usually be
+ achieved by using the number of individual reactors you would typically use to span the space and adding one or two. 
+This is not a particularly accurate rule of thumb, (sometimes you can get by with significantly fewer). This number of
+reactors will usually give you significantly better convergence than using individual reactors. How different this 
+result is from individual reactors and full interval convergence is incredibly system dependent not just on the 
+intervals choosen but the chemistry occuring.  
+
+If normalization of the ranged conditions is undesirable one can use a ``balanceSpecies``.  When a ``balanceSpecies`` is used
+instead of normalizing the mole fractions the concentration of the defined ``balanceSpecies`` is adjusted to maintain
```
alongd: 'is used, instead of normalizing the mole fractions, the concentration'

rmgpy/rmg/input.py
```diff
@@ -133,36 +133,51 @@ def adjacencyList(string):
...
+    for key,value in initialMoleFractions.iteritems():
+        if not isinstance(value,list):
+            initialMoleFractions[key] = float(value)
+            if value < 0:
+                raise InputError('Initial mole fractions cannot be negative.')
+        else:
+            if len(value) != 2:
+                raise InputError("Initial mole fraction values must either be a number or a list with 2 entries")
+            initialMoleFractions[key] = [float(value[0]),float(value[1])]
+            if len(value) > 2:
```
alongd: The condition `if len(value) != 2` assures that `len(value)` equals 2. I think that the `if len(value) > 2` check is unnecessary.

rmgpy/rmg/input.py
```diff
@@ -133,36 +133,51 @@ def adjacencyList(string):
...
+        T = Quantity(temperature)
+    else:
+        if len(temperature) != 2:
+            raise InputError('Temperature and pressure ranges can either be in the form of (number,units) or a list with 2 entries of the same format')
+        T = [Quantity(t) for t in temperature]
+        
+    if not isinstance(pressure,list):
+        P = Quantity(pressure)
+    else:
+        if len(pressure) > 2:
```
alongd: Should be `!=` instead of `>`

rmgpy/rmg/input.py
```diff
@@ -178,26 +193,72 @@ def simpleReactor(temperature,
...
+    else:
+        sensConditions = sensitivityMoleFractions
+        sensConditions['T'] = Quantity(sensitivityTemperature).value_si
+        sensConditions['P'] = Quantity(sensitivityPressure).value_si
+
+    
+    system = SimpleReactor(T, P, initialMoleFractions, nSimsTerm, termination, sensitiveSpecies, sensitivityThreshold,sensConditions)
     rmg.reactionSystems.append(system)
+    
+    rmg.balanceSpecies = balanceSpecies
```
alongd: Do we allow more than one balance species? If not, do we check that only one was entered per reactor?

documentation/source/users/rmg/input.rst
```diff
@@ -257,6 +257,61 @@ sensitivities for dln(C_i)/dln(k_j) > sensitivityThreshold  or dlnC_i/d(G_j) > s
...
+reactors.  If there is a particular reaction condition you expect to converge more slowly than the rest of the range 
+and are worried might be missed in theory there is virually no cost to using a single condition reactor at that 
+condition along with the range based reactor.  Interval convergence on most of the interval can usually be
+ achieved by using the number of individual reactors you would typically use to span the space and adding one or two. 
+This is not a particularly accurate rule of thumb, (sometimes you can get by with significantly fewer). This number of
+reactors will usually give you significantly better convergence than using individual reactors. How different this 
+result is from individual reactors and full interval convergence is incredibly system dependent not just on the 
+intervals choosen but the chemistry occuring.  
+
+If normalization of the ranged conditions is undesirable one can use a ``balanceSpecies``.  When a ``balanceSpecies`` is used
```
alongd: Perhaps comment that this applies only for gas phase reactors?

alongd:
Will [`self.Tmax = max([x.T for x in self.reactionSystems]).value_si`](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/rmg/main.py#L544) still function if T is a list?

mjohnson541:
Where does that occur?

mjohnson541:
Ok, I believe I've fixed all of the above issues.

mjohnson541:
Are there any other issues or can we merge this?

alongd:
Could you take a look at the Codacy report? I think there a some recommendations there we could accept. Everything else looks good to me.

mjohnson541:
Ok, I've fixed some of codacy's recommendations.

Issue #1327, opened 2018-03-22
Nondeterministic kinetics/thermo when running RMG-tests


goldmanm:
Sometimes when running tests in RMG-tests, the models give different values, which is not always intuitive.

[...]

Issue #332, opened 2015-02-04
calculateCyclicSymmetryNumber gave wrong result when dealing with 'C1(C(C(C(C(C1C2CCC2)C3CCC3)C4CCC4)C5CCC5)C6CCC6)C7CCC7'


KEHANG:
I'm trying to calculate the cyclic symmetry number for the following molecule: C1(C(C(C(C(C1C2CCC2)C3CCC3)C4CCC4)C5CCC5)C6CCC6)C7CCC7, whose structure is
![image](https://cloud.githubusercontent.com/assets/2739496/6044224/f166f8a8-ac5f-11e4-885e-a497ce9a10af.png)

My script is simply :
from rmgpy.molecule.symmetryTest import *

def main():
    molecule = Molecule().fromSMILES('C1(C(C(C(C(C1C2CCC2)C3CCC3)C4CCC4)C5CCC5)C6CCC6)C7CCC7')
    # calculate it 10 times
    for i in range(10):
        symNum = calculateCyclicSymmetryNumber(molecule)
        print 'SymNum: ', symNum
    print('********************')

if **name** == "**main**":
    main()

But the results is very strange: first they are not 12, which is the right answer; second, results are not reproducible, meaning running this script 10 times gives 10 different results. Enoch also confirmed that.

connie:
Please rerun it and check what happens after making a species object and using the generateResonanceIsomers function.

goldmanm:
I added this to a list of molecules that I'm writing unit tests for. Even with resonance isomers, I ran it multiple times and obtain a Total symmetry number of 3 each time. 

I am not sure if the unreliability is due to resonance  (especially since there aren't any radicals in the molecule), so I am not sure what is causing it.

mliu49:
Fixed by #1333.

PR #1333, opened 2018-03-28
Fix symmetry inconsistencies

goldmanm:
Previously, the cyclic symmetry algorithm assumed a cycle is planar,
and incorrectly identified some cyclic structures as symmetric
when they are flipped over. This commit looks at the groups attached
to the atom where flipping occurs and makes sure that it would not
result flipping the other groups attached to the ring before assigning
a symmetry value.

Thanks to Max for discussion.

This solves issue #332 and possibly part of #1327.

### DIFF
--- a/rmgpy/molecule/symmetry.py
+++ b/rmgpy/molecule/symmetry.py
@@ -398,7 +398,45 @@ def calculateCyclicSymmetryNumber(molecule):
                     break
                 min_index += 1
                 max_index += -1
-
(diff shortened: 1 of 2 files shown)


mliu49:
It might help to hear an explanation of the code that you added, since it's difficult to understand without being familiar with the symmetry algorithm.

I'm also confused about the  symmetry number for the snowflake in #332 (and by extension, the symmetry number for cyclohexane) and would like to discuss once you get back.

PR #1334, opened 2018-03-29
Identifier fixes

mliu49:
Very straightforward fixes related to identifier generation:
- If a molecule doesn't have atoms, then don't try generating an identifier.
- When counting elements based on InChI, add counts from multiple instances of the same element. This can occur with disconnected molecules (see unit test).

### DIFF
--- a/rmgpy/molecule/translator.py
+++ b/rmgpy/molecule/translator.py
@@ -509,6 +509,10 @@ def _write(mol, identifier_type, backend):
 
     Returns a string identifier of the requested type.
     """
+    # Check that the molecule is not empty
+    if not mol.atoms:
+        return ''
+
     for option in _get_backend_list(backend):
         if option == 'rdkit':
             try:
--- a/rmgpy/molecule/translator.py
+++ b/rmgpy/molecule/translator.py
@@ -345,6 +345,8 @@ def _rdkit_translator(input_object, identifier_type, mol=None):
             raise ValueError('Identifier type {0} is not supported for reading using RDKit.'.format(identifier_type))
         if rdkitmol is None:
             raise ValueError("Could not interpret the identifier {0!r}".format(input_object))
+        if mol is None:
+            mol = mm.Molecule()
         output = fromRDKitMol(mol, rdkitmol)
     elif isinstance(input_object, mm.Molecule):
         # We are converting from a molecule to a string identifier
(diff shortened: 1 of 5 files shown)


rmgpy/molecule/translator.py
```diff
@@ -509,6 +509,10 @@ def _write(mol, identifier_type, backend):
 
     Returns a string identifier of the requested type.
     """
+    # Check that the molecule is not empty
+    if not mol.atoms:
+        return ''
```
mjohnson541: My thought is that an empty molecule object is usually a really bad thing, so shouldn't we have a more easily recognizable string than ''?

rmgpy/molecule/util.py
```diff
@@ -48,7 +48,10 @@ def retrieveElementCount(obj):
                 element, count = match.groups()
                 if count is '':
                     count = 1
-                element_count[element] = int(count)
+                if element in element_count:
```
mjohnson541: How did not having this section create problems in disconnected inchi generation?

rmgpy/molecule/translator.py
mliu49: Do you have examples of when it would really bad? Currently, the reason I added this is for when you instantiate an empty Molecule instance and try to do something with it.
```
mol = Molecule()
mol.toSMILES()  # ValueError
mol.toInChI() # ValueError
mol.__repr__() # ValueError
mol.__str__() # ValueError
```
I feel like these should be able to succeed with an empty Molecule.

mjohnson541: Ok, we talked about this.

mjohnson541:
I mean if we're doing the same for SMILES and InChI I don't feel like we have any reason to make that one an error and the other ones empty strings.

mliu49:
Your last comment actually reminded me of another fix that I needed to make, so I'll go ahead an explain it here. Both `_rdkit_translator` and `_openbabel_translator` have the optional `mol` argument which is used when reading an identifier. I had added proper handling to `_openbabel_translator` if no `mol` argument was provided, but I forgot to implement it in `_rdkit_translator`.

mliu49:
@mjohnson541, I think this is ready to be merged, unless you have any other comments.

mjohnson541:
Oh, sorry I didn't realize you'd added in your new changes yet.  They look fine.

PR #1338, opened 2018-04-05
New reaction filtering and trimolecular reactions

cgrambow:
This PR slightly modifies the way in which RMG handles reaction filtering and it implements full support for trimolecular reaction families (reactions involving three reactants and/or three products).

 **Reaction Filtering**

If reaction filtering is turned on, RMG only allows two species (A and B) to react if

bimolecfilt

where the threshold rate constant was supposedly set to the diffusion limit for bimolecular reactions (1e7 m^3/mol/s). There are several issues with this value:

1. Using self-diffusion of oxygen as an example, its bimolecular diffusion rate constant can be calculated using Chapman-Enskog theory (to get the diffusivity) and the Smoluchowski equation (to get the diffusion rate constant). At a temperature of 1000 K and pressure of 1 atm, the diffusion rate constant is 5.7e11 m^3/mol/s; several orders of magnitude larger than the previous value!
2. The previous value is lacking theoretical temperature- and pressure-dependence. The diffusion rate should scale as temperature^1.5/pressure.
3. Several reactions (e.g., radical recombination and oxygen addition) are faster than 1e7 m^3/mol/s, which means they could have been filtered out erroneously with the previous algorithm.

I tested different values for the threshold, some results shown here (all for `eg1`):

| Tolerance | Threshold | Core species | Core reactions | Edge species | Edge reactions | Time (s) |
| --- | --- | --- | --- | --- | --- | --- |
| 0.1 | no filter | 22 | 56 | 324 | 1097 | 236 |
| 0.1 | 1e7 | 22 | 49 | 94 | 227 | 141 |
| 0.1 | 8.9e11 | 22 | 56 | 235 | 692 | 196 |
| 0.01 | no filter | 97 | 1158 | 2439 | 5939 | 2352 |
| 0.01 | 1e7 | 97 | 1158 | 2439 | 5939 | 2352 |
| 0.01 | **5e7** | 105 | 1605 | 3308 | 8125 | **3125** |
| 0.01 | **1e8** | 227 | 4759 | 12761 | 36497 | **37755** |
| 0.01 | 1e9 | 230 | 5791 | 19417 | 56786 | 59869 |

At a tolerance of 0.1, there does not seem to be a very significant increase in time if we use the correct bimolecular diffusion rate constant as a threshold. However, at a tolerance of 0.01, the effects are much more drastic. Notice the bold values in the above table: As we jump from a threshold of 5e7 to 1e8, the time required for RMG convergence increases by an order of magnitude and the model becomes significantly larger. Of the 3154 additional reactions in the core, 202 have rate constants greater than 1e7 m^3/mol/s. Nonetheless, for the `eg1` example, picking up the additional reactions does not seem to affect the major simulation profiles by a lot.

As a result of this analysis, I have opted to make the filtration threshold user-definable, but to retain the correct theoretical temperature- and pressure-dependence, such that the threshold is now calculated as

kthreshnew

where `filterThreshold` is the new user-definable parameter. It is set to a default value of 5e8, which corresponds to a threshold value slightly above 1e8 at 1000 K and 1 atm, such that fast reactions can now be picked up reliably. Use a lower value if you desire the old RMG behavior.

For liquid phase reactions, the threshold is now calculated based on an approximate diffusion limit in solution, which is often within the same order of magnitude as the previous threshold, such that liquid phase simulations do not change a lot compared to previously.

**Trimolecular Reactions**

Trimolecular reactions are also handled in this PR because their treatment also requires a filtration threshold, which should be consistent with the aforementioned analysis. For most gas phase systems, one would not expect trimolecular reactions to be of any significance, but there exist some reactions in the liquid phase where three reactants participate. Often this involves an acid catalyst, which first forms a complex with one of the reactants and then reacts with the other reactant. However, because we don't want to add a whole bunch of complexity to RMG by treating van der Waals complexes as separate species, implementing them as trimolecular reactions is easier.

Families with three products already worked prior to this PR, but some functionality was missing. Families with three reactants were previously not supported at all, but their implementation is relatively straightforward. I went through all of RMG and looked for code specific to one or two reactants and extended those sections to allow three reactants.

I added a `trimolecularReact` array, which is an array of flags indicating whether a triple of species is allowed to react, similar to `unimolecularReact` and `bimolecularReact`. To save time, this array is only used if families with three reactants or products are being used. If reaction filtering is turned off, this means that an additional cost to check all possible triples of species exists, which is `O(N^3)` in the number of species. However, because the vast majority of triples will not match any of the families, this cost should not be too bad. To make sure this is the case, I ran tests. Results for `eg1` with all families:

| Tolerance | Trimolecular | Core species | Core reactions | Edge species | Edge reactions | Time (s) |
| --- | --- | --- | --- | --- | --- | --- |
| 0.1 | no | 29 | 100 | 814 | 2681 | 495 |
| 0.1 | yes | 29 | 100 | 814 | 2681 | 528 |
| 0.04 | no | 53 | 612 | 5745 | 17492 | 3297 |
| 0.04 | yes | 53 | 612 | 5745 | 17492 | 3535 |

The increase in time is negligible and should have even less of an impact if reaction filtering is turned on. There will also be some additional memory requirement for storing the `trimolecularReact` array.

To filter trimolecular reactions and to extend diffusion limitations in the liquid phase to the trimolecular case, we require estimates of the diffusion limit rate constant, which is non-trivial compared to the bimolecular case. The appropriate analysis was recently completed and can be found in `Flegg (2016) in SIAM J. Appl. Math.`, which extends the Smoluchowski equation to arbitrary orders. I added sections to the documentation detailing the new equations. The new `trimolecularThreshold` array is used for filtering.


mliu49:
Have you had a chance to check with RMG-cat developers regarding their use of trimolecular reactions? I think it would be good to evaluate potential conflicts before merging.

mliu49:
@mazeau, I think you're currently working on merging current RMG-Py into RMG-cat, right? Could you comment on if/how RMG-cat currently deals with trimolecular reactions? If you have the time, could you also take a brief look at this PR to see if there are any major conflicts, i.e. in the general approach?

We could also schedule a Slack discussion if needed.

mazeau:
I wasn't aware of any trimolecular things before, but there might be?  I can take a look at your PR and schedule a slack call with you on Monday, when I'm back from vacation.

mjohnson541:
@cgrambow can you rebase this?  I'll take a look at it.

cgrambow:
Rebased

rmgpy/solver/base.pyx
```diff
@@ -621,6 +625,7 @@ cdef class ReactionSystem(DASx):
         sensitivityRelativeTolerance = simulatorSettings.sens_rtol
         filterReactions = modelSettings.filterReactions
         maxNumObjsPerIter = modelSettings.maxNumObjsPerIter
+        trimolecular = False if self.trimolecularThreshold is None else True
```
mjohnson541: I'm a bit confused.  In general structurally are you trying to infer that trimolecular reactions aren't happening because of the value of self.trimolecularThreshold or is trimolecular itself the input that determines this?  If it's the second I think you should create and check self.trimolecular instead of checking if self.trimolecularThreshold is None.

rmgpy/tools/generate_reactions.py
```diff
@@ -91,10 +91,11 @@ def execute(rmg, **kwargs):
     Returns an RMG object.
     """
     rmg.initialize(**kwargs)
-    
+
```
mjohnson541: Squash this in with the commit where you change enlarge

rmgpy/data/solvation.py
```diff
@@ -138,7 +138,7 @@ def processOldLibraryEntry(data):
     raise NotImplementedError()
 
 
-class SolventData():
+class SolventData(object):
```
mjohnson541: Is there a particular reason for this?

rmgpy/kinetics/diffusionLimited.py
```diff
@@ -115,18 +114,43 @@ def getDiffusionLimit(self, T, reaction, forward=True):
...
+            Dhat[i] = diffusivities[i+1] + Dbar
+
+            for j in range(i+1, N):
+                denom += 1.0 / (diffusivities[i]*diffusivities[j])
+
+        delta = np.sum(Dinv) / denom
+        k_diff = (np.prod(Dhat)**1.5
+                  * 4.0*constants.pi**(alpha+1.0)/math.gamma(alpha)
+                  * (radii/np.sqrt(delta))**(2.0*alpha)
+                  * constants.Na**(N-1.0))  # m^(3*(N-1))/mol^(N-1)/s
```
mjohnson541: So this entirely replaces the original calculation even for the 2 reactant case?

rmgpy/kinetics/diffusionLimitedTest.py
```diff
@@ -111,6 +139,14 @@ def testGetEffectiveRate2to1(self):
         self.assertTrue(effective_rate < self.intrinsic_rates[self.bi_uni_reaction])
         self.assertTrue(effective_rate >= 0.2 * self.intrinsic_rates[self.bi_uni_reaction])
 
+    def testGetEffectiveRate3to2(self):
+        """
+        Tests that the effective rate is limited for a 3 -> 2 reaction
+        """
+        effective_rate = diffusionLimiter.getEffectiveRate(self.tri_bi_reaction, self.T)
+        self.assertTrue(effective_rate < self.intrinsic_rates[self.tri_bi_reaction])
+        self.assertTrue(effective_rate >= 0.2 * self.intrinsic_rates[self.tri_bi_reaction])
```
mjohnson541: Where does the 0.2 come from?

documentation/source/theory/rmg/enlarger.rst
```diff
@@ -70,8 +70,36 @@ For a bimolecular reaction occuring between species A and B, this threshold is s
 
 :math:`R_{bimolecular} = k_{threshold}C_A C_B > \epsilon R_{char}` 
 
+where :math:`k_{threshold} = filterThreshold`. ``filterThreshold`` is set by the user in the input file and its
+default value is :math:`10^{8}`. This is on the same order of magnitude as the collision limit for two hydrogen atoms
```
mjohnson541: You should have units on this since they are relevant

rmgpy/data/kinetics/family.py
```diff
@@ -1752,9 +1752,11 @@ def __generateReactions(self, reactants, products=None, forward=True, prod_reson
                         # Iterate over each pair of matches (A, B)
                         for mapA in mappingsA:
                             for mapB in mappingsB:
-                                reactantStructures = [moleculeA, moleculeB]
+                                # Reverse the order of reactants in case we have a family with only one reactant tree
```
mjohnson541: If I understand this correctly this should probably be squashed in with some of the earlier stuff

documentation/source/users/rmg/input.rst
```diff
@@ -737,6 +737,8 @@ Miscellaneous options::
         saveSimulationProfiles=True,
         verboseComments=False,
         saveEdgeSpecies=True,
+        keepIrreversible=True,
+        trimolecularProductReversible=False,
```
mjohnson541: I don't think an option like this belongs in the options block.  It's a bit of an orphan option, but I think it probably best fits in either the database or generatedSpeciesConstraints blocks.

rmgpy/rmg/input.py
```diff
@@ -313,7 +313,7 @@ def model(toleranceMoveToCore=None, toleranceMoveEdgeReactionToCore=numpy.inf,to
           toleranceMoveEdgeReactionToSurface=numpy.inf, toleranceMoveSurfaceSpeciesToCore=numpy.inf, toleranceMoveSurfaceReactionToCore=numpy.inf,
           toleranceMoveEdgeReactionToSurfaceInterrupt=None,
           toleranceMoveEdgeReactionToCoreInterrupt=None, maximumEdgeSpecies=1000000, minCoreSizeForPrune=50, 
-          minSpeciesExistIterationsForPrune=2, filterReactions=False, ignoreOverallFluxCriterion=False,
+          minSpeciesExistIterationsForPrune=2, filterReactions=False, filterThreshold=1e8, ignoreOverallFluxCriterion=False,
```
mjohnson541: This commit is way out of order as the systems using this input were defined many commits ago.

mjohnson541: Are you certain you didn't say in the documentation that the default was 5e8?

rmgpy/solver/liquid.pyx
```diff
@@ -54,15 +54,16 @@ cdef class LiquidReactor(ReactionSystem):
     cdef public ScalarQuantity P    
     cdef public double V
     cdef public bint constantVolume
+    cdef public double viscosity
```
mjohnson541: Why is this necessary in this commit?

rmgpy/solver/base.pyx
```diff
@@ -920,6 +920,15 @@ cdef class ReactionSystem(DASx):
                     # Set the maximum bi/trimolecular rate by using the user-defined rate constant threshold
```
mjohnson541: You could have a gas phase simulation that was constantVolume.  You should instead check what the object is, probably with isinstance(self,LiquidReactor).

rmgpy/data/solvation.py
cgrambow: Since Python 2.3, fixes/features for classes have only been implemented for new-style classes, so it's generally bad practice to continue using old-style classes. Figured I'd fix it here.

rmgpy/kinetics/diffusionLimited.py
cgrambow: Yes. It simplifies exactly to the previous expression for 2 reactants.

rmgpy/kinetics/diffusionLimitedTest.py
cgrambow: I think it's basically just for checking if the new rate was reduced too much by the diffusion limit, which would be very unlikely. I used 0.2 because that's what was used in `testGetEffectiveRate2to1`.

documentation/source/theory/rmg/enlarger.rst
cgrambow: Good point

rmgpy/solver/liquid.pyx
cgrambow: It's not. Should have been part of the liquid phase commit, but somehow I lost that change. Will update.

rmgpy/data/kinetics/family.py
```diff
@@ -1389,12 +1391,6 @@ def __generateProductStructures(self, reactantStructures, maps, forward):
                 logging.info('{0}\n{1}\n'.format(struct, struct.toAdjacencyList()))
             raise
 
-        # If there are two product structures, place the one containing '*1' first
```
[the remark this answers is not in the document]
cgrambow: I did note that in the commit, but I'll make it clearer.

rmgpy/data/kinetics/common.py
```diff
@@ -405,4 +407,36 @@ def reduce_same_reactant_degeneracy(reaction, same_reactants=None):
...
+                    'are identical'.format(reaction, reaction.degeneracy)
+                )
+            else:
+                reaction.degeneracy *= 0.5
+                logging.debug(
+                    'Degeneracy of reaction {} was decreased by 50% to {} since two of the reactants '
+                    'are identical'.format(reaction, reaction.degeneracy)
+                )
+        else:
+            same_01 = reaction.reactants[0].isIsomorphic(reaction.reactants[1])
```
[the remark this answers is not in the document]
cgrambow: True. I don't remember why I wrote it this way.

rmgpy/data/kinetics/database.py
```diff
@@ -472,6 +472,26 @@ def generate_reactions_from_families(self, reactants, products=None, only_famili
                 same_reactants = True
             elif reactants[0].isIsomorphic(reactants[1]):
                 same_reactants = True
+        elif len(reactants) == 3:
+            same_01 = reactants[0] is reactants[1]
+            same_02 = reactants[0] is reactants[2]
```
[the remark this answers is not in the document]
cgrambow: Apparently I made this mistake multiple times.

rmgpy/data/kinetics/database.py
```diff
@@ -472,6 +472,26 @@ def generate_reactions_from_families(self, reactants, products=None, only_famili
...
+            elif same_02:
+                same_reactants = True
+                reactants[2] = reactants[2].copy(deep=True)
+            else:
+                same_01 = reactants[0].isIsomorphic(reactants[1])
+                same_02 = reactants[0].isIsomorphic(reactants[2])
+                if same_01 and same_02:
+                    same_reactants = 'all'
+                elif same_01 or same_02:
+                    same_reactants = True
```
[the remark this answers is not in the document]
cgrambow: Yeah, number of reactants is nicer.

rmgpy/data/kinetics/common.py
```diff
@@ -405,4 +407,36 @@ def reduce_same_reactant_degeneracy(reaction, same_reactants=None):
                 reaction.reactants[0].isIsomorphic(reaction.reactants[1])
             ):
         reaction.degeneracy *= 0.5
-        logging.debug('Degeneracy of reaction {} was decreased by 50% to {} since the reactants are identical'.format(reaction, reaction.degeneracy))
+        logging.debug(
```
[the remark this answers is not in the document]
cgrambow: I just noticed that RMG does duplicate isomorphism checking every time `same_reactants==False`, so actually we might get a nice speed-up once I change that.

documentation/source/users/rmg/input.rst
cgrambow: I think this option is very similar in spirit to the `keepIrreversible` option, so I am inclined to leave it where it is.

rmgpy/data/kinetics/family.py
```diff
@@ -1521,6 +1548,13 @@ def addReverseAttribute(self, rxn, react_non_reactive=True):
             sameReactants = False
             if len(rxn.products) == 2 and rxn.products[0].isIsomorphic(rxn.products[1]):
                 sameReactants = True
+            elif len(rxn.products) == 3:
+                same_01 = rxn.products[0].isIsomorphic(rxn.products[1])
+                same_02 = rxn.products[0].isIsomorphic(rxn.products[2])
+                if same_01 and same_02:
+                    sameReactants = 'all'
+                elif same_01 or same_02:
```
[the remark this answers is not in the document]
cgrambow: Done

rmgpy/data/kinetics/family.py
```diff
@@ -1593,6 +1627,26 @@ def calculateDegeneracy(self, reaction):
...
+            same_01 = reactants[0] is reactants[1]
+            same_02 = reactants[0] is reactants[2]
+            if same_01 and same_02:
+                same_reactants = 'all'
+                reactants[1] = reactants[1].copy(deep=True)
+                reactants[2] = reactants[2].copy(deep=True)
+            elif same_01:
+                same_reactants = True
+                reactants[1] = reactants[1].copy(deep=True)
+            elif same_02:
```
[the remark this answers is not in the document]
cgrambow: Done

rmgpy/data/kinetics/family.py
cgrambow: Done

rmgpy/data/kinetics/family.py
```diff
@@ -2142,12 +2312,104 @@ def getLabeledReactantsAndProducts(self, reactants, products):
...
+                                            else:
+                                                if same_02 and same_11 and same_20:
+                                                    return reactantStructures, productStructures
+                        else:
+                            continue
+            # if there're some mapping available but cannot match the provided products
+            # raise exception
+            if len(mappingsA)*len(mappingsB)*len(mappingsC) > 0:
+                raise ActionError('Something wrong with products that RMG cannot find a match!')
+            return (None, None)
```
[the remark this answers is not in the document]
cgrambow: Done

rmgpy/rmg/react.py
```diff
@@ -113,7 +113,7 @@ def deflate(rxns, species, reactantIndices):
             pass
 
 
-def reactAll(coreSpcList, numOldCoreSpecies, unimolecularReact, bimolecularReact, trimolecularReact):
+def reactAll(coreSpcList, numOldCoreSpecies, unimolecularReact, bimolecularReact, trimolecularReact=None):
```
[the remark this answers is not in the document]
cgrambow: Done

rmgpy/tools/generate_reactions.py
cgrambow: Done

rmgpy/solver/liquid.pyx
cgrambow: Moved to other commit.

rmgpy/solver/base.pyx
cgrambow: `LiquidReactor` and `SimpleReactor` now have methods to generate the threshold rate constants.

rmgpy/data/kinetics/family.py
```diff
@@ -1728,6 +1782,68 @@ def __generateReactions(self, reactants, products=None, forward=True, prod_reson
...
+                        if reactants[1] is not reactants[2]:
+                            # Reactants stored as A + C + B
+                            generate_products_and_reactions((0, 2, 1))
+                        if reactants[0] is not reactants[1]:
+                            # Reactants stored as B + A + C
+                            generate_products_and_reactions((1, 0, 2))
+                        if reactants[0] is not reactants[2]:
+                            # Reactants stored as C + B + A
+                            generate_products_and_reactions((2, 1, 0))
+                            if reactants[0] is not reactants[1]:
```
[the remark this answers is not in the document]
cgrambow: Added extra if-statement to hopefully cover all cases.

cgrambow:
Addressed comments, squashed things, rebased.

I should mention that this PR no longer implements the temperature- and pressure-dependence during filtering, because it is unclear whether or not the Smoluchowski equation is valid in the gas phase. Therefore, the changes to filtering are relatively minimal. `filterThreshold` is now exactly equal to the threshold rate constant and is set to `1e8 m3/mol/s`, which is on the same order as the collision limit for two hydrogen atoms at 1000 K. The previous RMG behavior for gas phase simulations is obtained exactly when `filterThreshold` is set to `1e7`. For liquid phase reactions, the changes are slightly more significant, because `filterThreshold` is not used and filtering is instead done based on the diffusion limit.

I still think that significant speed-up can be obtained if we implement filtering based on the maximum rates in each family, but someone would have to take the initiative to code that up. I currently don't have enough time to do so.

documentation/source/users/rmg/liquids.rst
```diff
@@ -189,18 +189,35 @@ where `A, B, E, S` and `L` are the same solute descriptors used in the Abraham m
...
+`D`\ :sub:`i` are the individual diffusivities and `σ` is the Smoluchowski radius, which would usually be fitted to
+experiment, but RMG approximates it as the sum of molecular radii. RMG uses the McGowan method for estimating
 radii, and diffusivities are estimated with the Stokes-Einstein equation using experimental solvent 
-viscosities (`\eta` (T)).  In a unimolecular to bimolecular reaction, for example, the forward rate 
-constant (`k`\ :sub:`f`\ ) can be slowed down if the reverse rate (`k`\ :sub:`r, eff`\ ) is diffusion 
-limited since the equilibrium constant (`K`\ :sub:`eq`\ ) is not affected by diffusion limitations. In cases 
-where both the forward and the reverse reaction rates are bimolecular, both diffusive limits are 
+viscosities (`\eta` (T)). In a unimolecular to bimolecular reaction, for example, the forward rate
+constant (`k`\ :sub:`f`\ ) can be slowed down if the reverse rate (`k`\ :sub:`r, eff`\ ) is diffusion-limited
+since theequilibrium constant (`K`\ :sub:`eq`\ ) is not affected by diffusion limitations. In cases
```
mliu49: Accidentally removed space?

rmgpy/rmg/main.py
```diff
@@ -327,13 +329,29 @@ def loadDatabase(self):
...
+                if len(family.forwardTemplate.products) > 2:
+                    family.reversible = False
+                    family.reverseTemplate = None
+                    family.reverseRecipe = None
+                    family.reverse = None
+
         # Determine if trimolecular families are present
         for family in self.database.kinetics.families.itervalues():
-            if len(family.forwardTemplate.reactants) > 2 or len(family.forwardTemplate.products) > 2:
+            if len(family.forwardTemplate.reactants) > 2:
```
mliu49: Does this assume that we won't have any families with 2 reactants going to 3 products?

rmgpy/rmg/input.py
mliu49: For PRs, I think GitHub lists commits by author date rather than commit date, which is rather inconvenient. It's possible that the actual commit order makes more sense.

mliu49: For PRs, I think GitHub lists commits by author date rather than commit date, which is rather inconvenient. It's possible that the actual commit order makes more sense.

rmgpy/data/kinetics/family.py
```diff
@@ -1373,12 +1381,6 @@ def __generateProductStructures(self, reactantStructures, maps, forward):
             productStructures = self.applyRecipe(reactantStructures, forward=forward)
             if not productStructures: return None
         except (InvalidActionError, KekulizationError):
-#            logging.error('Unable to apply reaction recipe!')
-#            logging.error('Reaction family is {0} in {1} direction'.format(self.label, 'forward' if forward else 'reverse'))
-#            logging.error('Reactant structures are:')
-#            for struct in reactantStructures:
-#                logging.error(struct.toAdjacencyList())
-            # If unable to apply the reaction recipe, then return no product structures
```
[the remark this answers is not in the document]
cgrambow: Yeah, I can keep the last line. I actually don't remember why I removed this code, but I vaguely remember discussing it with someone. Was that with you?

rmgpy/rmg/main.py
cgrambow: No, we can have families with 2 reactants to 3 products, but trimolecular will then only be turned on if `trimolecularProductReversible` is `True`.

cgrambow:
I added fixup commits addressing your comments, @mliu49.

mliu49:
Thanks for making the changes! They look great.

I added a commit reducing some code duplication in `getLabeledReactantsAndProducts`. You can squash it into the other commit if you'd like.

cgrambow:
Thanks for the addition. I squashed your commit and all of mine and rebased.

Issue #147, opened 2013-09-12
Handling duplicate PLOGs correctly


shamelmerchant:
I think there are few issues right now how RMG-Py handles them
1) Chemkin allows duplicate plogs having missing pressures we have strict check for it right now.

2) Sum of PLOGS - papers which have sum of Arrhenius equations must be input as
PLOG / 10 A1 n1 Ea1/
PLOG /10 A2 n2 Ea2/

and not 
PLOG / 10 A1 n1 Ea1/
DUPLICATE

PLOG / 10 A2 n2 Ea2/
DUPLICATE

These two notations have different interpolation consequences. Unfortunately Chemkin documentation is not very clear but Ray Speth tested it out and told me we should have the same pressure twice in single PLOG reaction rather than have two duplicates.

rwest:
@edames has just brought this up again (and offered to fix it?!)

enochd:
I have most of this in the form of a doc that i can send upon request, but the text below clarifies how chemkin interprets PLOG and how duplicate reactions are intended to be used. 

In order to illustrate the consequences of providing multiple PLOG expressions and denoting them as 'DUPLICATE' reactions, below are two cases that show how interpolation can arrive at different rate coefficients. These cases are manually computed and compared with rates computed by Chemkin Pro:

![3](https://cloud.githubusercontent.com/assets/5880480/5979824/098c57d8-a879-11e4-922f-c81330d71a4c.jpg)

In the following example, we want to represent a known reaction with rates that are 1 s-1 and 50 s-1 at 1 and 10 atm, respectively. We will compute the half-life of A in a batch reactor at 2atm. Note that the rate of this isomerization reaction is independent of temperature and only dependent on pressure.  Although this does not affect the simulation outcome since the reaction is defined as irreversible, the thermodynamic parameters for the reactant and product have been made equivalent. A constant temperature and pressure homogeneous reactor is simulated in Chemkin-Pro. The species profiles are saved and the half-life time is linearly interpolated from the output:

t1/2 (Case1, Chemkin-Pro) = 0.21 seconds
t1/2 (Case2, Chemkin-Pro) = 0.28 seconds

Below, we manually compute the half-life for comparison with Chemkin-Pro's simulation results. Using the original PLOG formulation introduced by James Miller (X. Gou, J.A. Miller, W. Sun and Y. Ju, http://engine.princeton.edu, 2011):

[...]

PR #1341, opened 2018-04-06
Fix multi pdep arrhenius calculation

mjohnson541:
This removes the requirement that PdepArrhenius objects in a MultiPdepArrhenius must have the same pressure lists by changing the getRateCoefficient function to evaluate each PdepArrhenius object independently.  Also adds a database test that checks that all library reactions can be evaluated.

### DIFF
--- a/testing/databaseTest.py
+++ b/testing/databaseTest.py
@@ -110,6 +110,12 @@ class TestDatabase():  # cannot inherit from unittest.TestCase if we want to use
             test.description = test_name
             self.compat_func_name = test_name
             yield test, library_name
+            
+            test = lambda x: self.kinetics_checkLibraryRatesCanBeEvaluated(library)
+            test_name = "Kinetics library {0}: check rates can be evaluated?".format(library_name)
+            test.description = test_name
+            self.compat_func_name = test_name
+            yield test, library_name
 
     def test_thermo(self):
         for group_name, group in self.database.thermo.groups.iteritems():
@@ -415,7 +421,14 @@ class TestDatabase():  # cannot inherit from unittest.TestCase if we want to use
                         continue
 
                     nose.tools.assert_false(speciesList[i].molecule[0].isIsomorphic(speciesList[j].molecule[0], initialMap), "Species {0} and species {1} in {2} database were found to be identical.".format(speciesList[i].label,speciesList[j].label,database.label))
-
+    
+    def kinetics_checkLibraryRatesCanBeEvaluated(self, library):
+        """
+        This test ensures that every library reaction has evaluable kinetics
+        """
+        for entry in library.entries.values():
+            entry.data.getRateCoefficient(1000.0,1.0)
+            
     def kinetics_checkReactantAndProductTemplate(self, family_name):
         """
         This test checks whether the reactant and product templates within a family are correctly defined.
--- a/rmgpy/kinetics/arrhenius.pyx
+++ b/rmgpy/kinetics/arrhenius.pyx
@@ -762,25 +762,9 @@ cdef class MultiPDepArrhenius(PDepKineticsModel):
         if P == 0:
             raise ValueError('No pressure specified to pressure-dependent MultiPDepArrhenius.getRateCoefficient().')
         
-        Plist1 = self.arrhenius[0].pressures.value_si
-        for arrh in self.arrhenius[1:]:
-            Plist2 = arrh.pressures.value_si
-            assert Plist1.shape[0] == Plist2.shape[0]
-            for i in range(Plist1.shape[0]):
-                assert 0.99 < (Plist2[i] / Plist1[i]) < 1.01            
-        
-        klow = 0.0; khigh = 0.0
+        k = 0
         for arrh in self.arrhenius:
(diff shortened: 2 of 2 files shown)


mliu49:
Are these two calculation approaches actually the same? I was looking through git history yesterday to understand why we have `MultiArrhenius` and `MultiPDepArrhenius`, and discovered that for a brief period, we actually had a `MultiKinetics` class:

1. 5821ad3d1304496be31d03922acba1a8803b6f86 (7/25/11): `MultiArrhenius` was replaced by `MultiKinetics`, which simply sums together rates in the list, allowing any type of kinetics to be used.
2. d5803cca091148fbcb84e1ec5a4b74286726d2fd (9/3/12): `MultiArrhenius` and `MultiPDepArrhenius` classes were added, with `MultiPDepArrhenius` having the fancier calculation method.
3. d386cb22b1b6594fec03a732047b1510ff38b0b2 (9/3/12): The old classes were removed.

Based on those commits, I don't think that this is an appropriate fix for this issue. At the very least, I think we need to look deeper into the difference between the calculation methods.

mjohnson541:
Suppose they were different, the point of having a MultiPdepArrhenius is to combine multiple PdepArrhenius rates together so they can be the kinetics for a single reaction object.  Why should the rate of a MultiPdepArrhenius object be different from the sum of the rates of the PdepArrhenius objects it contains?

I should also clarify that we have a test that checks that a particular MultiPdepArrhenius evaluates to particular values and both versions pass that test.

rwest:
I don't recall exactly, @mliu49, but I had enough of a deja-vu to make me search, leading me to this discussion which may be informative:
https://github.com/ReactionMechanismGenerator/RMG-Py/issues/147
https://github.com/ReactionMechanismGenerator/RMG-Py/pull/329

testing/databaseTest.py
```diff
@@ -415,7 +421,14 @@ def kinetics_checkAdjlistsNonidentical(self, database):
...
 
                     nose.tools.assert_false(speciesList[i].molecule[0].isIsomorphic(speciesList[j].molecule[0], initialMap), "Species {0} and species {1} in {2} database were found to be identical.".format(speciesList[i].label,speciesList[j].label,database.label))
-
+    
+    def kinetics_checkLibraryRatesCanBeEvaluated(self, library):
+        """
+        This test ensures that every library reaction has evaluable kinetics
+        """
+        for entry in library.entries.values():
+            entry.data.getRateCoefficient(1000.0,1.0)
```
cgrambow: Should we also add some additional checks here? For example, check that the rate is non-negative and within a reasonable range?

mjohnson541:
Do we want to proceed with this solution?

mliu49:
Yes, I think we can proceed with this. Do you want to add additional checks like @cgrambow suggested?

mjohnson541:
Yes, we can check that its nonnegative pretty easily, @cgrambow were you thinking of checking the diffusion and TST limits?

cgrambow:
Yeah, that probably makes the most sense.

alongd:
One of the reactions this reports is:
`root: ERROR: library reaction ONONO2 <=> NO2 + NO2 from library primaryNitrogenLibrary, exceeds the TST limit at 1000 K, 1 bar`
This is barrierless scission of:
image
I double-checked with the [source](https://doi.org/10.1016/B978-0-12-800345-9.00007-6) by M.C. Lin to make sure there are no typos.
Does it make sense that a high-P limit unimolecular decomposition reaction has a TST limit as the error suggests?

testing/databaseTest.py
```diff
@@ -443,14 +443,14 @@ def kinetics_checkLibraryRatesAreReasonable(self, library):
             if k < 0:
                 boo = True
                 logging.error('library reaction {0} from library {1}, had a negative rate at 1000 K, 1 bar'.format(rxn,library.label))
-            if len(rxn.reactants) == 1:
+            if len(rxn.reactants) == 1 and rxn.allow_max_rate_violation==False:
```
alongd: The test fails with `AttributeError: 'rmgpy.reaction.Reaction' object has no attribute 'allow_max_rate_violation'`, I think you should first check `hasattr`

mjohnson541: Sorry, I forgot to actually set the attribute Reaction.__init__

alongd:
@mjohnson541, I still get an error:
`TypeError: loadEntry() got an unexpected keyword argument 'allow_max_rate_violation'`
https://github.com/ReactionMechanismGenerator/RMG-database/pull/279 fails when I run the tests locally on this -Py branch

mliu49:
Any chance we could separate this from the collision limit check? It would be nice to fix the rate evaluation.

mliu49:
I went ahead and rebased this. Any final considerations before merging?

mjohnson541:
Together with the database maxviol branch there shouldn’t be any issues,
might be some issues merging them individually.  You could merge the unit
test separately and merge the rest of the RMG-Py branch first then the
database branch and then the unit test.  I think that would avoid unit
tests failing.

Best,
Matt

PR #1343, opened 2018-04-13
Fix Seed React Behavior

mjohnson541:
So apparently for quite a while seed mechanisms have been fully reacted after entering the model due to a major bug.  The primary problem was that filterReactions is default false and it wasn't being set until after the first enlargement step.  I fixed this, adjusted the algorithm so that we always run through the filter checks on the first iteration and prevented the algorithm from terminating when new reactions have been generated.


mliu49:
It seems like the reaction filter flags are first set during the initial enlarge step prior to entering the model settings for loop. I understand the first commit, but could you explain the reasoning for the second two a bit more?

mjohnson541:
So suppose you seed it in such a way that the seed includes all species the input file species can react to form, which may not be all that uncommon.  This makes the edge empty until the filter simulation/react enlargement takes place.  Since the edge is empty the simulation runs to completion with no returned objects, which usually sets it up to terminate the job.  Furthermore, since the model size hasn't changed the algorithm doesn't think it needs to update the react flags and do react enlargement (because it assumes this was done the iteration before (filter simulations simulate the species that will be simulated next)).  The second commit forces it to do the filter/react which brings new species into the edge and the third commit prevents it from terminating before running a simulation with these new species in the edge.

mjohnson541:
This new commit adds an allowInitialSpeciesReact option that is default True.  If one wants to restart mechanism generation with a core and edge seed one can set this to False to avoid reacting the original species.  It should be noted that this means the initial species can't be reacted with each other later (as they might have if they violated filter tolerances) (they can react with non-original core species though).

mjohnson541:
So I was just thinking, if we tracked the direction each reaction was generated in we could probably infer nearly everything that has been reacted from the core and edge seed.  It wouldn't even involve any isomorphism checks to figure it out.  That might be a better option than the allowInitialSpeciesReact option.

mjohnson541:
The issue is that when I set allowInitialSpeciesReact=False I'm effectively telling RMG that all the core species have been reacted.  However, if filterReactions was used likely all of them haven't really been reacted and you could potentially miss reactions.  If I add a flag indicating the direction each reaction was generated from and use that to infer what has been reacted we don't run into this problem and I think we only end up re-reacting things that don't produce any new reactions.  Which I think is probably small?

mjohnson541:
For now I've removed the fourth commit.  I figure we can deal with seed restarting in a separate PR.  Is this good?

rmgpy/rmg/main.py
```diff
@@ -540,9 +544,10 @@ def execute(self, **kwargs):
         self.Tmax = max([x.T for x in self.reactionSystems]).value_si
         
         # Initiate first reaction discovery step after adding all core species
-        if self.modelSettingsList[0].filterReactions:
+        if self.filterReactions:
             # Run the reaction system to update threshold and react flags
             for index, reactionSystem in enumerate(self.reactionSystems):
+
```
mliu49: Could you remove this empty line?

rmgpy/rmg/main.py
```diff
@@ -710,9 +715,15 @@ def execute(self, **kwargs):
...
                                                               reactionSystems=self.reactionSystems)
         
+                        
+                        oldEdgeSize = len(self.reactionModel.edge.reactions)
+                        oldCoreSize = len(self.reactionModel.core.reactions)
                         self.reactionModel.enlarge(reactEdge=True, 
-                                unimolecularReact=self.unimolecularReact, 
-                                bimolecularReact=self.bimolecularReact)
+                                    unimolecularReact=self.unimolecularReact, 
+                                    bimolecularReact=self.bimolecularReact)
```
mliu49: Could you undo these indentation changes and remove the empty line 718?

rmgpy/rmg/main.py
```diff
@@ -669,8 +674,8 @@ def execute(self, **kwargs):
                     # Add objects to enlarge to the core first
                     for objectToEnlarge in objectsToEnlarge:
                         self.reactionModel.enlarge(objectToEnlarge)
-                        
-                    if len(self.reactionModel.core.species) > numCoreSpecies:
+
```
mliu49: Could you remove this indentation change?

PR #1348, opened 2018-04-21
Adding resonance transitons for lone pairs

alongd:
This PR adds new resonance transitions and introduces the resonance filtration function based on heuristics.

### lone_pair_multiple_bond_resonance_structures
Resonance structures formed by lone electron pair - multiple bond shifts.
Example: image image image

### lone_pair_radical_multiple_bond_resonance_structures
Resonance structures formed by lone electron pair - radical - multiple bond shifts.
Example: image image image

### N5dc_resonance_structures
Resonance structures formed by radical and lone pair shifts mediated by an N5dc atom (not to be confused with `N5ddc_N5tc_resonance_structures`).
Example: image image image
(in the above example, the two `O` atoms are "different" - this addresses issue #1293)

### Other resonance modifications
* `adjacent_resonance_structures` was renamed `ally_delocalization_resonance_structures` and elaborated for heteroatoms
* `N5dd_N5ts_resonance_structures` was renamed `N5ddc_N5tc_resonance_structures` and elaborated for sulfur
* `lone_pair_radical_resonance_structures` was elaborated for heteroatoms
* Added respective resonance and pathfinder tests

## Resonance filtration
This module contains methods for filtering a list of Molecules representing a single Species,
keeping only the representative structures. 
The rules this module follows are (by order of importance):
1. Minimum overall deviation from the Octet Rule (also elaborated for Dectet and Duodectet for hypervalance heteroatoms)
2. Additional charge separation is only important for radical species, and is kept only if it makes a new radical site in the species
3. If a structure must have charge separation, negative charges will be assigned to more electronegative atoms, and vice versa
4. Opposite charges will be as close as possible to one another, and vice versa

## Other modifications introduced by this PR
* Added the `reactive` attribute to the Molecule class. If a non-representative molecule (one that would have been filtered out if formed by `generate_resonance_structures()`) is directly formed during the `enlarge` step, the non-representative Molecule is marked an `reactive=False`, and the respresentative structures are appended to the Molecule list. If, however, this species already exists (and doesn't contain the non-representative structure), the species are merged.
* Added various tests
* Added an electronegativity dictionary to the PeriodicSystem class
* An iPython notebook for resonance generation was added

PR #1349, opened 2018-04-23
Changing the level of aug-InChI (2 --> 1)

alongd:
Augmented InChI in RMG is currently being used in QM modules to name scratch files. It has also been suggested to be used in parallelization, but isn't implemented to the best of my knowledge.

Since augmented InChI Level 2 (specifying both _u_ and _p_ layer) is time-consuming to maintain (see #1348), this PR suggests to use augmented InChI Level 1 (only specifying _multiplicity_) instead.

### DIFF
--- a/rmgpy/molecule/molecule.py
+++ b/rmgpy/molecule/molecule.py
@@ -1429,7 +1429,7 @@ class Molecule(Graph):
         
         Separate layer with a forward slash character.
         """
-        return translator.toInChI(self, aug_level=2)
+        return translator.toInChI(self, aug_level=1)
         
     
     def toInChIKey(self):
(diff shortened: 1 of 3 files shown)


alongd:
@KEHANG, is aug-InChI being used in the Central Thermo Database?

KEHANG:
Yes, thermo central database uses aug-inchi. I think it's fine to change the default aug_level, but we should still allow level-2. To do that, we can add an new argument `aug_level` to molecule's `toAugmentedInChI()` and also species's `generate_aug_inchi()`.

alongd:
Thanks for the input, @KEHANG. I added these attributes.
I still need to check that we can regenerate the molecule from the level 1 aug-InChI.

PR #1350, opened 2018-04-24
Fix cantherm examples

mjohnson541:
This PR fixes two issues @cgrambow and I identified in the Cantherm examples and I've set them to the values @cgrambow thinks they should be.  The symmetry of the methyl rotors in the Toulene example have been fixed from 6 to 3 and the number of opticalIsomers in the 23dimethylpropoxy example have been changed from 1 to 2 to reflect the presence of a chiral carbon.

### DIFF
--- a/examples/cantherm/reactions/23dimethylpropoxy/dimetpropoxy.py
+++ b/examples/cantherm/reactions/23dimethylpropoxy/dimetpropoxy.py
@@ -8,7 +8,7 @@ externalSymmetry = 1
 
 spinMultiplicity = 2
 
-opticalIsomers = 1
+opticalIsomers = 2
 
 energy = {
     'M08SO': QchemLog('dimetpropoxy.out'),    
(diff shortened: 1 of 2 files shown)


mliu49:
Why is the methyl rotor symmetry 3 and not 6?

cgrambow:
You can think of it in two different ways: Either you plot the energy profile of the methyl rotation or you think of it geometrically. The energy profile has three identical minima, so symmetry is 3. Rotating the methyl by 120° results in a completely superimposable structure. 360°/120° is also 3.

mliu49:
The energy profile has 6 minima though, and every 60° rotation results in a superimposable structure...

image

cgrambow:
Oh, good point. I was just thinking of methyl rotors in general. In the toluene case, there is additional symmetry due to the plane of symmetry down the middle, so I guess 6 is more appropriate.

Issue #1345, opened 2018-04-18
RMG cannot finds reactions whose inverse reactions generating CO


xiaoruiDong:
Recently, I am dealing with the following reaction 

![image](https://user-images.githubusercontent.com/35771233/38956950-25a992f4-4327-11e8-87da-8087f9541a9d.png)

and no kinetics data has been found for this reaction. However its reverse reaction, belonging to the family 'R_Addition_COm', could be found. Other than this specific reaction and family, I also found that if the reaction does not match a library or training reaction, the reverse reaction from family 'R_Addition_COm', 'R_Addition_CSm', '1,2_Insertion_CO', '2+2 cycloaddition_CO', 'CO_Disproportionation' will not be found or matched to kinetics data. 

According to my discussion with Max and Alon, this may relates to the generation of reverse templates, but still need to be inspected.

mliu49:
From our debugging this morning, I can pretty confidently say that it's due to generation of product templates, at least in the case of `R_Addition_COm`.

The two reactant templates are:
```
# COm
1 *1 C2tc u0 p1 c-1 {2,T}
2 *3 O4tc u0 p1 c+1 {1,T}

# Y_rad
1 *2 R u1
```

And the automatically generated product template is:
```
# YC.=O
1 *1 [Cdc,Cd,CS,CO] u1 p0 c-1 {2,D} {3,S}
2 *3 [O2s,O2sc]     u0 p2 c+1 {1,D}
3 *2 R              u0 {1,S}
```

During product template generation, we apply the recipe to the reactant templates but don't do any updating of the resulting Group (see family.py line 1300). Thus, all attributes are unchanged unless specified by the reaction recipe.

I think there are two potential solutions: 1) we add charge modification to reaction recipes, or 2) we add a Group.update() method that can somehow identify how charges should change.

alongd:
Could tell RMG to use a product template, if available, otherwise generate it?

mliu49:
Addressed by #1356 and ReactionMechanismGenerator/RMG-database#268.

PR #1351, opened 2018-04-25
Added GAIN_CHARGE and LOSE_CHARGE to recipe actions

xiaoruiDong:
This request solves the issue #1345. Generally, it allows reactions whose products contain [C-]#[O+] to be generated. The main change is adding two recipe action: 'GAIN_CHARGE' and 'LOSE_CHARGE' (corresponding to method molecule.group.GroupAtom.__gain_charge and molecule.group.GroupAtom.__lose_charge) so that when forming templates, the charge of C and O will be removed. Other minor changes are also involved, but just to let RMG recognize and perform these actions. The new action currently have no impact on atomtype, however it could be changed if needed in the future.


rmgpy/data/kinetics/database.py
```diff
@@ -458,7 +458,6 @@ def generate_reactions_from_families(self, reactants, products=None, only_famili
         # Delete reactions which we could not find a reverse reaction for
         for i in reversed(to_delete):
             del reaction_list[i]
-
```
alongd: Could you remove this line deletion from this commit? Since this PR has no more changes to databse.py, it will create an unnecessary change to the history.

rmgpy/data/kinetics/family.py
```diff
@@ -1173,6 +1197,9 @@ def applyRecipe(self, reactantStructures, forward=True, unique=True):
         # because the two reactants are identical, they have the same tags
         # In this case, we must change the labels from '*' and '*' to '*1' and
         # '*2'
+
+
+
```
alongd: Why were these empty lines added?

rmgpy/molecule/atomtype.py
```diff
@@ -574,7 +574,7 @@ def getFeatures(self):
...
 atomTypes['O4sc'].setActions(incrementBond=['O4dc'],         decrementBond=[],               formBond=['O4sc'],      breakBond=['O4sc'],      incrementRadical=['O4sc'], decrementRadical=['O4sc'], incrementLonePair=['O2s','O2sc'],decrementLonePair=[])
 atomTypes['O4dc'].setActions(incrementBond=['O4tc'],         decrementBond=['O4sc'],         formBond=['O4dc'],      breakBond=['O4dc'],      incrementRadical=['O4dc'], decrementRadical=['O4dc'], incrementLonePair=['O2d'], decrementLonePair=[])
-atomTypes['O4tc'].setActions(incrementBond=[],               decrementBond=['O4dc'],         formBond=[],            breakBond=[],            incrementRadical=[],       decrementRadical=[],       incrementLonePair=['O2d'], decrementLonePair=[])
+atomTypes['O4tc'].setActions(incrementBond=[],               decrementBond=['O4dc'],         formBond=[],            breakBond=[],            incrementRadical=[],       decrementRadical=[],       incrementLonePair=[], decrementLonePair=[])
```
alongd: Thanks for correcting this! Could you add a short description to the commit message, so it is clear to everyone looking at this change in the future that we first decrement the bond of `O4tc` and only then loose its lone pair to correctly transition to `O2d`? It would even be better to make a separate commit just for this change.

rmgpy/molecule/group.py
```diff
@@ -277,6 +277,46 @@ def __losePair(self, pair):
...
+        chargeNum = []
+
+        if not self.charge:
+            raise ActionError('Unable to update GroupAtom due to GAIN_CHARGE action: Unknown charge status for GroupAtom "{0}".'.format(self.atomType))
+        else:
+            for x in self.charge:
+                chargeNum.append(x + charge)
+            self.charge = chargeNum
+
+        # Have not update __gainCharge to the molecule.atomtype
```
alongd: I think we should add the gainCharge and looseCharge actions to atomType.py where appropriate, even if in most (all?) cases the atomType remains the same. We could discuss this in person if you'd like.

alongd:
Replaced by #1356

PR #1355, opened 2018-05-02
Fix library labels when loading all libraries

amarkpayne:
Previously the first character in the library name was being cut-off


amarkpayne:
I was loading all of the libraries in the RMG-database when I came across this issue. I think the person who wrote this forgot that Python zero indexes, while `len(path)` starts from 1. Attached is an ipython notebook that should be helpful for reviewing this PR (change the file extension to .ipynb).


[Library_Labels_PR_Review.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/1965786/Library_Labels_PR_Review.txt)

amarkpayne:
This is only an issue if you input the path manually and include the final slash for the last directory (i.e. `path/to/folder/` as opposed to `path/to/folder`. If you use `os.path` this is not an issue.

PR #1356, opened 2018-05-03
Newest version of debug_CO

xiaoruiDong:
This pull request is an alternative for #1351. In this pull request, changes are fewer than the last pull request, where only method updateCharge() to the class Group and its relative call have been added. In general, this will make more sense as mentioned by Alon that once we know the valence electron number, bond number, lone pairs and radical electron numbers, we are able to update the charge. The concept of the method updateCharge() just follows this idea, and it will not been operated when lacking of information (for example lone pair condition is not given) or multiple information (for example, multiple charges, multiple lone pairs conditions) are assigned to the group.

Now, this method have only been applied for families where charged species involved, as others may not run into incorrect generation templates issue, and I have checked that it worked well for all five families. Besides, it also passes the test of RMG-Py and RMG-database. 

This new updateCharge() could also be applied for other purpose in the future, but will have an issue when the bonding information is incomplete. 
For example: 
Group with AdjacencyList:
1 Cd u0 p0 c0 {2,D}
2 Cd u0 p0 c0 {1,D}
which is used in unittest: MoleculeTest/testSubgraphIsomorphism. This will result in incorrect charge update (updating to +2) as bonding information is incomplete. So far, I have nothing to do with that.


rmgpy/molecule/group.py
```diff
@@ -1021,6 +1021,33 @@ def update(self):
...
+        """
+        Update the charge according to the valence_electron, total bond order, lone pairs
+        and radical electrons. When charged molecule is appearing in reactants, the product
+        template generated by applying recipe will result in wrong charge status. This
+        function is aimed to solve this problem
+        """
+        for atom in self.atoms:
+            if (len(atom.charge) == 1) & (len(atom.lonePairs) == 1) & (len(atom.radicalElectrons) == 1):
+                # if the charge of the group is not labeled, then no charge update will be
+                # performed. If there is a list of charge, no update either. Besides, this groupatom have to
```
alongd: 'If there is a list of charge, no update either.' - Do you mean a list of more than one entry?

rmgpy/molecule/group.py
```diff
@@ -1021,6 +1021,33 @@ def update(self):
...
+                # have enough information to be updated
+                atomtype = atom.atomType[0]
+                for element in allElements:
+                    if atomtype is atomTypes[element] or atomtype in atomTypes[element].specific:
+                        elementlabel = element
+                        bondOrder, lonePairs, radicalElectrons = 0, 0, 0
+                        import rmgpy.molecule.element as elements
```
alongd: I think that imports should generally be done either at the head of the file, or the start of the method.
Although it doesn't seem like the case here, an import in a for loop could get called several times needlessly.

rmgpy/molecule/group.py
```diff
@@ -1021,6 +1021,33 @@ def update(self):
...
+                        valence_electron = elements.PeriodicSystem.valence_electrons[elementlabel]
+                        for connectAtom, bond in atom.bonds.iteritems():
```
alongd: If your not using one of the iterators in the loop (here `connectAtom`), it is a convention to name it with an underscore (`_`) so future programmers immediately understand it is not used on purpose, and won't think something is missing.

rmgpy/molecule/group.py
```diff
@@ -1021,6 +1021,33 @@ def update(self):
...
+        for atom in self.atoms:
+            if (len(atom.charge) == 1) & (len(atom.lonePairs) == 1) & (len(atom.radicalElectrons) == 1):
+                # if the charge of the group is not labeled, then no charge update will be
+                # performed. If there is a list of charge, no update either. Besides, this groupatom have to
+                # have enough information to be updated
+                atomtype = atom.atomType[0]
+                for element in allElements:
+                    if atomtype is atomTypes[element] or atomtype in atomTypes[element].specific:
+                        elementlabel = element
+                        bondOrder, lonePairs, radicalElectrons = 0, 0, 0
```
alongd: Is there a reason to set `lonePairs` and `radicalElectrons` to zero here?

rmgpy/data/kinetics/family.py
```diff
@@ -1300,6 +1300,9 @@ def applyRecipe(self, reactantStructures, forward=True, unique=True):
         for struct in productStructures:
```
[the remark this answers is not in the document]
xiaoruiDong: Comments are added for this line

rmgpy/data/kinetics/family.py
```diff
@@ -1300,6 +1300,9 @@ def applyRecipe(self, reactantStructures, forward=True, unique=True):
...
             if isinstance(struct, Molecule):
                 struct.update()
+            if isinstance(struct, Group) & (label in ['1,2_insertion_co', 'r_addition_com', 'co_disproportionation',
+                                                      'intra_no2_ono_conversion', 'lone_electron_pair_bond']):
```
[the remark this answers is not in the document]
xiaoruiDong: intra no2 ono conversion and lone electron pair bond  are already added in this PR

rmgpy/molecule/group.py
xiaoruiDong: Yes... sorry for unclear word.

rmgpy/molecule/group.py
xiaoruiDong: import has been moved to the head of the file

rmgpy/molecule/group.py
xiaoruiDong: I think I just want to initialize these variables, which seems like useless.

xiaoruiDong:
I will try to add unittests. All the things you mentioned have been updated in my local branch and will be pull once tests are added. Thank you for important advices for my coding.

xiaoruiDong:
@alongd Hi Alon, the unittest has been added, named test_r_addition_com. The purpose for this test is to see whether we can generate correct product templates from charged species. The family used in this test could be alternated to other similar families.

rmgpy/molecule/group.py
```diff
@@ -1021,6 +1023,32 @@ def update(self):
         self.updateConnectivityValues()
         self.updateFingerprint()
 
+    def updateCharge(self):
```
alongd: Please use lowercase for function names

rmgpy/molecule/group.py
```diff
@@ -1021,6 +1023,32 @@ def update(self):
...
+        """
+        Update the charge according to the valence_electron, total bond order, lone pairs
+        and radical electrons. When charged molecule is appearing in reactants, the product
+        template generated by applying recipe will result in wrong charge status. This
+        function is aimed to solve this problem
```
alongd: I suggest to modify the wording a bit, e.g.: `Update partial charges according to the valence_electron, total bond order, lone pairs and radical electrons. This method is used for products of specific families with recipes that modify charges.`

rmgpy/molecule/group.py
```diff
@@ -1021,6 +1023,32 @@ def update(self):
...
+        """
+        for atom in self.atoms:
+            if (len(atom.charge) == 1) & (len(atom.lonePairs) == 1) & (len(atom.radicalElectrons) == 1):
```
alongd: `and` should be used instead of `&`

rmgpy/molecule/group.py
```diff
@@ -1021,6 +1023,32 @@ def update(self):
...
+                # if the charge of the group is not labeled, then no charge update will be
```
alongd: Where in this method is the atom labelling being checked?

rmgpy/molecule/group.py
```diff
@@ -1021,6 +1023,32 @@ def update(self):
...
+                # performed. If there multiple charges are assigned, no update either.
+                # Besides, this groupatom should have enough information to be updated
+                atomtype = atom.atomType[0]
+                for element in allElements:
+                    if atomtype is atomTypes[element] or atomtype in atomTypes[element].specific:
+                        elementlabel = element
```
alongd: What does this line do? You could continue using `element`

rmgpy/molecule/group.py
```diff
@@ -1021,6 +1023,32 @@ def update(self):
...
+                        bondOrder = 0
```
alongd: Please use lowercase names.
Same comment for other variables here

rmgpy/test_data/testing_database/kinetics/families/R_Addition_COm/NIST/reactions.py
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env python
+# encoding: utf-8
```
alongd: I think we don't have to add the NIST folder for the tests

alongd:
The temporary `.travis.yml` and `deploy.sh` changes should be in a separate commit, so we can loose them before merging this in (currently they are in `add updateCharge method to Group, to genereate templates with correct… `). Please correct

alongd:
Looking good! Please rebase this as well, and remove the temporary commit.

rmgpy/data/kinetics/family.py
```diff
@@ -1297,11 +1297,16 @@ def applyRecipe(self, reactantStructures, forward=True, unique=True):
...
 
         # If product structures are Molecule objects, update their atom types
+        # If product structures are Group objects and the reaction is in certain families
+        # (families with charged substances), the charge of structures will be updated
         for struct in productStructures:
             if isinstance(struct, Molecule):
                 struct.update()
             else:
                 struct.resetRingMembership()
+            if isinstance(struct, Group) and (label in ['1,2_insertion_co', 'r_addition_com', 'co_disproportionation',
```
mliu49: This could be lumped into the else above. You could even change it to `elif isinstance(struct, Group)` to eliminate the type assumption. E.g.:
```
    struct.update()
elif isinstance(struct, Group):
    struct.resetRingMembership()
    if label in [...]:
        struct.update_charge()
else:
    raise TypeError('Expecting Molecule or Group object, not {0}'.format(struct.__class__.__name__))
```

alongd:
Please also rebase

xiaoruiDong:
Yes, this has already been rebased with origin/master. @alongd

xiaoruiDong:
Sorry, let me correct it.

xiaoruiDong:
Max and Alon, I think now it is truly rebased. @alongd , @mliu49

PR #1358, opened 2018-05-06
"Edgeless"-RMG Part 1:  Moving Global Forbidden Structure Checking to Core Entrance

mjohnson541:
This moves global forbidden structure checking from when species are created in the edge to when a species attempts to enter the core.  In a test constructing a dodecane combustion model this reduced time spent checking forbidden structures by 148x and overall runtime (in the react regime) by 4.5x without impacting the output model.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1509,17 +1509,12 @@ class KineticsFamily(Database):
         Return ``True`` if the molecule is forbidden in this family, or
         ``False`` otherwise. 
         """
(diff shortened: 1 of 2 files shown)


rwest:
Interesting. Does the edge end up much bigger, and if so when is that a problem?

mjohnson541:
I tested it a while ago, but if I remember correctly the edge size increase was ~10% or less.  However, one of the future goals of the project this PR starts is to actually get rid of edge species altogether.

mjohnson541:
Break this apart into individual PRs

PR #1361, opened 2018-05-10
Check core reactions for collision rate limit violators

alongd:
This collision rate limit check is inspired by [Wang et at. 2017](https://doi.org/10.1016/j.combustflame.2017.08.005).
Currently we only output a warning along with a relevant log file if any core reactions violate this limit.

### DIFF
--- a/rmgpy/reaction.py
+++ b/rmgpy/reaction.py
@@ -1146,6 +1146,106 @@ class Reaction:
         except AttributeError:
             pass
 
+    def check_collision_limit_violation(self, t_min, t_max, p_min, p_max):
+        """
+        Warn if a core reaction violates the collision limit rate in either the forward or reverse direction
+        at the relevant extreme T/P conditions. Assuming a monotonic behaviour of the kinetics.
+        Returns a list with the reaction object and the direction in which the violation was detected.
+        """
+        conditions = [[t_min, p_min]]
+        if t_min != t_max:
+            conditions.append([t_max, p_min])
+        if self.kinetics.isPressureDependent() and p_max != p_min:
+            conditions.append([t_min, p_max])
+            if t_min != t_max:
+                conditions.append([t_max, p_max])
+        logging.debug("Checking whether reaction {0} violates the collision rate limit...".format(self))
+        violator_list = []
+        kf_list = []
+        kr_list = []
+        collision_limit_f = []
+        collision_limit_r = []
+        for condition in conditions:
+            if len(self.reactants) >= 2:
+                try:
+                    collision_limit_f.append(self.calculate_coll_limit(temp=condition[0], reverse=False))
+                except ValueError:
+                    continue
+                else:
+                    kf_list.append(self.getRateCoefficient(condition[0], condition[1]))
+            if len(self.products) >= 2:
+                try:
+                    collision_limit_r.append(self.calculate_coll_limit(temp=condition[0], reverse=True))
+                except ValueError:
+                    continue
+                else:
+                    kr_list.append(self.generateReverseRateCoefficient().getRateCoefficient(condition[0], condition[1]))
+        if len(self.reactants) >= 2:
+            for i, k in enumerate(kf_list):
+                if k > collision_limit_f[i]:
+                    ratio = k / collision_limit_f[i]
+                    condition = '{0} K, {1:.1f} bar'.format(conditions[i][0], conditions[i][1]/1e5)
+                    violator_list.append([self, 'forward', ratio, condition])
+        if len(self.products) >= 2:
+            for i, k in enumerate(kr_list):
+                if k > collision_limit_r[i]:
+                    ratio = k / collision_limit_r[i]
+                    condition = '{0} K, {1:.1f} bar'.format(conditions[i][0], conditions[i][1]/1e5)
+                    violator_list.append([self, 'reverse', ratio, condition])
+        return violator_list
+
+    def calculate_coll_limit(self, temp, reverse=False):
+        """
+        Calculate the collision limit rate for the given temperature
+        implemented as recommended in Wang et al. doi 10.1016/j.combustflame.2017.08.005 (Eq. 1)
+        """
+        reduced_mass = self.get_reduced_mass(reverse)
+        sigma, epsilon = self.get_mean_sigma_and_epsilon(reverse)
+        Tr = temp * constants.kB * constants.Na / epsilon
+        reduced_coll_integral = 1.16145 * Tr ** (-0.14874) + 0.52487 * math.exp(-0.7732 * Tr) + 2.16178 * math.exp(
+            -2.437887 * Tr)
+        k_coll = (math.sqrt(8 * math.pi * constants.kB * temp * constants.Na / reduced_mass) * sigma ** 2
+                  * reduced_coll_integral * constants.Na)
+        return k_coll
+
+    def get_reduced_mass(self, reverse=False):
+        """
+        Returns the reduced mass of the reactants if reverse is ``False``
+        Returns the reduced mass of the products if reverse is ``True``
+        """
+        if reverse:
+            mass_list = [spc.molecule[0].getMolecularWeight() for spc in self.products]
+        else:
+            mass_list = [spc.molecule[0].getMolecularWeight() for spc in self.reactants]
+        reduced_mass = reduce((lambda x, y: x * y), mass_list) / sum(mass_list)
+        return reduced_mass
+
+    def get_mean_sigma_and_epsilon(self, reverse=False):
+        """
+        Calculates the collision diameter (sigma) using an arithmetic mean
+        Calculates the well depth (epsilon) using a geometric mean
+        If reverse is ``False`` the above is calculated for the reactants, otherwise for the products
+        """
+        sigmas = []
+        epsilons = []
+        if reverse:
+            for spc in self.products:
+                trans = spc.getTransportData()
+                sigmas.append(trans.sigma.value_si)
+                epsilons.append(trans.epsilon.value_si)
+            num_of_spcs = len(self.products)
+        else:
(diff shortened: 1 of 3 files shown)


alongd:
@amarkpayne, could you take a look at this PR?

alongd:
@amarkpayne, I just rebased (conflicts were solved)

amarkpayne:
Thanks! I am looking over the code now

rmgpy/rmg/main.py
```diff
@@ -898,6 +902,27 @@ def execute(self, **kwargs):
                 
                 plot_sensitivity(self.outputDirectory, index, reactionSystem.sensitiveSpecies)
 
+        # Check all core reactions (in both directions) for collision limit violation
+        logging.info("Checking core reactions for collision rate limit violators...")
+        violators = []
+        for rxn in self.reactionModel.core.reactions:
+            f_violator, r_violator = rxn.check_collision_limit_violation(t_min=self.Tmin, t_max=self.Tmax,
+                                                            p_min=self.Pmin, p_max=self.Pmax)
+            if f_violator is not None:
```
amarkpayne: I think this statement can be simplified to `if f_violator:`

rmgpy/rmg/main.py
```diff
@@ -898,6 +902,27 @@ def execute(self, **kwargs):
...
+                violators.append(f_violator)
+            if r_violator is not None:
+                violators.append(r_violator)
+        if len(violators):
```
amarkpayne: This can be replaced with `if violators:` since empty lists have evaluate to False. This is not a big deal though

rmgpy/rmg/main.py
```diff
@@ -898,6 +902,27 @@ def execute(self, **kwargs):
...
+            logging.info("\n")
+            logging.warning("{0} CORE reactions violate the collision rate limit!"
+                            " See the 'collision_rate_violators.log' for details.\n\n".format(len(violators)))
+            with open('collision_rate_violators.log', 'w') as violators_f:
```
amarkpayne: Do we need to check if a file with this name already exists? Otherwise, this file will get overwritten if the user restarts the job (that might be okay for this file, I am just checking because I know that we do this for the RMG log file)

rmgpy/rmg/main.py
cgrambow: `if x` and `if x is not None` are used in different contexts. Check the PEP 8 recommendations [here](https://legacy.python.org/dev/peps/pep-0008/#programming-recommendations).

rmgpy/reaction.py
```diff
@@ -1144,6 +1144,121 @@ def ensure_species(self, reactant_resonance=False, product_resonance=True):
...
+        f_violator = None
+        r_violator = None
+        kf_list = []
+        kr_list = []
+        if len(self.reactants) == 2:
+            collision_limit = []
+            for condition in conditions:
+                try:
+                    collision_limit.append(self.calculate_coll_limit(temp=condition[0], reverse=False))
+                except ValueError:
```
amarkpayne: Under what circumstances does this try statement throw a ValueError? Is it something that is okay to ignore?

rmgpy/reaction.py
```diff
@@ -1144,6 +1144,121 @@ def ensure_species(self, reactant_resonance=False, product_resonance=True):
...
+        If reverse is ``False`` the above is calculated for the reactants, otherwise for the products
+        """
+        sigmas = []
+        epsilons = []
+        if reverse:
+            for spc in self.products:
+                trans = spc.getTransportData()
+                sigmas.append(trans.sigma.value_si)
+                epsilons.append(trans.epsilon.value_si)
+                num_of_spcs = len(self.products)
```
amarkpayne: Move this out of the for loop. Otherwise, `num_of_spcs` is set to the same value over and over again

rmgpy/reaction.py
```diff
@@ -1144,6 +1144,121 @@ def ensure_species(self, reactant_resonance=False, product_resonance=True):
...
+        else:
+            for spc in self.reactants:
+                trans = spc.getTransportData()
+                sigmas.append(trans.sigma.value_si)
+                epsilons.append(trans.epsilon.value_si)
+                num_of_spcs = len(self.reactants)
+        if any([x == 0 for x in sigmas + epsilons]):
+            raise ValueError
+        mean_sigmas = sum(sigmas) / num_of_spcs
+        mean_epsilons = reduce((lambda x, y: x * y), epsilons) ** (1 / sum(epsilons))
```
amarkpayne: For the geometric mean for the well depths, should the product be raised to `(1/len(epsilons))` ? I thought the geometric mean is raised to the inverse power of the number of elements, not their sum.

rmgpy/rmg/main.py
alongd: Thanks @cgrambow, it was helpful reading this piece. I'll leave the `is not None` statement as as

rmgpy/rmg/main.py
alongd: I think that since the model itself (chemkin/cantera formats) is being overwritten, and this output directly relates to the generated model, it's ok the overwrite the collision violators log too. A related question, though, is what if the user re-runs in the same folder, and only the first run generates the collision violators log file... I guess I should delete it if it exists and no violators were found?

rmgpy/reaction.py
alongd: When I tested this branch by inventing ridiculous rates, I got a `ValueError` from `calculate_coll_limit()`. It was a while ago, so I don't remember exactly what calculation caused it to crash (perhaps either `reduced_coll_integral` or `k_coll`?).
Since we decided that at the fist stage we only want to report rate violators, but not crash RMG, I decided to capture errors here and `pass` on them. This is just to be on the safe side and not crushing a job near its end for a reason we're not certain of.

rmgpy/reaction.py
alongd: done, thanks!

rmgpy/reaction.py
alongd: You're right, thanks for catching my mistake!

alongd:
@amarkpayne, thanks for your suggestions!
I addressed the comments in fixup commits.

I also relocated the collision rate violation checks to the `check_model` method.
Last, I slightly changed the order of imports in rmg/main.py according to [PEP 8 recommendations](https://www.python.org/dev/peps/pep-0008/?#imports). Slight modifications to `os` and `np` imports as well.

Let me know if you have additional comments, or whether we're good to squash.

alongd:
@amarkpayne, I simplified the code a bit, and added more relevant info to the collision violation log file

rmgpy/reaction.py
```diff
@@ -1144,6 +1144,99 @@ def ensure_species(self, reactant_resonance=False, product_resonance=True):
...
+        if t_min != t_max:
+            conditions.append([t_max, p_min])
+        if self.kinetics.isPressureDependent() and p_max != p_min:
+            conditions.append([t_min, p_max])
+            if t_min != t_max:
+                conditions.append([t_max, p_max])
+        logging.debug("Checking whether reaction {0} violates the collision rate limit...".format(self))
+        violator = None
+        kf_list = []
+        kr_list = []
```
amarkpayne: With the newest simplification, I don't thing kr_list is being used (are you missing the code that handles checking the reverse reaction? I like the direction of this code simplification, but I think it is only checking the forward direction now.

alongd: Can you take a look at line 1180 in this file? I think `kr_list` is used as expected

https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1361/files/fe94464ff3ad69d8b4c12e9950cc39e1dafede75#diff-eb892c0ffd531a3aa16ba76563c6ccc6R1180

alongd:
I guess we shouldn't merge this until the 2.2.0 release is out, right?

rmgpy/reaction.py
amarkpayne: Does anything append on to `kr_list` before line 1180 though? Maybe I am missing something

alongd: You're right. I'll fix it.

amarkpayne:
I think that was the plan for this PR, but I'll double check

rmgpy/rmg/main.py
```diff
@@ -945,6 +948,34 @@ def check_model(self):
...
+                violators_f.write('*** Collision rate limit violators report ***\n'
+                                  'The "violation factor" is the ratio of the rate coefficient to the collision limit'
+                                  ' rate at the relevant conditions\n\n')
+                for violator in violators:
+                    violators_f.write('{0}\n{1}\nDirection: {2}\nViolation factor: {3:.2f}\n'
+                                      'Violation condition: {4}\n\n'.format(
+                        violator[0], violator[1], violator[2], violator[3], violator[4]))
+        else:
+            if os.path.isfile('collision_rate_violators.log'):
```
mliu49: I think you could replicate the behavior of RMG.log and rename the existing log regardless of whether or not violators were found, i.e. move this check before `if violators:`. It might also be better to put OLD at the end of the filename so that it appears right next to the new file in the file list.

mliu49:
To confirm my understanding, this check is run once at the end of the RMG run, and only prints warnings and saves the log file of violations?

I think it could be included in v2.2 if it's ready.

alongd:
@mliu49, your understanding is correct.
From testing this out a bit, I expect we would start seeing many collision rate violators...
A simple ethane oxidation run at 1500-2000 K gave ~20 violators. Perhaps we shouldn't include it in the current binaries release, and just let the developers work with it and be aware of the output to improve the database towards the next release?

amarkpayne:
This PR will be merged after the next version release

alongd:
@amarkpayne, this branch is rebased and ready to merge once all tests pass

PR #1251, opened 2017-12-13
Convolutional Neural Network Based Thermo Estimator

KEHANG:
## Introduction

Ever since the very first thermo-estimator **Group Additivity** in RMG, we've been trying to improve it . The whole improvement process usually consists of three steps: 
- s1. identify which molecule might be badly estimated, 
- s2. conduct quantum calculations and 
- s3. figure out what new groups to add

The process is very time-consuming and requires good chemical intuition to do s1 and s3. This PR introduces a new way of estimating thermochemistry, which doesn't require s3 and has the potential  of telling whether a molecule is well estimated by giving uncertainty which might eliminate the need of human's doing s1 as well (will be included in the future PRs).

What **Convolutional Neural Network** does is to take a molecule and figure out a fingerprint from training data which consists of important features for thermochemistry prediction. For technical details, one can read [this paper](https://arxiv.org/abs/1509.09292) and [my post](http://kehang.github.io/basic_project/2017/04/18/machine-learning-in-molecular-property-prediction/).

## How it performs

A typical trained model with `polycyclic_2954_table_train` (a table hosted on RMG server with 2364 examples) and `cyclic_O_only_table_train` (a table hosted on RMG server with 20496 examples) gives following performance.

thermo estimation error (kcal/mol) | master GA | this PR 
----- | ----- | -----
small_cyclic_table | 3.42 | 2.19 
large_linear_polycyclic_table | 9.09 | 1.86
large_fused_polycyclic_table | 9.53 | 1.96
small_O_only_polycyclic_table | 5.30 | 0.99
large_linear_O_only_polycyclic_table | 6.94 | 1.66
large_fused_O_only_polycyclic_table | 10.32 | 1.63

More performance updates will be added soon with nitrogen, sulfur and radical datasets.

## How to use it
To set up the foundation for future related PRs, this PR includes cnn model framework for thermo-estimator, training script and evaluate script.

### To train an example model
```bash
cd examples/cnn/train

bash train.sh
```
### To evaluate performance for a testing model
```bash
cd examples/cnn/evaluate

bash evaluate.sh
```
## Future PRs

There will a couple more PRs coming after this is merged.
- uncertainty estimation
- train model with entropy and heat capacity
- integration with RMG so that simulation can use CNN model's predictions

PR #1362, opened 2018-05-10
Integrate MCNN (Molecular Convolutional Neural Network) thermo estimator with RMG

KEHANG:
This PR is an alternative version of PR https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1251.

Instead of adding all the CNN modules (data processing, model building, training, testing, loading parameters, and prediction) into RMG, we made a separate package of all these machine learning modules, called [DataDrivenEstimator](https://github.com/KEHANG/DataDrivenEstimator) (package short name: **`dde`**, key module: **`cnn_framework`**) and appended it to RMG's dependency list.

Thus this PR is very light, with only a handful of commits, including

- use **`cnn_framework`** to create a new thermo estimator: **`MCNNEstimator`**.

- load **`MCNNEstimator`** from RMG input file via **`mcnnEstimator()`**,

- an example **`minimal_mcnn`** to show the input file with **`MCNNEstimator`** turned on,

- currently **`MCNNEstimator`** will only be used when user specifically turn it on via input file, and
estimate thermo for **saturated non-aromatic polycyclics** as a starting point, as we discussed in the related RMG subgroup meeting. So user could opt out (the default option), and a developer can increase its responsibility in the future whenever he/she sees it fits. A performance dashboard mentioned later can guide such decision.

- latest weights files are placed in `rmgpy/mcnn/pretrained_models`. In the future, users are allowed to replace them with better weights trained via [DataDrivenEstimator](https://github.com/KEHANG/DataDrivenEstimator).

A brief introduction to [DataDrivenEstimator](https://github.com/KEHANG/DataDrivenEstimator). 
- we design **`DataDrivenEstimator`** in a way so that it contains current and future machine learning estimators developed by the team. Currently it only has MCNN thermo estimator, later it may have MCNN kinetic estimator, transfer learning feature, uncertainty etc.
- the MCNN thermo estimator currently supports resonance-isomer-free input; without bond order or atom-type features it also gets reasonably good prediction. 

Current status of package [DataDrivenEstimator](https://github.com/KEHANG/DataDrivenEstimator):

- thermo prediction performance dashboards can be found for [Hf298](http://kehangsblog.com/thermo_predictor/overall_performance/Hf298/), [S298](http://kehangsblog.com/thermo_predictor/overall_performance/S298/) and [Cp](http://kehangsblog.com/thermo_predictor/overall_performance/Cp/)

- the binary package of [DataDrivenEstimator v1.0.0](https://anaconda.org/RMG/dde) has been uploaded to RMG channel with Linux and Mac versions.

- The package is currently under my github account. After my graduation, it might be more convenient to transfer it over to `ReactionMechanismGenerator` organization.

### DIFF
--- a/environment_linux.yml
+++ b/environment_linux.yml
@@ -42,4 +42,4 @@ dependencies:
   - muq
   - pymongo
   - mpmath
-  
+  - dde
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -48,6 +48,7 @@ from rmgpy.thermo import NASAPolynomial, NASA, ThermoData, Wilhoit
 from rmgpy.molecule import Molecule, Bond, Group
 import rmgpy.molecule
 from rmgpy.species import Species
+from rmgpy.ml.estimator import MLEstimator
 
 #: This dictionary is used to add multiplicity to species label
 _multiplicity_labels = {1:'S',2:'D',3:'T',4:'Q',5:'V',}
@@ -1102,17 +1103,15 @@ class ThermoDatabase(object):
         Return the thermodynamic parameters for a given :class:`Species`
         object `species`. This function first searches the loaded libraries
         in order, returning the first match found, before falling back to
-        estimation via group additivity.
+        estimation via machine learning and then group additivity.
         
-        The method corrects for symmetry when the molecule uses HBI or 
-        group additivity. Libraries and direct QM calculations are already
-        corrected.
+        The method corrects for symmetry when the molecule uses machine
+        learning or group additivity. Libraries and direct QM calculations
+        are already corrected.
         
         Returns: ThermoData
         """
         from rmgpy.rmg.input import getInput
-        
-        thermo0 = None
         
         thermo0 = self.getThermoDataFromLibraries(species)
         
@@ -1127,6 +1126,12 @@ class ThermoDatabase(object):
         except Exception:
             logging.debug('Quantum Mechanics DB could not be found.')
             quantumMechanics = None
+
+        try:
+            ml_estimator, ml_settings = getInput('MLEstimator')
+        except Exception:
+            logging.debug('ML estimator could not be found.')
+            ml_estimator, ml_settings = None, None
             
         if quantumMechanics:
             original_molecule = species.molecule[0]
@@ -1177,7 +1182,7 @@ class ThermoDatabase(object):
                     thermo0 = quantumMechanics.getThermoData(original_molecule) # returns None if it fails
                 
         if thermo0 is None:
-            # Use group additivity methods to determine thermo for molecule (or if QM fails completely)
+            # First try finding stable species in libraries and using HBI
             for mol in species.molecule:
                 if mol.reactive:
                     original_molecule = mol
@@ -1219,17 +1224,29 @@ class ThermoDatabase(object):
                     species.molecule = newMolList
                     thermo0 = thermo[0][2]
 
-                else:
-                    # Did not find any saturated values in the thermo libraries, so try group additivity instead
-                    thermo0 = self.getThermoDataFromGroups(species)
+            if thermo0 is None:
+                # If we still don't have thermo, use ML to estimate it, but
+                # only if the molecule is made up of H, C, N, and O atoms and
+                # is not a singlet carbene. Also check ML settings.
+                if (ml_estimator is not None
+                        and all(a.element.number in {1, 6, 7, 8} for a in species.molecule[0].atoms)
+                        and species.molecule[0].getSingletCarbeneCount() == 0):
 
-            else:
-                # Saturated molecule, estimate it via groups since we've already checked libraries much earlier
+                    nheavy = sum(1 for atom in species.molecule[0].atoms if atom.isNonHydrogen())
+                    min_heavy = ml_settings['min_heavy_atoms'] or 0
+                    max_heavy = ml_settings['max_heavy_atoms'] or numpy.inf
+                    
+                    if min_heavy <= nheavy <= max_heavy:
+                        thermo0 = self.get_thermo_data_from_ml(species,
+                                                               ml_estimator,
+                                                               ml_settings['uncertainty_cutoffs'])
+
+            if thermo0 is None:
+                # And lastly, resort back to group additivity to determine thermo for molecule
                 thermo0 = self.getThermoDataFromGroups(species)
-                
-            # update entropy by symmetry correction
-            thermo0.S298.value_si -= constants.R * math.log(species.getSymmetryNumber())
 
+            # Update entropy by symmetry correction (not included in trained ML model)
+            thermo0.S298.value_si -= constants.R * math.log(species.getSymmetryNumber())
 
         # Make sure to calculate Cp0 and CpInf if it wasn't done already
         findCp0andCpInf(species, thermo0)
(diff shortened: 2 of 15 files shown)


rmgpy/rmg/input.py
```diff
@@ -274,6 +274,21 @@ def quantumMechanics(
                                         onlyCyclics = onlyCyclics,
                                         maxRadicalNumber = maxRadicalNumber,
                                         )
+
+def mcnnEstimator(mode='thermochemistry'):
```
mliu49: Perhaps change the input argument to `thermochemistry=True`? That way, in the future we could potentially add `kinetics=True`, etc.

The if logic would also be a bit easier since you could just do `if thermochemistry:` and have Python automatically handle unsupported arguments.

examples/rmg/minimal_mcnn/RMG.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# Make sure your PYTHONPATH includes the necessary paths to run RMG-Py 
+# Uncomment the following lines if your PYTHONPATH is not stored in 
+# your .bashrc file already. Be sure to modify the paths to the locations
+# of your code as necessary.
+
+# export PYTHONPATH=$PYTHONPATH:$HOME/RMG-Py/
+# export PYTHONPATH=$PYTHONPATH:$HOME/RMG-database/
+# export PYTHONPATH=$PYTHONPATH:$HOME/PyDAS/
+# export PYTHONPATH=$PYTHONPATH:$HOME/PyDQED/
```
mliu49: I think we can remove the PyDAS and PyDQED statements, here and below as well.

environment_linux.yml
```diff
@@ -41,3 +41,4 @@ dependencies:
   - mock
   - muq
   - pymongo
+  - dde
```
mliu49: Could you also add to meta.yaml, in the run section?

cgrambow:
This is ready for review. The new module is called `estimator` and provides the `MLEstimator` class to predict thermo with uncertainties using trained neural net models. The model parameters have been moved to the database and are part of PR ReactionMechanismGenerator/RMG-database#303, which should be merged alongside this. Users can easily specify new ML parameters by making a new folder in the database directory that contains `H298`, `S298`, and `Cp` subfolders, and then specifying the new folder using the `name` argument provided to the `mlEstimator` block in the input file.

If uncertainties are available for the ML model, they will be automatically calculated alongside the thermo predictions. They are then compared to the provided (or default) `UncertaintyCutoff` values and if any of the uncertainties are larger than their corresponding cutoff, then RMG falls back to group additivity instead. For the current PR, uncertainties are not yet available for enthalpy.

New options also include `minHeavyAtoms` and `maxHeavyAtoms`. Setting `minHeavyAtoms` to some medium value is generally recommended for good predictions. If the reviewer would like, I can also include other options; perhaps something like `polycyclicsOnly` would be desirable.

Because singlet carbenes are estimated with groups instead of HBI, the ML model is not used for them, because no singlet carbenes are present in the training set. Radicals are also not present in the training set, so they are estimated using ML + HBI. The ML model is only used for molecules containing HCNO atoms.

rmgpy/ml/estimator.py
```diff
@@ -0,0 +1,138 @@
...
+
+        return thermo
+
+    def get_thermo_data_for_species(self, species):
+        """
+        Return the set of thermodynamic parameters corresponding to a
+        given :class:`Species` object `species`.
+
+        The current ML estimator treats each resonance isomer
+        identically, i.e., any of the resonance isomers can be chosen.
```
mliu49: For my knowledge, how does this work? Are the graphs reduced to connectivity only before convolution?

rmgpy/ml/estimator_test.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def setUp(self):
+        models_path = os.path.join(settings['database.directory'], 'thermo', 'ml', 'main')
+        Hf298_path = os.path.join(models_path, 'H298')
+        S298_path = os.path.join(models_path, 'S298')
+        Cp_path = os.path.join(models_path, 'Cp')
+        self.ml_estimator = MLEstimator(Hf298_path, S298_path, Cp_path)
+
+    def test_get_thermo_data(self):
+        mol = Molecule().fromSMILES('C1C2C1C2')
```
mliu49: Add a docstring? Mainly to give nosetests something to print besides the test name.

rmgpy/ml/estimator.py
cgrambow: The ML models don't make use of bond order information and they don't differentiate between aromatic and non-aromatic atoms, so all resonance isomers look the same to the machine. We found that bond order features are not needed to train accurate models, which is quite nice.

rmgpy/data/thermo.py
```diff
@@ -1219,17 +1224,30 @@ def getThermoData(self, species, trainingSet=None):
...
-                else:
-                    # Did not find any saturated values in the thermo libraries, so try group additivity instead
-                    thermo0 = self.getThermoDataFromGroups(species)
+            if thermo0 is None:
+                # If we still don't have thermo, use ML to estimate it, but
+                # only if the molecule is made up of H, C, N, and O atoms and
+                # is not a singlet carbene. Also check ML settings.
+                if (ml_estimator is not None
+                        and all(a.isHydrogen() or a.isCarbon() or a.isNitrogen() or a.isOxygen()
+                                for a in species.molecule[0].atoms)
```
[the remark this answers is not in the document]
cgrambow: I'll go with the second option, which is more flexible if we add additional atoms in the future.

rmgpy/data/thermo.py
```diff
@@ -1219,17 +1224,30 @@ def getThermoData(self, species, trainingSet=None):
...
+                        and species.molecule[0].getSingletCarbeneCount() == 0):
+
+                    nheavy = sum(1 for atom in species.molecule[0].atoms if atom.isNonHydrogen())
+                    min_heavy = 0 if ml_settings['min_heavy_atoms'] is None else ml_settings['min_heavy_atoms']
+                    max_heavy = numpy.inf if ml_settings['max_heavy_atoms'] is None else ml_settings['max_heavy_atoms']
```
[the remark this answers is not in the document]
cgrambow: I always forget that I can do this. Thanks for the suggestion.

rmgpy/data/thermo.py
```diff
@@ -1470,8 +1520,8 @@ def estimateRadicalThermoViaHBI(self, molecule, stableThermoEstimator ):
         if not isinstance(thermoData_sat, ThermoData):
             thermoData_sat = thermoData_sat.toThermoData()
         
-        
-        if not stableThermoEstimator == self.computeGroupAdditivityThermo:
+        if not (stableThermoEstimator == self.computeGroupAdditivityThermo
+                or isinstance(stableThermoEstimator.__self__, MLEstimator)):
```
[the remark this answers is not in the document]
cgrambow: For some reason that doesn't work, i.e., when I do that, then it enters that if-statement even though `stableThermoEstimator` is the thermo estimation method from `MLEstimator`. I think the reason is that you are trying to compare the bound method of the class `MLEstimator` to a method of an instance of that class and that doesn't work. You don't have that issue for `computeGroupAdditivityThermo` because `self` refers to the class instance.

rmgpy/data/thermo.py
```diff
@@ -1404,6 +1422,38 @@ def getThermoDataFromGroups(self, species):
...
+        ML estimator. Also compare the estimated uncertainties to the
+        user-defined cutoffs. If any of the uncertainties are larger
+        than their corresponding cutoffs, return None.
+
+        The entropy is not corrected for the symmetry of the molecule.
+        This should be done later by the calling function.
+        """
+        if species.molecule[0].isRadical():
+            thermo = [self.estimateRadicalThermoViaHBI(mol, ml_estimator.get_thermo_data) for mol in species.molecule]
+            thermo0 = min(thermo, key=lambda tdata: tdata.H298)
```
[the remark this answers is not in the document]
cgrambow: See fixup commit. I only sort it in the case of radicals, because I would like to avoid making additional predictions for all resonance isomers if not necessary. Is it also important to sort the non-radical ones?

rmgpy/ml/estimator_test.py
cgrambow: Done.

cgrambow:
I addressed your changes. Let me know what you think.

rmgpy/data/thermo.py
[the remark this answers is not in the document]
mliu49: I think that's ok. All of the resonance structures would end up with the same thermo anyway, right?

The changes I'm making to resonance structure generation in #1498 should make this sorting less important as well.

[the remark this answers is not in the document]
cgrambow: Yup, that's right.

mliu49:
Ok, the minimal example crashed with the error you documented in #1495. I think this PR can move on though, since it isn't a direct cause of that bug.

I also noticed some logging which may be unnecessary:
- When loading the predictors, the entire predictor_model file is logged ([here](https://github.com/ReactionMechanismGenerator/DataDrivenEstimator/blob/master/dde/input.py#L75))
- keras also logs and prints some stuff when creating loading the model architecture (the printed output does not show up in RMG.log)
- There's a weird None which is logged, possibly from keras

These aren't done by the code in RMG-Py, so it's independent of this PR, but it's something to look into when you get an opportunity.

cgrambow:
Yeah, the database PR can be merged first. Yeah, some of that logging might be done by `dde` and some by Keras. Not sure where the None comes from.

mliu49:
Ok, you can go ahead and rebase.

PR #1365, opened 2018-05-15
Merge v2.1.9 updates and new git workflow

mliu49:
Contains version number changes and release notes for versions 2.1.1 through 2.1.9. RMG releases 2.1.1 through 2.1.8 were created retroactively to help stay organized and have tags to refer to. Binaries are not planned for those versions.

RMG release 2.1.9 is in progress. This PR contains also changes to the conda recipe to fix issues which have appeared since the last release. After this PR is merged into `master`, the same release branch will be merged into the new `stable` branch. A binary release will be build on that branch.

Following this PR, we will officially move to the new git workflow, which integrates some features of git-flow. Development will continue as usual on the `master` branch. Release branches will occasionally be created from `master` and merged into both `master` and `stable`. Release tags will reference the `stable` branch only. The requirement for strictly linear history (i.e. rebasing before merging) will be relaxed for these branches, provided that the merge can be performed cleanly.

For reviewing this PR, proofread the release notes and skim through the changes to the conda recipe. Testing the conda recipe is optional, since that'll happen when I build the final binary.


meta.yaml
```diff
@@ -1,7 +1,7 @@
 # For conda build
 package:
   name: rmg
-  version: "0" # set by build.sh, which gets from rmgpy/__init__.py
+  version: 2.1.9
```
KEHANG: could use `version: {{ environ.get('GIT_DESCRIBE_TAG', '') }}` once you've set the tag

setup.py
```diff
@@ -157,10 +157,6 @@ def getCanthermExtensionModules():
 ################################################################################
 
 ext_modules = []
-if 'install' in sys.argv:
-    # This is so users can still do simply `python setup.py install`
-    ext_modules.extend(getMainExtensionModules())
-    ext_modules.extend(getSolverExtensionModules())
```
KEHANG: I'm not very clear about this removal. Could you explain a little bit more?

meta.yaml
mliu49: That's true, I'll change it. It didn't work during testing because the tags aren't on this branch, but it should work in general.

PR #1368, opened 2018-05-18
Adjust atom sorting algorithm for Molecules

mliu49:
Use negative of connectivity value to prioritize atoms with high connectivity.

Also eliminate unnecessary `getAtomConnectivityValue` method.

This is a pretty minor fix which somewhat restores prior behavior. In 8abbee9a9856bfd3cbac9d74da8e3d6a856cbefc, I restored a previously implemented improvement to the atom sorting method. However, I failed to notice that `Molecule.sortAtoms()` method sorts in reverse, whereas the `Graph.sortVertices()` method sorts in the forward direction. Therefore, we need to take the negative of `getVertexConnectivityValue` to maintain the same result.

### DIFF
--- a/rmgpy/molecule/molecule.py
+++ b/rmgpy/molecule/molecule.py
@@ -212,11 +212,13 @@ class Atom(Vertex):
                     return False
             return True
     
-    def getDescriptor(self):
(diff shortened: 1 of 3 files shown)


mliu49:
@goldmanm, have you had a chance to look at this?

PR #1369, opened 2018-05-19
Characteristic Rate Based Termination Criterion

mjohnson541:
Until now RMG has had two termination criteria:  time and conversion.  The fundamental problem with these criteria is that before we run RMG we don't know what the right value is for them.  We don't know how long the chemistry takes.  We don't know if the chemistry continues after conversion of the reactants to something else.  We don't even know if a specified conversion can even be reached.  

This creates issues, terminating using only time can cause you to simulate past where the chemistry is really happening at points where the characteristic rate is very small, which can cause RMG to add undesirable species.  Terminating using conversion will terminate much earlier than desired if the reactant breaks down faster than the downstream chemistry and will never terminate if the specified termination conversion is unachievable.  

This PR specifies a criterion that terminates the run exactly when the chemistry stops rather than relying on user guesses for times and conversions.  The criterion is the characteristic rate at a given time divided by the maximum characteristic rate so far observed in the run.  When this ratio drops below the specified value (a fraction of the maximum characteristic rate) the run terminates.  This termination criterion has none of the issues with time or conversion based termination and is theoretically robust to any chemical system.

### DIFF
--- a/documentation/source/users/rmg/input.rst
+++ b/documentation/source/users/rmg/input.rst
@@ -214,9 +214,13 @@ Currently, RMG can only model constant temperature and pressure systems. Future 
 will allow for variable temperature and pressure. To define a reaction system we need to
 define the temperature, pressure and initial mole fractions of the reactant species. The
 initial mole fractions are defined using the label for the species in
-the species block. Every reaction system can have its termination criterion based on
-species conversion or termination time or both. When both termination criterion are specified
-the model generation will stop when either of the termination criterion is satisfied.
+the species block. Reaction system simulations terminate when one of the specified termination
+criteria are satisfied.  Termination can be specied to occur at a specific time, at a specific
+conversion of a given initial species or to occur at a given terminationRateRatio, which is the
+characteristic flux in the system at that time divided by the maximum characteristic flux observed so far
+in the system (measure of how much chemistry is happening at a moment relative to the main chemical process).  
+
+
 
 The following is an example of a simple reactor system::
 
@@ -232,6 +236,7 @@ The following is an example of a simple reactor system::
 			'CH4': 0.9,
 		},
 		terminationTime=(1e0,'s'),
+		terminationRateRatio=0.01,
 		sensitivity=['CH4','H2'],
 		sensitivityThreshold=0.001,
 
--- a/rmgpy/rmg/input.py
+++ b/rmgpy/rmg/input.py
@@ -38,7 +38,7 @@ from rmgpy import settings
 
 from rmgpy.molecule import Molecule
 from rmgpy.quantity import Quantity
-from rmgpy.solver.base import TerminationTime, TerminationConversion
+from rmgpy.solver.base import TerminationTime, TerminationConversion, TerminationRateRatio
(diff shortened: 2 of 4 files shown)


documentation/source/users/rmg/input.rst
```diff
@@ -214,9 +214,13 @@ Currently, RMG can only model constant temperature and pressure systems. Future
...
 define the temperature, pressure and initial mole fractions of the reactant species. The
 initial mole fractions are defined using the label for the species in
-the species block. Every reaction system can have its termination criterion based on
-species conversion or termination time or both. When both termination criterion are specified
-the model generation will stop when either of the termination criterion is satisfied.
+the species block. Reaction system simulations terminate when one of the specified termination
+criteria are satisfied.  Termination can be specied to occur at a specific time, at a specific
+conversion of a given initial species or to occur at a given terminationRateRatio, which is the
+characteristic flux in the system at that time divided by the maximum characteristic flux observed in the
+system (measure of how much chemistry is happening at a moment relative to the main chemical process).
```
alongd: the maximum characteristic flux observed in the system *so far*

rwest:
If this does what it sets out to do (i.e. the motivation provided in the original pull request) then I am a big fan! Thank you!

mjohnson541:
In all my tests it has worked effectively as a termination criterion by itself.  I haven't really tested it on the cases where the other criteria work poorly yet, but those situations shouldn't feel different to this criterion than a typical RMG run.  It feels a bit harder to make sense of since time and conversion are things that chemical engineers are familiar with, but I think that's just an illusion since we typically have no idea how long an arbitrary reaction system will take to react and conversion is a horrible way to think about multi step processes.  

So I tested this on the minimal example (Ethane pyrolysis) and an Ethane oxidation example.  

For the minimal example with filter (with no other termination criteria):  
0.001->62 species, 484 reactions
0.01-> 25 species, 63 reactions
0.003-> 27 species, 68 reactions
0.002->28 species, 78 reactions
0.03->15 species, 23 reactions
90% conversion, 1e6 sec -> 25 species, 63 reactions

For Ethane Combustion at 700 K (with no other termination criteria):
0.01->80 species, 518 reactions
0.1->67 species, 396 reactions
Fittingly I initially guessed the time poorly for this one.  
90% conversion, 1e6 sec -> 64 species, 384 reactions

documentation/source/users/rmg/input.rst
mjohnson541: added

mjohnson541:
I think this ready to merge.

rwest:
Now this has been around for 10 months... does it work as intended? Is it helpful? is 0.01 a good value?

mjohnson541:
Yes, I haven't run into any situations where it doesn't work yet and 0.01 seems to be a good value.  

I'm not sure if anyone else is actually using it though.  It hasn't caught on as fast as ranged reactors, probably because it doesn't simplify people's input file writing as much.

mjohnson541:
It's absolutely more robust than conversion and time.

alongd:
I use it! (seems to be working fine)

Issue #1367, opened 2018-05-17
ValueError: invalidObjects could not be filled during resurrection process


davidfarinajr:
I've added some chlorine rules etc to my database (as discussed a bit in https://github.com/ReactionMechanismGenerator/RMG-database/pull/257 - will open a PR once we have it closer to working), and I am trying to build a model for chloromethane combustion. After one iteration it fails:

```python
initializing surface ...
surface initialization complete
 DASPK--  AT T (=R1) AND STEPSIZE H (=R2) THE                                    
      In above,  R1 =  0.3886834855475D-38   R2 =  0.6996302739854D-38
 DASPK--  ERROR TEST FAILED REPEATEDLY OR WITH ABS(H)=HMIN                       
Error: Trying to step from time 0.0 to 1e-12 resulted in a solver (DASPK) error
Resurrecting Model...
Error: Model Resurrection has failed
Error: Core species names: ['Ar', 'He', 'Ne', 'N2', 'CH3Cl(1)', 'H2(2)', 'O2(3)', 'HCl(4)', 'Cl(6)', 'CH3(5)']
Error: Core species moles: array([  9.60000000e-01,   0.00000000e+00,   0.00000000e+00,
         0.00000000e+00,   2.00000000e-02,   1.00000000e-02,
         1.00000000e-02,   0.00000000e+00,   1.11803399e-16,
         1.11803399e-16])
Error: Volume: 0.09752875656000003
Error: Core species net rates: array([  0.00000000e+00,   0.00000000e+00,   0.00000000e+00,
         0.00000000e+00,  -2.94934944e+23,   0.00000000e+00,
         0.00000000e+00,   0.00000000e+00,   2.94934944e+23,
         2.94934944e+23])
Error: Edge species net rates: array([], dtype=float64)
Error: Network leak rates: array([], dtype=float64)
Traceback (most recent call last):
  File "/Users/DavidFarina/Code/RMG-Py//rmg.py", line 173, in <module>
    main()
  File "/Users/DavidFarina/Code/RMG-Py//rmg.py", line 167, in main
    rmg.execute(**kwargs)
  File "/Users/DavidFarina/Code/RMG-Py/rmgpy/rmg/main.py", line 695, in execute
    simulatorSettings = simulatorSettings,
  File "rmgpy/solver/base.pyx", line 520, in rmgpy.solver.base.ReactionSystem.simulate
[...]

Issue #1258, opened 2018-01-15
DASPK error after attempted model resurrection


cgrambow:
I am trying to run two separate hydrocarbon oxidation simulations, which run fine by themselves. Upon adding a phenolic species to each job, I get DASPK errors for both jobs after ~160 species. Both errors look like this:

```
Error: Trying to step from time 0.0 to 100000.0 resulted in a solver (DASPK) error
Resurrecting Model...
Error: Model Resurrection has failed
```
[...]
```
Traceback (most recent call last):
  File "/data1/cgrambow/Code/RMG-Py/rmg.py", line 173, in <module>
    main()
  File "/data1/cgrambow/Code/RMG-Py/rmg.py", line 167, in main
    rmg.execute(**kwargs)
  File "/data1/cgrambow/Code/RMG-Py/rmgpy/rmg/main.py", line 702, in execute
    simulatorSettings = simulatorSettings,
  File "rmgpy/solver/base.pyx", line 498, in rmgpy.solver.base.ReactionSystem.simulate (build/pyrex/rmgpy/solver/base.c:24537)
  File "rmgpy/solver/base.pyx", line 678, in rmgpy.solver.base.ReactionSystem.simulate (build/pyrex/rmgpy/solver/base.c:15263)
ValueError: invalidObjects could not be filled during resurrection process
 DASPK--  AT T (=R1) AND STEPSIZE H (=R2) THE
      In above,  R1 =  0.7945984920517D+05   R2 =  0.1598236102956D-04
 DASPK--  NONLINEAR SOLVER FAILED TO CONVERGE
 DASPK--  REPEATEDLY OR WITH ABS(H)=HMIN
```

Evidently, model resurrection was not able to fix the simulation. Any ideas on how to avoid the error?

mjohnson541:
So if you have the latest model resurrection related commit this should only happen if your edge is entirely empty.  Is that the case?

cgrambow:
Maybe? Two lines above the traceback, it prints
```
Error: Edge species net rates: array([], dtype=float64)
Error: Network leak rates: array([], dtype=float64)
```
but the latest log about the edge after model enlargement was
```
The model edge has 157570 species and 409040 reactions
```

mjohnson541:
[...]

PR #1371, opened 2018-05-19
Allow Resurrection from Failed Filter Simulations

mjohnson541:
This adds a try except that allows model runs to continue after failed filter simulations.  I believe this caused #1367 and #1258.  It worked as expected on the example in #1367.  However, its worth noting that nothing is reacted until a regular or a filter simulation succeeds and changes the react flags.  In the example in #1367 the run doesn't recover which results in the resurrection process consuming species from the edge until the edge is empty and then crashing.

### DIFF
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -723,7 +723,8 @@ class RMG(util.Subject):
                             # Run a raw simulation to get updated reaction system threshold values
                             # Run with the same conditions as with pruning off
                             if not resurrected:
-                                reactionSystem.simulate(
+                                try:
+                                    reactionSystem.simulate(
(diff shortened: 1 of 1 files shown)


rmgpy/rmg/main.py
```diff
@@ -735,9 +736,14 @@ def execute(self, **kwargs):
                                         simulatorSettings = simulatorSettings,
                                         conditions = self.rmg_memories[index].get_cond()
                                     )
-                                self.updateReactionThresholdAndReactFlags(
+                                    self.updateReactionThresholdAndReactFlags(
```
alongd: I think here we're capturing exceptions from `simulate()`. Perhaps `updateReactionThresholdAndReactFlags`  should be relocated to an `else` clause after the try-except?

PR #1370, opened 2018-05-19
Extend saveTrainingReactions to take list input

mjohnson541:
This PR includes some improvements to the saveTrainingReaction function that were useful in converting the rules in the database to training reactions.  Note this is not interdependent with the rules to training reactions pull request itself, although it was used to run the script used to create that PR.

PR #1373, opened 2018-05-22
Revamped training reaction ipython notebook

mliu49:
This PR moves the `convertKineticsLibraryToTrainingReactions.ipynb` notebook from RMG-database to RMG-Py. This was done to make it easier to keep the script up to date with RMG-Py.

The notebook was streamlined substantially by converting most of the code to functions and moving them into a separate file. New steps were also added to handle the common issue of having multiple matches and offer summary to review the reactions to be added.

Some improvements were also added to `KineticsFamily.saveTrainingReactions` to make it easier to use.

This PR includes changes from ReactionMechanismGenerator/RMG-database#125 and ReactionMechanismGenerator/RMG-database#241.

### DIFF
--- a/ipython/kinetics_library_to_training_tools.py
+++ b/ipython/kinetics_library_to_training_tools.py
@@ -1,0 +1,425 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+###############################################################################
+#                                                                             #
+# RMG - Reaction Mechanism Generator                                          #
+#                                                                             #
+# Copyright (c) 2002-2018 Prof. William H. Green (whgreen@mit.edu),           #
+# Prof. Richard H. West (r.west@neu.edu) and the RMG Team (rmg_dev@mit.edu)   #
+#                                                                             #
+# Permission is hereby granted, free of charge, to any person obtaining a     #
+# copy of this software and associated documentation files (the 'Software'),  #
+# to deal in the Software without restriction, including without limitation   #
+# the rights to use, copy, modify, merge, publish, distribute, sublicense,    #
+# and/or sell copies of the Software, and to permit persons to whom the       #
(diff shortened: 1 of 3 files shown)


mjohnson541:
If you aren't aware you only merged in the first commit from #1370

mliu49:
I changed the default value for `reference` back to `None` and dropped the second commit. Was there a reason you had changed it to an empty string?

rmgpy/data/kinetics/family.py
```diff
@@ -755,48 +755,45 @@ def saveTrainingReactions(self, reactions, reference=None, referenceType='', sho
...
-        speciesDict = Database().getSpecies(directory_file)
+        species_dict = Database().getSpecies(dictionary_path)
 
-        # add new unique species with labeledAtoms into speciesDict
+        # Add new unique species with labeledAtoms into species_dict
         for rxn in reactions:
             for spec in (rxn.reactants + rxn.products):
-                for ex_spec_label in speciesDict:
-                    ex_spec = speciesDict[ex_spec_label]
+                for ex_spec in species_dict.itervalues():
```
amarkpayne: Good catch!

ipython/kinetics_library_to_training_tools.py
```diff
@@ -0,0 +1,425 @@
+#!/usr/bin/env python
```
amarkpayne: Since this file does not have a `main()` function that would get executed if you were to run this as a script, should we move this file to `rmgpy/tools/` where code related to this is stored?

mliu49: I considered that, but also wanted to keep it close to the ipython notebook. I can move it if you think it would be better.

mliu49:
@amarkpayne, this has been rebased and is ready to merge.

PR #1374, opened 2018-05-22
New recommended families system and improved family loading

mliu49:
This implements a new system for recommended families and specifying kinetics families which offers much more flexibility than the current method.

## Recommended families
The current dictionary of recommended families will be replaced by sets, still located in the `recommended.py` file. Multiple sets can be defined, each specifying some subset of kinetics families. Items in these sets must correspond to an existing family, but there is no hard requirement that every family is part of a set.

## Family loading
Loading of kinetics families has been rewritten to take advantage of sets. Users can specify any combination of the sets defined in `recommended.py` and individual families. The union of all specified items is then loaded. The `'all'` and `'none'` options still work, and an empty list `[]` is also now accepted.

Requesting the inverse of the specified families can still be done (using `!`), and will simply take the union of all the specified families or sets and subtract the result from the list of all families (determined by the directories present on disk).

## Transition
The newly implemented methods are fully backwards compatible with the existing `recommended.py` format. Thus, this PR can be merged before replacing the `recommended.py` file in RMG-database. 

When we do update the recommended families, I think it would be good to have a discussion about what sets we would like to define.


rmgpy/data/kinetics/database.py
```diff
@@ -107,86 +107,101 @@ def load(self, path, families=None, libraries=None, depositories=None):
...
-            self.recommendedFamilies = local_context['recommendedFamilies']
-        except Exception, e:
-            raise DatabaseError('Error while reading list of recommended families from {0}/recommended.py.\n{1}'.format(filepath,e))
-        for recommended in self.recommendedFamilies.values():
-            if not isinstance(recommended, bool):
-                raise DatabaseError("recommendedFamilies dictionary should contain only True or False values")
+        import imp
+
+        # Load the recommended.py file as a module
+        rec = imp.load_source('rec', filepath)
```
cgrambow: Dynamically loading the module is a really good idea. The old code was so ugly... You might want to catch possible exceptions here and raise `DatabaseError`s, similarly to what the previous code did.

rmgpy/data/kinetics/database.py
```diff
@@ -107,86 +107,101 @@ def load(self, path, families=None, libraries=None, depositories=None):
...
+        # For backward compatibility, check for old-style recommendedFamilies dictionary
+        if hasattr(rec, 'recommendedFamilies'):
+            default = set()
+            for family, recommended in rec.recommendedFamilies.iteritems():
+                if recommended:
+                    default.add(family)
+            self.recommendedFamilies = {'default': default}
+        else:
+            recommended = {}
+            for option in [attribute for attribute in dir(rec) if not attribute.startswith('_')]:
```
cgrambow: I think this would be slightly more readable if you used `rec.__dict__.iteritems()`.

rmgpy/data/kinetics/database.py
```diff
@@ -107,86 +107,101 @@ def load(self, path, families=None, libraries=None, depositories=None):
...
-                    if label not in dirs:
-                        raise DatabaseError('Family {0} not found on disk.'.format(label))
+        all_families = set(dirs)
+
+        # Convert input to a list to simplify processing
+        if not isinstance(families, list):
+            families = [families]
+
+        # Check for ! syntax, which allows specification of undesired families
+        if all([item.startswith('!') for item in families]):
```
cgrambow: You can remove the square brackets around `item.startswith('!') for item in families`, because `all` accepts generator expressions.

rmgpy/data/kinetics/database.py
```diff
@@ -107,86 +107,101 @@ def load(self, path, families=None, libraries=None, depositories=None):
...
+            inverse = True
+            families = [family[1:] for family in families]
+        elif not any([item.startswith('!') for item in families]):
```
cgrambow: Same here.

rmgpy/data/kinetics/database.py
mliu49: Thanks, I didn't know about `__dict__`. It seems to give the same set of attributes as `dir()` in this case:
```
>>> dir(rec)
['__builtins__', '__doc__', '__file__', '__name__', '__package__', 'default', 'pah']
>>> rec.__dict__.keys()
['__builtins__', '__file__', '__package__', 'default', 'pah', '__name__', '__doc__']
```
I used the list comprehension is to filter out the private/special attributes and only keep the actual variables that have been defined. Using `__dict__` would remove the need for `getattr` though.
```
recommended = {}
for attribute, value in rec.__dict__.iteritems():
    if not attribute.startswith('_'):
        recommended[attribute] = value
```

And with `__dict__`, this could actually be done with a dictionary comprehension:
```
recommended = {attribute: value for attribute, value in rec.__dict__.iteritems() if not attribute.startswith('_')}
```

Let me know what you prefer.

cgrambow: I like the dict comprehension

rmgpy/data/kinetics/database.py
mliu49: Added

rmgpy/data/kinetics/database.py
mliu49: Thanks, fixed.

rmgpy/data/kinetics/database.py
mliu49: Fixed.

mliu49:
Thanks for reviewing! I also made some other changes to error handling in the fixup commit.

cgrambow:
Looks good to me. You can squash and rebase.

PR #1376, opened 2018-05-25
Added sensitivity analysis to Cantherm

alongd:
Now users can run a Cantherm job (`KineticsJob` or `PressureDependenceJob`) and see to which wells or transitions states the calculated rates are most sensitive to (and hopefully later improve them).
The output of the SA is saved into a `sensitivity` folder in the output directory with a text file tabulating all normalized sensitivity coefficients per reaction (in both direction if the network reaction is reversible), and figures visualizing the sensitivity coefficients per network reaction at all conditions defined for the sensitivity analysis.

Test, example, and some documentation also added.

### DIFF
--- a/documentation/source/users/cantherm/input.rst
+++ b/documentation/source/users/cantherm/input.rst
@@ -611,6 +611,20 @@ This is also acceptable::
 
     kinetics('H + C2H4 <=> C2H5')
 
+If a sensitivity analysis is desired, simply add the conditions at which to calculate sensitivity coefficients
+in the following format, e.g.::
+
+    kinetics(
+        label = 'HSOO <=> HOOS',
+        Tmin = (500,'K'), Tmax = (3000,'K'), Tcount = 15,
+        sensitivity_conditions = [(1000, 'K'), (2000, 'K')]
+    )
+
+The output of a sensitivity analysis is saved into a ``sensitivity`` folder in the output directory. A text file, named
+with the reaction label, delineates the semi-normalized sensitivity coefficients ``dln(r)/dE0`` in units of ``mol/J``
+at all requested conditions. A horizontal bar figure is automatically generated per reaction with subplots for both the
+forward and reverse direction at all conditions.
+
 Examples
 ========
 
--- a/rmgpy/cantherm/kinetics.py
+++ b/rmgpy/cantherm/kinetics.py
@@ -48,7 +48,7 @@ from rmgpy.molecule.draw import MoleculeDrawer, createNewSurface
 
 ################################################################################
 
-class KineticsJob:
+class KineticsJob(object):
     """
     A representation of a CanTherm kinetics job. This job is used to compute 
     and save the high-pressure-limit kinetics information for a single reaction.
--- a/rmgpy/cantherm/kinetics.py
+++ b/rmgpy/cantherm/kinetics.py
@@ -32,6 +32,7 @@ import os.path
 import numpy
 import string
 import logging
+from sensitivity import KineticsSensitivity as sa
 
 from rmgpy.cantherm.output import prettify
 
@@ -57,7 +58,8 @@ class KineticsJob:
                  Tmin=None, 
                  Tmax=None,
                  Tlist=None,
-                 Tcount=0):
+                 Tcount=0,
+                 sensitivity_conditions=None):
         
         if Tmin is not None:
             self.Tmin = quantity.Quantity(Tmin)
@@ -90,6 +92,11 @@ class KineticsJob:
         
         self.reaction = reaction
         self.kunits = None
+
+        if sensitivity_conditions is not None:
+            self.sensitivity_conditions = [quantity.Quantity(condition) for condition in sensitivity_conditions]
+        else:
+            self.sensitivity_conditions = None
     
     @property
     def Tmin(self):
@@ -117,8 +124,7 @@ class KineticsJob:
         
     def execute(self, outputFile=None, plot=True):
         """
-        Execute the kinetics job, saving the results to the given `outputFile`
-        on disk.
+        Execute the kinetics job, saving the results to the given `outputFile` on disk.
         """
         if self.Tlist is not None:
             self.generateKinetics(self.Tlist.value_si)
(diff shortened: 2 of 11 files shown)


documentation/source/users/cantherm/input.rst
```diff
@@ -611,6 +611,14 @@ This is also acceptable::
 
     kinetics('H + C2H4 <=> C2H5')
 
+If a sensitivity analysis is desired, simply add the conditions at which to calculate sensitivity coefficients
```
goldmanm: it might be useful to add the syntax for sensitivity using pressureDependence method (in addition to the kinetics method used here). 

we might also want to describe what the output of sensitivity analysis looks like in the documentation (including the folder name, how the sensitivity coefficients are normalized, etc.)

rmgpy/cantherm/pdepTest.py
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env python
```
goldmanm: It seems like the actual operation being tested is inside the `setUp` method and not within any particular test itself. 

These also seem more like functional tests than unit tests, since you are running an entire module and checking for file output. Maybe this file should explicitly mention that it is a functional test, and it should be specified with the attribute `@attr('functional')` above the class line like in `testGenerateReactions.py`.

.gitignore
```diff
@@ -43,6 +43,7 @@ nbproject/*
 .coverage
 testing/*
 htmlcov/*
+rmgpy/test_data/Cantherm/tst1/*
```
goldmanm: It might be better to have a `tearDown` method that cleans these files instead of adding them to gitignore

rmgpy/cantherm/kinetics.py
```diff
@@ -43,6 +43,7 @@
 
 import rmgpy.quantity as quantity
 import rmgpy.constants as constants
+from rmgpy.molecule.draw import MoleculeDrawer, createNewSurface
```
goldmanm: I am a little confused by the functionality of this. Is drawing the PES optional? If it is, should we describe that in documentation? If not, how much time is consumed by this operation relative to calculating the TST equation? When processing many reactions, are separate PES files created or just one with everything drawn on it (which I have seen cantherm do before)?

alongd: It's fairly fast, so I included it by default (will not run if Cairo isn't installed, though).
I ran two `Kinetics` calcs including PES draw and SA for one of them in just under 10 seconds on my PC.
I just verified that it saves different plots for each PES.

alongd:
@goldmanm, thanks for your comments!
I addressed them with fixups (and dropped the gitignore one, so force-pushed), let me know if you have additional input for me.

rmgpy/cantherm/sensitivity.py
```diff
@@ -207,3 +209,176 @@ def plot(self):
...
+
+        for entry in wells + transition_states:
+            if entry in wells:
+                logging.info("\n\nPerturbing well '{0}' by {1}:".format(entry, self.perturbation))
+            else:
+                logging.info("\n\nPerturbing TS '{0}' by {1}:".format(entry.label, self.perturbation))
+            self.perturb(entry)
+            self.job.execute(outputFile=None, plot=False, print_summary=False)  # run the perturbed job
+            self.unperturb(entry)
+            if isinstance(entry, TransitionState):
```
mjohnson541: Is this done because E0 always needs to be positive to prevent negative values near zero?

alongd:
@cgrambow, thanks for your offline comments yesterday.

Since in the normalized sensitivity calculation (`dln(r)/dln(E0)`) I multiply by `E0` (which might be negative, or even zero...) I got wrong coefficients (both in sign and in magnitude).
I now modified the calculation to semi-normalized coefficients (`dln(r)/dE0`), like the thermo sensitivity in RMG.

Now the figures make more sense. Here's the example we discussed:
image
I get the following sensitivities with tunnelling:
```
  Sensitivity of network reaction reactant <=> product :
=====================================================================================
| Well or TS           | Temperature (K) | Pressure (bar) | Sensitivity coefficient |
|----------------------+-----------------+----------------+-------------------------|
| reactant             | 1000.0          |     1.00       | +1.46e-04               |
| product              | 1000.0          |     1.00       | -3.07e-06               |
| (TS) TS1             | 1000.0          |     1.00       | -1.43e-04               |
=====================================================================================
```
image

cgrambow:
Awesome! Those make a lot more sense.

rmgpy/cantherm/kinetics.py
```diff
@@ -295,3 +298,435 @@ def plot(self, outputDirectory):
...
+            thisWellWidth = max(wellWidth, rect[2])
+            E0 = 0.001 * well.E0
+            y = y_E0 - E0 * Eslope
+            coordinates[i] = [x + 0.5 * thisWellWidth, y]
+            x += thisWellWidth + wellSpacing
+        width = x + padding - wellSpacing
+
+        # Determine the rectangles taken up by each well
+        # We'll use this to merge columns safely so that wells don't overlap
+        wellRects = []
```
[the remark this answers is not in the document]
alongd: I believe so. Right, in complex networks we do have overlap and reading might be hard

rmgpy/cantherm/sensitivity.py
```diff
@@ -0,0 +1,209 @@
...
+        self.conditions = self.job.sensitivity_conditions
+        self.f_rates = [self.job.reaction.kinetics.getRateCoefficient(condition.value_si)
+                        for condition in self.conditions]
+        kr = self.job.reaction.generateReverseRateCoefficient()
+        self.r_rates = [kr.getRateCoefficient(condition.value_si) for condition in self.conditions]
+        self.f_sa_rates = {}
+        self.r_sa_rates = {}
+        self.f_sa_coefficients = {}
+        self.r_sa_coefficients = {}
+        self.perturbation = quantity.Quantity(1, 'kcal/mol')
```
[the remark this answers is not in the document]
alongd: I also think this shouldn't be hard-coded.
I changed the Class, so this is now an input parameter. We also iterate several times using `try - except` while capturing errors, and slightly decreasing the perturbation for better network stability

rmgpy/cantherm/sensitivity.py
alongd: I removed this part, so now we use semi-normalized coefficients, and E0 isn't needed

alongd:
@mjohnson541, thanks for your comments. I think I addressed everything, changes are already squashed.
If there are no more comments, I think this can get merged once the tests pass

PR #1387, opened 2018-06-02
Fix pruning for ranged reactors

mjohnson541:
This PR adds a pruning counter to fix issues with ranged reactors and pruning.  The pruning counter counts simulations up to max(nSimsTerm for each reactionSystem) * # of reaction systems (after this number of simulations the reactor with the largest nSimsTerm value can be assumed to have sampled the whole space).  When the counter is below this value pruning will not occur, once pruning happens the counter is reset at 0.

Should help solve #1387.


mjohnson541:
Found a better solution

Issue #1386, opened 2018-06-01
Excessive pruning with ranged reactors


mliu49:
I've run into an interesting issue when using the new ranged reactors feature. Essentially, RMG will eventually run the simulations at conditions where the previous chemistry becomes unimportant and then immediately prune almost the entire edge. In this case, the run completely successfully with no species or reactions in the edge.

Some logging statements towards the end of the run:
```
After model enlargement:
    The model core has 394 species and 21749 reactions
    The model edge has 2256 species and 3646 reactions
... (running simulation)
Pruning 1788 species whose rate ratios against characteristic rate did not exceed the minimum threshold of 0.0005
... (running simulation)
Pruning 468 species whose rate ratios against characteristic rate did not exceed the minimum threshold of 0.0005
... (running simulation)
MODEL GENERATION COMPLETED

The final model core has 394 species and 21749 reactions
The final model edge has 0 species and 0 reactions
```

mjohnson541:
So I think the best solution to this is to increase minSpeciesIterationsForPrune to be proportional to max(nSimsTerm for each reactor) * # reactors.  This can be done directly in the current input file or I could internally multiply the given value by those factors if we think that's more convenient.

mjohnson541:
I think I've worked out the problem.  Prior to this we had say six reactors we ran and then we checked maxEdgeSpeciesRateRatios variables on each reactor.  Now that attributes is being overwritten every time run a simulation, which means that regardless of how many cycles you wait in ranged reactors pruning is being done only for the last condition tested in the reactor.

mliu49:
Fixed by #1394.

PR #1388, opened 2018-06-02
Fix Ranged Reactor Pruning

mjohnson541:
Redefine an iteration for purposes of pruning as a full cycle of reactor runs.  

This calculates the number of reactor runs before pruning should be allowed for the species
as minSpeciesIterationsForPrune * max(nSimsTerm for each reactor) * # of reactors.  

This is an option for solution of #1386.

### DIFF
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -569,6 +569,12 @@ class RMG(util.Subject):
         self.nSimsTerms = [0 for i in xrange(len(self.reactionSystems))]
         
         self.rmg_memories = []
+        
+        maxNSimsTerm = 0
+        for rsys in self.reactionSystems:
+            if rsys.nSimsTerm > maxNSimsTerm:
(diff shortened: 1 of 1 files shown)


rmgpy/rmg/main.py
```diff
@@ -569,6 +569,12 @@ def execute(self, **kwargs):
         self.nSimsTerms = [0 for i in xrange(len(self.reactionSystems))]
         
         self.rmg_memories = []
+        
+        maxNSimsTerm = 0
+        for rsys in self.reactionSystems:
+            if rsys.nSimsTerm > maxNSimsTerm:
+                maxNSimsTerm = rsys.nSimsTerm
```
mliu49: Perhaps use `max(rsys.nSimsTerm for rsys in self.reactionSystems)`?

mjohnson541:
replaced by #1394

PR #1394, opened 2018-06-07
Ranged Pruning

mjohnson541:
This PR updates pruning to work properly with ranged reactors and solve #1386.  Instead of tracking how many times a ranged reactor is completed ranged reactors are simply run through nSimsTerm times each reactor cycle.  Furthermore, maxEdgeSpeciesRateRatios is now only reset for each reactor at the beginning of the reactor loop allowing ranged reactors to store the maxEdgeSpeciesRateRatios of all the runs they did that reactor cycle instead of just the last cycle.  This also solves the minimum number of iterations to pruning issues as the number of iterations is defined as the number of reactor cycles run.  I've also added an example for testing pruning and updated the documentation to reflect the changes to how ranged reactors operate.

### DIFF
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -826,14 +826,14 @@ class RMG(util.Subject):
                             self.reactionModel.thermoFilterDown(maximumEdgeSpecies=modelSettings.maximumEdgeSpecies)
                         
                         maxNumSpcsHit = len(self.reactionModel.core.species) >= modelSettings.maxNumSpecies
-    
-                        if maxNumSpcsHit: #breaks the while loop 
-                            break
                         
                         if not reactorDone:
                             self.done = False
                         
                         self.saveEverything()
+                        
+                        if maxNumSpcsHit: #breaks the while loop 
+                            break
                     
                 if not self.done: # There is something that needs exploring/enlarging
 
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -606,9 +606,8 @@ class RMG(util.Subject):
                 
         self.Tmax = max(Tlist)
         
-        self.nSimsTerms = [0 for i in xrange(len(self.reactionSystems))]
-        
         self.rmg_memories = []
+        
         # Initiate first reaction discovery step after adding all core species
         if self.filterReactions:
             # Run the reaction system to update threshold and react flags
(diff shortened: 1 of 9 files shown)


rwest:
Could we consider changing the name of the "nSimsTerm" input parameter, as we are changing it's meaning? I found the name not very clear intuitive or self-documenting when I first came across it. 
Maybe not worth it - just a suggestion.

mjohnson541:
I agree, it breaks input back compatibility, but the feature hasn't been around that long so its probably worth it.

mjohnson541:
Can we try to get this merged?  I'm starting to do stuff in other PR's that will create some really dangerous merge conflicts whenever this or the other one needs rebased.

mliu49:
The test case I was using finished, and this branch does much better than master at keeping a reasonable edge. I think it's still smaller than I would normally expect with standard reactors, but it's definitely better than the behavior on master.

rmgpy/rmg/main.py
```diff
@@ -801,9 +802,9 @@ def execute(self, **kwargs):
                                 logging.warn('Reaction thresholds/flags for Reaction System {0} was not updated due to resurrection'.format(index+1))
```
mliu49: Should this line also be indented? I think the arguments and closing parenthesis in the function calls above should also be indented.

rmgpy/rmg/main.py
```diff
@@ -817,27 +818,23 @@ def execute(self, **kwargs):
...
-                    self.saveEverything()
+                        maxNumSpcsHit = len(self.reactionModel.core.species) >= modelSettings.maxNumSpecies
+    
+                        if maxNumSpcsHit: #breaks the while loop 
+                            break
+                        
+                        if not reactorDone:
+                            self.done = False
+                        
+
```
mliu49: Could you only leave one empty line?

rmgpy/rmg/main.py
```diff
@@ -688,111 +687,113 @@ def execute(self, **kwargs):
                 
                 for index, reactionSystem in enumerate(self.reactionSystems):
                     
-                    reactionSystem.prunableSpecies = prunableSpecies
+                    reactionSystem.prunableSpecies = prunableSpecies   #these lines reset pruning for a new cycle
                     reactionSystem.prunableNetworks = prunableNetworks
+                    reactionSystem.reset_max_edge_species_rate_ratios()
```
mliu49: Why is this needed in the new setup?

mjohnson541: This resets the maxEdgeSpeciesRateRatios array to zero, this isn't done in simulate anymore since we need to accumulate the values across all ranged reactor simulations.

rmgpy/rmg/main.py
mjohnson541: done

rmgpy/rmg/main.py
mjohnson541: done

rmgpy/rmg/main.py
```diff
@@ -688,122 +687,124 @@ def execute(self, **kwargs):
...
                                 )
-                            else:
-                                self.updateReactionThresholdAndReactFlags(
+                                else:
+                                    self.updateReactionThresholdAndReactFlags(
                                         rxnSysUnimolecularThreshold = reactionSystem.unimolecularThreshold,
                                         rxnSysBimolecularThreshold = reactionSystem.bimolecularThreshold,
                                         rxnSysTrimolecularThreshold = reactionSystem.trimolecularThreshold,
                                         skipUpdate = True
                                 )
```
mliu49: Could you fix the indentation of this line and lines 791-794?

mliu49:
I think you can also consider adding #1407 to this PR since it's only one commit.

mjohnson541:
Done

mjohnson541:
Is this ready to merge?

PR #1401, opened 2018-06-12
New Rank Definitions

mjohnson541:
This makes some improvements to database write.  Redefines ranks to make them more specific identifiers of accuracy and changes the rank we give averaged kinetics accordingly.

### DIFF
--- a/documentation/source/users/rmg/kinetics.rst
+++ b/documentation/source/users/rmg/kinetics.rst
@@ -21,30 +21,58 @@ In the case where multiple rules or training set reactions fall under the same
 template node, we use a user-defined rank to determine the priority of kinetic
 parameters
 
-+-------+------------------------------------------------------+
-|Rank   |Example methods                                       |
-+=======+======================================================+
-|Rank 1 |Experiment                                            |
-+-------+------------------------------------------------------+
-|Rank 2 |High level calculation: CCSD(T)-F12, MS-Tor, 2d rotors|
-+-------+------------------------------------------------------+
-|Rank 3 |Mid level calculation:  CBS-QB3, 1-d rotors           |
-+-------+------------------------------------------------------+
-|Rank 4 |Low level calculation: no rotor treatment             |
-+-------+------------------------------------------------------+
-|Rank 5 |User's estimates without supporting methods           |
-+-------+------------------------------------------------------+
-|Rank 10|Averaged value from child nodes' rate rules           |
-+-------+------------------------------------------------------+
-|Rank 0 |Untrusted method and never used in generation         |
-+-------+------------------------------------------------------+
++-------+--------------------------------------------------------------------------------------+
+|Rank   |Example methods                                                                       |
++=======+======================================================================================+
+|Rank 1 |Experiment/FCI                                                                        |
++-------+--------------------------------------------------------------------------------------+
+|Rank 2 |W4/HEAT with very good (2-d if necessary) rotors                                      |
++-------+--------------------------------------------------------------------------------------+
+|Rank 3 |CCSD(T)-F12/cc-PVnZ with n>2 or CCSD(T)-F12/CBS with good (2-d if necessary) rotors   |
++-------+--------------------------------------------------------------------------------------+
+|Rank 4 |CCSD(T)-F12/DZ, with good (2-d if necessary) rotors                                   |
++-------+--------------------------------------------------------------------------------------+
+|Rank 5 |CBS-QB3 with 1-d rotors                                                               |
++-------+--------------------------------------------------------------------------------------+
+|Rank 6 |Double-hybrid DFT with 1-D rotors                                                     |
++-------+--------------------------------------------------------------------------------------+
+|Rank 7 |Hybrid DFT (w/ dispersion) (rotors if necessary)                                      |
++-------+--------------------------------------------------------------------------------------+
+|Rank 8 |B3LYP & lower DFT (rotors if necessary)                                               |
++-------+--------------------------------------------------------------------------------------+
+|Rank 9 |Group Additivity                                                                      |
++-------+--------------------------------------------------------------------------------------+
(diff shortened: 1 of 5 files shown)


documentation/source/users/rmg/kinetics.rst
```diff
@@ -24,26 +24,36 @@ parameters
 +-------+------------------------------------------------------+
 |Rank   |Example methods                                       |
 +=======+======================================================+
-|Rank 1 |Experiment                                            |
+|Rank 1 |Experiment/FCI                                        |
 +-------+------------------------------------------------------+
-|Rank 2 |High level calculation: CCSD(T)-F12, MS-Tor, 2d rotors|
+|Rank 2 |W4/HEAT with very good (2-d if necessary) rotors      |
```
alongd: Could we add a simple glossary or references to explain terms like  FCI / W4 / HEAT ?

documentation/source/users/rmg/kinetics.rst
```diff
@@ -24,26 +24,36 @@ parameters
...
+|Rank 10|Direct Estimate/Guess                                 |
++-------+------------------------------------------------------+
+|Rank 11|Average of Rates                                      |
++-------+------------------------------------------------------+
+|Rank 0 |General Estimate (Never used in generation)           |
 +-------+------------------------------------------------------+
 
 The rank of 0 is assigned to kinetics that are generally default values for top level nodes 
 that we have little faith in.  It is never used in generation and its value will in fact be overriden
 by averages of its child nodes, which generates an averaged rate rule with rank 10.
```
alongd: This is no longer updated (rank 10 for average)

alongd: Strange that both `0` and `11` are considered inaccurate...

documentation/source/users/rmg/kinetics.rst
mjohnson541: That would be a nice idea, I don't really feel qualified to do that myself though.

documentation/source/users/rmg/kinetics.rst
mjohnson541: fixed

documentation/source/users/rmg/kinetics.rst
alongd: @cgrambow, could you assist in writing a short explanation or adding links, so we know how to use the new ranks?

cgrambow: Here are some bullet points about the acronyms. Let me know if you want this in a different format or in more detail.
- FCI (full configuration interaction): Exact solution to Schrödinger equation within the chosen basis set and Born-Oppenheimer approximation; possible for about 12 electrons with reasonably sized basis set (cost grows factorially with number of electrons)
- Wn (Weizmann-n): Composite methods often with sub-kJ/mol accuracies; W1 is possible for about 9 heavy atoms; W1 aims to reproduce CCSD(T)/CBS; W4 aims to reproduce CCSDTQ5/CBS; https://aip.scitation.org/doi/full/10.1063/1.2348881
- HEAT (high accuracy extrapolated ab initio thermochemistry): Also sub-kJ/mol accuracies; essentially CCSDTQ with various corrections; similar in cost to Wn I think; https://aip.scitation.org/doi/10.1063/1.1811608; https://aip.scitation.org/doi/full/10.1063/1.2835612
- CBS (complete basis set): Typically obtained by extrapolating to the complete basis set limit, i.e., successive cc-pVDZ, cc-pVTZ, cc-pVQZ, etc. calculations with some extrapolation formula
- CCSD(T)-F12: Coupled cluster with explicit electron correlation; chemical accuracy (1 kcal/mol) possible with double-zeta basis sets; https://aip.scitation.org/doi/full/10.1063/1.3054300

Some notes:
- CCSD(T)-F12/D2 should be CCSD(T)-F12/DZ, where DZ stands for double-zeta and refers to any double-zeta basis set, e.g., cc-pVDZ, def2-SVP, 6-31G*, ...
- Write B3LYP instead of B3lyp
- Maybe write "rotors if necessary" for the DFT methods?

mjohnson541: Added

mjohnson541:
Ok, I've fixed those issues and a few @mliu49 found with the database PR.  Is this ready to merge?

alongd:
Yes, looking good!
Where will `CCSD(T)-F12a//cc-pVTZ-F12` (w/ good rotors) fall in the new rank? Is it `CCSD(T)-F12/DZ`?
(and what about `QZ`?)

cgrambow:
I would probably say Rank 3. Maybe we should change the description to be CCSD(T)-F12/nZ with n>2.

alongd:
perhaps n>= 2

cgrambow:
n=2 would still be rank 4

cgrambow:
If doing TZ calculations, it might be a good idea to always do the corresponding DZ calculation and to extrapolate to CBS from those two data points.

alongd:
@mjohnson541, could you change the description of rank 3 to `CCSD(T)-F12/cc-PVnZ with n>2, or CCSD(T)-F12/CBS with good (2-d if necessary) rotors`?

mjohnson541:
Done, I also added in some fixes related to reading and writing the database

mjohnson541:
Is this ready to merge?

alongd:
Yes. Just a practical question: all QM rankings specify `with good rotors` or `rotors if necessary`.
Were would a QM calculation at some arbitrary level w/o rotors fall in these ranks?

mjohnson541:
So the error is balanced between the energy and that of the A factor (rotors).  Ideally each rank should correspond to a certain amount of error.  If the rotors aren't important they don't need to be done even for the high level calculations.  However, if the rotor error is greater than the energy error we need to downgrade to a more suitable rank.  This is highly system dependent so I can't think of a better way than requiring the user to make the call as to where the calculation fits.

PR #1402, opened 2018-06-13
Save and parse Arkane species with all statmech properties using YAML format

alongd:
We'd like to be able to conveniently use previously calculated statmech data in Arkane (prev. Cantherm). The purpose is to have an online repository of such files, from which users can pick the desired species at the appropriate level of theory (if available). This PR sets the ground by allowing Arkane to save and parse such files.

### DIFF
--- a/arkane/common.py
+++ b/arkane/common.py
@@ -31,10 +31,227 @@
 import numpy
 import os.path
 import logging
+import time
+import string
 
+import yaml
+try:
+    from yaml import CDumper as Dumper, CLoader as Loader, CSafeLoader as SafeLoader
+except ImportError:
+    from yaml import Dumper, Loader, SafeLoader
+
+from rmgpy.rmgobject import RMGObject
+from rmgpy import __version__
+from rmgpy.quantity import ScalarQuantity, ArrayQuantity
+from rmgpy.molecule.element import elementList
+from rmgpy.molecule.translator import toInChI, toInChIKey
+from rmgpy.statmech.conformer import Conformer
+from rmgpy.statmech.rotation import LinearRotor, NonlinearRotor, KRotor, SphericalTopRotor
+from rmgpy.statmech.torsion import HinderedRotor, FreeRotor
+from rmgpy.statmech.translation import IdealGasTranslation
+from rmgpy.statmech.vibration import HarmonicOscillator
+from rmgpy.pdep.collision import SingleExponentialDown
+from rmgpy.transport import TransportData
+from rmgpy.thermo import NASA, Wilhoit
 import rmgpy.constants as constants
 
+from arkane.pdep import PressureDependenceJob
+
 ################################################################################
+
+
+class ArkaneSpecies(RMGObject):
+    """
+    A class for archiving an Arkane species including its statmech data into .yml files
+    """
+    def __init__(self, species=None, conformer=None, author='', level_of_theory='', model_chemistry='',
+                 frequency_scale_factor=None, use_hindered_rotors=None, use_bond_corrections=None, atom_energies='',
+                 chemkin_thermo_string='', smiles=None, adjacency_list=None, inchi=None, inchi_key=None, xyz=None,
+                 molecular_weight=None, symmetry_number=None, transport_data=None, energy_transfer_model=None,
+                 thermo=None, thermo_data=None, label=None, datetime=None, RMG_version=None):
+        if species is None and conformer is None:
+            # Expecting to get a `species` when generating the object within Arkane,
+            # or a `conformer` when parsing from YAML.
+            raise ValueError('No species or conformer was passed to the ArkaneSpecies object')
+        if conformer is not None:
+            self.conformer = conformer
+        if label is None and species is not None:
+            self.label = species.label
+        else:
+            self.label = label
+        self.author = author
+        self.level_of_theory = level_of_theory
+        self.model_chemistry = model_chemistry
+        self.frequency_scale_factor = frequency_scale_factor
+        self.use_hindered_rotors = use_hindered_rotors
+        self.use_bond_corrections = use_bond_corrections
+        self.atom_energies = atom_energies
+        self.chemkin_thermo_string = chemkin_thermo_string
+        self.smiles = smiles
+        self.adjacency_list = adjacency_list
+        self.inchi = inchi
+        self.inchi_key = inchi_key
+        self.xyz = xyz
+        self.molecular_weight = molecular_weight
+        self.symmetry_number = symmetry_number
+        self.transport_data = transport_data
+        self.energy_transfer_model = energy_transfer_model  # check pdep flag
+        self.thermo = thermo
+        self.thermo_data = thermo_data
+        if species is not None:
+            self.update_species_attributes(species)
+        self.RMG_version = RMG_version if RMG_version is not None else __version__
+        self.datetime = datetime if datetime is not None else time.strftime("%Y-%m-%d %H:%M")
+
+    def __repr__(self):
+        """
+        Return a string representation that can be used to reconstruct the object
+        """
+        result = '{0!r}'.format(self.__class__.__name__)
+        result += '{'
+        for key, value in self.as_dict().iteritems():
+            if key != 'class':
+                result += '{0!r}: {1!r}'.format(str(key), str(value))
+        result += '}'
+        return result
+
+    def update_species_attributes(self, species=None):
+        """
+        Update the object with a new species (while keeping non-species-dependent attributes unchanged)
+        """
+        if species is None:
+            raise ValueError('No species was passed to ArkaneSpecies')
+        self.label = species.label
+        if species.molecule is not None and len(species.molecule) > 0:
+            self.smiles = species.molecule[0].toSMILES()
+            self.adjacency_list = species.molecule[0].toAdjacencyList()
+            try:
+                inchi = toInChI(species.molecule[0], backend='try-all', aug_level=0)
+            except ValueError:
+                inchi = ''
+            try:
+                inchi_key = toInChIKey(species.molecule[0], backend='try-all', aug_level=0)
+            except ValueError:
+                inchi_key = ''
+            self.inchi = inchi
+            self.inchi_key = inchi_key
+            if species.conformer is not None:
+                self.conformer = species.conformer
+                self.xyz = self.update_xyz_string()
+            self.molecular_weight = species.molecularWeight
+            if species.symmetryNumber != -1:
+                self.symmetry_number = species.symmetryNumber
+            if species.transportData is not None:
+                self.transport_data = species.transportData  # called `collisionModel` in Arkane
+            if species.energyTransferModel is not None:
+                self.energy_transfer_model = species.energyTransferModel
+            if species.thermo is not None:
+                self.thermo = species.thermo.as_dict()
+                thermo_data = species.getThermoData()
+                h298 = thermo_data.getEnthalpy(298) / 4184.
+                s298 = thermo_data.getEntropy(298) / 4.184
+                cp = dict()
+                for t in [300,400,500,600,800,1000,1500,2000,2400]:
+                    temp_str = '{0} K'.format(t)
+                    cp[temp_str] = '{0:.2f}'.format(thermo_data.getHeatCapacity(t) / 4.184)
+                self.thermo_data = {'H298': '{0:.2f} kcal/mol'.format(h298),
+                                    'S298': '{0:.2f} cal/mol*K'.format(s298),
+                                    'Cp (cal/mol*K)': cp}
+
+    def update_xyz_string(self):
+        if self.conformer is not None and self.conformer.number is not None:
+            # generate the xyz-format string from the Conformer coordinates
+            xyz_string = '{0}\n{1}'.format(len(self.conformer.number.value_si), self.label)
+            for i, coorlist in enumerate(self.conformer.coordinates.value_si):
+                for element in elementList:
+                    if element.number == int(self.conformer.number.value_si[i]):
+                        element_symbol = element.symbol
+                        break
+                else:
+                    raise ValueError('Could not find element symbol corresponding to atom number {0}'.format(
+                        self.conformer.number.value_si[i]))
+                xyz_string += '\n{0} {1} {2} {3}'.format(element_symbol,
+                                                         coorlist[0],
+                                                         coorlist[1],
+                                                         coorlist[2])
+        else:
+            xyz_string = ''
+        return xyz_string
+
+    def save_yaml(self, path):
+        """
+        Save the species with all statMech data to a .yml file
+        """
+        if not os.path.exists(os.path.join(os.path.abspath(path),'ArkaneSpecies', '')):
+            os.mkdir(os.path.join(os.path.abspath(path),'ArkaneSpecies', ''))
+        valid_chars = "-_.()<=>+ %s%s" % (string.ascii_letters, string.digits)
+        filename = os.path.join('ArkaneSpecies',
+                                ''.join(c for c in self.label if c in valid_chars) + '.yml')
+        full_path = os.path.join(path, filename)
+        content = yaml.dump(data=self.as_dict(), Dumper=Dumper)
+        # remove empty lines from the file (multi-line strings have excess new line brakes for some reason):
+        content = content.replace('\n\n', '\n')
+        with open(full_path, 'w') as f:
+            f.write(content)
+        logging.debug('Dumping species {0} data as {1}'.format(self.label, filename))
+
+    def load_yaml(self, path, species, pdep=False):
+        """
+        Load the all statMech data from the .yml file in `path` into `species`
+        `pdep` is a boolean specifying whether or not jobList includes a pressureDependentJob.
+        """
+        logging.info('Loading statistical mechanics parameters for {0} from .yml file...'.format(species.label))
+        with open(path, 'r') as f:
+            data = yaml.safe_load(stream=f)
+        try:
+            if species.label != data['label']:
+                logging.warning('Found different labels for species: {0} in input file, and {1} in the .yml file. '
+                                'Using the label "{0}" for this species.'.format(species.label, data['label']))
+        except KeyError:
+            # Lacking label in the YAML file is strange, but accepted
+            logging.debug('Did not find label for species {0} in .yml file.'.format(species.label))
+        try:
+            class_name = data['class']
+        except KeyError:
+            raise KeyError("Can only make objects if the `class` attribute in the dictionary is known")
+        if class_name != 'ArkaneSpecies':
+            raise KeyError("Expected a ArkaneSpecies object, but got {0}".format(class_name))
+        del data['class']
+        class_dict = {'ScalarQuantity': ScalarQuantity,
+                      'ArrayQuantity': ArrayQuantity,
+                      'Conformer': Conformer,
+                      'LinearRotor': LinearRotor,
+                      'NonlinearRotor': NonlinearRotor,
+                      'KRotor': KRotor,
+                      'SphericalTopRotor': SphericalTopRotor,
+                      'HinderedRotor': HinderedRotor,
+                      'FreeRotor': FreeRotor,
+                      'IdealGasTranslation': IdealGasTranslation,
+                      'HarmonicOscillator': HarmonicOscillator,
+                      'TransportData': TransportData,
+                      'SingleExponentialDown': SingleExponentialDown,
+                      'Wilhoit': Wilhoit,
+                      'NASA': NASA,
+                      }
+        self.make_object(data=data, class_dict=class_dict)
+        if pdep and (self.transport_data is None or self.energy_transfer_model is None):
+            raise ValueError('Transport data and an energy transfer model must be given if pressure-dependent '
+                             'calculations are requested. Check file {0}'.format(path))
+        if pdep and self.smiles is None and self.adjacency_list is None\
+                and self.inchi is None and self.molecular_weight is None:
+            raise ValueError('The molecular weight was not specified, and a structure was not given so it could '
+                             'not be calculated. Specify either the molecular weight or structure if '
+                             'pressure-dependent calculations are requested. Check file {0}'.format(path))
+        logging.debug("Parsed all YAML objects")
+
+################################################################################
+
+
+def is_pdep(jobList):
+    for job in jobList:
+        if isinstance(job, PressureDependenceJob):
+            return True
+    return False
 
 
 def check_conformer_energy(Vlist,path):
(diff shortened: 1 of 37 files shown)


mliu49:
If this is going to be a brand new file type, I strongly suggest using an established format such as yaml, which you can parse using pyyaml to give a python dict. Using exec on user provided input files is a security risk since we could be executing arbitrary code. It would be best to convert all of our input files and database files to not require exec, but it'll take time.

Regarding the idea of having a database of statmech info, it would again be best if it's not a plain-text database like RMG-database is currently. Kehang's thermo central database uses mongoDB, which is an option, especially since it seems like it would overlap or be a part of this future statmech database. There are also tons of other options, but we should definitely put some thought into the database design.

alongd:
I changed the format into `.yml`.
Still, we need to use `eval()` (instead of `execute()`), because loading the YML file only gives dictionaries of strings, and the objects must be reconstructed.
I suggest we add more thorough checks to the constructor functions I added to common.py that the string we're evaluating is indeed what we think it is (contains several key words, under a certain maximum length, etc.). Other suggestions are welcomed!

rmgpy/cantherm/data/HSOO.yml
```diff
@@ -0,0 +1,43 @@
...
+  HinderedRotor(inertia=(4.80315,'amu*angstrom^2'), symmetry=1, fourier=([[0.305495,-0.899654,-0.432911,-0.0818798,-0.0102349],[0.000181646,0.000219011,0.000120536,1.95099e-05,-1.8673e-05]],'kJ/mol'))],
+  spinMultiplicity=2, opticalIsomers=1, mass=([31.9721,1.00783,15.9949,15.9949],'amu'),number=[16,1,8,8],
+  coordinates=([[-6.95052,3.36369,0.00013063],[-7.09087,5.91987,-0.000118519],[-3.50502,3.51303,0.000114712],[-2.45803,1.33214,0.000328714]],'angstroms'))
+datetime: 2018-06-15 11:32
+energy_transfer_model: SingleExponentialDown(alpha0=(0.5718,'kcal/mol'), T0=(300,'K'),
+  n=0.85)
+frequency_scale_factor: 0.975
+label: HSOO
+level_of_theory: CCSD(T)-F12a/cc-pVTZ-F12//B3LYP/6-311G(2d,d,p), rotors at B3LYP/6-311G(2d,pd)
+model_chemistry: CCSD(T)-F12/cc-pVTZ-F12
```
mliu49: I think the `level_of_theory` and `model_chemistry` may be a bit confusing/repetitive. This would require some refactoring of existing code, but you could consider renaming `model_chemistry` to `level_of_theory_energy`, then create two new attributes called `level_of_theory_geometry` and `level_of_theory_rotors`.

rmgpy/cantherm/input.py
```diff
@@ -124,8 +139,16 @@ def species(label, *args, **kwargs):
             else:
                 raise TypeError('species() got an unexpected keyword argument {0!r}.'.format(key))
             
-        if structure: spec.molecule = [structure]
+        if structure:
+            spec.molecule = [structure]
         spec.conformer = Conformer(E0=E0, modes=modes, spinMultiplicity=spinMultiplicity, opticalIsomers=opticalIsomers)
+        if molecularWeight is None:
+            if is_pdep():
```
mliu49: Does this work if the species block comes before the pdep block?

Also, if you refactor the molecular weight implementation based on my other comment, would that eliminate the need for this if block?

rmgpy/cantherm/input.py
```diff
@@ -388,27 +411,111 @@ def loadInputFile(path):
...
+            job.cantherm_species.species_dictionary['atom_energies'] = atom_energies
     
     return jobList
+
+
+def load_species_from_database_file(job):
+    """
+    Load the species with all statMech data from the .yml file
+    """
+    global jobList
```
mliu49: It doesn't seem like you use jobList in this function.

rmgpy/cantherm/input.py
```diff
@@ -388,27 +411,111 @@ def loadInputFile(path):
...
+            try:
+                mol = Molecule().fromInChI(data['InChI'])
+            except ValueError:
+                try:
+                    number = np.ndarray(shape=(len(job.species.conformer.number.value_si)))
+                    number[:] = [num for num in job.species.conformer.number.value_si]
+                    coordinates = np.ndarray(shape=(len(number), 3))
+                    for i, coorlist in enumerate(job.species.conformer.coordinates.value_si):
+                        coordinates[i] = [coor for coor in coorlist]
+                    mol = Molecule().fromXYZ(atomicNums=number, coordinates=coordinates)
```
mliu49: This section seems oddly coded. Was it done intentionally? It seems like the following would be sufficient:
```
number = job.species.conformer.number.value_si.tolist()
coordinates = job.species.conformer.coordinates.value_si.tolist()
```

rmgpy/cantherm/statmech.py
```diff
@@ -200,6 +206,12 @@ def load(self):
         each conformer and appends them to the list of conformers on the
         species object.
         """
+
+        filename, file_extension = os.path.splitext(self.path)
+        if 'yml' in file_extension or 'yaml' in file_extension:
+                self.load_species_from_database = True
+                return
```
mliu49: It seems like you could just call `load_species_from_database_file` here instead of `main.py`, and it would also eliminate the need for the `load_species_from_database` attribute.

rmgpy/cantherm/thermo.py
```diff
@@ -188,6 +188,30 @@ def save(self, outputFile):
                 f.write('\n')
         f.close()
 
+        return chem_string
+
+    def save_species_to_database_file(self, path):
```
mliu49: I think this may be more appropriate as a method of `CanthermSpecies`. Let me know what you think.

rmgpy/cantherm/common.py
```diff
@@ -43,6 +63,296 @@ def checkConformerEnergy(Vlist,path):
     Vlist = numpy.array(Vlist, numpy.float64)
     Vdiff = (Vlist[0] - numpy.min(Vlist))*constants.E_h*constants.Na/1000
     if Vdiff >= 2: #we choose 2 kJ/mol to be the critical energy
-        logging.warning('the species corresponding to ' + str(os.path.basename(path)) + ' is different in energy from the lowest energy conformer by ' + "%0.2f" % Vdiff + ' kJ/mol. This can cause significant errors in your computed rate constants. ')
+        logging.warning('the species corresponding to ' + str(os.path.basename(path)) +
+                        ' is different in energy from the lowest energy conformer by ' + "%0.2f" % Vdiff +
+                        ' kJ/mol. This can cause significant errors in your computed rate constants. ')
+
+
+def determine_molecular_weight(species):
```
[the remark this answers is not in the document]
alongd: Thanks for the comment!
I'll implement it in the Species `__init__()`:
```python
def __init__(self, ..., molecularWeight=None, ...):
        if self.molecularWeight is None:
            self._molecularWeight = quantity.Mass(self.molecule[0].getMolecularWeight(), 'kg/mol')
        else:
            self.molecularWeight = molecularWeight
```
Shall I still change the syntax of `molecularWeight`? I can leave it untouched so #1329 will have less conflicts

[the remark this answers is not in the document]
alongd: On second though, your recommendation seems better if for some reason an empty Species object is first created, then a structure is assigned.

rmgpy/cantherm/common.py
```diff
@@ -43,6 +63,296 @@ def checkConformerEnergy(Vlist,path):
...
+        if species.molecularWeight is not None:
+            d['molecularWeight'] = [species.molecularWeight.value, species.molecularWeight.units]  # decode
+        if species.energyTransferModel is not None:
+            d['energy_transfer_model'] = species.energyTransferModel
+
+        if species.thermo is not None:
+            d['thermo'] = species.thermo
+        d['chemkin_thermo_string'] = chemkin_thermo_string
+
+        self.species_dictionary = d
```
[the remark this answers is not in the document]
alongd: I agree.
I think I still need something like the update() method, though, to only append the species attributes to this object (the object is initiated with non-species-specific attributes such as `author` and `level_of_theory` which are kept)

rmgpy/cantherm/data/HSOO.yml
```diff
@@ -0,0 +1,43 @@
...
+molecularWeight: [64.96973070847088, amu]
+thermo: NASA(polynomials=[NASAPolynomial(coeffs=[3.80904,0.0201786,-7.63088e-05,1.6953e-07,-1.4166e-10,13862.2,10.1747],
+  Tmin=(10,'K'), Tmax=(363.844,'K')), NASAPolynomial(coeffs=[4.45296,0.00805632,-5.54113e-06,1.76698e-09,-2.12266e-13,13848.8,8.17796],
+  Tmin=(363.844,'K'), Tmax=(3000,'K'))], Tmin=(10,'K'), Tmax=(3000,'K'), E0=(115.254,'kJ/mol'),
+  Cp0=(33.2579,'J/(mol*K)'), CpInf=(78.9875,'J/(mol*K)'))
+transport_data: TransportData(epsilon=(3625.11,'J/mol'), sigma=(5.7,'angstrom'))
+use_atom_corrections: true
+use_bond_corrections: false
+use_hindered_rotors: true
```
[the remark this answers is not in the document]
alongd: At least `use_bond_corrections` is crucial, the rest are just "nice to have".

`use_bond_corrections`: This might impact the decision of the user of whether or not to use this file. If the user wants to use this species in a PES with other species calculated at different levels of theory, it is important/recommended to use bond corrections. On the other hand, if all species were calculated at the same level, it is better practice not to use bond corrections (which don't capture all bond types, and are often ambiguous for TSs).

`use_hindered_rotors`: This could also be directly inferred from the Conformer object. However, for a quick assessment, it's nice to have it explicitly written.

`use_atom_corrections`: Can be inferred from the `atom_energies` attribute, I'm OK with removing it since using this feature is less common.

rmgpy/cantherm/data/HSOO.yml
alongd: I agree that it's a bit confusing this way, but I'm not sure that `model_chemistry` and `level_of_theory_energy` are interchangeable. Since `model_chemistry` is being used by Cantherm to apply the respective bond and atom corrections, users could specify a "close enough" `model_chemistry` to the actual `level_of_theory_energy` they used. I'm not saying it's a good thing to do, but might happen. I want to report here both, so first we know what real `level_of_theory_energy` was used, but also which type of corrections were implemented.

Perhaps a good approach to solve this confusion would be to only dump `model_chemistry` into the YAML file if it's different than the energy level of theory.

rmgpy/cantherm/input.py
alongd: modified

rmgpy/cantherm/input.py
alongd: removed

rmgpy/cantherm/statmech.py
alongd: I agree, will change accordingly

rmgpy/cantherm/thermo.py
alongd: I agree

rmgpy/cantherm/input.py
mliu49: Oh, I misinterpreted it, I thought initially that the conformer attributes were `np.ndarray` objects, and you were converting them into lists.

If the conformer attributes are `np.ndarray`, it seems you could use them directly. If they're lists, then couldn't you use `np.array(list_input)` to convert it to an `np.ndarray`?

alongd:
@mliu49, Thanks for all the valuable comments!
I addressed them in fixup commits. At some point I didn't bother correctly naming the fixup commits due to the relocation of methods between files, I think I'll re-commit it all instead of squashing.
Let me know if you have any additional comments.

alongd:
@mliu49, can I squash this branch?

mliu49:
Yeah, you can go ahead and squash. I guess this might get replaced by the new method though, it that turns out ok.

rmgpy/rmg_obj.py
```diff
@@ -0,0 +1,52 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE #
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER      #
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING     #
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER         #
+# DEALINGS IN THE SOFTWARE.                                                   #
+#                                                                             #
+###############################################################################
+import numpy as np
+
+class RMG_Obj(object):
```
mliu49: A couple comments on naming:
- Class names should be CamelCase, though I think we can make an exception for RMG. I would probably call this `RMGObject`.
- Module names are ideally one word. Perhaps something like `object`, `base`, or `common`?
- Variable names should be lowercase. I think `D` should be renamed to something else, like `output` or even `output_dict`.

Also, could you end the file with a newline?

mjohnson541:
They're our own objects right?  So can't we just make them inherit from RMGObject?

rmgpy/rmgobject.py
```diff
@@ -0,0 +1,51 @@
...
+
+class RMGObject(object):
+    """
+    This class provides a general as_dict method to help with yml file construction
+    for other objects in rmgpy to inherit
+    """
+    def __init__(self):
+        pass
+    def as_dict(self):
+        D = self.__dict__.copy()
```
alongd: Thanks for this addition, @mjohnson541 !
I think that `self.__dict__` [contains methods as well](https://stackoverflow.com/questions/61517/python-dictionary-from-an-objects-fields), and we should first check `if not callable(value)` before adding entries to the output dictionary.

alongd:
I rewrote this branch using the `as_dict()` and `make_object()` methods we discussed for YAML loading and parsing.
Thanks @mjohnson541 and @mliu49 for the discussion and help with writing the methods.
The branch is now ready for another review cycle.

I'll note that since cytonized classes cannot inherit from non-cytonized classes, and vice versa, I made two parent classes: `RMGObject` which is cytonized, and `RMGPYObject` which isn't.
Quantity (Scalar & Array) and the thermo classes (NASA & Wilhoit) were not set to inherit from the parent objects. I think it's OK for quantity, but we could theoretically modify the way attributes are saved in the thermo classes so that they too can inherit from `RMGObject`. But I didn't think it was that important when I implemented it.

mliu49:
Are you sure that Python classes can't subclass Cython classes? It should be possible according to the [Cython documentation](https://media.readthedocs.org/pdf/cython/stable/cython.pdf) (section 3.5).

alongd:
Thanks for the guidance, @mliu49 .
Now only one RMGObject class (cythonized) is implemented.

alongd:
@mliu49, the MCH test has timed out in this PR. I restarted the `push` tests, and got the same behaviour. Could you take a look?

mliu49:
I wouldn't expect this branch to affect RMG jobs, so I'm not sure what would cause it to take so long. My only suggestion would be to run it locally to see what it does.

alongd:
It ran fine locally within a reasonable time (we had some rates violating the collision limit, but that should be meaningless for the test)
```Saving current model core to Chemkin file...
Chemkin file contains 323 reactions.
Saving current model to verbose Chemkin file...
Chemkin file contains 323 reactions.
Saving current model core and edge to Chemkin file...
Chemkin file contains 2197 reactions.
Saving current core and edge to verbose Chemkin file...
Chemkin file contains 2197 reactions.
Updating RMG execution statistics...
    Execution time (DD:HH:MM:SS): 00:00:06:11
    Memory used: 989.93 MB

Making seed mechanism...
Performing final model checks...


Warning: 4 CORE reactions violate the collision rate limit!
See the 'collision_rate_violators.log' for details.



MODEL GENERATION COMPLETED

The final model core has 81 species and 322 reactions
The final model edge has 654 species and 1872 reactions

RMG execution terminated at Thu Jul 12 22:30:49 2018
```

alongd:
@mliu49 and @mjohnson541, I think this branch is ready for another review cycle, getting us closer to a YAML RMG future

alongd:
@mliu49, I rebased the Arkane YAML file PR.
All comments from above were implemented. I think we should be good to go, unless you have additional comments. Please take a look whenever you got time. Thanks!

arkane/statmech.py
```diff
@@ -175,29 +176,41 @@ def __init__(self, species, path):
...
         species object.
         """
-        logging.info('Loading statistical mechanics parameters for {0}...'.format(self.species.label))
-        
         path = self.path
         TS = isinstance(self.species, TransitionState)
+        _, file_extension = os.path.splitext(path)
+        if file_extension in ['.yml', '.yaml']:
+            if TS:
+                raise ValueError('Loading transition states from a YAML file is still unsupported.')
```
mliu49: This could be a `NotImplementedError`.

rmgpy/rmgobject.pyx
```diff
@@ -0,0 +1,113 @@
+# cython: embedsignature=True, cdivision=True
```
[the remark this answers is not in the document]
alongd: Thanks for alerting me, it's a result of copy-pasting another pyx file as template. Removed

rmgpy/rmgobject.pyx
```diff
@@ -0,0 +1,113 @@
...
+            val = getattr(self, attr)
+            if val is not None and not callable(val) and val != '':
+                output_dict[attr] = val
+        for key, val in output_dict.iteritems():
+            if isinstance(val, list) and isinstance(val[0], RMGObject):
+                output_dict[key] = [v.as_dict() for v in val]
+            elif not isinstance(val, (int, float, str, dict)):
+                if isinstance(val, np.ndarray):
+                    output_dict[key] = val.tolist()
+                elif val is not None:
```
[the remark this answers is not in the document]
alongd: You're right, it cannot be `None`. But it doesn't have to be an RMGObject either: some of our classes have tailored `as_dict` and `make_object` methods and do not inherit from RMGObject (e.g., scalar and array quantities). I'll try removing the condition.

rmgpy/quantity.py
```diff
@@ -395,8 +462,12 @@ def getValue(self):
         The numeric value of the array quantity, in the given units.
         """
         return self.value_si * self.getConversionFactorFromSI()
+
     def setValue(self, v):
-        self.value_si = numpy.array(v) * self.getConversionFactorToSI()
+        if isinstance(v, float):
+            self.value_si = numpy.array([v]) * self.getConversionFactorToSI()
```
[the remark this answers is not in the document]
alongd: I'm not sure (it was written a while ago), but I think I had an instance like that when playing around with this branch. Maybe an ArrayQuantity with only one entry saved to YAML (such as a biatomic molecule's frequency) is read as a float? I'm also OK with removing this addition, and adding it back only if we encounter a related bug.

arkane/common.py
```diff
@@ -31,12 +31,233 @@
...
+        """
+        if not os.path.exists(os.path.join(os.path.abspath(path),'ArkaneSpecies', '')):
+            os.mkdir(os.path.join(os.path.abspath(path),'ArkaneSpecies', ''))
+        valid_chars = "-_.()<=>+ %s%s" % (string.ascii_letters, string.digits)
+        filename = os.path.join('ArkaneSpecies',
+                                ''.join(c for c in self.label if c in valid_chars) + '.yml')
+        full_path = os.path.join(path, filename)
+        with open(full_path, 'w') as f:
+            yaml.dump(data=self.as_dict(), stream=f, canonical=False)
+        # remove empty lines from the file (multi-line strings have excess new line brakes for some reason):
```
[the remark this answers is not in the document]
alongd: Thanks for the link above to the solution on stackoverflow re extra line breaks. The solution there uses a string representer, but we no longer use those since we don't `evaluate` any more. I'll implement your suggestion here.

alongd:
@mliu49, thanks for the helpful comments!
I addressed them all either in the comments above or as fixup commits.
Let me know if there's anything else you think we should change

alongd:
rebased. @mliu49, let me know when I can squash.

alongd:
@mliu49, let me know when to squash the fixups to finalize the PR

arkane/input.py
```diff
@@ -214,12 +214,11 @@ def species(label, *args, **kwargs):
...
                 # If a structure was given, simply calling spec.molecularWeight will calculate the molecular weight
-                spec.molecularWeight = spec.molecularWeight
-            elif is_pdep(jobList):
-                # If one of the jobs is pdep and no molecular weight is given or calculated, raise an error
-                raise ValueError("No molecularWeight was entered for species {0}. Since a structure wasn't given"
-                                 " as well, the molecularWeight, which is important for pressure dependent jobs,"
-                                 " cannot be reconstructed.".format(spec.label))
+                if spec.molecularWeight is None and is_pdep(jobList):
+                    # If one of the jobs is pdep and no molecular weight is given or calculated, raise an error
+                    raise ValueError("No molecularWeight was entered for species {0}. Since a structure wasn't given"
```
mliu49: Hmm, my suggestion might not have been correct, since this now misses the case where a structure is not provided. One possibility:
```
if molecularWeight is not None:
    spec.molecularWeight = molecularWeight
elif spec.molecularWeight is None and is_pdep(jobList):
    raise ValueError()
```
There shouldn't be any need to check whether structure exists, because the the Species.molecularWeight property checks if it has a molecule attribute, which will be empty if structure was empty.

rmgpy/rmgobject.pyx
```diff
@@ -1,10 +1,11 @@
-# cython: embedsignature=True, cdivision=True
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
```
mliu49: This actually isn't necessary either, since this is technically not a python file.

arkane/qchem.py
```diff
@@ -273,12 +273,13 @@ def loadEnergy(self, frequencyScaleFactor=1.):
         """
         E0 = None
         with open(self.path, 'r') as f:
-            for line in f:
+            lines = f.readlines()
+            for line in lines:
```
mliu49: The original implementation is preferable because it treats the file as an iterator, so only one line is loaded at a time. If you use readlines(), then the entire file has to be loaded into memory.

alongd: OK, I think this isn't related to this PR (sorry) so I removed it, we could discuss it at another time (I think it might be necessary for running Arkane through ARC). Thanks for bringing this up, I removed the commit.

Issue #1293, opened 2018-02-16
accurate degeneracy w/ resonance structures


goldmanm:
@alongd mentioned a case where degeneracy is not accurate in ReactionMechanismGenerator/RMG-database#248.

Currently, in the degeneracy algorithm, if all the molecules in a reaction are isomorphic but not identical, the reactions are considered degenerate. However, if a molecule has resonance structures which don't interact with the reacting site, this algorithm may still treat them as isomorphic but not identical and increase the degeneracy. 

To correct for this issue, we should only check that the atom ids of reacting sites are identical in the calculate degeneracy method instead of checking the entire molecule. 

to describe the weird comparison terms:

isomorphic = checks that the graph is the same, looking at bond order, isotope, atom type, etc.
identical = almost the same as isomorphic but also checks the atomID.

mliu49:
I hope you don't mind me fixing your link.

This is an interesting observation, but I don't fully understand your explanation. I would definitely like to look into this more.

alongd:
I'd like to add my understanding of the situation as well, as recently discussed with @goldmanm 

Our discussion primarily concerned the `[NH] + [O-][N+]=O <=> [NH][N+](=O)[O-]` birad-R recombination reaction as an example:
![image](https://user-images.githubusercontent.com/16158262/36362081-6248962a-14ff-11e8-8360-0c79436791c4.png) + ![image](https://user-images.githubusercontent.com/16158262/36362073-58a54258-14ff-11e8-9d57-9431ee4fb9b2.png) <=> ![image](https://user-images.githubusercontent.com/16158262/36362062-468a89c0-14ff-11e8-915f-74a0eaeb7008.png)

NO2 is described by two distinctive contributing Lewis structures, but if we account for isomorphic structures as well, we get four:

![image](https://user-images.githubusercontent.com/16158262/36325750-d74bed6c-1326-11e8-85ff-5e5c21613235.png)
![image](https://user-images.githubusercontent.com/16158262/36325755-db703f06-1326-11e8-845b-623bf6836ba4.png)
![image](https://user-images.githubusercontent.com/16158262/36325758-de863aa6-1326-11e8-8826-b5e3219998b0.png)
![image](https://user-images.githubusercontent.com/16158262/36325762-e07458ac-1326-11e8-89d1-479f6fac1ec2.png)

For the above reaction, only the first two are relevant.
[...]

PR #1410, opened 2018-06-27
Added a generate_N5dc_radical_resonance_structures resonance transition

alongd:
### Motivation or Problem
We're currently not describing the resonance path of
[image <-> image]

This resembles to the transition mentioned in #1293, mainly since this is also probably only important for correct degeneracy determination.

### Description of Changes
Added resonance transition, pathfinder modifications, and tests

### Testing
Compare resonance structures for `N[N+]([O-])=O` on master and this branch

(Edit: One of the drawings in the resonance transition above was wrong, now fixed)

### DIFF
--- a/rmgpy/molecule/resonance.py
+++ b/rmgpy/molecule/resonance.py
@@ -43,7 +43,8 @@ Currently supported resonance types:
     - ``generate_lone_pair_multiple_bond_resonance_structures``: multiple bond shift with lone pair
     - ``generate_lone_pair_radical_multiple_bond_resonance_structures``: multiple bond and radical shift with lone pair and radical
     - ``generate_N5ddc_N5tc_resonance_structures``: shift between nitrogen with two double bonds and single + triple bond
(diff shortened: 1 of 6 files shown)


rmgpy/molecule/resonance.py
```diff
@@ -43,7 +43,8 @@
     - ``generate_lone_pair_multiple_bond_resonance_structures``: multiple bond shift with lone pair
     - ``generate_lone_pair_radical_multiple_bond_resonance_structures``: multiple bond and radical shift with lone pair and radical
     - ``generate_N5ddc_N5tc_resonance_structures``: shift between nitrogen with two double bonds and single + triple bond
-    - ``generate_N5dc_resonance_structures``: shift between radical and lone pair mediated by an N5dc atom
+    - ``generate_N5dc_radical_resonance_structures``: shift between radical and lone pair mediated by an N5dc atom
+    - ``generate_N5dc_resonance_structures``:
```
mliu49: Missing description?

alongd: Done. Sorry for missing that!

Issue #1335, opened 2018-03-30
Updating quantities


mazeau:
Quantities stopped working and I got 
```
ValueError: ufunc <ufunc 'positive'> not supported by quantities
                please file a bug report at https://github.com/python-quantities
```

but uninstalled and reinstalled from pip

```
pip install --upgrade git+git://github.com/python-quantities/python-quantities.git@master
```

and now it works.

https://github.com/python-quantities/python-quantities/issues/128

Should we upgrade version in conda channel?

mliu49:
Sure, we can definitely update quantities in the RMG conda channel. @mazeau, are you running into this with current RMG-Py or other code?

mazeau:
on the cat branch I believe

mliu49:
I updated the version in the conda channel, but it seems to have broken one of the quantity tests. Will look into the cause.

mliu49:
Ok, a bug was introduced in quantities v0.12.0 (the version I added to our conda channel), and was fixed in v0.12.1. I'll test the newer version locally this time before updating the conda package.

mliu49:
I updated the quantities package in the RMG channel to v0.12.1 a few weeks ago, so this should be resolved.

PR #1416, opened 2018-06-29
Duplicate fixes

mliu49:
### Motivation or Problem
We were not properly marking duplicates for reaction libraries in the edge and irreversible reactions.

### Description of Changes
Add duplicate checking when loading reaction libraries and update `markDuplicateReactions` to account for irreversible reactions.

### Testing
Check that the edge mechanism from the new sulfur test in RMG-tests can be loaded in Chemkin.

It would probably also be good to check RMG-tests to see if this has any noticeable effect on performance.

### DIFF
--- a/rmgpy/chemkinTest.py
+++ b/rmgpy/chemkinTest.py
@@ -381,6 +381,51 @@ multiplicity 2
         self.assertTrue(isinstance(rtest.kinetics, MultiArrhenius))
         self.assertTrue(all(isinstance(k, Arrhenius) for k in rtest.kinetics.arrhenius))
 
+    def test_mark_duplicate_reactions(self):
+        """Test that we can properly mark duplicate reactions for Chemkin."""
+        s1 = Species().fromSMILES('CC')
+        s2 = Species().fromSMILES('[CH3]')
+        s3 = Species().fromSMILES('[OH]')
+        s4 = Species().fromSMILES('C[CH2]')
+        s5 = Species().fromSMILES('O')
+        s6 = Species().fromSMILES('[H]')
+
+        # Try initializing with duplicate=False
+        reaction_list = [
+            Reaction(reactants=[s1], products=[s2, s2], duplicate=False, kinetics=Arrhenius()),
+            Reaction(reactants=[s1], products=[s2, s2], duplicate=False, kinetics=Arrhenius()),
+            Reaction(reactants=[s1, s3], products=[s4, s5], duplicate=False, kinetics=Arrhenius()),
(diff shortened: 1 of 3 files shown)


alongd:
@mliu49, can you check why this fails?
The tests are too verbos, the raw log has many `ImportError`s and `AttributeError`s

mliu49:
It seems like new numpy packages were released, which somehow broke a lot of things. I restarted the build to see if it was a temporary incompatibility issue. If not, I'll look more into what the conflict is.

mliu49:
1. Anaconda uploaded new numpy packages (1.14)
2. Anaconda now splits the numpy package into numpy and numpy-base (no idea why and what the difference is)
3. We are currently restricted to numpy 1.11 (#1417), but could install numpy-base 1.14
4. The current quantities is incompatible with whatever modules are in numpy-base 1.14 (similar to #1335)
5. Updating quantities seems to fix the issue
6. However, we're still installing numpy 1.11 and numpy-base 1.14, which may or may not cause other issues

rmgpy/rmg/model.py
```diff
@@ -1591,6 +1591,14 @@ def addReactionLibraryToEdge(self, reactionLibrary):
                 # Don't add to the edge library reactions that were already processed
                 self.addReactionToEdge(rxn)
 
+        if self.saveEdgeSpecies:
+            from rmgpy.chemkin import markDuplicateReaction
+            newEdgeReactions = self.edge.reactions[numOldEdgeReactions:]
+            checkedReactions = self.core.reactions + self.edge.reactions[:numOldEdgeReactions]
+            for rxn in newEdgeReactions:
+                markDuplicateReaction(rxn, checkedReactions)
```
alongd: Could you remind me why we don't need to `checkedReactions()` when adding a seed library to core?

rmgpy/chemkinTest.py
```diff
@@ -381,6 +381,51 @@ def test_process_duplicate_reactions(self):
         self.assertTrue(isinstance(rtest.kinetics, MultiArrhenius))
         self.assertTrue(all(isinstance(k, Arrhenius) for k in rtest.kinetics.arrhenius))
 
+    def test_mark_duplicate_reactions(self):
```
alongd: Excellent test implementation

rmgpy/rmg/model.py
mliu49: In `addSeedMechanismToCore`, we call `self.markChemkinDuplicate` which calls `Chemkin.markDuplicateReactions` to process all reactions and mark duplicates. Here, we assume that the existing reactions have already been marked properly, so we just check the new reactions against the previously checked ones.

PR #1418, opened 2018-07-02
Added lone pair / multiple bond resonance pathway in a 3-atom system

alongd:
### Motivation or Problem
RMG should capture resonance transitions of lone pair and non-adjacent multiple bond shifts (similar to the allyl resonance pathway, only for lone pairs)

Example:

[image <-> image]

### Description of Changes
Added the resonance method, added the pathway, modified the resonance filtration function accordingly

### Testing
resonance, pathway, and filtration tests added

### Reviewer Tips
The tests are pretty comprehensive, but we should see what effect we get in RMG-tests particularly for N/S


alongd:
The Huckel-Lewis Configuration-Interaction method predicts the following relative resonance weights for the localized structures in the above example:
image

alongd:
@mliu49, This PR is now updated and rebased. I made some improvements to the resonance structure filtration algorithm as well.
Continuing our offline discussion with @rgillis8 and @mjohnson541 yesterday, I'm currently allowing dectet on sulfur to conform to our current GAV and families.
So DMSO for example is still represented in RMG as image, not as image

Thank you all for joining the discussion, it was important and very helpful.

rmgpy/molecule/pathfinder.py
```diff
@@ -243,28 +243,49 @@ def compute_atom_distance(atom_indices, mol):
...
+def find_lone_pair_multiple_bond_paths(atom1):
+    """
+    Find all the delocalization paths between lone electron pair and multiple bond in a 3-atom system
+    `atom1` indicates the delocalized lone pair site.
+    """
+    cython.declare(paths=list, atom2=Atom, atom3=Atom, bond12=Bond, bond23=Bond)
+
+    # No paths if atom1 has no lone pairs or cannot lose it
+    if atom1.lonePairs <= 0 or not (atom1.isCarbon() or is_atom_able_to_lose_lone_pair(atom1)):
+        return []
```
mliu49: Is the isCarbon check necessary? Or could it just be
`atom1.lonePairs <= 0 or not is_atom_able_to_lose_lone_pair(atom1)`?

rmgpy/molecule/resonance.py
```diff
@@ -315,6 +315,13 @@ def _generate_resonance_structures(mol_list, method_list, keep_isomorphic=False,
         # Move to the next resonance structure
         index += 1
 
+    # check net charge
+    for mol in mol_list:
+        if mol.getNetCharge() != 0:
+            raise ResonanceError('Resonance generation gave a net charged molecule:\n{0}'
+                                 'Ions are not yet supported in RMG.'.format(
+                mol.toAdjacencyList()))
```
mliu49: You have a newline before {0}, so it seems like this would print like:
```
Resonance generation gave a net charged molecule:
smilesstringIons are not yet supported in RMG.
```

Could you double-check? Also, could you indent the last line to match the prior line?

rmgpy/molecule/translatorTest.py
```diff
@@ -341,29 +335,28 @@ def test_singlet_vs_closed_shell(self):
     #         aug_inchi = 'InChI=1S/C6H5/c1-2-4-6-5-3-1/h1-5H/u1'
     #         self.compare(adjlist, aug_inchi)
 
-    def test_C5H6_triplet_singlet(self):
+    def test_C5H6_singlet(self):
         """
-        n-C5 chain with 2 unpaired electrons at the terminal carbon atoms,
-        and 2 carbon atoms with each a lone pair, next to a terminal
-        carbon atom.
+        n-C5 chain with 1 lone pair at the central carbon atom
```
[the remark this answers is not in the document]
alongd: Yes, I changed some tests here because they were testing non-representative structures, and when generating resonance structures in `compare()`, the desired ones got filtered out. I figured we might as well test the properties we're interested in using representative structures.

rmgpy/molecule/pathfinder.py
```diff
@@ -243,28 +243,49 @@ def compute_atom_distance(atom_indices, mol):
...
     
     # No paths if atom1 is not a radical
     if atom1.radicalElectrons <= 0:
         return []
 
-    # Find all delocalization paths
     paths = []
     for atom2, bond12 in atom1.edges.items():
         # Vinyl bond must be capable of gaining an order
-        if (bond12.isSingle() or bond12.isDouble()) and atom1.radicalElectrons:
```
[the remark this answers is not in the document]
alongd: Because it's being checked several lines above:
`    if atom1.radicalElectrons <= 0:   return []`
I don't mind keeping it here, perhaps for better code readability? Or perhaps keeping the condition here and removing the above `<= 0` check?

rmgpy/molecule/pathfinder.py
alongd: I agree, it doesn't look necessary

rmgpy/molecule/resonance.py
alongd: I thought printing an adjList automatically ends with a new line. Trying it our I get:
```
Resonance generation gave a net charged molecule:
1 S u0 p1 c0 {2,D} {3,D}
2 O u0 p2 c0 {1,D}
3 O u0 p2 c0 {1,D}
Ions are not yet supported in RMG.
```

rmgpy/rmg/model.py
```diff
@@ -1492,7 +1492,7 @@ def addSeedMechanismToCore(self, seedMechanism, react=False):
         for spec in self.newSpeciesList:            
             if spec.reactive:
                 submit(spec,self.solventName)
-
+                spec.generate_resonance_structures()
```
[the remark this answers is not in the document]
alongd: Tracing `submit` down in `thermoengine.py` and `scoop_framework\util.py`, I don't see any resonance structure generation call. I'd like to believe that back when I added it, it was out of necessity and that I saw non-representative structures entering the core through seed libraries (it was a while ago). So I think it doesn't call it, but I might be wrong.

alongd:
Thanks for the review, @mliu49 !
I think I addressed all comments. I also added some additional improvements to the structure filtration. Specifically, removing the N5t atomType, which in hind-sight doesn't seem physical (N val 5 with a triple bond has to be charged), helped simplifying the conditions in the filtration octet deviation check.
https://github.com/ReactionMechanismGenerator/RMG-database/pull/294 should be merged before this PR.

rmgpy/rmg/model.py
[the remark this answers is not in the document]
mliu49: Submit calls evaluator on each species, which calls generate_resonance_structures ([here](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/thermo/thermoengine.py#L173)).

rmgpy/molecule/pathfinder.py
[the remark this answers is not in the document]
mliu49: Ah, I didn't notice that. This is fine.

rmgpy/molecule/resonance.py
mliu49: You're right, for some reason I thought it was including the SMILES.

alongd:
@mliu49, I think this branch is clean and ready to be merged right after https://github.com/ReactionMechanismGenerator/RMG-database/pull/294. Thanks!

alongd:
Another minor change added to this PR: The octet deviation condition for on-the-fly structure filtration is now more permissive, allowing structure rearrangement. (For example, HSO3 radical isn't explored correctly without this fix when trans-octet structures aren't allowed.)

mliu49:
What was the reasoning for allow_trans_octet? I don't quite understand what trans refers to.

alongd:
It has been a while since we discussed it.
Principally, there are two approaches for representative Lewis structure drawing of third row elements: either you obey the octet rule as much as possible and often end up with formal charge separation, or you allow these "trans-octet" representations.
Here's an example for DMSO:
image, image
The S in the left structure has 1 lp and 3 S bonds, so 8 shared electrons
The S in the right structure has 1 lp, 1 D bond and 2 S bonds, so 10 shared electrons, hens a dectet, or trans-octet.
We should generally only adhere to one kind of representation to be consistent

mliu49:
Right, I remember that, I was wondering why you changed the naming from higher_octet to trans_octet?

alongd:
In the literature this state is referred to as "trans-octet" or "expanded octet" (https://pubs.acs.org/doi/abs/10.1021/ed069p791). Some commented that "expanded octet" isn't a correct name (https://pubs.acs.org/doi/abs/10.1021/ed063p28). I'm open to any recommendation of naming dectet and duodectet collectively.

mliu49:
Ah, ok. I've never heard of trans-octet, and a google search also doesn't return anything. I would be slightly more in favor of expanded octet, despite the second paper claiming that it's misnamed (without giving any reason).

rmgpy/molecule/pathfinder.py
```diff
@@ -319,22 +319,23 @@ def find_adj_lone_pair_radical_delocalization_paths(atom1):
...
-                        or (atom2.isOxygen() and atom2.lonePairs in [2, 3] and not atom1.isOxygen())  #avoid RO[::O.] <-> R[:O.+][:::O-], see RMG-Py #1223
-                        or (atom2.isSulfur() and atom2.lonePairs in [1, 2, 3])):
-                    paths.append([atom1, atom2])
+    if atom1.radicalElectrons >= 1 and\
+            ((atom1.isCarbon() and atom1.lonePairs == 0)
+             or (atom1.isNitrogen() and atom1.lonePairs in [0, 1])
+             or (atom1.isOxygen() and atom1.lonePairs in [1, 2])
+             or (atom1.isSulfur() and atom1.lonePairs in [0, 1, 2])):
+        for atom2 in atom1.edges.keys():
+            if ((atom2.isCarbon() and atom2.lonePairs == 1)
```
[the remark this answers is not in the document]
alongd: yes

rmgpy/molecule/filtration.py
```diff
@@ -210,40 +211,37 @@ def charge_filtration(filtered_list, charge_span_list):
...
+                if atom.radicalElectrons and int(atom.sortingLabel) not in rad_sorting_list:
+                    rad_sorting_list.append(int(atom.sortingLabel))
+                for atom2, bond in atom.bonds.iteritems():
+                    # check if bond is multiple, store only from one side (atom1 < atom2) for consistency
+                    if atom2.sortingLabel > atom.sortingLabel and bond.isDouble() or bond.isTriple():
+                        mul_bond_sorting_list.append((int(atom.sortingLabel), int(atom2.sortingLabel)))
+        # Find unique radical and multiple bond sites in charged_list and append to unique_charged_list:
+        unique_charged_list = []
+        for mol in charged_list:
+            unique_charged_list.extend(find_unique_sites_in_charged_list(mol, rad_sorting_list, mul_bond_sorting_list))
```
mliu49: The approach here seems a bit complex. Could you move the outer for loop into `find_unique_sites_in_charged_list` and have it directly return `unique_charged_list`? Then I think you would be able to use append instead of extend.

alongd: As discussed off-line, the code here is left as is since the helper function's `return` assists in breaking out of two `for` loops

alongd:
The tests look OK to me. Here are some key impressions:
- The nitrogen test has one species with a new resonance structure, the test treats it as a different species
- The sulfur test has three additional species in the test model, plus one species with different localized structures
- Looks like rates from `Singlet_Carbene_Intra_Disproportionation` aren't deterministic, as reflected in the liquid_oxidation and eg1 tests.
- I didn't understand the warning in the superminimal test `WARNING: Charges were rearranged`, looks like an RDKit or openbabel thing.

Other than the superminimal warning, everything looks good to go. @mliu49 Could you take a look?

PR #1423, opened 2018-07-13
Corrects a geometry reading error for QChem files

rgillis8:
While working with QChem for thermo calculations I notice that arkane(adjective) Arkane/Cantherm was not loading the atoms and geometries from a geometry log file.

This commit corrects that, at least for the case I was looking at, and raises an error if, after loading the geometry the atom vector remains empty.

### DIFF
--- a/rmgpy/cantherm/qchem.py
+++ b/rmgpy/cantherm/qchem.py
@@ -33,6 +33,7 @@ import numpy
 import logging
 import os.path
 import rmgpy.constants as constants
+from rmgpy.exceptions import InputError
 from rmgpy.cantherm.common import checkConformerEnergy
 from rmgpy.statmech import IdealGasTranslation, NonlinearRotor, LinearRotor, HarmonicOscillator, Conformer
 ################################################################################
@@ -115,35 +116,41 @@ class QchemLog:
         """
         atom = []; coord = []; number = []; 
 
-        f = open(self.path, 'r')
-        line = f.readline()
-        while line != '':
-            if 'Final energy is' in line:
-                print 'found a sucessfully completed Qchem Geometry Optimization Job'
-                line = f.readline()
-                atom = []; coord = []
+
+        with open(self.path) as f:
+            log = f.read().splitlines()
+
+        #First check that the Qchem job file (not necessarily a geometry optimization)
+        #has successfully completed, if not an error is thrown
+        completed_job = False
+        for line in reversed(log):
+            if 'Total job time:' in line:
+                logging.debug('Found a sucessfully completed Qchem Job')
+                completed_job = True
                 break
-            line = f.readline()
-        found = 0           
-        while line != '':        
+
+        if not completed_job:
+            raise InputError('Could not find a successfully completed Qchem job in Qchem output file {0}'.format(self.path))
+
+        #Now look for the geometry.
+        #Will return the final geometry in the file under Standard Nuclear Orientation.
+        geometry_flag = False
+        for i in reversed(xrange(len(log))):
+            line = log[i]
             if 'Standard Nuclear Orientation' in line:
-                found += 1
-                for i in range(3): line = f.readline() # skip  lines
-                while '----------------------------------------------------' not in line:
-                    data = line.split()
-                    atom.append((data[1]))
-                    coord.append([float(data[2]), float(data[3]), float(data[4])])
-                    line = f.readline()
-                # Read the next line in the file    
-                line = f.readline()
-            # Read the next line in the file
-            line = f.readline()
-            if found ==1: break
-        line = f.readline()
-        #print coord
-        f.close()
+                atom, coord, number = [], [], []
+                for line in log[(i+3):]:
+                    if '------------' not in line:
+                        data = line.split()
+                        atom.append(data[1])
+                        coord.append([float(c) for c in data [2:]])
+                        geometry_flag = True
+                    else:
+                        break
+                if geometry_flag:
+                    break
+
         coord = numpy.array(coord, numpy.float64)
-        mass = numpy.array(coord, numpy.float64)
         # Assign appropriate mass to each atom in molecule
         # These values were taken from "Atomic Weights and Isotopic Compositions" v3.0 (July 2010) from NIST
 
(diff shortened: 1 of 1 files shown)


rmgpy/cantherm/qchem.py
```diff
@@ -115,19 +115,30 @@ def loadGeometry(self):
...
+        #First check that the Qchem job file (not necessarily a geometry optimization)
+        #has successfully completed, if not an error is thrown
+        completedjob = 0;
         f = open(self.path, 'r')
         line = f.readline()
         while line != '':
-            if 'Final energy is' in line:
-                print 'found a sucessfully completed Qchem Geometry Optimization Job'
+            if 'Total job time:' in line:
+                print('Found a sucessfully completed Qchem Job')
```
alongd: I think that the `print` here should be `logging.debug` instead (although originally it was a print statement as well)

rmgpy/cantherm/qchem.py
```diff
@@ -115,19 +115,30 @@ def loadGeometry(self):
...
+                completedjob = 1
                 line = f.readline()
                 atom = []; coord = []
                 break
             line = f.readline()
-        found = 0           
+        f.close()
+        if completedjob != 1:
+            raise Exception('Could not find a successfully completed Qchem job')
```
alongd: Could you add `self.path` to the output message to make it more useful in case there are several QChem jobs in a single run?

Also, please use `InputError` instead of `Exception` (you'll have to import it: `from rmgpy.exceptions import InputError`). There are three other `Exception`s in this file, would you mind changing them as well?

rmgpy/cantherm/qchem.py
```diff
@@ -115,19 +115,30 @@ def loadGeometry(self):
...
+
+        #Restart looking for the geometry
+        #Will return the final geometry in the file under Standard Nuclear Orientation.
+        f = open(self.path, 'r')
+        line = f.readline()
         while line != '':        
             if 'Standard Nuclear Orientation' in line:
-                found += 1
+                atom = []; coord = []; number = [];
```
alongd: The `atom = []; coord = []` in L128 is no longer necessary.

rmgpy/cantherm/qchem.py
```diff
@@ -173,7 +181,9 @@ def loadGeometry(self):
                 number.append('17')
             else:
                 raise NotImplementedError('Atomic atom {0:d} not yet supported in loadGeometry().'.format(atom[i]))
-        number = numpy.array(number, numpy.int)       
+        number = numpy.array(number, numpy.int)
+        if number == [] or coord == [] or mass == []:
+            raise Exception('Unable to read the numbers and types of atoms from the geometry log file.')
```
alongd: Perhaps this should be a `ValueError`

rgillis8:
Round 2 corrections are completed. Let me know if you see anything else.

rmgpy/cantherm/qchem.py
```diff
@@ -115,19 +116,29 @@ def loadGeometry(self):
         """
         atom = []; coord = []; number = []; 
 
+        #First check that the Qchem job file (not necessarily a geometry optimization)
+        #has successfully completed, if not an error is thrown
+        completedjob = 0;
```
cgrambow: Please make this variable `True` or `False` instead of `0` or `1`

rmgpy/cantherm/qchem.py
```diff
@@ -115,19 +116,29 @@ def loadGeometry(self):
...
         f = open(self.path, 'r')
         line = f.readline()
         while line != '':
-            if 'Final energy is' in line:
-                print 'found a sucessfully completed Qchem Geometry Optimization Job'
+            if 'Total job time:' in line:
+                logging.debug('Found a sucessfully completed Qchem Job')
+                completedjob = 1
                 line = f.readline()
```
cgrambow: This line can be removed.

rmgpy/cantherm/qchem.py
```diff
@@ -138,9 +149,6 @@ def loadGeometry(self):
...
             # Read the next line in the file
             line = f.readline()
-            if found ==1: break
-        line = f.readline()
-        #print coord
         f.close()
         coord = numpy.array(coord, numpy.float64)
         mass = numpy.array(coord, numpy.float64)
```
cgrambow: This line can be removed. Very curious as to why it even exists.

rmgpy/cantherm/qchem.py
```diff
@@ -173,7 +181,9 @@ def loadGeometry(self):
                 number.append('17')
             else:
                 raise NotImplementedError('Atomic atom {0:d} not yet supported in loadGeometry().'.format(atom[i]))
-        number = numpy.array(number, numpy.int)       
+        number = numpy.array(number, numpy.int)
+        if number == [] or coord == [] or mass == []:
```
cgrambow: I suggest just checking `if not (number and coord and mass)` instead of `if number == []`, etc. since `if number` evaluates to `False` if `number` is an empty list.

rmgpy/cantherm/qchem.py
alongd: I agree this is strange. But needs to be double-checked since the function returns `mass` (and we check that it's not empty)

cgrambow: `mass` gets set in the lines after this one, so the `mass` in this line can be removed without issue.

rmgpy/cantherm/qchem.py
rgillis8: This is a little more complicated because if the arrays are not empty then it is ambiguous to ask their truth value (which a python error message was kind enough to tell me). I might could make it work by using a combination of any() calls, but then it seems like beyond what we're asking it.

cgrambow: Oh right, I forgot that happens with numpy arrays. Maybe we should instead check if `len(array) == 0`. Mostly I just brought this up because it seems a little weird to me to compare a list to an array.

rgillis8:
Round 3.

rgillis8:
Round 4.

rgillis8:
It worked for the example I had. I don't use Qchem much and so I didn't have many examples to try though...

cgrambow:
If it works for one of your examples without erroring, then it's probably fine for everything. I have run similar parsing code for many Q-Chem jobs.

cgrambow:
If it works for one of your examples without erroring, then it's probably fine for everything. I have run similar parsing code for many Q-Chem jobs. Just wanted to make sure that the final geometry in the log file is returned.

PR #1424, opened 2018-07-13
RMG-Py v2.2.0 release

mliu49:
This pull request is to the `stable` branch, and represents the official v2.2.0 release of RMG-Py.

Reviews are welcome for the most recent commits which are not already on master. Changes made include a number of fixes to the binary creation process and a few documentation fixes. It would be great if the following commits in particular could be double-checked:
- 8b739107078365eeece9a8f532ca4559ab590776: release notes for v2.2.0
- 5f173aaf4b84b06d914cd07ca2d4b7331758002b: new pdf documentation

This branch also brings automatic deployment! Travis builds on the `stable` branch will now automatically compile Linux and macOS binaries and upload them to Anaconda Cloud. Parts of the build script have been tested separately, but the final version will only be able to be tested once this branch is merged.

Important note: The RMG-database PR must be merged first in order for its binaries to be uploaded before the RMG-Py one is created, since the RMG-Py binary requires the equivalent version of RMG-database. I will do the merging once we think this is ready.

Side note: I'm planning to also re-implement automatic updating of the online documentation, but will do so on the master branch.


mliu49:
Another note: I just realized that I'll need to merge this manually in order to properly tag the release. I'll leave these PRs up for review though.

Issue #735, opened 2016-07-30
Round trip SMILES generation with Nitrogen broken


rwest:
Starting with this adjacency list

```
multiplicity 2
1 N u1 p0 c+1 {2,S} {3,S} {4,S}
2 O u0 p3 c-1 {1,S}
3 H u0 p0 c0 {1,S}
4 H u0 p0 c0 {1,S}
```

I generate the SMILES string `[NH2+][O-]`
But starting from that SMILES string I generate the adjacency list

```
1 N u0 p1 c+1 {2,S} {3,S} {4,S}
2 H u0 p0 c0 {1,S}
3 H u0 p0 c0 {1,S}
4 O u0 p3 c-1 {1,S}
```

which is invalid and causes the error `InvalidAdjacencyListError: Invalid valency for atom N (N3s) with 0 unpaired electrons, 1 pairs of electrons, 1 charge, and bonds [S,S,S].
`

alongd:
Nitrogen SMILES indeed require out attention.
Here is another example involving N val 5 where RMG + OB fail to recognize the radical site.
Starting with this adjacency list
```
multiplicity 2
1 O u0 p2 c0 {2,S} {4,S}
2 N u0 p1 c0 {1,S} {3,D}
3 O u0 p2 c0 {2,D}
4 N u1 p0 c0 {1,S} {5,T}
5 N u0 p1 c0 {4,T}
```
The generated SMILES string is `O(N=O)[N]#N`. Going back to AdjList from that SMILES string gives:

```
1 O u0 p2 c0 {2,S} {3,S}
2 N u0 p1 c0 {1,S} {4,D}
3 N u0 p0 c0 {1,S} {5,T}
4 O u0 p2 c0 {2,D}
5 N u0 p1 c0 {3,T}
```

(atom 3 here has a valance of 4, not 5 due to the missing rad)

alongd:
One idea is to develop augmented SMILES in RMG. This is even more important when it comes to sulfur.
This might look something like:
`O[:S](=O)[::O.]`, or
`O[S.](=O)=[::O]`
(where ':' denotes a lone pair, '.' denotes a radical)

alongd:
Just documenting an even stranger SMILES round trip problem:
Both of the following structures
![image](https://user-images.githubusercontent.com/16158262/42704530-2ae30cea-869e-11e8-875b-c3932537d68d.png)
[...]

PR #1425, opened 2018-07-13
Remove formulas with multiple structures from SMILES lookup dictionaries

cgrambow:
### Motivation or Problem
@alongd just noticed in #735 that two nitrogen/oxygen molecules are not exported correctly as SMILES when using RMG. The issue is that both of those molecules correspond to the same chemical formula, which is only associated with a single SMILES in `RADICAL_LOOKUPS`, which is different from both of the structures.

### Description of Changes
In addition to `NO2`, I have removed other potentially offending formulas. For example, `CH2O` can correspond to both formaldehyde and hydroxymethylene. I also removed `CO2` and `N2O`, which probably have some kind of alternate cyclic structures, although someone may want to verify that.

### Other Comments
Why are these lookups needed? SMILES generation is very fast, so can we just get rid of all of them?

### DIFF
--- a/rmgpy/molecule/translator.py
+++ b/rmgpy/molecule/translator.py
@@ -90,7 +90,6 @@ SMILES_LOOKUPS = {
 MOLECULE_LOOKUPS = {
     'N2': 'N#N',
     'CH4': 'C',
-    'CH2O': 'C=O',
     'H2O': 'O',
     'C2H6': 'CC',
     'H2': '[H][H]',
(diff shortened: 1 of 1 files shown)


alongd:
I think this is only important for structures for which we get wrong SMILES (so the recent observed behaviour is ironic...).
If this is true, then we can definitely remove species like CO2 (which aren't problematic). We should also add a comment to the code of the purpose of these lookup lists for future devs.
Also, if we implement it by fingerprinting, then by isIsomorphic if the fingerprint matches, we'll be certain of the species we represent.

mliu49:
I'm going to include this in the release.

alongd:
I agree re N2O and CO2:
image
image

mliu49:
Moving this comment to here since the issue was primarily about nitrogen SMILES.

History of the SMILES lookups:
- 0a33279c8c90d437fa5ca32a508e90a5f8ab33c8
- 3a1c5144af19b8df9676b222638dcd7b65f8d8f6

They should definitely only be included if there's only a single interpretation of the chemical formula. They were originally included for speed-up, though I agree that it seems a bit unnecessary.

alongd:
OK, is this good to go?

mliu49:
If you need this on master right away, then you can merge this PR. I'm going to cherry-pick it to v2.2.0 which will be merged into master after it's first merged into stable, which will likely happen over the weekend. Git should be able to figure it out though if you merge it here as well.

alongd:
No, not urgent

mliu49:
Incorporated into the v2.2.0 release and #1426.

PR #1433, opened 2018-07-20
Clear vertex mapping after isomorphism

### LINKED ISSUE #1344: Molecule.isIsomorphic() assigns atom.mapping, sometimes breaking symmetry
The only symptom for this I'm currently aware of is getting a different symmetryNumber for the same Molecule after running isIsomorphic().
I only saw this for a certain type of molecules (carbon rings with all double bonds).
Looking at image for example:
image
Just by running isIsomorphic the symmetryNumber is now wrong (should be 6).
The molecules still have the same object id, same for all their atoms. THe atoms also have the same Atom.id

mliu49:
### Motivation or Problem
Fixes #1344. Previously, the mapping attribute was causing the symmetry algorithm to return unexpected results. This actually affects normal jobs, since the symmetry number is generally determined at the end of thermo estimation, following many calls to isomorphism. As a result, the symmetry corrections were sometimes incorrect.

### Description of Changes
At the end of `vf2.isomorphism`, clear the `Vertex.mapping`  and `Vertex.terminal` attributes. These are used solely be the vf2 algorithm, so there is no need for them to persist.

### Testing
Check that the examples in #1344 now work as expected. Thermo estimates are expected to be different in RMG-tests, but should be limited to entropy values.

### DIFF
--- a/rmgpy/data/thermoTest.py
+++ b/rmgpy/data/thermoTest.py
@@ -424,7 +424,6 @@ class TestThermoAccuracy(unittest.TestCase):
                 self.assertAlmostEqual(Cp, thermoData.getHeatCapacity(T) / 4.184, places=1,
                                        msg="Cp{3} error for {0}. Expected {1} but calculated {2}.".format(smiles, Cp, thermoData.getHeatCapacity(T) / 4.184, T))
 
-    @work_in_progress
     def testSymmetryNumberGeneration(self):
         """
         Test we generate symmetry numbers correctly.
--- a/rmgpy/molecule/vf2.pyx
+++ b/rmgpy/molecule/vf2.pyx
@@ -166,6 +166,14 @@ cdef class VF2:
             graph1.restore_vertex_order()
             graph2.restore_vertex_order()
 
+        # We're done, so clear the mappings to prevent downstream effects
+        for vertex1 in graph1.vertices:
+            vertex1.mapping = None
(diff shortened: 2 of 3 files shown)


alongd:
Thanks @mliu49, the `C1=C=C=1` example in #1344 works as expected on this branch

mliu49:
Travis is failing because the work_in_progress symmetry tests are now passing, which is good. If we can confirm that the test values are correct, then I'll remove the work_in_progress decorator. @goldmanm @cgrambow

mliu49:
I looked through the cases in the symmetry test, and I think the values are correct, so I removed the work_in_progress decorator. I think this should be ready to merge once travis passes and we see how much of a difference this has on RMG-tests.

cgrambow:
Why were the work in progress tests failing and are now passing? We observed that we only get weird symmetry numbers if thermo is calculated in between, which is not done for the wip tests.

mliu49:
We do generate thermo for this [test](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/data/thermoTest.py#L428). It's an odd test for sure, but we generate the thermo before checking the symmetry number.

cgrambow:
Oh, I see. I was looking at the WIP tests in `symmetryTest.py`. I couldn't figure out from the Travis errors, which tests it was referring to.

PR #1438, opened 2018-07-25
Bug fix: Loading a species' thermo from the database in Cantherm

alongd:
### Motivation or Problem
Currently we attempt to load the thermo data for a species if no thermo was specified and the structure is given. However, we shouldn't do that in two cases:
- If the species has a path directing to a species.py file (or YAML soon).
- If the species is a bath gas in a pdep job, and the database isn't loaded

### Description of Changes
Cantherm now first checks for a species path before attempting to load it's thermo from the db. It will also output a warning if the database isn't loaded (and no thermo is given) instead of cruching.

### Testing
The acetyl+O2 network example crashes on master, and should complete on this branch

### DIFF
--- a/rmgpy/cantherm/input.py
+++ b/rmgpy/cantherm/input.py
@@ -37,12 +37,11 @@ import logging
 import numpy as np
 
 from rmgpy import settings
-from rmgpy.exceptions import InputError
+from rmgpy.exceptions import InputError, DatabaseError
 from rmgpy.data.rmg import RMGDatabase
 from rmgpy.data.rmg import getDB
 
 from rmgpy.rmg.model import CoreEdgeReactionModel
-from rmgpy.rmg.react import react
 
 from rmgpy.species import Species, TransitionState
 from rmgpy.quantity import Quantity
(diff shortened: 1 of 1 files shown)


alongd:
@mjohnson541, could you take a look?

rmgpy/cantherm/input.py
```diff
@@ -197,29 +198,36 @@ def species(label, *args, **kwargs):
...
-            if spec.molecule == []:
-                raise InputError('Thermo, E0 and structure not specified cannot estimate properties of species {0}'.format(spec.label))
-            logging.info('No E0 or thermo found, estimating thermo and E0 of species {0} using RMG-Database'.format(spec.label))
-            db = getDB('thermo')
-            spec.thermo = db.getThermoData(spec)
-            if spec.thermo.E0 is None:
-                th = spec.thermo.toWilhoit()
-                spec.conformer.E0 = th.E0
-                spec.thermo.E0 = th.E0
+        if spec.thermo is None and spec.conformer.E0 is None and path is not None:
```
mjohnson541: Shouldn't this be `and path is None:`?

alongd: You're right. Fixed and squashed.

alongd:
@mjohnson541, this should be ready to merge once tests pass

PR #1444, opened 2018-08-09
Chemkin flexibility

goldmanm:
### Motivation or Problem
RMG puts many more restrictions on chemkin output files than Chemkin actually does. It would be great to have more SMILES output.

### Description of Changes
This PR increases the flexibility of outputs to allow common SMILES characters like `=`, `[` and `]`, allowing RMG to output more smiles characters than before.

It fixes some of the parsing issues this creates.

### Testing
unittests run. travis tests. 

My biggest concern is character line limits in Chemkin reaction lines (currently 52 char) may be closer to being hit if RMG outputs `<=>` for reversible reactions instead of `=`. For third-body colliders this could cause an issue. 

Each species can maximally be 10 characters long. For bimolecular reactions this leads to 40 + ' <=> '(5 char) + 2' + '(6 char). This is 51 characters, which is still within the limit...though it doesn't include third body colliders, but these typically involve a unimolecular reaction, so it is probably not important...unless specific third body colliders with long names are able to be automatically generated. Since we're not planning on adding that functionality, we should be fine using `<=>` instead of `=`.

### Reviewer Tips
make sure the additions make sense and that

### DIFF
--- a/rmgpy/chemkin.pyx
+++ b/rmgpy/chemkin.pyx
@@ -1373,7 +1373,7 @@ def getSpeciesIdentifier(species):
     if species.index == -1:
         # No index present -- probably not in RMG job
         # In this case just return the label (if the right size)
-        if len(label) > 0 and not re.search(r'[^A-Za-z0-9\-_,\(\)\*#]+', label):
+        if len(label) > 0 and not re.search(r'[^A-Za-z0-9\-_,\(\)\*#=\[\]]+', label):
             if len(label) <= 10:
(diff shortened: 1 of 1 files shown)


goldmanm:
On chemkin version 15151, having an `=` in species name makes the program fail, in contrast with what is said in [this link](http://akrmys.com/public/chemkin/CKm_inp.html.en). 

I'll close this PR, since it does not work with chemkin.

PR #1445, opened 2018-08-09
Allow labeled atoms from generateReactions

goldmanm:
### Motivation or Problem
Generate reactions automatically removes atom labels, which may not be desired in all applications. Previously, people would have to take the output reaction and re-add atom labels, taking over twice as much work

### Description of Changes
Added option to prevent removing labeled atoms in generateReactions. When this is used, the reactant molecule object is copied since the atom labels might be different for each reaction in the list of reactions returned from generateReactions.

This also makes the default __repr__ output to be the full adjacency list when there are labeled atoms in the molecule. This allows proper loading and unloading of molecule objects.

### Testing
unittests and will do server rmg-tests

### Reviewer Tips
ensure code is well written

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1572,7 +1572,7 @@ class KineticsFamily(Database):
         else:
             raise NotImplementedError("Not expecting template of type {}".format(type(struct)))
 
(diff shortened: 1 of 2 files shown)


JacksonBurns:

PR #1449, opened 2018-08-15
Added S atom corrections at CCSD(T)-F12/aug-cc-pVnZ to Cantherm's statmech

alongd:
Also updated documentation

### DIFF
--- a/documentation/source/users/cantherm/input.rst
+++ b/documentation/source/users/cantherm/input.rst
@@ -91,6 +91,7 @@ Model Chemistry                                  AEC   BC   SOC  Freq Scale Supp
 ``'BMK/6-311G(2d,d,p)'``                          v    v    v               H, C, N, O, P, S
 ``'B3LYP/6-311+G(3df,2p)'``                       v    v    v    v (0.967)  H, C, N, O, P, S
 ``'B3LYP/6-31G**'``                               v    v         v (0.961)  H, C, O, S
+``'MRCI+Davidson/aug-cc-pV(T+d)Z'``               v         v               H, C, N, O, S
 ================================================ ===== ==== ==== ========== ====================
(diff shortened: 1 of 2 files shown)


alongd:
Thanks for the helpful comment.
I calculated using this S correction enthalpies for small molecules for which I found experimental data:

| Species  | Experimental H298 | Experimental uncertainty | H298 at CCSD(T)-F12a/aug-cc-pVTZ | Difference |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| SO  | 5.01  | 1.30 | 7.78 | 2.77 |
| SO2  | -296.81  | 0.20 | -288.99 | 10.59 |
| H2S | -20.60 | 0.50 | -21.32 | -0.72 |

(units are kJ/mol, data taken from CCCBDB)

It's hard to see the effect of all atom corrections in small species. The SO2 deviation is significant. Also, SO2 has unique bonds that probably were not fitted to in the O correction.

I can remove this commit of S atom energy correction, since we're uncertain of its behaviour.

I added unfitted `MRCI+Davidson/aug-cc-pV(T+d)Z` atom corrections, we could keep those.

cgrambow:
I suggest adding a comment specifying that S is unfitted and the others are not or removing S from those model chemistries. Perhaps someone else has opinions on this matter.

For `MRCI`, are all atoms `(T+d)` or just sulfur?

alongd:
I added a comment re S being unfitted for CCSD(T)-F12.
Re MRCI, I used `(T+d)` for all atoms, but I think it makes a difference only for S.
I added the molpro input file to the commit message to document it.

alongd:
If there are no further objections, I think this PR can be merged. Thanks @cgrambow for the review and suggestions

PR #1450, opened 2018-08-15
Adjust file paths for the collision violators log

mliu49:
### Motivation or Problem
The collision violators log file was being written to the execution directory, unlike the other output files.

### Description of Changes
Modify the file paths so that it is saved to the output directory insteady

### Testing
Run an RMG job from a different directory from the location of the input file, e.g.
```
python rmg.py examples/rmg/minimal/input.py
```
from the root RMG directory and confirm that the log is saved in `examples/rmg/minimal` instead of the root directory.

### DIFF
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -976,17 +976,19 @@ class RMG(util.Subject):
                 violators.extend(violator_list)
                 num_rxn_violators += 1
         # Whether or not violators were found, rename 'collision_rate_violators.log' if it exists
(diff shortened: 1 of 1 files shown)


alongd:
The code looks good. Did the minimal example have collision violators when you ran it? I didn't see the violators' output file

mliu49:
Ah, that's a good point. I wasn't testing with the minimal example, but with the ignition delay ipython demo. You could try using that as well, or adding FFCM as a seed mech to minimal, since I remember it having a couple of violations.

PR #1459, opened 2018-08-30
Reaction generation using multiprocessing

ajocher:
### Motivation or Problem
Scoop has been found to increase the memory consumption until the OS's out of memory killer kills processes and the reaction generation fails. Consequently, python's multiprocessing module is implemented for MAC and Linux and a deprecation warning is added to scoop references.

### Description of Changes
Multiprocessing is implemented for reaction generation in rmgpy/rmg/react.py. The processes are spawned and closed on a single node within the function 'react()'. The number of processes is determined based on the ratio of currently available RAM and currently used RAM. The user can input the maximum number of allowed processes from the command line. For each reaction generation the number of processes will be the minimum value either being the number of allowed processes due to user input or the value obtained by the RAM ratio. The RAM limitation is employed, because multiprocessing is forking the base process and the memory limit (SWAP + RAM) might be exceeded when using too many processes for a base process large in memory.

Multiprocessing is employed from the command line using the -n command and the maximum number of processes the user want to use, here 4. The default is 1 process, only an integer value smaller than the number of available cpu is allowed.
_python rmg -n 4 input.py_

### Testing
The implementation has been tested for the superminimal, minimal and PDD cases on both MAC and RMG-server, ranging from 1 to 24 processes on a single node. An example submission script for the RMG-server is attached here:
[submit.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/2336887/submit.txt)

### Reviewer Tips
Try your own cases and document the changes in run time and memory consumption.


rmg.py
```diff
@@ -82,6 +82,10 @@ def parse_command_line_arguments(command_line_args=None):
     parser.add_argument('-t', '--walltime', type=str, nargs=1, default='00:00:00:00',
                         metavar='DD:HH:MM:SS', help='set the maximum execution time')
 
+    # Add option to select max number of processes for reaction generation
+    parser.add_argument('-n', '--maxProc', type=int, nargs=1, default=1,
+                        help='used by multiprocessing in react.py')
```
mliu49: This may not be the most helpful info, since users may not know what react.py is. Perhaps `max number of processors to use during reaction generation` or something similar.

rmgpy/rmg/main.py
```diff
@@ -474,6 +479,15 @@ def initialize(self, **kwargs):
...
         self.wallTime = int(data[-1]) + 60 * int(data[-2]) + 3600 * int(data[-3]) + 86400 * int(data[-4])
 
+        global maxProc
+        try:
+            maxProc = kwargs['maxProc']
+        except KeyError:
+            pass
+
+        if maxProc > psutil.cpu_count():
+            raise ValueError('Invalid format for user defined maximum number of procesors {0}; should be an integer and smaller or equal to your available number of cpu {1}'.format(maxProc, psutil.cpu_count()))
```
mliu49: It doesn't seem like you're checking the format here?

rmgpy/rmg/model.py
```diff
@@ -674,7 +675,15 @@ def enlarge(self, newObject=None, reactEdge=False,
...
-            spcs = [self.retrieveNewSpecies(rxn) for rxn in rxns]
+            #spcs = [self.retrieveNewSpecies(rxn) for rxn in rxns]
+
+            # get new species and save in spcs
+            spcs = []
+            for rxn in rxns:
+                spcs.extend(rxn.reactants)
+                spcs.extend(rxn.products)
+
+            ensure_independent_atom_ids(spcs, resonance=True)
```
mliu49: Why do you need to ensure_independent_atom_ids here?

rmgpy/rmg/react.py
```diff
@@ -54,13 +55,29 @@ def react(*spcTuples):
...
+    from rmgpy.rmg.main import maxProc
+    
+    # Available RAM memory (kB)
+    mem = psutil.virtual_memory()
+    user_mem = mem.available
+    
+    tmp = divmod(user_mem, resource.getrusage(resource.RUSAGE_SELF).ru_maxrss)
+    tmp2 = min(maxProc, tmp[0])
+    procNum = max(1, tmp2)
+    print 'For reaction generation {0} processes are used.'.format(procNum)
```
mliu49: We like to avoid print statements in favor of using the logging module. So you would just replace `print` with `logging.info()` or maybe `logging.debug` depending on whether you want this to be printed normally.

Something you should consider is how many times this will be printed (seems like it will be a lot) and whether it's useful info for a typical user.

rmg.py
```diff
@@ -83,7 +83,7 @@ def parse_command_line_arguments(command_line_args=None):
                         metavar='DD:HH:MM:SS', help='set the maximum execution time')
 
     # Add option to select max number of processes for reaction generation
-    parser.add_argument('-n', '--maxProc', type=int, nargs=1, default=1,
+    parser.add_argument('-n', '--maxproc', type=int, nargs=1, default=1,
```
mliu49: It would be ideal if you could put this changes with the prior commit. I can show you how to do that if you'd like.

rmgpy/scoop_framework/util.py
```diff
@@ -160,6 +172,8 @@ def submit_(func, *args, **kwargs):
     returns the return value of the called function, or
     when SCOOP is loaded, the future object.
     """
+    warnings.warn("The option scoop is no longer supported"\
+     " and may be removed in Version: 2.3 ", DeprecationWarning)
```
mliu49: For all of these warnings, could you indent the second line to line up with the " in the first line? Also, the slashes shouldn't be necessary.

rmgpy/rmg/react.py
```diff
@@ -57,14 +59,21 @@ def react(*spcTuples):
...
     from rmgpy.rmg.main import maxproc
-    
-    # Available RAM memory (kB)
-    mem = psutil.virtual_memory()
-    usermem = mem.available
-    
-    tmp = divmod(usermem, resource.getrusage(resource.RUSAGE_SELF).ru_maxrss)
+   
+    # Get available RAM (GB)and procnum dependent on OS
+    if platform == "linux" or platform == "linux2":
```
mliu49: You could use `platform.startswith('linux')` here to check for both.

rmgpy/rmg/react.py
```diff
@@ -57,14 +59,21 @@ def react(*spcTuples):
...
+        # linux
+        memoryavailable = psutil.virtual_memory().free / (1000.0 ** 3)
+        memoryuse = psutil.Process(os.getpid()).memory_info()[0]/(1000.0 ** 3)
+    elif platform == "darwin":
+        # OS X
+        memoryavailable = psutil.virtual_memory().available/(1000.0 ** 3)
+        memoryuse = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss/(1000.0 ** 3) 
+
```
mliu49: I think you should add handling for Windows as well. It doesn't have to be equivalent handling, but you could simply add an `else` for any other platform and perhaps only use one processor in that case.

rmg.py
ajocher: Done.

rmgpy/rmg/main.py
ajocher: Changed to 'Invalid input ...'

rmgpy/rmg/model.py
ajocher: Otherwise Species.getResonanceHybrid() fails due to not having valid atom ids. We call ensure_independent_atom_ids in generate_reactions_from_families, however, this is not working for multiprocessing. Using ensure_independent_atom_ids here fixes that issue.

rmgpy/scoop_framework/util.py
ajocher: Done. Got the template from here:
https://github.com/ReactionMechanismGenerator/RMG-Py/wiki/RMG-Contributor-Guidelines
I guess it should be changed accordingly?

rmgpy/rmg/react.py
ajocher: Done. It will be printed for each time we are spawning a pool of worker. It might be of interest as the number of processes/worker might differ from the user input, depending on the available RAM and the memory consumption of the RMG base process before spawning the workers.

rmgpy/rmg/react.py
ajocher: Thanks! Done.

rmgpy/rmg/react.py
ajocher: Done, but couldn't test it. Is someone with a Windows OS volunteering?

ajocher:
[presentation1209.pdf](https://github.com/ReactionMechanismGenerator/RMG-Py/files/2372875/presentation1209.pdf)

mliu49:
Hmm, something seems to have gone wrong with your rebase. This branch now includes a ton of commits from master.

mliu49:
I added a commit refactoring the code for family splitting. Let me know what you think, and whether it matches your original design.

rmgpy/rmg/model.py
```diff
@@ -674,22 +693,81 @@ def enlarge(self, newObject=None, reactEdge=False,
...
+                        procnum = max(1, int(tmp2))
+                    else:
+                        # Everything else
+                        procnum = 1
+                
+                    # Execute multiprocessing map. It blocks until the result is ready.
+                    # This method chops the iterable into a number of chunks which it
+                    # submits to the process pool as separate tasks.
+                    p = Pool(processes=procnum)
+                    p.map(calculate_thermo_parallel,spcs)
```
mliu49: I have a couple concerns about the overall implementation for parallel QM. Normally, thermo is calculated for one species at a time in `self.processNewReactions()` below. Here, you determine all species which would be calculated using QM and pre-calculate their thermo in parallel, then in `processNewReactions`, it sees the thermo and does not re-calculate it.

1. The `calculate_thermo_parallel` function seems to skip a lot of processing done in `ThermoDatabase.getThermoData` and `thermoEngine.processThermoData`, which can have a significant effect on the final values.
2. It appears that you're using SMILES to determine whether species are unique? This is not safe because different resonance structures will have different SMILES. Also, generating SMILES for every new molecule will be relatively time consuming.

Minor suggestions, which may or may not be relevant after resolving the above issues:
- Since the procnum determination based on RAM is done both here and in react.py, it might be useful to turn it into a function somewhere and import it here.
- I would also split this section off as a new method, since `enlarge` is already a really big method.

ajocher: The thought behind the implemented procedure is that the QM files should be generated in parallel and then looked up for one species at a time in self.processNewReactions() below. Testing showed that calculating QM thermo for one species at a time is more time consuming than in a bulk up here. 

I use SMILES for comparison and then generate resonance structures within calculate_thermo_parallel() to generate QM files for each resonance structure. What would be a better way to ensure that the same QM file is not generated twice/overlapping during the parallel generation?

rmgpy/rmg/model.py
```diff
@@ -57,14 +62,39 @@
...
+    spc.generate_resonance_structures()
+    original_molecule = spc.molecule[0]
+
+    if quantumMechanics.settings.onlyCyclics and not original_molecule.isCyclic():
+        pass
+    else: 
+        if original_molecule.getRadicalCount() > quantumMechanics.settings.maxRadicalNumber:
+            pass
+        else: 
+            print 'Not too many radicals: do a direct QM calculation.'
```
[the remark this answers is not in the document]
ajocher: Logging, done.

rmgpy/rmg/model.py
```diff
@@ -569,17 +599,19 @@ def makeNewPDepReaction(self, forward):
...
-            
+
         # Set reaction index and increment the counter
         forward.index = self.reactionCounter + 1
         self.reactionCounter += 1
 
         return forward
 
+
+
```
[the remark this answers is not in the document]
ajocher: Done.

rmgpy/rmg/model.py
ajocher: The SMILES comparison is now substituted with isIsomorphic comparison. 

The procnum determination is now a function and imported for QMTP parallel and reaction generation in parallel.

And the parallel QM files writing section is split off outside of enlarge.

rmgpy/rmg/model.py
```diff
@@ -1171,11 +1171,12 @@ def thermoFilterSpecies(self, spcs):
         """
         Tmax = self.Tmax
         for spc in spcs:
-            G = spc.thermo.getFreeEnergy(Tmax)
-            if G > self.Gfmax:
-                Gn = (G-self.Gmax)/(self.Gmax-self.Gmin)
-                logging.info('Removing species {0} with Gibbs energy {1} from edge because it\'s Gibbs number {2} is greater than the toleranceThermoKeepSpeciesInEdge of {3} '.format(spc,G,Gn,self.toleranceThermoKeepSpeciesInEdge))
-                self.removeSpeciesFromEdge(self.reactionSystems,spc)
+            if spc.thermo:
```
mliu49: As we discussed briefly, I don't think this is an appropriate fix, since this means that we will simply never do thermo pruning because thermo is generated later. Instead, it might make sense to move when we do the thermo pruning.

rmgpy/rmg/react.py
```diff
@@ -32,14 +32,48 @@
...
+from sys import platform
 
 from rmgpy.data.rmg import getDB
-from rmgpy.scoop_framework.util import map_
+from multiprocessing import Pool
 
+################################################################################
 
-def react(*spcTuples):
+def determine_procnum_from_RAM():
```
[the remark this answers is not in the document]
ajocher: Moved to rmgpy.rmg.main

rmgpy/rmg/main.py
```diff
@@ -458,14 +458,25 @@ def initialize(self, **kwargs):
...
+        global maxproc
+        try:
+            maxproc = kwargs['maxproc']
+        except KeyError:
+            pass
+
+        if maxproc > psutil.cpu_count():
+            raise ValueError("""Invalid input for user defined maximum number of processes {0}; 
+            should be an integer and smaller or equal to your available number of 
+            processes {1}""".format(maxproc, psutil.cpu_count()))
```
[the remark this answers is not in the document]
ajocher: done

rmgpy/rmg/model.py
```diff
@@ -1953,28 +2008,25 @@ def getSpecies(self, obj):
...
         """
-        for obj in itertools.chain(deflatedRxn.reactants, deflatedRxn.products):
-            if isinstance(obj, int):
-                return self.getSpecies(obj)
-        raise Exception("No core species were found in either reactants or products of {0}!".format(deflatedRxn))
-
+        for obj in itertools.chain(rxn.reactants, rxn.products):
+            for spc in self.core.species:
+                if obj.isIsomorphic(spc):
+                    return spc
```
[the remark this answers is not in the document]
ajocher: done

rmgpy/rmg/react.py
```diff
@@ -124,14 +88,11 @@ def react_species(species_tuple, only_families=None):
     return reactions
 
 
-def react_all(core_spc_list, numOldCoreSpecies, unimolecularReact, bimolecularReact, trimolecularReact=None):
+def react_all(core_spc_list, numOldCoreSpecies, unimolecularReact, bimolecularReact, procnum, trimolecularReact=None):
```
mliu49: Maybe add a default value of 1 for procnum?

rmgpy/rmg/model.py
```diff
@@ -687,6 +687,13 @@ def enlarge(self, newObject=None, reactEdge=False,
             logging.info('Generating thermo for new species...')
             self.applyThermoToSpecies(procnum)
 
+        # Do thermodynamic filtering
+        try:
+            if not numpy.isinf(self.toleranceThermoKeepSpeciesInEdge) and spcs != []:
+                self.thermoFilterSpecies(spcs)
+        except:
```
mliu49: What is the exception that this is expecting?

rmgpy/rmg/model.py
```diff
@@ -687,6 +687,13 @@ def enlarge(self, newObject=None, reactEdge=False,
             logging.info('Generating thermo for new species...')
             self.applyThermoToSpecies(procnum)
 
+        # Do thermodynamic filtering
+        try:
+            if not numpy.isinf(self.toleranceThermoKeepSpeciesInEdge) and spcs != []:
```
mliu49: I think `spcs` is not properly defined here - replacing it with `self.newSpeciesList` should give the intended behavior.

rmgpy/rmg/model.py
```diff
@@ -872,9 +879,6 @@ def processNewReactions(self, newReactions, newSpecies, pdepNetwork=None, genera
                     if rxn in self.edge.reactions:
                         self.edge.reactions.remove(rxn)
 
-            if not numpy.isinf(self.toleranceThermoKeepSpeciesInEdge) and spcs != []: #do thermodynamic filtering
```
mliu49: You can also remove the creation of the `spcs` variable above.

rmgpy/rmg/react.py
ajocher: done

rmgpy/rmg/model.py
ajocher: correctly defining self.newSpeciesList allows the removal of try and except, thanks!

rmgpy/rmg/model.py
ajocher: done

rmgpy/rmg/model.py
ajocher: done

rmgpy/rmg/react.py
```diff
@@ -53,118 +54,98 @@ def react(*spcTuples):
...
-    for rxn in rxns:
-        deflateReaction(rxn, molDict)
-        try:
-            deflateReaction(rxn.reverse, molDict) 
-        except AttributeError:
-            pass
-
-
-def reactAll(coreSpcList, numOldCoreSpecies, unimolecularReact, bimolecularReact, trimolecularReact=None):
+def react_all(core_spc_list, numOldCoreSpecies, unimolecularReact, bimolecularReact, procnum=1, trimolecularReact=None):
```
mliu49: I didn't notice this before, but I think it would be better to move `procnum` to be the last argument.

ajocher: Added fixup to "Moving determine_procnum_from_RAM() to rmgpy.rmg.main."

rmgpy/rmg/reactTest.py
```diff
@@ -100,7 +101,7 @@ def testReactAll(self):
                 ]
 
         N = len(spcs)
-        rxns = react_all(spcs, N, np.ones(N), np.ones([N, N]), np.ones([N, N, N]))
+        rxns = react_all(spcs, N, np.ones(N), np.ones([N, N]), 2, np.ones([N, N, N]))
```
mliu49: This also needs to be updated following the change in argument position.

ajocher: Done, but still haven't found a solution for the error in liquidTest.py due to "fixing" the thermo labeling.

mliu49:
I changed the species re-labeling to only be done for new species created from reaction generation. Turns out that thermo generation for input species was happening earlier than we thought.

I also removed the whitespace changes to clean up the PR a bit.

mliu49:
I think this is ready code-wise. Are there any additional tests that you think should be run?

ajocher:
> I think this is ready code-wise. Are there any additional tests that you think should be run?

Just re-tested, QMTP, pdep, and thermo filtering in serial and parallel and seems to work fine. Ready to go from my side.

alongd:
Should we update the documentation re parallelization?
http://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/running.html?highlight=parallel

ajocher:
It is updated. Where do you think it needs more update?

PR #1462, opened 2018-09-01
Move Forbidden Structure Checking from Edge to Core (Edgeless RMG Part 1)

mjohnson541:
This moves global forbidden structure checking from when species are created in the edge to when a species attempts to enter the core. In a test constructing a dodecane combustion model this reduced time spent checking forbidden structures by 148x and overall runtime (in the react regime) by 4.5x without impacting the output model.

### DIFF
--- a/rmgpy/rmg/model.py
+++ b/rmgpy/rmg/model.py
@@ -53,6 +53,9 @@ from rmgpy.data.kinetics.family import KineticsFamily, TemplateReaction
 from rmgpy.data.kinetics.library import KineticsLibrary, LibraryReaction
 
 from rmgpy.kinetics import KineticsData, Arrhenius
+
+from rmgpy.data.rmg import getDB
+        
 import rmgpy.data.rmg
 from .react import reactAll
 
@@ -1052,6 +1055,28 @@ class CoreEdgeReactionModel:
 
         assert spec not in self.core.species, "Tried to add species {0} to core, but it's already there".format(spec.label)
 
+        forbidden_structures = getDB('forbidden')
+        
+        # check RMG globally forbidden structures
+        if forbidden_structures.isMoleculeForbidden(spec.molecule[0]):
+            
+            rxnList = []
+            if spec in self.edge.species:
+
+                #remove forbidden species from edge
+                logging.info("Species {0} was Forbidden and not added to Core...Removing from Edge.".format(spec))
+                self.edge.species.remove(spec)
+                # Search edge for reactions that contain forbidden species
+                for rxn in self.edge.reactions:
+                    if spec in rxn.reactants or spec in rxn.products:                        
+                        rxnList.append(rxn)
+                
+                #Remove any reactions that are globally forbidden from Edge
+                for rxn in rxnList:
+                    self.edge.reactions.remove(rxn)
+                    logging.info("Removing Forbidden Reaction from Edge: {0}".format(rxn))
+                return []
+        
         # Add the species to the core
         self.core.species.append(spec)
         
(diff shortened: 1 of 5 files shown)


rmgpy/rmg/model.py
```diff
@@ -1052,6 +1055,29 @@ def addSpeciesToCore(self, spec):
...
+                # If species was in edge, remove it
+                logging.info("Species {0} was Forbidden and not added to Core...Removing from Edge.".format(spec))
+                self.edge.species.remove(spec)
+                # Search edge for reactions that now contain only core species;
+                # these belong in the model core and will be moved there
+                for rxn in self.edge.reactions:
+                    if spec in rxn.reactants or spec in rxn.products:                        
+                        rxnList.append(rxn)
+                
+                # Move any identified reactions to the core
```
mliu49: This comment isn't correct anymore.

mjohnson541: fixed

mjohnson541:
Those are the two RMG-tests that take multiple species at a time.  Under those conditions this PR can significantly change model trajectories, which changes that amount of work that needs done during model generation.  This is certainly true for the liquid oxidation test case.

mliu49:
I tested with naphthalene and acetylene pyrolysis, terminating at 80 species. The stats are below. Overall speedup was approximately 7%, with slightly less reduction for isMoleculeForbidden . However, the models were not the same; each had 4 core species which were not in the other model. This run only took one species at a time, so I'm not sure what caused the divergence.

To sum up, I don't think this change provides as much benefit as we expected, but it also doesn't do any harm. I would be ok with merging it.

|                          | master   | checkForbidCore |
|--------------------------|----------|-----------------|
| core spcs                | 81       | 81              |
| core rxns                | 147      | 144             |
| edge spcs                | 5063     | 4885            |
| edge rxns                | 13497    | 12538           |
| total time               | 14:09:00 | 13:28:35        |
| isMoleculeForbidden time | 7518     | 7011            |

mjohnson541:
I've run the dodecane test again.  The model size achieved is drastically different (about twice the original size). 
For both models:  
The final model core has 306 species and 4971 reactions
The final model edge has 25571 species and 339901 reactions

For the original model isMoleculeForbidden was the largest tottime category at 11290 sec out of 98485. sec total.  For the model on this branch isMoleculeForbidden took 1216 sec out of 87555 sec total.  

This represents about an order of magnitude reduction in forbidden structure checking time (not exactly sure why it isn't closer to the ratio of the edge to the core of 83.5) eliminating it as the function taking up the most time and about a 12% overall speed up.  

It seems we've simply reduced the amount of time we spend on forbidden structure checking since I tested this last, but, as this converts forbidden structure checking from the largest runtime sink to a relatively minor runtime sink I think this is a good start for the Edgeless RMG project, as a part of which I intend to eliminate or severely reduce the runtime contributions from the next four contributors:  __generateReactions, ensure_species, simulate and applyRecipe.

mliu49:
I reran the tests on the correct branches, but these results are still not very helpful. This branch gave fewer core reactions, but many more edge species and reactions. The two models only have 50 species in common, which means the change to forbidden structure checking somehow caused the models to diverge onto very different paths. As a result, despite the reduction in forbidden structure time, the overall model generation time was longer.

Do you have any thoughts on why they would have diverged? The database branch is the same for both.

|                          | master   | checkForbidCore |
|--------------------------|----------|-----------------|
| core spcs | 81 | 81 |
| core rxns | 132 | 122 |
| edge spcs | 4802 | 5103 |
| edge rxns | 13887 | 12417 |
| total time | 9:47:52 | 14:32:28 |
| isMoleculeForbidden time | 3946 | 1656 |
| isMoleculeForbidden calls | 24.7 million | 19.5 million |
| applyRecipe time | 6171 | 13200 |
| applyRecipe calls | 4.1 million | 7.7 million |

mjohnson541:
Are we ready to merge this after your most recent test?

mliu49:
Yeah, you can go ahead and rebase.

mjohnson541:
Rebased, ready to merge.

PR #1463, opened 2018-09-02
Allow setting of the frequency at which edge fluxes are checked (Edgeless RMG Part 2)

mjohnson541:
Currently when RMG simulates a system at every time step it calculates both the core and edge fluxes.  This PR seperates these calculations and allows the user to specify the frequency (implicitly the number of time steps between each calculation of edge fluxes) at which the edge fluxes are calculated.  Since the edge is so much larger than the core this in theory significantly reduces simulation time.  I've measured slight but measurable improvements in small jobs such as the minimal example and ethane combustion example and on an extended minimal example I achieved between a factor of 2 and factor of 3 reduction in simulation time.  The option this provides is of course primarily useful when model construction is in the simulate regime.  

(This PR is independent of Edgeless RMG Part 1).

### DIFF
--- a/documentation/source/users/rmg/input.rst
+++ b/documentation/source/users/rmg/input.rst
@@ -580,6 +580,26 @@ For example ::
 		toleranceMoveSurfaceReactionToCore=5.0,
 	)
 
+Advanced Setting: Reducing Edge Checking
+-----------------------------------------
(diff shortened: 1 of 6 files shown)


JacksonBurns:

PR #1465, opened 2018-09-05
BugFix: Providing an exception to raise in Cantherm's pdep.py

### LINKED ISSUE #1464: Call to `raise` without exception in cantherm.pdep
### Bug Description
There is a call to `raise` in `cantherm.pdep` ([here](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/cantherm/pdep.py#L270)) which does not have an exception provided. PyCharm flagged this as an error with red squiggles.

alongd:
Fixes #1464


alongd:
I'm also fine with either way.
In the current implementation, though, it should be clearer to the user whether the raised error is InvalidMicrocanonicalRateError or ModifiedStrongCollisionError

PR #1466, opened 2018-09-06
allow parsing of Trees with molecule objects

goldmanm:
### Motivation or Problem
previously, the decendTree method would treat molecule objects in
trees as group objects, throwing an error. Though it is entirely possible
to have molecule objects decend through the tree

### Description of Changes
 This fix uses isomorphism
of molecule objects instead of using subgraph isomorphism


mjohnson541:
I remember having a conversation about this PR is it still necessary or can we close it?

PR #1467, opened 2018-09-12
Fix Uncertainty after RangedReactor refactoring

goldmanm:
The Local uncertainty ipynb was not able to run properly after
the nSims parameter was added to the middle of the arguments when
initiating SimpleReactor, since the arguments sent to the
SimpleReactor from the uncertainty module were all shifted.

This commit/PR fixes this by using keyword arguments in the
uncertainty module.

### Testing
Run local uncertainty module before and after this PR to make sure the problem is fixed.

### DIFF
--- a/rmgpy/solver/liquid.pyx
+++ b/rmgpy/solver/liquid.pyx
@@ -63,7 +63,7 @@ cdef class LiquidReactor(ReactionSystem):
     cdef public dict sensConditions
     
 
-    def __init__(self, T, initialConcentrations, nSims=None, termination=None, sensitiveSpecies=None, sensitivityThreshold=1e-3, sensConditions=None, constSPCNames=None):
+    def __init__(self, T, initialConcentrations, nSims=1, termination=None, sensitiveSpecies=None, sensitivityThreshold=1e-3, sensConditions=None, constSPCNames=None):
         
         ReactionSystem.__init__(self, termination, sensitiveSpecies, sensitivityThreshold)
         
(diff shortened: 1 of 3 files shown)


mliu49:
In input.py, the simple reactor has nSims=6 by default, and the liquid reactor has nSims=4. Perhaps we can go with the same defaults for the actual reactor classes? @mjohnson541

goldmanm:
I guess I am confused exactly under what circumstances nSims is used.

Does the default nSims parameter impact results that don't use ranged reactors? and why it is necessary to have this parameter if ranged reactors are not used?

goldmanm:
Jim mentioned that as written, this commit makes uncertainty work. Maybe it can be coded in a better way to interface with other parts of the code.

mliu49:
Right, I don't think it matters, so I only suggested using the same default for consistency. You should also update LiquidReactor so that sensitivity works with that as well.

mjohnson541:
The primary use of nSims is for the algorithm to figure out how many times the reactor has to complete before it can terminate the run in execute.  In input.py if the reaction isn't ranged I change nSims to 1 regardless of the input value so it probably makes the most sense to assign it a default value of nSims=1.  I think the value should only matter if execute is run.

mjohnson541:
This looks good to me once the default nSims is added for the liquid reactor I'm comfortable with merging.

mliu49:
To summarize, @goldmanm, default of 1 is fine, but please add the default to the `__init__` for LiquidReactor as well. I think this can be merged after you do that and rebase.

mliu49:
Bump. @goldmanm, could you update this branch and address my last comment?

goldmanm:
Thanks for the reminder. I just implemented the suggestions.

mliu49:
I'm not sure what's going on with RMG-tests. I will go ahead and merge this PR anyway, since this shouldn't affect anything.

PR #1470, opened 2018-09-24
Add option to ignore atom type errors when making molecules

cgrambow:
### Motivation or Problem
RMG's atom types are relatively limited when dealing with non-standard molecules and an `AtomTypeError` might be raised upon calling `Molecule.updateAtomTypes`. Often, it is desirable to use the RMG API simply to create arbitrary molecule objects and use useful RMG methods that may not be as readily available in other cheminformatics packages. One possible application, is to use RMG's isomorphism checks to compare molecules that have been converted to their respective single-bonded versions.

### Description of Changes
This PR implements a new option, `raise_atomtype_exception` for many methods in the `Molecule` class and functions in the `translator` and `converter` modules, which enables setting the `raiseException` argument that was already available in `Molecule.updateAtomTypes.`

### Question for Reviewer
A typical application for me is to read in two molecules and check if they are isomorphic. Would having a generic atom type potentially yield unexpected behavior? I usually do something along the lines of the following:
```python
mol1 = Molecule().fromSMILES(smi, raise_atomtype_exception=False).toSingleBonds()
mol2 = Molecule().fromXYZ(nums, coords, raise_atomtype_exception=False).toSingleBonds()
if not mol1.isIsomorphic(mol2):
    # do something
```
Note that I don't usually need `raise_atomtype_exception` in the call to `toSingleBonds` because it seems that RMG's atom types are general enough to deal with all molecules that have only single bonds.

### Final Comment
If you think, this PR has serious potential for unexpected behavior, then we don't have to merge it, but it would be nice to find a way to avoid frequently getting an `AtomTypeError`.

### DIFF
--- a/rmgpy/molecule/converter.pxd
+++ b/rmgpy/molecule/converter.pxd
@@ -30,8 +30,8 @@ cimport rmgpy.molecule.molecule as mm
 
 cpdef to_rdkit_mol(mm.Molecule mol, bint remove_h=*, bint return_mapping=*, bint sanitize=*)
 
-cpdef mm.Molecule from_rdkit_mol(mm.Molecule mol, object rdkitmol)
+cpdef mm.Molecule from_rdkit_mol(mm.Molecule mol, object rdkitmol, bint raise_atomtype_exception=?)
 
 cpdef to_ob_mol(mm.Molecule mol, bint return_mapping=*)
 
-cpdef mm.Molecule from_ob_mol(mm.Molecule mol, object obmol)
+cpdef mm.Molecule from_ob_mol(mm.Molecule mol, object obmol, bint raise_atomtype_exception=?)
--- a/rmgpy/molecule/converter.py
+++ b/rmgpy/molecule/converter.py
@@ -107,7 +107,7 @@ def to_rdkit_mol(mol, remove_h=True, return_mapping=False, sanitize=True):
     return rdkitmol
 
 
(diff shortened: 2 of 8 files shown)


amarkpayne:
@mjohnson541 I am interested in reviving this PR, as there are a number of molecules in our reference species database that will throw AtomType errors when all we need to do is perform isomorphism checks and use other feature of the rmgpy.molecule API. This similarly affects ARC when doing calculations on species containing AtomTypes we don't yet support (kind of defeats a key future use of ARC if we can't use it to calculate new types of chemistry. @alongd are there ways around this in ARC that I am not aware of? I ran a calculation that failed because it could not perform the isomorphism check). What are your thoughts on this? My plan is just to rebase this on top of master and fix the conflicts, but would you prefer another strategy to fix this problem?

Another (but almost certainly worse) solution I thought of would be to store a module level variable that the `update_atomtypes` function reads from to determine if AtomType errors should be raised. This would allow the user to set this value in their scripts at run-time depending on if their task should globally ignore AtomType errors. It is a simple solution in that we don't have to chain this argument up a bunch of functions, but it is also way less flexible. The user could either always ignore AtomType error for all molecules or enforce it for all of them.

mjohnson541:
I think this approach is perfectly reasonable, as a chem-informatics software I think this is a capability we should have.

alongd:
We have ways around it in ARC, but it aint pretty... I'm in favor.

PR #1471, opened 2018-09-27
 Removed two redundant resonance pathways

alongd:
Two of our resonance pathways are redundant:
Both `generate_N5ddc_N5tc_resonance_structures` and `generate_N5dc_resonance_structures` are sub-classes of `generate_adj_lone_pair_multiple_bond_resonance_structures`.

To visualize this, pathways (2) and (3) are sub-classes of pathway (1):
screenshot from 2018-09-26 20-57-12
Each of the three has a lone pair creating a multiple bond, and an adjacent multiple bond creating a lone pair. (2) and (3) are just more specific.

A minor change to the O+ filtration criterion was also implemented (relocating it from the octet rule criterion to the electronegativity criterion, where it really belongs).

Tests were modified accordingly.

To the reviewer:
Try generating resonance structures specific for pathways (2) and (3) on this branch.


rmgpy/molecule/filtrationTest.py
```diff
@@ -68,21 +68,27 @@ def basic_filtration_test(self):
...
         2 O u0 p3 c-1 {1,S}
         3 O u0 p1 c+1 {1,T}
         """
         mol = Molecule().fromAdjacencyList(adj)
         octet_deviation = get_octet_deviation(mol)
-        self.assertEqual(octet_deviation, 1)
+        self.assertEqual(octet_deviation, 0)
         self.assertEqual(mol.vertices[2].atomType.label, 'O4tc')
+        mol_list = generate_resonance_structures(mol)
+        assert len(mol_list) == 2
```
mliu49: Use `self.assertEquals` instead?

alongd:
Thanks @mliu49!
done

mliu49:
The sulfur test seems to have some notable differences. Could you take a look?

alongd:
Thanks for pointing out the sulfur tests.
It seems to me that the differences in this test are due to the relocation of the O+ heuristic to its "natural" place within the electronegativity check rather than the octet rule. On master we currently penalize via the octet score molecules with O+, and this gets caught in the on-the-fly filtration. Here's an example for one of the species that got added in the test, `HO4S(51)`:

on master we only have:
image
And on this branch we have:
image <-> image
There's an intermediate structure with O+ that isn't further explored on master.
I think we should be comfortable merging this in.

PR #1473, opened 2018-09-27
Miscellaneous Cantherm fixes

alongd:
This fixes a bug where the Explorer tool in CanTherm reads `reaction.kinetics.comment` for input file reactions even if `reaction` has not `kinetics` attribute.
Also, deleted a comment saying we're not loading the database until after the input file is read, when in fact we are.


alongd:
I didn't intend to push the auto-calc Mw commit, but since it's here, it might be a good opportunity to pull it in as well. Principally, it calculates species' Mw upon calling this property, as suggested by @mliu49. Useful to avoid specifying the Mw in CanTherm's input files if the structure is given. Originally part of the #1402 PR.

alongd:
This PR contains miscellaneous Cantherm fixes.
@mjohnson541, I could split it into several PRs if you feel it would be more appropriate.
Otherwise, I think it's ready for review.

alongd:
@amarkpayne, could you assist with reviewing this PR and take a look at the QChem symmetry changes, as we discussed offline?

amarkpayne:
@alongd Of course! I am taking a look at this now

rmgpy/cantherm/qchem.py
```diff
@@ -187,15 +190,13 @@ def loadGeometry(self):
     
     def loadConformer(self, symmetry=None, spinMultiplicity=0, opticalIsomers=1, symfromlog=None, label=''):
```
amarkpayne: I tried running two tests on some QChem files I had to see if Cantherm had the expected behavior:

1) Specify a symmetry number in the Cantherm input file and look at the symmetry number listed in the output.

This test was successful. As I changed the symmetry number in the Cantherm input file, the number in the output file changed as well, even though the QChem log file did specify a value.



2) Don't specify a symmetry number in the Cantherm input file and see if the QChem value is shown in the output.

This test failed. No matter what I set the number in the log to, Cantherm had an output of 1. I believe this is because in our code `symmetry` being `None` does not set the value of `symfromlog` to `True`. The other part of the problem is that the QChem 'Rotational Symmetry Number' occurs after the 'Eigenvalues --' line (seems to be true for QChem 4.4 as well as the latest version on NERSC), so `symmetry` is still `None` when lines 250-251 set the symmetry to a value of 1. When Cantherm finally does read in the symmetry number off of the 'Rotational Symmetry Number line', setting the `symmetry` variable no longer has an effect because the rotors are already defined.

alongd: Thanks for the observations!
If Symmetry in `None`, I think that `symfromlog` is being set to `True` in statmech.py (https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/cantherm/statmech.py#L251)
The other part you pointed out seems problematic, I'll address it in a fixup commit

amarkpayne: Gotcha, good catch! I should have checked statmech.py

alongd:
Thanks @amarkpayne! I squashed both of our fixups into one commit

alongd:
@mjohnson541, this branch is rebase (and squashed), ready for review

alongd:
Added support in Cantherm to all elements when reading geometry (Fixes #1494)

rmgpy/cantherm/common.py
```diff
@@ -46,3 +46,171 @@ def checkConformerEnergy(Vlist,path):
         logging.warning('the species corresponding to ' + str(os.path.basename(path)) + ' is different in energy from the lowest energy conformer by ' + "%0.2f" % Vdiff + ' kJ/mol. This can cause significant errors in your computed rate constants. ')
 
 
+def get_element_mass(input_element, isotope=None):
```
mjohnson541: These large dictionaries should be stored outside of the function declaration, otherwise they get constructed every time you call this function.  It probably doesn't matter that much here, but it's such bad practice I think we should avoid it so future developers don't copy it.

alongd: Thanks! Fixed and squashed.
The QChem stuff was approved by @amarkpayne. I think this could be merged after tests pass

PR #1478, opened 2018-10-04
Travis improvements

mliu49:
### Description of Changes
Streamline RMG-tests deployment method by using personal access token instead of ssh key. The token is provided in Travis settings as an encrypted environment variable, and can be changed on-the-fly in case we want to use a different GitHub account or if the token is somehow compromised.

This also adds automatic documentation deployment as a Travis build stage. This stage is run on the master branch, so every merge will trigger a documentation rebuild.

### Testing
I've checked that the documentation deployment works (you can take a look at the result on the test_gh_pages branch). RMG-tests deployment is also easily verified by checking that we get an RMG-tests build on this branch. You may also notice that the deployment commits and pushes to GitHub are now attributed to RMGdev.


mliu49:
Just realized that this will trigger documentation builds for pull requests as well, which is not desired. This should not be merged while I look for a solution.

mliu49:
Should be ok now.

PR #1484, opened 2018-10-11
Clear atom labels from products after reaction generation

mliu49:
### Motivation or Problem
After reaction generation, product labels are currently retained even though they are no longer used, while reactant labels are being cleared.

### Description of Changes
Clear atom labels from reaction products as well, at the end of `KineticsFamily.__generateReactions`.

Also change an unrelated docstring which was incorrect. (Let me know if this should go in a separate PR.)

### Testing
Check RMG-tests for any unanticipated effects. This change should not affect model generation.


mjohnson541:
Can you rebase this?

mliu49:
Rebased and ready to merge.

Issue #1482, opened 2018-10-05
Incorrect axis and cyclic symmetry determination


mliu49:
### 1) Axis Symmetry

This was discovered with `C=[C]C=CCC`, which should have a symmetry number of 3 from the methyl rotor only. Using either of its resonance structures alone gives the correct result, but using the resonance hybrid (with bond orders of `[2, 1.5, 1.5, 1, 1]`) gives a symmetry number of 6. I've determined that the extra factor of 2 comes from `calculateAxisSymmetryNumber`, but had a hard time figuring out exactly what was going wrong.

### 2) Cyclic Symmetry

This was discovered with 'c1ccccc1[O]', which should have a symmetry number of 2. RMG currently does get the correct value using the resonance hybrid, but I'm not sure if it's for the right reason. I made a change to the set of resonance structures generated for this species, which changed the resonance hybrid and gave a new symmetry number of 12.

Looking into this a bit more, I noticed that the Kekule structure for this species also has a symmetry number of 12, even though the bonds are different. My current hypothesis is that vf2.feasible may not be properly comparing the atoms. It relies on having mapping and terminal attributes, which are assigned by the isomorphism function, which is not used in this case.

goldmanm:
For the first problem, the solutions lies in this incorrect line of code `rmgpy/molecule.symmetry.py` written over 8 years ago in 11d2b27da66ceb5a5725ea91dbecad. Can you find what's wrong with it?

```python
# Do nothing if less than two cumulated bonds
if len(bonds) < 1: continue
```

Changing the `1` to a `2`, then causes errors estimating ethylene and formaldehyde, which we'll have to fix before resolving this issue.

mliu49:
It seems like that was an intentional change, though the comment was not updated. See 8d72a1b33cf4aafbc4d3b037ba9cf9229fec4547.

goldmanm:
fixed by #1487

PR #1487, opened 2018-10-16
Symmetry fixes

mliu49:
### Motivation or Problem
This fixes the second issue in #1482. Cyclic symmetry numbers were not being properly determined because the `vf2.feasible` method wasn't working properly out of context.

### Description of Changes
A new method called `_indistinguishable` was created in `rmgpy.molecule.symmetry` to perform the task that `vf2.feasible` was being used for, without the additional infrastructure of `vf2`. The new method checks connectivity values, atom equivalence, and nearest neighbor equivalence to determine if two atoms are feasibly indistinguishable.

### Testing
All unit tests should be passing, and two tests previously marked as `@work_in_progress` are also now passing. Any additional test cases which are found be necessary should probably be added as new unit tests.

### Reviewer Tips
Review the new `_indistinguishable` method to see that it makes sense.

Also, @goldmanm, if you figure out the problem with the first issue in #1482, you can add the fix to this PR.

### DIFF
--- a/rmgpy/species.py
+++ b/rmgpy/species.py
@@ -494,13 +494,17 @@ class Species(object):
             oldAtoms = atomsFromStructures[structureNum]
 
             for index1, atom1 in enumerate(oldAtoms):
+                # make bond orders average of resonance structures
                 for atom2 in atom1.bonds:
                     index2 = oldAtoms.index(atom2)
 
                     newBond = newMol.getBond(newAtoms[index1], newAtoms[index2])
                     oldBondOrder = oldMol.getBond(oldAtoms[index1], oldAtoms[index2]).getOrderNum()
                     newBond.applyAction(('CHANGE_BOND',None,oldBondOrder / numResonanceStructures / 2))
-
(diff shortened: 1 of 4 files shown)


rmgpy/species.py
```diff
@@ -494,13 +494,17 @@ def getResonanceHybrid(self):
...
                     index2 = oldAtoms.index(atom2)
 
                     newBond = newMol.getBond(newAtoms[index1], newAtoms[index2])
                     oldBondOrder = oldMol.getBond(oldAtoms[index1], oldAtoms[index2]).getOrderNum()
                     newBond.applyAction(('CHANGE_BOND',None,oldBondOrder / numResonanceStructures / 2))
-
+                # set radicals in resonance hybrid to maximum of all structures
+                if atom1.radicalElectrons > 0:
+                    newAtoms[index1].radicalElectrons = max((atom1.radicalElectrons,
+                                                             newAtoms[index1].radicalElectrons))
```
mliu49: Really minor, but you don't need the extra parenthesis to make a tuple since `max` accepts a list of arguments.

PR #1489, opened 2018-10-17
Output a supporting information file in Cantherm

alongd:
This PR outputs a `supporting_information.csv` file with rotational constants and unscaled frequencies (inc. imaginary frequency if a TS) for all statmech jobs. This is useful when preparing manuscripts using Cantherm's output.

**Review:** 
Run a Cantherm job using statmech (e.g., the C2H4 example). Also try this out for a reaction / network where multiple species are outputted into the same supporting info file (inc. TS), e.g., the CH3OH+HCO example.
I think it is OK to always save the file to the output directory, but let me know if you think this should be activated with a flag.

Tagging @sarahkha as a potential first user

### DIFF
--- a/rmgpy/cantherm/main.py
+++ b/rmgpy/cantherm/main.py
@@ -39,6 +39,7 @@ import sys
 import logging
 import argparse
 import time
+import csv
 try:
     import matplotlib
     matplotlib.rc('mathtext', default='regular')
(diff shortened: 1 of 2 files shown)


mjohnson541:
Can you rebase?

alongd:
Thanks! rebased.

PR #1497, opened 2018-10-22
Fix: rotationalConstant supporting info

alongd:
Mono and diatomic species threw errors while generating the supporting_info_file recently implemented in Cantherm, since the former doesn't have rotational constants / frequencies, and the latter has a `rotationalConstant` in the form of a float (cannot be iterated). This fix checks for such cases and handles them.


alongd:
@mjohnson541, this fix is ready for review

PR #1498, opened 2018-10-22
Make Kekule structures non-representative and refactor aromatic resonance

mliu49:
### Motivation or Problem
With recent changes to resonance structure generation and the implementation of resonance structure filtration, I felt that aromatic resonance structure generation could be refactored to fit better with the new infrastructure. In particular, some filtration which was being done for aromatic species could be relocated to the filtration module for clarity and improved code organization.

Additionally, removing Kekule structures as a representative aromatic resonance structure has been on the to do list for a while. With changes enabling benzene bond reactivity and the new capability to mark non-representative structures as non-reactive, I also felt this was a good opportunity to do that as well. The benefit of doing so would primarily be to reduce the number of resonance structures for aromatics, and also to avoid cases where the Kekule structure may give poor thermo or kinetics estimates.

### Description of Changes
Changes to aromatic resonance algorithm:
- Split generate_aromatic_resonance structures into two functions. One does the actual conversion to benzene bonds, one is the wrapper function which finds the structure with the most aromatic rings. The new function are named `generate_aromatic_resonance_structure` and `generate_optimal_aromatic_resonance_structure`. This is a backwards incompatible change. If desired, I can also add a function declaration to map the old `generate_aromatic_resonance_structures` name to call `generate_optimal_aromatic_resonance_structure`.
- Resonance structure filtration moved to new `aromaticity_filtration` method in the filtration module. This does similar filtering as before, removing all non-aromatic (i.e. with no benzene bonds) structures for polycyclic aromatics. For monocyclic aromatics, the new function also removes Kekule structures, which is different from before.

Changes related to Kekule structure demotion:
- Remove `generate_opposite_kekule_structure` function, which is no longer necessary.
- Remove some unit tests for checking Kekule structure generation.

Other changes:
- Updated a number of unit tests based on differences resonance structure output.
- Update R_Addition_MultipleBond in testing database
- Only use representative resonance structures when generating resonance hybrid

### Testing
See how RMG-tests does, particularly for the aromatics test. The other tests should be unaffected. I might run a longer test case as well.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -48,7 +48,7 @@ from rmgpy import settings
 from rmgpy.reaction import Reaction
 from rmgpy.kinetics import Arrhenius
 from rmgpy.molecule import Bond, GroupBond, Group, Molecule
-from rmgpy.molecule.resonance import generate_aromatic_resonance_structures
+from rmgpy.molecule.resonance import generate_optimal_aromatic_resonance_structures
 from rmgpy.species import Species
 
 from .common import saveEntry, ensure_species, find_degenerate_reactions, generate_molecule_combos,\
@@ -1934,7 +1934,7 @@ class KineticsFamily(Database):
                 if prod_resonance:
                     for i, product in enumerate(products0):
                         if product.isCyclic:
-                            aromaticStructs = generate_aromatic_resonance_structures(product)
+                            aromaticStructs = generate_optimal_aromatic_resonance_structures(product)
                             if aromaticStructs:
                                 products0[i] = aromaticStructs[0]
 
--- a/rmgpy/data/thermoTest.py
+++ b/rmgpy/data/thermoTest.py
@@ -293,7 +293,9 @@ multiplicity 2
 """)
         spec.generate_resonance_structures()
 
-        self.assertTrue(arom.isIsomorphic(spec.molecule[1]))  # The aromatic structure should be the second one
+        self.assertTrue(arom.isIsomorphic(spec.molecule[0]))  # The aromatic structure should be the first one
+        # Move the aromatic structure to the end for testing
+        spec.molecule.append(spec.molecule.pop(0))
 
         initial = list(spec.molecule)  # Make a copy of the list
         thermo = self.database.getThermoData(spec)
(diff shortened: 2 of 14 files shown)


alongd:
The PR looks excellent technically, and I don't have comments on the code. Thanks!
Some questions:
1.  I think it is fine not to add a mapping function for `generate_aromatic_resonance_structure` \ `s`. Do you have a specific place in mind that uses this as an API?
2. Did you end up running the longer test case you mentioned?
3. Two species were added to the edge of the aromatics test. The test overall looks good.
    - The reaction `rxn: HC2(5) + [CH]1C=CC=CC1(6) <=> ethyne(2) + benzene(1)		origin: Disproportionation` that was added to the edge doesn't involve any of the new species (there are other disprops like that). Can we hypothesize why we didn't see it on master? `[CH]1C=CC=CC1` has the same resonance structures on both master and this branch.
    - The only rxn involving the new species `[CH]=CC=CC#C(9)` is the ring opening `rxn: [CH]=CC=CC#C(9) <=> [c]1ccccc1(3)		origin: Intra_R_Add_Endocyclic`. Isn't it strange that master doesn't have it?
4. Looking at the resonance structures of naphthalene, on master we get:
image <=> image <=> image
and on this branch we get:
image <=> image
Why are we keeping the mixed aromatic / non-aromatic structure? wouldn't it be enough for reactivity to only consider the fully aromatic structure?

mliu49:
Thanks for reviewing!

1. There are a couple scripts in the wip repo which use it, but it wouldn't be hard to update them. I might be the most frequent user of that function on its own.
2. I have been running some tests with this branch, and through that discovered that disproportionation reactions were not being generated for aromatic resonance structures, which will be fixed in ReactionMechanismGenerator/RMG-database#293, so that should be merged before or at the same time as this.
3. I think those are on master but missing on this branch. The disproportionation reactions are all missing because of the issue I mentioned in the last point.
4. Yes, I have been wondering about this myself. I'm not sure if keeping the Clar structure is actually helpful. I might change the criteria for generating Clar structures to species with 3 or more rings (instead of 2 or more).

mliu49:
I think this is ready for final review/merge. The missing disproportionation reactions will be fixed by ReactionMechanismGenerator/RMG-database#305, so that should be merged with or before this.

I only made one new change since the previous review, which is the last commit, where I realized that I could simplify the code a bit more.

For Clar structures for two ring aromatics, I think I will keep them for now. There were a couple of odd side-effects of changing that which will need further study.

mliu49:
I think this is ready to be merged. I added one new commit to address some reactions which were not being generated due to product templates not having benzene bonds. I think the remaining differences in the aromatics test are acceptable.

PR #1500, opened 2018-10-29
Improve kekulization algorithm for reacting PAHs

mliu49:
### Motivation or Problem
The kekulization module could not properly handle cases where bonds in different aromatic rings were modified as a result of a reaction. For example, Diels_alder_addition of acetylene to the bay site of phenanthrene, which results in bond change in all three rings of phenanthrene.

### Description of Changes
This updates the kekulization algorithm to do initial bond processing at once at the start of the entire procedure instead of waiting until the start of kekulizing each ring.

Additionally, naming in the entire kekulization module has been redone to conform to PEP-8. Compatibility issues are not expected to be significant because the only outward facing method is `kekulize`, which was not modified.

### Testing
Unit tests should pass, RMG-tests should not show any differences from master. This PR is necessary for ReactionMechanismGenerator/RMG-database#293 to pass.

### Reviewer Tips
Skim the code, ask questions, verify RMG-tests.

### DIFF
--- a/rmgpy/molecule/kekulize.pyx
+++ b/rmgpy/molecule/kekulize.pyx
@@ -149,7 +149,7 @@ cdef class AromaticRing(object):
 
     DO NOT use outside of this module. This class does not do any aromaticity perception.
     """
-    cdef list atoms
+    cdef public list atoms, resolved, unresolved
     cdef set endoBonds, exoBonds
     cdef public int endoDOF, exoDOF
 
@@ -159,6 +159,8 @@ cdef class AromaticRing(object):
         self.exoBonds = exoBonds
         self.endoDOF = endoDOF
         self.exoDOF = exoDOF
+        self.resolved = []
+        self.unresolved = []
 
     cpdef update(self):
         """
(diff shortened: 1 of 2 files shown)


rmgpy/molecule/kekulize.pyx
```diff
@@ -150,52 +150,52 @@ cdef class AromaticRing(object):
...
-        self.exoDOF = exoDOF
+                exo_dof += 1
+        self.endo_dof = endo_dof
+        self.exo_dof = exo_dof
 
-        self.processBonds()
+        self.process_bonds()
 
-    cpdef tuple processBonds(self):
+    cpdef tuple process_bonds(self):
```
mjohnson541: This particular name change is non-backwards compatible (I think all the other ones are fine, I consider the cdef as pretty much internal since they can't be called directly from python).  Is it conceivable that a user could be using it currently?

mliu49: I don't expect anyone to be using this directly. The only thing which should be used externally from this module should be the main `kekulize` function.

I think the only reason I made these cpdef was so that I could test them.

PR #1501, opened 2018-10-30
Parse MRCI energies from molpro log files

alongd:
Added MRCI/aug-cc-pV(T+d)Z Atom Corrections to Cantherm's statmech


alongd:
Added `MRCI/aug-cc-pVTZ`, `MRCI+Davidson/aug-cc-pVTZ`, and `MRCI/aug-cc-pV(T+d)Z` (we already had `MRCI+Davidson/aug-cc-pV(T+d)Z`)

alongd:
I added the respective model chemistries to the documentation, as well as MRCI parsing to molpro.py.
All MRCI AEC were recalculated using uncontracted basis sets (S for H, SPD for C/N/O, SPDF for S) as recommended by @phalgunlolur.
@cgrambow, this should be ready for review now.

rmgpy/cantherm/molpro.py
```diff
@@ -260,58 +260,89 @@ def loadConformer(self, symmetry=None, spinMultiplicity=0, opticalIsomers=1, sym
         f.close()
         return Conformer(E0=(E0*0.001,"kJ/mol"), modes=modes, spinMultiplicity=spinMultiplicity, opticalIsomers=opticalIsomers)
 
-    def loadEnergy(self,frequencyScaleFactor=1.):
+    def loadEnergy(self, frequencyScaleFactor=1., model_chemistry=''):
```
cgrambow: Why is `model_chemistry` needed here? Shouldn't we be able to read this information from the log file?

rmgpy/cantherm/molpro.py
```diff
@@ -260,58 +260,89 @@ def loadConformer(self, symmetry=None, spinMultiplicity=0, opticalIsomers=1, sym
...
+                        or 'CCSD(T)-F12b total energy  ' in line):
+                        E0 = float(line.split()[-1])
+                        break
                 if 'Electronic Energy at 0' in line:
-                    E0=float(line.split()[-2])
+                    E0 = float(line.split()[-2])
                     break
                 line=f.readline()
-
+        elif mrci and 'mrci+davidson' in model_chemistry.lower():
```
cgrambow: See previous comment. Does Molpro always print the Davidson corrected energy if using MRCI? If not, I think we should remove `model_chemistry` in this function.

rmgpy/cantherm/molpro.py
```diff
@@ -260,58 +260,89 @@ def loadConformer(self, symmetry=None, spinMultiplicity=0, opticalIsomers=1, sym
...
+            while line != '':
+                if '(Davidson, relaxed reference)' in line:
+                    E0 = float(line.split()[3])
+                    break
+                line=f.readline()
+        elif mrci and 'mrci' in model_chemistry.lower():  # 'mrci+davidson' is not in the model chemistry
+            while line != '':
+                if 'MRCI            MULTI           HF-SCF' in line:
```
cgrambow: Will these exact number of spaces be consistent across different versions of Molpro? Perhaps it would be more prudent to just check if `MRCI`, `MULTI`, and `HF-SCF` are all in the line.
```suggestion
                if all(w in line for w in ('MRCI', 'MULTI', 'HF-SCF')):
```

rmgpy/cantherm/statmech.py
```diff
@@ -391,7 +391,7 @@ def load(self):
         if E0_withZPE is None:
             # The E0 that is read from the log file is without the ZPE and corresponds to E_elec
             if E0 is None:
-                E0 = energyLog.loadEnergy(self.frequencyScaleFactor)
+                E0 = energyLog.loadEnergy(self.frequencyScaleFactor, self.modelChemistry)
```
cgrambow: This will crash for non-Molpro log files.

rmgpy/cantherm/statmech.py
```diff
@@ -732,8 +732,16 @@ def applyEnergyCorrections(E0, modelChemistry, atoms, bonds,
             elif modelChemistry == 'b3lyp/6-311+g(3df,2p)':  # Calculated atomic energies
                 atomEnergies = {'H':-0.502155915123 + SOC['H'], 'C':-37.8574709934 + SOC['C'], 'N':-54.6007233609 + SOC['N'], 'O':-75.0909131284 + SOC['O'], 'P':-341.281730319 + SOC['P'], 'S':-398.134489850 + SOC['S']}
 
-            elif modelChemistry == 'MRCI+Davidson/aug-cc-pV(T+d)Z':  # Calculated atomic energies (unfitted)
-                atomEnergies = {'H':-0.49982118 + SOC['H'], 'C':-37.78321274 + SOC['C'], 'N':-54.51729444 + SOC['N'], 'O':-74.97847534 + SOC['O'], 'S':-397.6571654 + SOC['S']}
+            elif modelChemistry == 'mrci/aug-cc-pvtz':  # Calculated atomic energies using an uncontracted basis set (unfitted)
+                atomEnergies = {'H':-0.49982132 + SOC['H'], 'C':-37.77733057 + SOC['C'], 'N':-54.50685339 + SOC['N'], 'O':-74.95843922 + SOC['O'], 'S':-397.64713217 + SOC['S']}
```
cgrambow: Are SOCs needed for MRCI? @phalgunlolur

rmgpy/cantherm/molpro.py
alongd: We sometimes use a similar level of theory for model_chemistry, so to avoid errors I think it's best to pass it

rmgpy/cantherm/statmech.py
alongd: We believe so

rmgpy/cantherm/statmech.py
alongd: I also added it in gaussian.py and qchem.py

alongd:
@cgrambow, I removed all MRCI atom corrections from this branch - we couldn't get thermo right in any configuration we tried (contracted/uncontracted basis sets). Probably the only way to go is fitting AE and/or BAC to experimental data. Hopefully we'll get to that one day.
I only left the MRCI E0 reading feature, and made it more elegant, as you suggested. The only drawback of not passing model chemistry to this method is the need to deciding which energy to read, MRCI or MRCI+Q. I gave priority to MRCI+Q if it exists, if not then MRCI E0 is read.
Let me know what you think.

arkane/molpro.py
```diff
@@ -263,60 +263,85 @@ def loadConformer(self, symmetry=None, spinMultiplicity=0, opticalIsomers=1, sym
...
 
-    def loadEnergy(self,frequencyScaleFactor=1.):
+    def loadEnergy(self, frequencyScaleFactor=1.):
         """
-        Return the f12 energy in J/mol from a Molpro Logfile of a CCSD(T)-f12 job.
-        This function determines which energy (f12a or f12b) to use based on the basis set,
+        Return either the f12 or MRCI energy in J/mol from a Molpro Logfile.
+        If the MRCI job outputted the MRCI+Davidson energy, the latter is returned.
+        For CCSD(T)-f12, the function determines which energy (f12a or f12b) to use based on the basis set,
         which it will parse out of the Molpro file. For the vtz and dtz basis sets f12a is
```
cgrambow: I just noticed that this should be `vdz` and not `dtz`. Might as well change that as part of this PR.
```suggestion
        which it will parse out of the Molpro file. For the vtz and vdz basis sets f12a is
```

arkane/molpro.py
```diff
@@ -263,60 +263,85 @@ def loadConformer(self, symmetry=None, spinMultiplicity=0, opticalIsomers=1, sym
...
-                        or 'RHF-RCCSD(T)-F12b energy' in line
-                        or 'CCSD(T)-F12b total energy  ' in line):
-                    E0=float(line.split()[-1])
-                    break
-                if 'Electronic Energy at 0' in line:
-                    E0=float(line.split()[-2])
+        E0 = None
+        with open(self.path, 'r') as f:
+            lines = f.readlines()
+            # Determine whether this is b12a or f12b according to the basis set, or whether this is MRCI.
```
cgrambow: `f12a`, not `b12a`
```suggestion
            # Determine whether this is f12a or f12b according to the basis set, or whether this is MRCI.
```

arkane/molpro.py
```diff
@@ -263,60 +263,85 @@ def loadConformer(self, symmetry=None, spinMultiplicity=0, opticalIsomers=1, sym
...
+                    # We should know the method by this point, so break if possible, but don't throw an error yet
+                    if any([mrci, f12a, f12b]):
+                        break
+            else:
+                raise ValueError('Could not determine type of calculation. Currently, CCSD(T)-F12a, CCSD(T)-F12b,'
+                                 ' MRCI, MRCI+Davidson are supported')
+            # Search for E0
+            for line in lines:
+                if f12a:
+                    if ('RHF-UCCSD(T)-F12a energy' in line
```
cgrambow: Can these all be grouped into a single check? I.e., just this:
```suggestion
                    if 'CCSD(T)-F12a energy' in line:
```

arkane/molpro.py
```diff
@@ -263,60 +263,85 @@ def loadConformer(self, symmetry=None, spinMultiplicity=0, opticalIsomers=1, sym
...
+                            or 'RHF-RCCSD(T)-F12a energy' in line
+                            or 'CCSD(T)-F12a total energy  ' in line):
+                        E0 = float(line.split()[-1])
+                        break
+                    if 'Electronic Energy at 0' in line:
+                        E0 = float(line.split()[-2])
+                        break
+                elif f12b:
+                    if ('RHF-UCCSD(T)-F12b energy' in line
```
cgrambow: Same here.
```suggestion
                    if 'CCSD(T)-F12b energy' in line
```

arkane/molpro.py
```diff
@@ -263,60 +263,85 @@ def loadConformer(self, symmetry=None, spinMultiplicity=0, opticalIsomers=1, sym
...
+                        break
+                elif mrci:
+                    # First search for MRCI+Davidson energy
+                    if '(Davidson, relaxed reference)' in line:
+                        E0 = float(line.split()[3])
+                        logging.debug('Found MRCI+Davidson energy in molpro log file {0}, using this value'.format(
+                            self.path))
+                        break
+            if E0 is None and mrci:
+                # No Davidson correction is given, search for MRCI energe
```
cgrambow:
```suggestion
                # No Davidson correction is given, search for MRCI energy
```

arkane/molpro.py
alongd: I don't know. Are there cases where we'll have `CCSD(T)-F12a total energy  ` and no `CCSD(T)-F12a energy`? I can probably group the first two checks, perhaps we should leave the third?

alongd: OK. Perhaps we should also check F12 with a VQZ basis: If not specifying F12b in the input file, but just F12, will it also output something like `RHF-UCCSD(T)-F12a energy`? If so, we should differentiate them when parsing

cgrambow: Yeah, I still meant that we should differentiate. I.e., have two tests like we currently do, one that does `if CCSD(T)-F12a in line and energy in line` and one that does `if CCSD(T)-F12b in line and energy in line`.

alongd:
Test added, and the conditions were changes as suggested.
Thanks for the comments, I hope this is good to go now, let me know otherwise.

alongd:
done & squashed

PR #1503, opened 2018-10-30
Remove extra quotes from pretty repr of references classes

mliu49:
### Motivation or Problem
Extra quotes are being added when generating the repr for Reference objects because using `{0!r}` for strings automatically includes quotes.

### Description of Changes
Remove extra quotes being added in the `toPrettyRepr` methods of Reference classes.

### Testing
To verify that this works, simply call `repr` on any string to see that it adds a set of quotes.
```
>>> repr('abcd')
"'abcd'"
>>> 'title = {0!r}'.format('abcd')
"title = 'abcd'"
```

### DIFF
--- a/rmgpy/data/reference.py
+++ b/rmgpy/data/reference.py
@@ -100,11 +100,11 @@ class Reference:
         reconstruct the object.
         """
         string = u'Reference(\n'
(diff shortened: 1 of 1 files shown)


mliu49:
It does affect database writing by getting rid of the extra quotes. It doesn't affect database reading.

ReactionMechanismGenerator/RMG-database#304 was done with this change in place, so I'm pretty sure its safe.

PR #1504, opened 2018-11-06
fix diffusion limited rate calculations

mjohnson541:
The original algorithm used an inexact criterion for deciding whether the rate was limited in the forward or reverse direction.  This PR causes it to be calculated in both directions and the the direction that results in the slowest forward rate to be used.

### DIFF
--- a/rmgpy/kinetics/diffusionLimited.py
+++ b/rmgpy/kinetics/diffusionLimited.py
@@ -84,13 +84,12 @@ class DiffusionLimited(object):
                 k_diff = self.getDiffusionLimit(T, reaction, forward=True)
                 k_eff = k_forward*k_diff/(k_forward+k_diff)
             else: # 2 or 3 products
-                if Keq > 1.0: # forward rate is faster and thus limited
-                    k_diff = self.getDiffusionLimit(T, reaction, forward=True)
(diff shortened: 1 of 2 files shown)


rmgpy/kinetics/diffusionLimited.py
```diff
@@ -84,13 +84,15 @@ def getEffectiveRate(self, reaction, T):
...
-                else: # reverse rate is faster and thus limited
-                    k_diff = self.getDiffusionLimit(T, reaction, forward=False)
-                    k_eff_reverse = k_reverse*k_diff/(k_reverse+k_diff)
-                    k_eff = k_eff_reverse * Keq
+                kf_diff = self.getDiffusionLimit(T,reaction,forward=True)
+                krev_diff = self.getDiffusionLimit(T,reaction,forward=False)
+                kff = k_forward*kf_diff/(k_forward+kf_diff)
+                krevr = k_reverse*krev_diff/(k_reverse+krev_diff)
+                kfr = Keq*krevr
+                if kff > kfr:
```
cgrambow: To make this a bit simpler, you could just do `k_eff = min(kff, kfr)`.

mjohnson541:
Ok, that change was made and the documentation has been made more clear.

PR #1511, opened 2018-11-20
 Added an Arkane test that runs all its examples 

alongd:
This new unit test runs all of Arkane's examples, and reports if any of them crashed (reporting back isn't mandatory - the test will crash anyway - but it does no harm). It took just over 4 minutes to run locally. I couldn't add it to commonTest, which might have been a good place for it, since it messed with the global jobLog list of other tests. Instead it is in mainTest.


arkane/mainTest.py
```diff
@@ -0,0 +1,110 @@
...
+        self.base_path = os.path.join(os.path.dirname(os.path.dirname(rmgpy.__file__)), 'examples', 'arkane')
+        self.failed = []
+        self.example_types = ['species', 'reactions', 'networks']
+
+    def test_arkane_examples(self):
+        for example_type in self.example_types:
+            example_type_path = os.path.join(self.base_path, example_type)
+            examples = []
+            for (dirpath, dirnames, filenames) in os.walk(example_type_path):
+                examples.extend(dirnames)
```
mliu49: I think you can use `os.listdir(example_type_path)` here since all the examples are in subdirectories.

arkane/mainTest.py
```diff
@@ -0,0 +1,110 @@
...
+                arkane = Arkane(inputFile=os.path.join(path, 'input.py'), outputDirectory=path)
+                arkane.plot = True
+                arkane.execute()
+                with open(os.path.join(path, 'arkane.log'), 'r') as f:
+                    log = f.readlines()
+                for line in log[::-1]:
+                    if 'execution terminated' in line:
+                        break
+                else:
+                    self.failed.append([example_type, example + 'FreeRotor'])
```
mliu49: Why do you append `'FreeRotor'` to the example name?

arkane/mainTest.py
```diff
@@ -0,0 +1,110 @@
...
+        error_message = 'Arkane example(s) failed: '
+        for type_name_tuple in self.failed:
+            error_message += '{1} in {0}; '.format(type_name_tuple[0], type_name_tuple[1])
```
mliu49: I think you can directly unpack the tuple (though it's actually a list) here: `for example_type, example_name in self.failed`

arkane/mainTest.py
```diff
@@ -0,0 +1,110 @@
...
+        self.assertTrue(len(self.failed) == 0, error_message)
```
mliu49: Other options:
- `self.assertFalse(self.failed, error_message)`
- `self.assertEqual(len(self.failed), 0, error_message)`

arkane/mainTest.py
```diff
@@ -0,0 +1,110 @@
...
+        self.extensions_to_delete = ['pdf', 'csv', 'txt', 'inp']
+        self.files_to_delete = ['arkane.log', 'output.py']
+        self.files_to_keep = ['README.txt']  # files to keep that have extentions marked for deletion
+        self.base_path = os.path.join(os.path.dirname(os.path.dirname(rmgpy.__file__)), 'examples', 'arkane')
+        self.example_types = ['species', 'reactions', 'networks']
+        for example_type in self.example_types:
+            example_type_path = os.path.join(self.base_path, example_type)
+            examples = []
+            for (dirpath, dirnames, filenames) in os.walk(example_type_path):
+                examples.extend(dirnames)
```
mliu49: Similar to before, I think you can do `os.listdir(example_type_path)` here.

arkane/mainTest.py
```diff
@@ -0,0 +1,110 @@
...
+                break
+            for example in examples:
+                # clean working folder from all previous test output
+                example_path = os.path.join(example_type_path, example)
+                dirs = [d for d in os.listdir(example_path)
+                        if not os.path.isfile(os.path.join(example_path, d))]
+                for d in dirs:
+                    shutil.rmtree(os.path.join(example_path, d, ''))
+                files = [f for f in os.listdir(example_path) if os.path.isfile(os.path.join(example_path, f))]
```
mliu49: It might be a bit cleaner to loop through the values of `os.listdir` and use an if statement to determine whether its a file or directory:
```
for name in os.listdir(example_path):
    if os.path.isfile:
        check filename and remove file
    else:
        remove directory
```

arkane/mainTest.py
```diff
@@ -0,0 +1,110 @@
...
+
+################################################################################
+
+
+class TestArkaneExamples(unittest.TestCase):
+    """
+    Run all of Arkane's examples, and report which one failed
+    """
+    @classmethod
+    def setUp(self):
```
mliu49: I think this should be named `setUpClass(cls)` with the cls in place of self. Similarly for `tearDownClass(cls)`.

arkane/mainTest.py
alongd: It's a leftover from a different version of the test, fixed

alongd:
Thanks, I addressed all comments in a fixup commit

arkane/mainTest.py
```diff
@@ -35,29 +35,31 @@
...
+        cls.base_path = os.path.join(os.path.dirname(os.path.dirname(rmgpy.__file__)), 'examples', 'arkane')
+        cls.failed = []
+        cls.example_types = ['species', 'reactions', 'networks']
 
     def test_arkane_examples(self):
         for example_type in self.example_types:
             example_type_path = os.path.join(self.base_path, example_type)
             examples = []
-            for (dirpath, dirnames, filenames) in os.walk(example_type_path):
+            for (dirpath, dirnames, filenames) in os.listdir(example_type_path):
```
mliu49: Sorry, I wasn't clear about using `os.listdir`. You should be able to replace
```
examples = []
for (dirpath, dirnames, filenames) in os.walk(example_type_path):
    examples.extend(dirnames)
    break
for example in examples:
    ...
```
with
```
for example in os.listdir(example_type_path):
    ...
```
similar to what you did in the `tearDownClass` method.

arkane/mainTest.py
```diff
@@ -71,39 +73,38 @@ def test_arkane_examples(self):
...
-                        os.remove(os.path.join(example_path, f))
+                for name in os.listdir(example_path):
+                    if os.path.isfile(name):
+                        extension = name.split('.')[-1]
+                        if name in cls.files_to_delete or\
+                                (extension in cls.extensions_to_delete and name not in cls.files_to_keep):
+                            os.remove(os.path.join(example_path, name))
+                    else:
+                        # This is a sub-directory. remove.
+                    shutil.rmtree(os.path.join(example_path, name, ''))
```
mliu49: This should be indented by one more level.

alongd:
The code looks good to me. Thanks!

PR #1512, opened 2018-11-20
 Output unscaled frequencies in Arkane's supporting information file 

alongd:
Previously, non-imaginary frequencies outputted to the supporting information file in Arkane were extracted from the HarmonicOscillator mode, which later gets multiplied by the frequency scaling factor. With this fix, unscaled frequencies are reported.

How to test:
Run any species in arkane (e.g., the C2H4 example) and compare the reported frequencies in the supporting information file generated on this branch and on master with the frequencies reported in the QM log file (easier if the QM software is QChem, where the units are already in cm-1)

### DIFF
--- a/arkane/gaussian.py
+++ b/arkane/gaussian.py
@@ -165,6 +165,7 @@ class GaussianLog:
         """
 
         modes = []
+        unscaled_frequencies = []
         E0 = 0.0
 
         f = open(self.path, 'r')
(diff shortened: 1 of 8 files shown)


alongd:
@mjohnson541, we'd like to add this to the next release. Could you take a look?

alongd:
Thanks! done.

PR #1513, opened 2018-11-21
Allow hasStatMech to assess atomic molecules

goldmanm:
Previously, hasStatMech would return True for any molecule with one atom, even if the E0 parameter was None. This raises issues when creating pdep networks from NASA polynomials in arkane since E0 is necessary. At the same time, this method does not check E0 parameters even for polyatomic
molecules.

This commit reworks hasStatMech to differentiate between atomic molecules, which it only checks the E0 parameter, and polyatomic molecules, which it checks both E0 and the number of modes.

I have tested the arkane example methoxy, which deals with atomic hydrogen, and this commit results in no change in that result. This commit does result in a fix when the user inputs an atomic molecule into the arkane input file given its NASA polynomial.

### Reviewer Tips

Look over the code and let me know if anything should be added/fixed/etc.

### DIFF
--- a/rmgpy/species.py
+++ b/rmgpy/species.py
@@ -328,7 +328,12 @@ class Species(object):
         Return ``True`` if the species has statistical mechanical parameters,
         or ``False`` otherwise.
         """
-        return self.conformer is not None and (len(self.conformer.modes) > 0 or (len(self.molecule) > 0 and len(self.molecule[0].atoms) == 1))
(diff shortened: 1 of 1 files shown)


alongd:
@goldmanm, the code looks good, strange that the tests failed. Could you rebase and push, so we'll give the tests a second trial?

goldmanm:
@alongd Seems like the tests passed this time.

PR #1519, opened 2018-12-09
Added another parsing phrase to QChem loadEnergy()

alongd:
Added another parsing phrase in Arkane for QChem log files.
Parsing a line in the following format:
` Total energy in the final basis set = -328.6420045839`
Also some pythonic modifications to this block.
An input file for example and review of the code is attached.
[output.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/2660465/output.txt)

### DIFF
--- a/arkane/qchem.py
+++ b/arkane/qchem.py
@@ -42,10 +42,10 @@ from arkane.common import check_conformer_energy, get_element_mass
 ################################################################################
 
 
-class QchemLog:
+class QChemLog:
     """
-    Represent an output file from Qchem. The attribute `path` refers to the
-    location on disk of the Qchem output file of interest. Methods are provided
+    Represent an output file from QChem. The attribute `path` refers to the
+    location on disk of the QChem output file of interest. Methods are provided
     to extract a variety of information into Arkane classes and/or NumPy
     arrays.
     """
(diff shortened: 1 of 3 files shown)


arkane/qchem.py
```diff
@@ -270,29 +270,20 @@ def loadEnergy(self, frequencyScaleFactor=1.):
...
-#                ZPE = float(line.split()[4]) * 4184
-#                scaledZPE = ZPE * frequencyScaleFactor
-#                print 'ZPE is ' + str(ZPE)
-            # Read the next line in the file
+        with open(self.path, 'r') as f:
             line = f.readline()
-    
-        # Close file when finished
-        f.close()
+            while line != '':
```
mliu49: While you're making it more pythonic, the following is a lot simpler if you're going to read the whole file:
```
    for line in f:
        # Do something
```

alongd: Thanks for the comment!

alongd:
@rgillis8, this branch now passes all tests, I also incorporated @mliu49's suggestion. All of Arkane's examples were run on this branch and terminated successfully. I think this PR should be ready for a final review, whenever you got time. Thanks!

alongd:
@rgillis8, I added a minor change (lines 277-281), splitting the two checks into different loops, giving priority to `Final energy is` over `Total energy in the final basis set`.
(This matters, for example, if a freq job is executed after the sp job in the same run)
Could you take a final look? Thanks!

alongd:
Thanks @rgillis8!
I just rebased, I think it needs an additional approval after the tests are done

PR #1522, opened 2018-12-14
Allow a BAC dictionary to be assigned to a StatMechJob object

alongd:
This small change allows us to assign a BAC dictionary to a StatMechJob object after the object is created. It is useful when working with scripts that pass arguments into Arkane. Specifically, this is how ARC passes BAC.

For the reviewer:
Try creating a `StatMechJob` object on this branch, then pass a BAC dictionary to it using the `.bonds` attribute.
```python
stat_mech_job = StatMechJob(spec, input_file_path)
bac_dict = {'C-C': 1, 'C-H':4}
stat_mech_job.bonds = bac_dict
```

### DIFF
--- a/arkane/statmech.py
+++ b/arkane/statmech.py
@@ -174,6 +174,7 @@ class StatMechJob(object):
         self.applyBondEnergyCorrections = True
         self.atomEnergies = None
         self.supporting_info = [self.species.label]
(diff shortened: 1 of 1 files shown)


alongd:
Thanks! Just rebased

PR #1523, opened 2018-12-14
Corrected 'ccsd(t)-f12/cc-pvtz-f12' frequency scaling factor

alongd:
Previously, the auto-determined frequency scaling factor for `ccsd(t)-f12/cc-pvtz-f12` was derived from [Truhlar 2010](https://pubs.acs.org/doi/abs/10.1021/ct100326h), but looking there again I don't see it or a similar level in the paper (should have been in Table 8). The previous value of 0.984 seems to be taken from `CCSD(T)/jul-cc-pVTZ` in that table. I think that a better estimation for the frequency scaling factor is for `ccsd(t)/cc-pvtz` from [CCCBDB](http://cccbdb.nist.gov/vibscalejust.asp).

@cgrambow, would you agree?


alongd:
Thanks for catching my mistake!
I changed the comment to reflect the actual source, and also added `wb97x-d/aug-cc-pvtz`. Please take a look.

PR #1524, opened 2018-12-14
Machine learning estimator settings

cgrambow:
### Motivation or Problem
Adds additional option to the ML estimator to allow for better fine-tuning of when to use machine learning or group additivity estimates.

### Description of Changes
Implements new options in `rmgpy.rmg.input.mlEstimator` and adds relevant methods in other sections of RMG to make those options work. The biggest changes are related to the `minCycleOverlap` option.

### Testing
Added new unit tests.

### Reviewer Tips
Check unit tests and perhaps run a simple model with some of the options turned on.


documentation/source/users/rmg/input.rst
```diff
@@ -613,6 +613,48 @@ The following is an example of the quantum mechanics options ::
...
 
+.. _mlEstimation:
+
+Machine Learning-based Thermo Estimation
+========================================
+
+This block is used to turn on the machine learning module to estimate thermodynamic parameters.
+These calculations are only run if the molecule is not included in a specified thermo library.
+There are a number of different settings that can be specified to tune the estimator so that it only tries to estimate
+some species. This is useful, because the machine learning model may perform poorly for some molecules and group
```
mliu49: Minor, but this comma is unnecessary.

cgrambow:
I made the requested changes and split the unit test into two different ones. Let me know if you'd like it split up in a different way.

mliu49:
You could possibly move the setup code for the mlEstimator to `setUpClass`. Otherwise, I'm ok with the split.

cgrambow:
Done.

I also just realized that the `get_element_count` method won't work if one of `C`, `O`, or `N` is not present in the molecule, so I changed it to use `dict.get` with a default of 0 instead.

mliu49:
I added a commit to for using the mlEstimator with the thermoEstimator script which I was using to test this PR.

I think this will be ready to merge after tests pass.

PR #1526, opened 2018-12-19
Determine the rotor symmetry number from the potential scan

alongd:
### Motivation or Problem
We'd like to allow Arkane to automatically determine rotor symmetry numbers. It is essential for automated thermo and rate codes using Arkane, like ARC. This feature is purposely added to Arkane and not ARC to benefit all users (if it works correctly, it could assist users to avoid making errors).

### Description of Changes
A function for rotor symmetry determination was added to statmech.py.
The primary criterion used for determining whether a scan has symmetry or not is a comparison between peak/valley height difference and the worst peak/valley resolution.
This feature is able to differentiate between the following challenging cases, the one on the left is _not_ symmetric (although it might look like it is at first sight), and the second has a symmetry number of 3:
screenshot from 2018-12-19 11-29-47

### Testing
Tests for the rotors plotted above for ethylamine were added. A `statmechTest.py` file was created.

### Reviewer Tips
Try running Arkane with two arbitrary rotor scans, one with symmetry and one without, and see (in the log file) whether the determined symmetry is correct.


arkane/commonTest.py
```diff
@@ -270,24 +270,6 @@ def testTransitionStateStatmech(self):
         job.load()
 
 
-class TestStatmech(unittest.TestCase):
```
amarkpayne: Can you move this change to the next commit where you add statmechTest.py? The rest of this commit can probably be combined with one of your earlier syntax commits as well.

documentation/source/users/arkane/input.rst
```diff
@@ -306,8 +306,10 @@ scan in the following format::
 
 The ``Energy`` can be in units of ``kJ/mol``, ``J/mol``, ``cal/mol``, ``kcal/mol``, ``cm^-1`` or ``hartree``.
 
-The ``symmetry`` parameter will usually equal either 1, 2 or 3. Below are examples of internal rotor scans with these
-commonly encountered symmetry numbers. First, ``symmetry = 3``:
+The ``symmetry`` parameter will usually equal either 1, 2 or 3. It could be determined automatically by Arkane
+(by simple not specifying it altogether), however it is always better to explicitly specify it if it is known. If it is
```
amarkpayne: change `by simple not specifying` to `by simply not specifying`

documentation/source/users/arkane/input_pdep.rst
```diff
@@ -351,8 +351,10 @@ scan in the following format::
 
 The ``Energy`` can be in units of ``kJ/mol``, ``J/mol``, ``cal/mol``, ``kcal/mol``, ``cm^-1`` or ``hartree``.
 
-The ``symmetry`` parameter will usually equal either 1, 2 or 3. Below are examples of internal rotor scans with these
-commonly encountered symmetry numbers. First, ``symmetry = 3``:
+The ``symmetry`` parameter will usually equal either 1, 2 or 3. It could be determined automatically by Arkane
```
amarkpayne: Is there a way in .rst files to add a section of text from another file? That why, future updates to this in arkane/input.rst will automatically update the same section in input_pdep.rst. I am not sure if this is possible, but I will do some digging to find out.

arkane/statmech.py
```diff
@@ -1163,3 +1174,74 @@ def assign_frequency_scale_factor(model_chemistry):
...
+                    worst_peak_resolution = max(e - energies[i - 1], e - energies[i + 1])
+                peaks.append(e)
+            elif e < energies[i - 1] and e < energies[i + 1]:
+                # this is a local valley
+                if any([diff > worst_valley_resolution for diff in [energies[i - 1] - e, energies[i + 1] - e]]):
+                    worst_valley_resolution = max(energies[i - 1] - e, energies[i + 1] - e)
+                valleys.append(e)
+    # The number of peaks and valley must always be the same (what goes up must come down), if it isn't then there's
+    # something seriously wrong with the scan
+    if len(peaks) != len(valleys):
```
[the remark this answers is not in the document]
alongd: Above we define `peaks, valleys = list(), [energies[0]]`, so we take the first point into consideration, which is also represented by the last point (assuming a 360 deg scan). In a normal case both these points represent a valley, if it's a peak then this scan is problematic. This is checked [elsewhere](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/arkane/common.py#L49) (only a warning is raised).

documentation/source/users/arkane/input_pdep.rst
alongd: Interesting, I don't know

alongd:
@amarkpayne, I addressed the comments, let me know what your thoughts are on the ABAB case.

documentation/source/users/arkane/input_pdep.rst
amarkpayne: It appears to be possible with the `.. include ::` statement, but it involves moving the text to another file. I am not sure if it is worth doing then, as it will clutter the documentation source files a little bit.

alongd: Let's leave it as is for now, we want to change the pdep doc of Arkane to only include pdep stuff and avoid this repetition.

arkane/statmech.py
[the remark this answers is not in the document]
amarkpayne: Gotcha, sorry that I missed this

arkane/statmech.py
```diff
@@ -1163,3 +1174,74 @@ def assign_frequency_scale_factor(model_chemistry):
...
+        raise InputError('Rotor of species {0} between pivots {1} does not have the same number'
+                         ' of peaks and valleys.'.format(label, pivots))
+    min_peak = min(peaks)
+    max_peak = max(peaks)
+    min_valley = min(valleys)
+    max_valley = max(valleys)
+    # Criterion 1: worst resolution
+    if max_peak - min_peak > worst_peak_resolution:
```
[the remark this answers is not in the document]
amarkpayne: As we discussed offline, while this scenario would indeed have a symmetry number of two, we could not come up with a physical system that would have this property. Therefore, this technique should work in most if not all cases. No changes necessary.

arkane/statmechTest.py
```diff
@@ -0,0 +1,82 @@
...
+#                                                                             #
+###############################################################################
+
+"""
+This script contains unit tests of the :mod:`arkane.main` module.
+"""
+
+import unittest
+import os
+import shutil
```
amarkpayne: My IDE suggests that this import statement is not needed, and I only see it in this file once. Can it be deleted?

examples/arkane/species/C2H5/ethyl.py
```diff
@@ -19,14 +19,15 @@
...
-frequencies = Log('ethyl_cbsqb3.log')
+frequencies = Log('ethyl_b3lyp.log')
 
 """pivot are the two atoms that are attached to the rotor
 top contains the atoms that are being rotated including one of the atoms from pivots
 symmetry is the symmetry number of the scan
-fit is fit of the scan data. It defaults to 'best', but can also be assigned as 'cosine' or 'fourier'"""
-rotors = [
-    HinderedRotor(scanLog=Log('ethyl_scan_72.log'), pivots=[1,2], top=[1,3,4], symmetry=6, fit='best')
-]
```
amarkpayne: Is it worth doing the same thing to `examples/arkane/species/Toluene_Hindered_Rotor`?

alongd: I wanted to leave one example in with the explicit symmetry number (although it's well documented) for users who would like to specify it. But I'm open to other thoughts.

alongd:
Done!
Thanks for the review and the helpful comments. Let me know if you have anything else in mind

PR #1536, opened 2019-01-18
RMG-Cat!!!

rwest:
I'm hopefully going to continue trying to tidy up this pull request via some rebasing to make it easier to review in chunks, but however this is sliced up in terms of commits, I think this PR contains all the actual changes that will be included in merging RMG-Cat, so if anyone brave wants to start reviewing, here at least is something we can start to discuss...

(Oh, it needs the cat version of the RMG-database too)


rwest:
Replaced by #1537 which comes from the ReactionMechanismGenerator fork rather than the rwest fork.

PR #1539, opened 2019-01-21
Update copyright dates

mliu49:
Very straightforward. Just double-check that no unintentional changes were made.

The shebangs were removed from chemkin.pyx and chemkin.pxd because those are now cython files.


amarkpayne:
Everything looked good, thanks @mliu49!

PR #1540, opened 2019-01-23
BugFix: Empty list as val in RMGObject.as_dict() and .make_object()

alongd:
### Motivation or Problem
When RMGObject encounters an empty list, it cannot index it (e.g., call `val[0]`). This occures when running atomic species where the frequencies are just an empty list `[]`.

### Description of Changes
We shouldn't bother dumping and parsing empty lists, they will get their default values when loading the object.

### Testing
Files for testing (for carbon atom) are attached
[C_arkane_input.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/2787635/C_arkane_input.txt)
[input.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/2787636/input.txt)
[sp.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/2787637/sp.txt)

### Reviewer Tips
After successfully generating a YAML file for C, try feeding that YAML file back to Arkane to see that it reads it correctly.

### DIFF
--- a/rmgpy/rmgobject.pyx
+++ b/rmgpy/rmgobject.pyx
@@ -55,14 +55,13 @@ cdef class RMGObject(object):
             if val is not None and not callable(val) and val != '':
                 output_dict[attr] = val
         for key, val in output_dict.iteritems():
-            if isinstance(val, list) and isinstance(val[0], RMGObject):
+            if isinstance(val, list) and val and isinstance(val[0], RMGObject):
                 output_dict[key] = [v.as_dict() for v in val]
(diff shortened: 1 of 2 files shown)


mliu49:
I added a new test module for RMG object. One of the tests is actually failing, which I'm glad was caught. If the value is a numpy array, then directly checking the truth value is ambiguous and raises an exception.

mliu49:
The fix looks good! You can go ahead and squash.

alongd:
done!

PR #1545, opened 2019-02-02
Network Explorer Part 2

mjohnson541:
This PR improves the network explorer tool.  It enables use of bimolecular reactant channels and changes the definition of explore_tol.  Before explore_tol was an explicit rate coefficient the leak flux of the network was compared to.  This PR uses the sum of the coefficients for the net reactions from the source to all other channels to scale the leak rate coefficient.  This means that the network will be expanded if the leak rate of the network is larger than explore_tol times the characteristic rate coefficient.  Where the charactersitic rate coefficient is the sum of the rate coefficients for all (net) reactions taking the source channel to any other channel in the network.  

Check that the unittests pass, run the methoxy network example and take a look at how the new explore_tol affects network generation.


mjohnson541:

arkane/explorer.py
```diff
@@ -143,10 +143,19 @@ def execute(self, outputFile, plot, format='pdf', print_summary=True, speciesLis
                     self.source[i] = spc
         
         # react initial species
-        flags = np.array([s.molecule[0].getFormula()==form for s in reaction_model.core.species])
+        if len(self.source) == 1:
+            flags = np.array([s.molecule[0].getFormula()==form for s in reaction_model.core.species])
+            biflags = np.zeros((len(reaction_model.core.species),len(reaction_model.core.species)))
+        elif len(self.source) == 2:
+            flags = np.array([s.molecule[0].getFormula()==form for s in reaction_model.core.species])
```
alongd: If `flags` is being set the same regardless of the condition, perhaps set it before the if statement? Or should it be set to zeroes if `len(self.source)==2`?

arkane/explorer.py
```diff
@@ -193,7 +193,14 @@ def execute(self, outputFile, plot, format='pdf', print_summary=True, speciesLis
             incomplete = False
             for T in Tlist:
                 for P in Plist:
-                    if network.getLeakCoefficient(T=T,P=P) > self.explore_tol:
+                    kchar = 0.0 #compute the characteristic rate coefficient by summing all rate coefficients from the reactant channel
+                    for rxn in self.network.netReactions:#reaction_model.core.reactions+reaction_model.edge.reactions:
+                        if set(rxn.reactants) == set(self.source) and len(rxn.products) == 1 and rxn.products[0].molecule[0].getFormula() == form:
```
alongd: Can we have `len(rxn.products) == 1` but not `rxn.products[0].molecule[0].getFormula() == form`?

arkane/explorer.py
```diff
@@ -193,95 +194,103 @@ def execute(self, outputFile, plot, format='pdf', print_summary=True, speciesLis
             incomplete = False
             for T in Tlist:
                 for P in Plist:
-                    kchar = 0.0 #compute the characteristic rate coefficient by summing all rate coefficients from the reactant channel
-                    for rxn in self.network.netReactions:#reaction_model.core.reactions+reaction_model.edge.reactions:
-                        if set(rxn.reactants) == set(self.source) and len(rxn.products) == 1 and rxn.products[0].molecule[0].getFormula() == form:
-                            kchar += rxn.kinetics.getRateCoefficient(T=T,P=P)
-                        elif set(rxn.products) == set(self.source) and len(rxn.reactants) == 1 and rxn.reactants[0].molecule[0].getFormula() == form:
-                            kchar += rxn.generateReverseRateCoefficient(network_kinetics=True).getRateCoefficient(T=T,P=P)
```
alongd: I think this chunk was added in a commit which is part of this PR:
`use the sum of the rate coefficients for all net rates from the`, https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1545/commits/f236722653d56b7628ac8aca340c22b18ce73f0e

Can you merge these instead of deleting the code?

arkane/mainTest.py
```diff
@@ -53,15 +54,16 @@ def setUpClass(cls):
...
         cls.base_path = os.path.join(os.path.dirname(os.path.dirname(rmgpy.__file__)), 'examples', 'arkane')
         cls.failed = []
-        cls.example_types = ['species', 'reactions', 'networks']
+        cls.example_types = ['species', 'reactions', 'explorer', 'networks']
 
     def test_arkane_examples(self):
         for example_type in self.example_types:
             example_type_path = os.path.join(self.base_path, example_type)
-            for example in os.listdir(example_type_path):
+            for example in sorted(os.listdir(example_type_path)):
```
alongd: Can you check whether https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1565 solved this problem?

mjohnson541: It doesn't solve the issue

arkane/explorer.py
mjohnson541: Fixed

arkane/explorer.py
mjohnson541: Fixed

arkane/explorer.py
mjohnson541: I think this is fixed too.

mjohnson541:
I think this should be ready.

alongd:
@mjohnson541, I reviewed by commits, I see that the final code doesn't include the line I commented on. Please ignore my comment above. Maybe you'd like to polish the commits so there won't be additions and deletions of the same code blocks in a single PR?

mjohnson541:
I could do that, but the order in which the commits were applied wouldn't make sense any longer.

mjohnson541:
It would mean applying two major changes at once in one commit.

mjohnson541:
@alongd pdf overwritting issue is fixed can you merge?

PR #1550, opened 2019-02-08
List all errors in each database test before failing

### LINKED ISSUE #1549: databaseTest could collect all errors instead of reporting just the first
### Motivation or Problem
It takes 8 minutes for me to run `make test-database` locally, and ~30 minutes on travis. When there are many mistakes to be fixed in the database, it's frustrating to have them listed one at a time and have to run the whole test suite again to reveal the next problem.

### Desired Solution
Some of the tests (checking units, checking rates are reasonable) already collect up a list of failures and report them all. This is great. It'd be nice if all the other tests (nodes can be placed in trees, nodes are not duplicates, etc. etc.) did the same thing: log all the problems (with `logging.error`) and then fail the test at the end, rather than failing the test at the first problem.

mjohnson541:
Since the database tests take a while to run this makes it much easier to debug database related issues.  

Check that each test is unlikely to give a false pass and run the database tests to ensure there aren't any false fails.  

Fixes #1549

### DIFF
--- a/testing/databaseTest.py
+++ b/testing/databaseTest.py
@@ -294,10 +294,19 @@ class TestDatabase():  # cannot inherit from unittest.TestCase if we want to use
         """
         family = self.database.kinetics.families[family_name]
         expectedNumberNodes = len(family.getRootTemplate())
+        tst = []
         for label, entries in family.rules.entries.iteritems():
             for entry in entries:
                 nodes = label.split(';')
-                nose.tools.assert_equal(len(nodes), expectedNumberNodes, "Wrong number of groups or semicolons in family {family} rule {entry}.  Should be {num_nodes}".format(family=family_name, entry=entry, num_nodes=expectedNumberNodes))
+                tst.append((len(nodes), expectedNumberNodes, "Wrong number of groups or semicolons in family {family} rule {entry}.  Should be {num_nodes}".format(family=family_name, entry=entry, num_nodes=expectedNumberNodes)))
+
+        boo = False
+        for item in tst:
+            if item[0] != item[1]:
+                boo = True
+                logging.error(item[2])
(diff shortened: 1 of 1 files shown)


testing/databaseTest.py
```diff
@@ -294,10 +294,19 @@ def kinetics_checkCorrectNumberofNodesInRules(self, family_name):
...
+        tst = []
         for label, entries in family.rules.entries.iteritems():
             for entry in entries:
                 nodes = label.split(';')
-                nose.tools.assert_equal(len(nodes), expectedNumberNodes, "Wrong number of groups or semicolons in family {family} rule {entry}.  Should be {num_nodes}".format(family=family_name, entry=entry, num_nodes=expectedNumberNodes))
+                tst.append((len(nodes), expectedNumberNodes, "Wrong number of groups or semicolons in family {family} rule {entry}.  Should be {num_nodes}".format(family=family_name, entry=entry, num_nodes=expectedNumberNodes)))
+
+        boo = False
+        for item in tst:
+            if item[0] != item[1]:
```
mliu49: For all of these tests, my initial thought is that the check should occur when you're iterating through the entries, and you should just store a list of error messages that need to be printed. Then at the end (here), if the list is not empty, log all of the stored messages and fail the test.

It seems cleaner that way. Let me know what you think.

rwest:
Just a note that this commit may have made its way into master via the rmg-cat merge, as it had previously been cherry-picked onto the cat branch.

https://github.com/ReactionMechanismGenerator/RMG-Py/commit/0eb061b32fcbc1bbdf0f73d9d49e7caa17fc2159

mliu49:
Closing because this was merged into master with #1537.

PR #1552, opened 2019-02-20
Allow RDKit to ignore hydrogen bonds

mjohnson541:
This PR lets RMG ignore hydrogen bonds when converting a molecule to an RDKit mol.


alongd:
Looking great! Can you add test/s?

mjohnson541:
Done

PR #1558, opened 2019-03-04
Allow cosolvents to be defined in a solvent library

alongd:
### Motivation or Problem
We need to define a mixture of solvents in the solvent library

### Description of Changes
We now parse both a list of descriptors or just a descriptor as a solvent molecule. The descriptors could be either SMILES or adjLists.
A new method, `get_structure` was created in `SolventLibrary` 

### Testing
Could be tested using -db PR https://github.com/ReactionMechanismGenerator/RMG-database/pull/318

### DIFF
--- a/rmgpy/data/solvation.py
+++ b/rmgpy/data/solvation.py
@@ -277,28 +277,26 @@ class SolventLibrary(Database):
                   shortDesc='',
                   longDesc='',
                   ):
-        spc = molecule
         if molecule is not None:
-            try:
-                spc = Species().fromSMILES(molecule)
-            except:
-                logging.debug("Solvent '{0}' does not have a valid SMILES '{1}'" .format(label, molecule))
-                try:
-                    spc = Species().fromAdjacencyList(molecule)
-                except:
-                    logging.error("Can't understand '{0}' in solute library '{1}'".format(molecule, self.name))
-                    raise
-            spc.generate_resonance_structures()
+            if not isinstance(molecule, list):
+                molecule = [molecule]
+            spc_list = []
+            for mol in molecule:
+                spc0 = Species(label=label)
+                spc0.set_structure(mol)
+                spc_list.append(spc0)
+        else:
+            spc_list = None
 
         self.entries[label] = Entry(
-            index = index,
-            label = label,
-            item = spc,
-            data = solvent,
-            reference = reference,
-            referenceType = referenceType,
-            shortDesc = shortDesc,
-            longDesc = longDesc.strip(),
+            index=index,
+            label=label,
+            item=spc_list,
+            data=solvent,
+            reference=reference,
+            referenceType=referenceType,
+            shortDesc=shortDesc,
+            longDesc=longDesc.strip(),
         )
 
     def load(self, path):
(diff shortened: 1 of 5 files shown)


alongd:
Force pushed a small correction to the test. Should be ready for review.
@oscarwumit, could you change the corresponding database PR to have list representations so we could check this branch?

rmgpy/data/solvation.py
```diff
@@ -324,7 +320,22 @@ def getSolventStructure(self, label):
         Get a solvent's molecular structure as SMILES or adjacency list from its name
         """
         return self.entries[label].item
-        
+
+    def get_structure(self, label, molecule):
```
mjohnson541: I understand the desire to reduce code duplication, but this really shouldn't be a bound function of SolventLibrary.  I suggest adding a similar function to species.py or just having the code duplication.

oscarwumit:
> Force pushed a small correction to the test. Should be ready for review.
> @oscarwumit, could you change the corresponding database PR to have list representations so we could check this branch?

I updated the molecule representation of co-solvent to list.

alongd:
Thanks, @mjohnson541, addressed the comment and added a test.
(reorganized commits and force pushed)

rmgpy/species.py
```diff
@@ -707,9 +707,27 @@ def generateEnergyTransferModel(self):
             alpha0 = (300*0.011962,"kJ/mol"),
             T0 = (300,"K"),
             n = 0.85,
-        ) 
+        )
+
+    def get_structure(self, structure):
```
mliu49: It seems like this should be called `set_structure` or `set_molecule`?

rmgpy/data/solvation.py
```diff
@@ -277,28 +277,26 @@ def loadEntry(self,
...
-                logging.debug("Solvent '{0}' does not have a valid SMILES '{1}'" .format(label, molecule))
-                try:
-                    spc = Species().fromAdjacencyList(molecule)
-                except:
-                    logging.error("Can't understand '{0}' in solute library '{1}'".format(molecule, self.name))
-                    raise
-            spc.generate_resonance_structures()
+            if not isinstance(molecule, list):
+                molecule = [molecule]
+            spc = []
```
mliu49: I would name this variable something like `spcs` or `spc_list` for consistency with expectations.

alongd:
Thanks @mliu49 , I addressed the comments in fixup commits

mliu49:
Thanks! You can squash and rebase. This PR looks ok to me code-wise, though I'm not fully aware of the context for this change. It seems like @mjohnson541 have been resolved, so I think it could be merged if you think it's ready.

alongd:
Thanks! Squashed and pushed.

The context is that we'd like to define a binary solvent mixture, e.g. 50% water and 50% methanol. We can get Abraham parameters from the mixture from literature (https://github.com/ReactionMechanismGenerator/RMG-database/pull/318), but the solvent library forced us to define only one molecular structure for the solvent. Merging this PR will allow us to define something like:
```python
entry(
    index = 27,
    label = "methanol_50_water_50",
    molecule = ["CO", "O"],
    solvent = SolventData(
    ...
```

PR #1561, opened 2019-03-13
Enable Automatic Isodesmic Reaction Generation for Arkane

amarkpayne:
Based heavily on ideas and algorithms provided by 

Buerger, P., Akroyd, J., Mosbach, S., & Kraft, M. (2018). A systematic method to estimate and validate enthalpies of
formation using error-cancelling balanced reactions. Combustion and Flame (Vol. 187).
https://doi.org/10.1016/j.combustflame.2017.08.013

Dobek, F. J., Ranasinghe, D. S., Throssell, K., & Petersson, G. A. (2013). Evaluation of the heats of formation of
corannulene and C60 by means of inexpensive theoretical procedures. Journal of Physical Chemistry A, 117(22), 4726–4730.
https://doi.org/10.1021/jp404158v


davidfarinajr:
How did you generate the yaml files? When do you run the reference species through arkane to get the NASA polynomials? Do you have a script to make them?

rwest:
We've been looking into the YAML thing a bit more.
Seems Arkane already has a way to save and load YAML representations of everything, like NASA polynomials, etc. eg https://github.com/ReactionMechanismGenerator/RMG-Py/blob/bb7175c45df27223a3259dd3e9396b287a6566f3/arkane/common.py#L217

But the yaml looks different from that generated by the yaml Dumper in 
https://github.com/ReactionMechanismGenerator/RMG-Py/blob/81c7cee7b27f5b66df3e7f0ab44bde202884f48a/arkane/benchmark.py#L53

Should we pick one horse to back? Re-use existing I/O routines?

amarkpayne:
I think it is probably best if the yaml files follow the already established Arkane yaml format as much as possible. I'll take a look at this and update this PR.

arkane/isodesmic.py
```diff
@@ -392,6 +392,28 @@ def multiple_error_canceling_reaction_search(self, n_reactions_max=20, milp_soft
...
+
+        """
+        reaction_list = self.multiple_error_canceling_reaction_search(n_reactions_max, milp_software)
+        h298_list = np.zeros(len(reaction_list))
+
+        for i, rxn in enumerate(reaction_list):
+            h298_list[i] = rxn.calculate_target_thermo().value_si
+
+        return ScalarQuantity(np.median(h298_list), 'J/mol')
+
```
davidfarinajr: median seems sensible, although it would be nice if we could incorporate some uncertainty for each reaction and put greater weight on reactions with lower uncertainty (uncertainty related to the known high-level hf of the reference species in the reaction or the low-level dft calc)

arkane/reference.py
```diff
@@ -322,6 +324,53 @@ def load(self, paths=''):
 
             self.reference_sets[set_name] = reference_set
 
+    def extract_model_chemistry(self, model_chemistry, sets=None):
+        """
+        Return a list of ErrorCancelingSpecies objects from the reference species in the database that have entries for
+        the requested model chemistry
+
+        Args:
+            model_chemistry (str): String that describes the level of chemistry used to calculate the low level data
```
davidfarinajr: We have to be careful with naming convention for `model_chemistry`.  Maybe convert everything to lowercase?  Also, perhaps have a `model_chemistries` attribute for the ReferenceDatabase class that is a list of all the model chemistries in that reference set, and then assert that the `model_chemistry` argument for `extract_model_chemistry` method is in `model_chemistries`

arkane/isodesmic.py
```diff
@@ -141,5 +143,102 @@ def calculate_target_thermo(self):
         return ScalarQuantity(target_thermo, 'J/mol')
 
 
+class SpeciesConstraints(object):
+    """
+    A class for defining and enumerating constraints to ReferenceSpecies objects for error canceling reactions
+    """
+
+    def __init__(self, target, reference_species, conserve_bonds=True, conserve_ring_size=True):
+        """
```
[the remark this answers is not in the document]
davidfarinajr: ReferenceDatabase has a method `extract_model_chemistry` which return a list of `ErrorCancelingSpecies` with desired model_chemistry.  This list can then be passed to `SpeciesConstraints` as `reference_species` argument.  I think it might be also useful to have a method in the `ReferenceSpecies` class that converts `ReferenceSpecies` object to `ErrorCancelingSpecies` object.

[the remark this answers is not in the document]
amarkpayne: Thanks for the suggestion, this is actually a much better way to do it. I have also renamed the `reference_species` argument to `reference_list` to avoid confusion

arkane/reference.py
amarkpayne: Good point. We will have to standardize this, ideally with what Arkane currently uses/will use in the future.

arkane/reference.py
```diff
@@ -0,0 +1,430 @@
...
+                             'ReferenceDataEntry object')
+
+    @property
+    def calculated_data(self):
+        return self._calculated_data
+
+    @calculated_data.setter
+    def calculated_data(self, value):
+        if not value:
+            self._calculated_data = None
```
davidfarinajr: I think, if not value, `self._calculated_data` should be an empty dictionary `{}` instead of `None` (same for self._reference_data in line 123). `self.calculated_data[arkane_species.model_chemistry] = calc_data` from `update_from_arkane_spcs` method throws an error if the ReferenceSpecies is initialized without calculated_data.

arkane/reference.py
```diff
@@ -0,0 +1,430 @@
...
+    """
+    def __init__(self, conformer, thermo, thermo_data, t1_diagnotic=None, fod=None):
+        """
+
+        Args:
+            conformer (rmgpy.statmech.Conformer): Conformer object generated from an Arkane job. Stores many peices of
+                information gained from quantum chemistry calculations, including coordinates, frequencies etc.
+            thermo (HeatCapacityModel): `NASA` or `Wilhoit` thermo object to store the fitted polynomials
+            thermo_data (rmgpy.thermo.ThermoData): Actual thermochemistry values calculated using statistical mechanics
+                at select points. Arkane fits a heat capacity model to this data
```
davidfarinajr: Since `thermo` has a `.toThermoData()` function, perhaps we should make the `thermo_data` argument optional, and if `thermo_data` is passed to CalculatedDataEntry(), ensure that `thermo.toThermoData()` is consistent with `thermo_data`.  Alternatively, maybe we should just get rid of the `thermo_data` argument since we can get `thermo_data` from `thermo`

arkane/reference.py
```diff
@@ -0,0 +1,430 @@
...
+
+
+class CalculatedDataEntry(RMGObject):
+    """
+    A class for storing a single entry of statistical mechanical and thermochemistry information calculated at a single
+    model chemistry or level of theory
+    """
+    def __init__(self, conformer, thermo, thermo_data, t1_diagnotic=None, fod=None):
+        """
+
```
davidfarinajr: typo in argument `t1_diagnotic` -> `t1_diagnostic`

arkane/reference.py
amarkpayne: Done, thanks!

arkane/reference.py
amarkpayne: I think the relationship is actually the other way around. ThermoData stores the rigorously correct (up to RRHO) data that we then fit polynomials to. Therefore, I'd rather keep the ThermoData objects since we can derive NASAPolynomials from this data. I have removed the thermo attribute but will add a method to obtain NASAPolynomials from the ThermoData

arkane/reference.py
amarkpayne: Fixed, thanks!

amarkpayne:
Thanks for agreeing to help review this, @alongd ! Please let me know how I can help. As a final note, there are 4 commits on top of this PR that are marked as cherry-picked/temporary. The cherry-picked ones can be removed when #1470 is merged. The temporary commit to change the database branch will be removed when the twin database PR is merged. The last temporary commit will be dealt with shortly, but it should not impact the review.

arkane/reference.py
```diff
@@ -0,0 +1,180 @@
...
+                species = Species(smiles=smiles)
+            elif inchi:
+                species = Species(inchi=inchi)
+            elif adjacency_list:
+                species = Species().from_adjacency_list(adjacency_list)
+            else:
+                raise ValueError('Either an rmgpy species object, smiles string, InChI string, or an adjacency list '
+                                 'must be given to create a ReferenceSpecies object')
+
+        super(ReferenceSpecies, self).__init__(species=species, label=label, **kwargs)
```
cgrambow: Replace with `super().__init__(...)`

arkane/reference.py
```diff
@@ -0,0 +1,180 @@
...
+    def reference_data(self):
+        return self._reference_data
+
+    @reference_data.setter
+    def reference_data(self, value):
+        if not value:
+            self._reference_data = {}
+        elif isinstance(value, dict):
+            if all(isinstance(source, str) for source in value.keys()):
+                if all(isinstance(data_entry, ReferenceDataEntry) for data_entry in value.values()):
```
cgrambow: If these if-statements are not true, `self._reference_data` does not get set and no error is raised. Is that desired?

arkane/reference.py
```diff
@@ -0,0 +1,180 @@
...
+    def calculated_data(self):
+        return self._calculated_data
+
+    @calculated_data.setter
+    def calculated_data(self, value):
+        if not value:
+            self._calculated_data = {}
+        elif isinstance(value, dict):
+            if all(isinstance(source, str) for source in value.keys()):
+                if all(isinstance(data_entry, CalculatedDataEntry) for data_entry in value.values()):
```
cgrambow: Same here

arkane/reference.py
```diff
@@ -0,0 +1,180 @@
...
+        pass
+
+
+class CalculatedDataEntry(RMGObject):
+    """
+    A class for storing a single entry of statistical mechanical and thermochemistry information calculated at a single
+    model chemistry or level of theory
+    """
+    def __init__(self):
+        super(CalculatedDataEntry, self).__init__()
```
cgrambow: `super().__init__()`

arkane/ess/qchem.py
```diff
@@ -264,8 +264,8 @@ def load_conformer(self, symmetry=None, spin_multiplicity=0, optical_isomers=Non
                         rot.append(rotation)
 
                     inertia = []
-
-        modes = mmass + rot + freq
+        # Take only the lat modes found (in the event of multiple jobs)
```
cgrambow: Typo

arkane/ess/qchem.py
```diff
@@ -264,8 +264,8 @@ def load_conformer(self, symmetry=None, spin_multiplicity=0, optical_isomers=Non
...
+        modes = mmass[-1:] + rot[-1:] + freq[-1:]
```
cgrambow: Why are these lists in the first place if you only want the last one? I'm skeptical that information that may be wanted in other cases isn't being discarded here.

arkane/ess/qchem.py
```diff
@@ -264,8 +264,8 @@ def load_conformer(self, symmetry=None, spin_multiplicity=0, optical_isomers=Non
                         rot.append(rotation)
```
cgrambow: This isn't your code but why is `rot.append(rotation)` not in the for-loop? Only the rotation for `i=2` gets used.

arkane/reference.py
```diff
@@ -245,5 +250,62 @@ def thermo_data(self, value):
...
+    A class for loading and working with database of reference species, located at RMG-database/input/reference_sets/
+    """
+    def __init__(self):
+        """
+        Attributes:
+            self.reference_sets (dict): {'set name': [ReferenceSpecies, ...], ...}
+        """
+        self.reference_sets = {}
+
+    def load(self, paths=''):
```
cgrambow: Why is `''` the default instead of `None`?

arkane/referenceTest.py
```diff
@@ -137,5 +138,36 @@ def test_calculated_data_entry(self):
...
+    """
+    Test that the ReferenceDatabase class functions properly
+    """
+
+    def test_load_main_reference_set(self):
+        """
+        Test that the main reference set can be loaded properly
+        """
+        database = ReferenceDatabase()
+        database.load()
```
cgrambow: Consider separating into a setUpClass method if you think that additional methods will get added in the future.

arkane/data/species/reference_species_example.yml
```diff
@@ -0,0 +1,164 @@
+RMG_version: 2.4.0
```
cgrambow: Update or remove?

arkane/isodesmic.py
```diff
@@ -421,5 +443,13 @@ def __init__(self, target, reference_set):
         super(IsodesmicScheme, self).__init__(target, reference_set, conserve_bonds=True, conserve_ring_size=False)
 
 
+class IsodesmicRingScheme(ErrorCancelingScheme):
+    """
+    A stricter form of the traditional isodesmic reaction scheme where the number of each ring size is also conserved
+    """
+    def __init__(self, target, reference_set):
+        super(IsodesmicRingScheme, self).__init__(target, reference_set, conserve_bonds=True, conserve_ring_size=True)
```
cgrambow: `super().__init__(...)`

arkane/reference.py
```diff
@@ -121,10 +121,8 @@ def reference_data(self):
     def reference_data(self, value):
         if not value:
             self._reference_data = {}
-        elif isinstance(value, dict):
-            if all(isinstance(source, str) for source in value.keys()):
-                if all(isinstance(data_entry, ReferenceDataEntry) for data_entry in value.values()):
-                    self._reference_data = value
+        elif isinstance(value, dict) and _is_valid_reference_data(value):
+            self._reference_data = value
```
cgrambow: I think I commented on this in an outdated version (commits aren't in order), but I think this change might fix my previous comment.

arkane/isodesmicTest.py
```diff
@@ -50,13 +50,14 @@ class TestErrorCancelingReactionAndSpecies(unittest.TestCase):
     Tests that ErrorCancelingReaction objects and ErrorCancelingSpecies object are properly implemented
     """
 
-    def setUp(self):
+    @classmethod
```
cgrambow: I didn't see this commit until later and I made some comments about this during my review. This should be squashed.

arkane/isodesmic.py
```diff
@@ -293,6 +298,11 @@ def _find_error_canceling_reaction(self, reference_subset, milp_software='lpsolv
         solution = None
 
         if milp_software == 'pyomo':
+            # Check that pyomo is available
+            if not pyo:
```
cgrambow: Check for `None` is done using `if pyo is None`

arkane/isodesmic.py
```diff
@@ -293,6 +298,11 @@ def _find_error_canceling_reaction(self, reference_subset, milp_software='lpsolv
...
+                raise ImportError('Can not import optional package pyomo. Either install this dependency with '
```
cgrambow: 'Cannot' is one word

arkane/isodesmicTest.py
```diff
@@ -241,11 +248,12 @@ def test_multiple_error_canceling_reactions(self):
         rxn_str2 = '<ErrorCancelingReaction 1.0*C=CC + 1.0*CCCC == 1.0*C=CCC + 1.0*CCC >'
         self.assertTrue(any(rxn_string in reaction_string for rxn_string in [rxn_str1, rxn_str2]))
 
-        # pyomo is slower, so don't test as many
-        reaction_list = scheme.multiple_error_canceling_reaction_search(n_reactions_max=5, milp_software='pyomo')
-        self.assertEqual(len(reaction_list), 5)
-        reaction_string = reaction_list.__repr__()
-        self.assertTrue(any(rxn_string in reaction_string for rxn_string in [rxn_str1, rxn_str2]))
+        if self.pyo:
```
cgrambow: Again, check that not `None`.

arkane/isodesmicTest.py
```diff
@@ -258,8 +266,9 @@ def test_calculate_target_enthalpy(self):
         self.assertEqual(target_thermo.value_si, 115000.0)
         self.assertIsInstance(rxn_list[0], ErrorCancelingReaction)
 
-        target_thermo, _ = scheme.calculate_target_enthalpy(n_reactions_max=3, milp_software='pyomo')
-        self.assertEqual(target_thermo.value_si, 115000.0)
+        if self.pyo:
```
cgrambow: Same

arkane/reference.py
```diff
@@ -399,6 +399,8 @@ def thermo_data(self, value):
                 self._thermo_data = value
             else:
                 raise ValueError('thermo_data for a CalculatedDataEntry object must be an rmgpy ThermoData object')
+        else:
```
cgrambow: Squash this commit

arkane/reference.py
```diff
@@ -369,6 +381,32 @@ def thermo_data(self, value):
             else:
                 raise ValueError('thermo_data for a CalculatedDataEntry object must be an rmgpy ThermoData object')
 
+    @property
+    def electronic_energy(self):
+        return self._electronic_energy
+
+    @electronic_energy.setter
+    def electronic_energy(self, value):
+        if value:
```
cgrambow: Still not sure why `if value` is needed

arkane/data/species/reference_species_example.yml
```diff
@@ -19,157 +19,86 @@ Properties:
     units: amu
     value: 39.05592359170969
   multiplicity: 2
-RMG_version: 2.4.0
+  symmetry_number: 2.0
```
cgrambow: Squash commit

arkane/data/species/supporting_information.csv
```diff
@@ -0,0 +1,2 @@
+Label,Symmetry Number,Number of optical isomers,Symmetry Group,Rotational constant (cm-1),"Calculated Frequencies (unscaled and prior to projection, cm^-1)",Electronic energy (J/mol),"E0 (electronic energy + ZPE, J/mol)",E0 with atom and bond corrections (J/mol),Atom XYZ coordinates (angstrom),T1 diagnostic,D1 diagnostic
```
cgrambow: What is the purpose of this commit?

arkane/reference.py
```diff
@@ -93,7 +93,8 @@ def __init__(self, species=None, smiles=None, adjacency_list=None, inchi=None, r
             elif inchi:
                 species = Species(inchi=inchi)
             elif adjacency_list:
-                species = Species().from_adjacency_list(adjacency_list)
+                species = Species().from_adjacency_list(adjacency_list, raise_atomtype_exception=False,
```
cgrambow: Oh, this is great. Does this mean it'll now work for all species in the reference dataset?

arkane/common.py
```diff
@@ -142,13 +149,34 @@ def __repr__(self):
...
+            elif attr in self._properties:
+                new_dictionary['Properties'].update({attr: value})
+            else:
+                new_dictionary.update({attr: value})
+
+        return new_dictionary
+
+    def make_object(self, data, class_dict):
+        data = expand_formatted_blocks(data)
+        super(ArkaneSpecies, self).make_object(data, class_dict)
```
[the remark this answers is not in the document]
amarkpayne: Thanks for the suggestion! I should have changed this when upgrading to python 3

arkane/isodesmic.py
```diff
@@ -139,5 +141,102 @@ def calculate_target_thermo(self):
...
+    def _get_constraint_map(self):
+        # Enumerate all of the constraints in the target molecule to initialize the constraint mapping
+        constraint_map = {label: i for i, label in enumerate(self.target.molecule.get_element_count().keys())}
+        if self.conserve_bonds:
+            j = len(constraint_map)
+            constraint_map.update(
+                {label: j + i for i, label in enumerate(self.target.molecule.enumerate_bonds().keys())})
+        if self.conserve_ring_size:
+            j = len(constraint_map)
+            possible_rings_sizes = set(map(lambda x: '{0}_ring'.format(len(x)),
```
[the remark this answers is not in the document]
amarkpayne: I switched to a set comprehension. I want to keep the `_ring` strings though because this are human readable descriptions of the features

arkane/ess/qchem.py
amarkpayne: Good question. This and the point about `rot.append(rotation)` makes me think that this code could use a second glance and a cleanup/fixup PR of its own. I'll talk with @alongd about this and open a separate issue.

arkane/reference.py
amarkpayne: No reason other than I set the type hint for this variable a string or list. I'll make this change though

arkane/reference.py
amarkpayne: This is one of the ones where it could be None or an empty string if read from the supporting information file

arkane/data/species/supporting_information.csv
amarkpayne: The supporting information file is the only way to get information like unscaled frequencies, electronic energy, and T1 diagnostic easily from an Arkane job after the fact

arkane/reference.py
amarkpayne: It should, yes, though I think there are some edge cases for raise_atomtype_exception that I need to update

amarkpayne:
It wasn't easy, but after a few hours of effort and learning the glorious capabilities of `git add -p` in edit mode, I have finally recommited the commits in this PR to squash a lot of the fixups. I verified before pushing that the PR is exactly the same as before, with the exception of two minor fixes I caught along the way.

@alongd this PR should be ready for your review now. Let me know how I can help!

amarkpayne:
Previously this PR was failing tests due to memory errors on Travis that mostly seem to be unrelated to this PR. After making some adjustments, this PR should now be passing, so I will stop pushing to this branch until the next round of review comments. 

@davidfarinajr this branch is stable now if you want to take a look to see how we can merge in your additional changes in a future PR.

arkane/reference.py
```diff
@@ -0,0 +1,638 @@
...
+    A class for loading and working with database of reference species, located at RMG-database/input/reference_sets/
+    """
+    def __init__(self):
+        """
+        Attributes:
+            self.reference_sets (Dict[str, ReferenceSpecies]): {'set name': [ReferenceSpecies, ...], ...}
+        """
+        self.reference_sets = {}
+
+    def load(self, paths=None, ignore_incomplete=True):
```
davidfarinajr: It would be nice if we had a `save` function too for the database.  It could just iterate through the species in the database and call the `save_yaml` method for each species.

arkane/statmech.py
```diff
@@ -497,7 +521,7 @@ def load(self, pdep=False, plot=False):
                                            multiplicity=conformer.spin_multiplicity)
             else:
                 bond_corrections = 0
-            e_electronic_with_corrections = e_electronic + atom_corrections + bond_corrections
+            e_electronic_with_corrections = e_electronic + atom_corrections + bond_corrections + isodesmic_correction
```
davidfarinajr: This isodesmic correction is at 298K (which implicitly includes the atom thermal corrections).  For an accurate `E0`, I think we should be subtracting the atom thermal corrections here, then add them back to calculate `H298K`.  When I did this, my `H0` (which I assume is E0?) was in better agreement with literature (ATcT).  This is related to PR #1871. If E0 != H0 though, then this isn't necessary.

arkane/statmech.py
```diff
@@ -701,6 +725,35 @@ def load(self, pdep=False, plot=False):
...
+
+            scheme = IsodesmicRingScheme(target=ErrorCancelingSpecies(self.species.molecule[0],
+                                                                      (uncorrected_thermo, 'J/mol'),
+                                                                      self.modelChemistry),
+                                         reference_set=reference_db.extract_model_chemistry(self.modelChemistry))
+            isodesmic_thermo, self.isodesmicReactionList = scheme.calculate_target_enthalpy()
+
+            # Set the difference as the isodesmic EO correction and re-run the statmech job
+            self.isodesmicCorrection = isodesmic_thermo.value_si - uncorrected_thermo
+            self.load(pdep)
```
alongd: what's the role of this line?

arkane/util.py
```diff
@@ -73,3 +77,46 @@ def determine_qm_software(fullpath):
             raise InputError(f'The file at {fullpath} could not be identified as a '
                              'Gaussian, Molpro, QChem, or TeraChem log file.')
     return software_log
+
+
+def read_supporting_information(path):
```
alongd: since Arkane wrote the SI file, do we really need to parse it?

arkane/common.py
```diff
@@ -142,13 +149,34 @@ def __repr__(self):
         result += '}'
         return result
 
+    def as_dict(self):
+        dictionary = super(ArkaneSpecies, self).as_dict()
+        new_dictionary = {'Identifiers': {}, 'Properties': {}}
+
+        # Add additional formatting for identifiers, and properties
+        for attr, value in dictionary.items():
```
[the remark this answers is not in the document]
alongd: My only concern is that we might want a different arrangement in the future, adding or changing categories, which will break compatibility. I'm OK with leaving it as is if you don't think it's an issue

arkane/common.py
```diff
@@ -339,6 +352,26 @@ def replace_yaml_syntax(content, label=None):
     return content
 
 
+def expand_formatted_blocks(data_dictionary):
+    """
+    Unpacks attributes contained in formatted blocks (if present) from an ArkaneSpecies dictionary obtained from a
+    YAML file
+    Args:
```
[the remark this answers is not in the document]
alongd: Add an empty line before `Args:`?

arkane/util.py
alongd: But isn't this YAML file being saved before Arlane termiates? Or did I miss the expected workflow?

arkane/common.py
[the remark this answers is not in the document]
amarkpayne: You are correct that if we decide that we want to change the name of (or get rid of for that matter) the Identifiers or Properties block, then we will have to either leave this code in or break backwards compatibility.

On the other hand, if we want to add/remove items from these blocks then we can do so without breaking backwards compatibility. This is because the code for reading in YAML files relies on the `expand_formatted_blocks` function, which takes in whatever is underneath these blocks (it does not need to know which items are allowed to be underneath the identifiers block).

I personally like having all of the Identifiers and Properties grouped together, as otherwise they can be difficult to find. However, I don't want to cause us to have to keep legacy code if we want to change this later.

I'll let you @alongd (as the principle designer of the ArkaneSpecies YAML format) decide if you think you will want to keep this change going forward. This would be simple for me to remove from this PR, and it can always be added into another PR later. Let me know what you decide, I am happy with either decision. Sorry for all of the back and forth

arkane/util.py
amarkpayne: The csv file is being saved before Arkane terminates, and contains information not found in the species YAML file. I added this in case the user ran an Arkane job say a month ago and then realized that they wanted to update a reference species with that information (they would only be able to use the information contained in these files, and the CSV file is the only place that has this information).

That being said, I have decided to cut this function from the PR. We would have to add back in the `util.py` file just for this, and we would have to maintain this function whenever we change the format. If the user wants to add in this information they can always do so by hand or write a script for themselves.

The preferred way going forward would be to have Arkane update reference species files before terminating, but this is will be included in a future PR.

amarkpayne:
@alongd I think this PR is ready for its final review. Since you last reviewed I have done the following:

- Squashed all fixup commits
- Updated the Arkane documentation relevant for this PR on how to use isodesmic reactions. Further documentation needs to be added in future PRs once the final touches have been added, specifically on how to add new model chemistries into Arkane using the new Reference Species database. I have made a note of this on the GitHub project kanban board
- Dropped the CSV reader functionality
- Marked all comments on this review as resolved as appropriate. Besides the comment on the identifiers and properties block (which needs addressing) I believe all unresolved comments relate to requested functionality that will be the subject of future PRs. I am in the progress of updating the project kanban board with these so that they do not get lost in time.

Let me know what you think about Identifiers/Properties. If everything else looks good, then we should give the database PR a quick look and merge that in so that I remove the temporary commits  from this one. Thanks again for all of your help with this!

amarkpayne:
@alongd I have removed the temporary commit from this PR. Let me know if there are any other comments or changes you would like me to make.

arkane/ess/qchem.py
alongd: I'll make a new PR for this

alongd:
The code looks great, thanks!!
Could you please add Arkane thermo calc examples?

amarkpayne:
I had added an example before, but we were encountering memory issues. I created a new example, and tried to make the QM output files as small as possible (though I don't think they contribute much anyways). It ran fine on my local machine, so let's hope that Travis handles it okay.

amarkpayne:
Thanks for all of the hard work you put in to reviewing and help me fine tune this PR @alongd, @davidfarinajr and @cgrambow !! I really appreciate it

PR #1565, opened 2019-03-21
Initialize the global RMGDatabase object upon calling it

alongd:
Initialize the global `RMGDatabase` object upon calling it  even (and especially) if it already exists.
This solves cases when using RMG's API where the database doesn't get appropriately updated.

### DIFF
--- a/rmgpy/data/rmg.py
+++ b/rmgpy/data/rmg.py
@@ -50,7 +50,8 @@ database = None
 
 ################################################################################
 
-class RMGDatabase:
+
+class RMGDatabase(object):
     """
     The primary class for working with the RMG database.
     """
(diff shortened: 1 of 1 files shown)


mliu49:
I think this looks good. Do we want the re-initialization message to be debug or warning? Personally, I'm leaning more towards warning since I think users should be aware that it is happening.

alongd:
done!

mjohnson541:
Looks good to me!

rmgpy/data/rmg.py
```diff
@@ -62,15 +63,12 @@ def __init__(self):
...
         # Store the newly created database in the module.
         global database
-#        assert database is None, "Should only make one instance of RMGDatabase because it's stored as a module-level variable."
-        if database is None:
-            database = self
-        else:
-            logging.warning("Should only make one instance of RMGDatabase because it's stored as a module-level variable!")
-            logging.warning("Unexpected behaviour may result!")
+        if database is not None:
+            logging.warn('An instance of RMGDatabase already exists. Re-initializing it.')
```
mliu49: I think we should use `logging.warning`. `logging.warn` is just a pointer to `logging.warning`, and is now deprecated in Python 3.

alongd:
good to know, done

alongd:
I can approve that the RMG/ARC feature works as expected when cherry-picking this commit.
In fact, and strangely, the implementation of resetting the database after each call to RMG or ARC via something like
```python
def clear_rmg_database():
    """Set the RMG database object instance to None"""
    rmg_db = getDB()
    rmg_db.thermo = None
    rmg_db.transport = None
    rmg_db.forbiddenStructures = None
    rmg_db.kinetics = None
    rmg_db.statmech = None
    rmg_db.solvation = None
```
was problematic, and in the second iteration RMG comlained about missing components in the database (e.g., can't check forbidden structures since `database.forbiddenStructures` was `None`; when I manually initialized it with the correct object, it complained that a different part was None). I thought I did use this function in an order that makes sense.

In any case, I'm very happy with the implementation in this PR.

PR #1566, opened 2019-03-26
Corrected `raise Error as e` syntax

alongd:
Corrected all instances of `raise Error, e:` to the correct syntax `raise Error as e:`. The former won't be supported in Py3. 
Also, if `e` isn't used, it was deleted.
See https://www.python.org/dev/peps/pep-0008/#id51


rmgpy/rmg/model.py
```diff
@@ -354,7 +354,7 @@ def makeNewSpecies(self, object, label='', reactive=True, checkForExisting=True)
         try:
             spec = Species(index=speciesIndex, label=label, molecule=[molecule], reactive=reactive,
                  thermo=object.thermo, transportData=object.transportData)
-        except AttributeError, e:
+        except AttributeErro:
```
mliu49: Letter missing.

rmgpy/rmg/model.py
```diff
@@ -1952,7 +1952,7 @@ def getSpecies(self, obj):
             try:
                 spc = self.indexSpeciesDict[obj]
                 return spc
-            except KeyError, e:
+            except KeyError as e:
```
mliu49: This seems like an unnecessary try/except clause...

alongd:
OK, let me know if that's OK and I'll squash

mliu49:
Thanks! The fixups look good.

alongd:
Should be good to go

PR #1571, opened 2019-04-03
Auto optical isomer ex symmetry arkane

goldmanm:
### Motivation or Problem
Currently Arkane requires optical isomer input.

### Description of Changes
This PR allows Arkane to estimate optical isomer and symmetry using the qm symmetry package.

### Testing
* I wrote test methods to ensure that the arkane
* I have not tested compatibility with ARC, which may cause an issue since the redundant `symfromlog` boolean was removed.

### Reviewer Tips
Run an arkane calculation with & without the symmetry and optical isomer parameter.

### DIFF
--- a/arkane/gaussianTest.py
+++ b/arkane/gaussianTest.py
@@ -54,7 +54,7 @@ class GaussianTest(unittest.TestCase):
         """
 
         log = GaussianLog(os.path.join(os.path.dirname(__file__),'data','ethylene.log'))
-        conformer, unscaled_frequencies = log.loadConformer(symfromlog=True)
+        conformer, unscaled_frequencies = log.loadConformer()
         E0 = log.loadEnergy()
         
         self.assertTrue(len([mode for mode in conformer.modes if isinstance(mode,IdealGasTranslation)]) == 1)
@@ -81,7 +81,7 @@ class GaussianTest(unittest.TestCase):
         """
 
         log = GaussianLog(os.path.join(os.path.dirname(__file__),'data','oxygen.log'))
-        conformer, unscaled_frequencies = log.loadConformer(symfromlog=True)
+        conformer, unscaled_frequencies = log.loadConformer()
         E0 = log.loadEnergy()
         
         self.assertTrue(len([mode for mode in conformer.modes if isinstance(mode,IdealGasTranslation)]) == 1)
@@ -109,7 +109,7 @@ class GaussianTest(unittest.TestCase):
         """
 
         log = GaussianLog(os.path.join(os.path.dirname(__file__),'data','ethylene_G3.log'))
-        conformer, unscaled_frequencies = log.loadConformer(symfromlog=True)
+        conformer, unscaled_frequencies = log.loadConformer()
         E0 = log.loadEnergy()
         
         self.assertTrue(len([mode for mode in conformer.modes if isinstance(mode,IdealGasTranslation)]) == 1)
--- a/arkane/gaussianTest.py
+++ b/arkane/gaussianTest.py
@@ -37,6 +37,7 @@ import rmgpy.constants as constants
 from external.wip import work_in_progress
 
 from arkane.gaussian import GaussianLog
+from arkane.statmech import determine_qm_software
 
 ################################################################################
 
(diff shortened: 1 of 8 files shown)


alongd:
@goldmanm, many of Arkane's tests still use the `symfromlog` argument, could you resolve it so the tests will pass?

goldmanm:
@alongd, fixes for the tests should be back up. I was trying to fix a codacy issue and ended up pushing outdated version of the branch. force pushing is dangerous.

arkane/log.py
```diff
@@ -77,3 +83,36 @@ def loadNegativeFrequency(self):
...
+            )
+            settings = type(str(''), (), dict(symmetryPath=str('symmetry'), scratchDirectory=scr_dir))()  # Creates anonymous class
+            pgc = PointGroupCalculator(settings, unique_id, qmdata)
+            pg = pgc.calculate()
+            if pg is not None:
+                optical_isomers = 2 if pg.chiral else 1
+                symmetry = pg.symmetryNumber
+                logging.debug("Symmetry algorithm found {0} optical isomers and a symmetry number of {1}".format(optical_isomers,symmetry))
+            else:
+                logging.warning("Symmetry algorithm errored when computing optical isomers. Setting to 1")
```
alongd: I think this should be a `logging.error` level, and the text should tell the user to try re-running and supply the missing values (unless we want to raise an error?)
Also, the current message only refers to opt isomers, symmetry could be problematic as well

arkane/log.py
```diff
@@ -77,3 +83,36 @@ def loadNegativeFrequency(self):
...
+                optical_isomers = 1
+                symmetry = 1
+            return optical_isomers, symmetry
+        finally:
```
alongd: nice implementation!

arkane/statmech.py
```diff
@@ -253,17 +253,13 @@ def load(self, pdep=False):
 
         try:
             linear = local_context['linear']
-            symfromlog = False
         except KeyError:
             externalSymmetry = None
```
alongd: (not introduced in this PR): Could you also remove `externalSymmetry = None` here? The next try except block sets `externalSymmetry` to either the user input or None anyway. Let's fix it while we're here.
I'm thinking:
```
try:
    linear = local_context['linear']
except KeyError:
    logging.error('some helpful message')
    raise
```
I have a function for automatically determining whether a species is linear (using xyz) + tests in ARC, I'll migrate it from ARC here if `linear` isn't given by the user instead of raising an error.

arkane/molpro.py
```diff
@@ -265,8 +267,6 @@ def loadConformer(self, symmetry=None, spinMultiplicity=0, opticalIsomers=None,
 
         # Close file when finished
         f.close()
-        if opticalIsomers is None:
```
alongd: Maybe merge this and the `Get loadConformer to automatically estimate optical isomers` commits instead of deleting?

arkane/gaussianTest.py
```diff
@@ -130,5 +130,25 @@ def testLoadEthyleneFromGaussianLog_G3(self):
...
+        """
+        Uses a Gaussian03 log file for oxygen (O2) to test that its
+        molecular degrees of freedom can be properly read.
+        """
+
+        log = GaussianLog(os.path.join(os.path.dirname(__file__),'data','oxygen.log'))
+        optical, symmetry = log.get_optical_isomers_and_symmetry_number()
+        self.assertEqual(optical,1)
+        self.assertEqual(symmetry,2)
+
```
alongd: Could you also test `self.assertIsInstance(log, GaussianLog)`?

arkane/log.py
goldmanm: I wonder how making this an error might cause issues when integrating with ARC (or other automated software generation). I guess we just have to integrate the code in a way such that a factor of two error does not destroy the whole RMG run.

goldmanm: Changed to logging.error

arkane/gaussianTest.py
goldmanm: Great idea! I made the assertIsInstance an entirely new test, using the `determine_QM_software` method

arkane/molpro.py
goldmanm: I don't really understand what two commits you'd like me to merge.

alongd: commit https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1571/commits/dfe4c606e559b61be95121874dd56ea1884ba707 added the lines
```python
if opticalIsomers is None:
    opticalIsomers = self.get_optical_isomers_and_symmetry_number()[0]
```
and commit https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1571/commits/7f7a91e213920d67d94bba30f35f40bd02bf19cd removes them.
Maybe merge these two, or don't add the lines to begin with if they're deleted here?

goldmanm: Oh yeah. My goal behind that was to separate commits which make small changes that each commit can function properly on their own, which is why there were lines added and removed. I'll combine these two commits as you suggested.

goldmanm: Just merged those two commits with the commit between them, which removed `symfromlog` parameter.

alongd:
@goldmanm, let's rebase and merge this PR, then do the respective ARC PR

Issue #1574, opened 2019-04-08
Arkane Error for inadequate information for TST


cjmcgill:
### Bug Description
I am getting the below error for one of my reactions when submitted to Arkane. The problem seems to be from either the reactant MeTPA or the transition state. When I swap out for a different transition state this runs, but I also get the failure with other reactions using MeTPA elsewhere. Thermo calculations of the MeTPA and the transition state both go as normal. These reactions are likely to have negative apparent activation energies, in case that's a contributor.

```
The species in reaction MeTPA + S22 <=> EG + S24 do not have adequate information for TST, using default kinetics values.

Intel MKL ERROR: Parameter 6 was incorrect on entry to DGELSD.
Traceback (most recent call last):
  File "../../RMG-Py/arkane.py", line 60, in <module>
    arkane.execute()
  File "/home/cjmcgill/RMG/RMG-Py/arkane/main.py", line 298, in execute
    job.execute(outputFile=outputFile, plot=self.plot)
  File "/home/cjmcgill/RMG/RMG-Py/arkane/kinetics.py", line 137, in execute
    self.save(outputFile)
  File "/home/cjmcgill/RMG/RMG-Py/arkane/kinetics.py", line 264, in save
    kineticsrev = Arrhenius().fitToData(Tlist, numpy.array(krevs), kunits=self.krunits)
  File "rmgpy/kinetics/arrhenius.pyx", line 134, in rmgpy.kinetics.arrhenius.Arrhenius.fitToData  File "rmgpy/kinetics/arrhenius.pyx", line 166, in rmgpy.kinetics.arrhenius.Arrhenius.fitToDataIndexError: index 0 is out of bounds for axis 0 with size 0

```

### Installation Information
Describe your installation method and system information.
 - OS: Windows 10 in Ubuntu subenvironment
 - Installation method: source install with conda
 - RMG version information:
   
   - RMG-Py: 2.1.0-1519-gae79625

alongd:
I think Arkane can only calculate rates if the energy of the TS is higher than both the reactants and the products wells. If by negative apparent activation energy you mean that the expected Ea in the final Arhhenius fit is negative, that might be OK (I think?), as long as the physical barrier is positive (as described above). Let us know what you're wells and TS energies are. Otherwise, we'll need to play around with your inputs to see what's going on (at least I don't have a better guess from looking at the trace)

mjohnson541:
I'm not sure about 
```
[...]

PR #1575, opened 2019-04-09
Improve error/warning messages concerning energy barriers in Arkane

alongd:
Added a warning if a reaction barrier is too high in Arkane.
Also enhanced a related error message in tunnelling.
Minor: corrected reciprocal T axis units in Arkane plots (`1000/K` is strange, should be either `1/K` or `K^-1`)

### DIFF
--- a/arkane/kinetics.py
+++ b/arkane/kinetics.py
@@ -242,7 +242,7 @@ class KineticsJob(object):
                 except (SpeciesError, ZeroDivisionError):
                     k = reaction.getRateCoefficient(T)
                     kappa = 0
-                    logging.info("The species in reaction {} do not have adequate information for TST, "
+                    logging.info("The species in reaction {0} do not have adequate information for TST, "
                                  "using default kinetics values.".format(reaction))
                 tunneling = reaction.transitionState.tunneling
                 ks.append(k)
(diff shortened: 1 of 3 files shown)


alongd:
Addresses #1574

cjmcgill:
The 1000/K axis is a pretty common convention to get the axis numbers into a more readily readable range, numbers in the ones or tenths digit instead of being in the thousandths digit and a bunch of leading zeros. The 1000/K axis is strangely artificial though  and even though it makes the results more readable, it does not make them any more intelligible.

Might I suggest instead adding a second axis, either in K or C? Something like what this link describes doing in Origin. Then you can keep either 1000/K or 1/K on the primary axis to keep the linearity of the Arrhenius plot, but the secondary axis makes it more easily understood which is probably your goal.

[http://blog.originlab.com/graphing/display-celsius-as-top-axis-in-arrhenius-plot](http://blog.originlab.com/graphing/display-celsius-as-top-axis-in-arrhenius-plot)

alongd:
Thanks for the suggestion, I agree it looks better to have a dual T axis.
I'll clarify that my original comment was just for the axis units (should be `1/K` and not `1000/K` as we currently have), I did not mean to change the axis scaling (we'll keep the primary axis at `1000/T`, not `1/T`).
Again, thanks for the helpful comment, I'll try to implement it!

cjmcgill:
Oh I see! Yeah, having the units be 1000/K makes it ambiguous if the transform has already been done or done twice. Weird minor error. Carry on!

alongd:
@goldmanm, could you take a look?

alongd:
@goldmanm, could you take a look at this PR?

alongd:
@goldmanm, could you take a look at this PR?

goldmanm:
@alongd, I don't know how I missed the messages before. Looking at it now.

PR #1577, opened 2019-04-16
Custom filter criteria per reaction family

ajocher:
### Motivation or Problem
By filtering reactions we add a pre-filtering step before the step of reacting species together (slow), which prevents species from reacting together when the reactions are expected to be negligible throughout the simulation. Currently, unimolecularThreshold, bimolecularThreshold, and trimolecularThreshold are binary arrays storing flags for whether a species or a pair of species are above a reaction threshold. For a unimolecular rate, this threshold is set to True if the unimolecular rate of reaction 𝑘 reaction  k for a species A 
image 
at any given time 𝑡 t in the reaction system, where
image
For a bimolecular reaction occuring between species A and B, this threshold is set to True if the bimolecular rate
image
where 𝑘𝑡ℎ𝑟𝑒𝑠ℎ𝑜𝑙𝑑=filterThreshold is set by the user in the input file and its default value is 
image.
Similarly, for a trimolecular reaction, the following expression is used:
image
where
image
The threshold values can be refined for each reaction family to speed reaction generation up.

### Description of Changes
In this pull request we will refine the threshold values for uni-/bi- and trimolecular reactions by evaluating an Arrhenius fit for each reaction family at the user defined reactor temperature. The Arrhenius fits were obtained by extracting the highest reaction rates from all training reactions of one family for a discrete set of temperatures. The ipython notebook is RMG-Py/ipython/kinetics_training_analysis.ipynb.

### Testing
Tested superminimal and minimal example with reaction filtering turned on. The latter one also with pdep.

### Reviewer Tips
This pull request shouldn't dramatically change the core model of your own test case, but speed reaction generation up.

Modified Travis to simultaneously update RMG-Py and RMG-database. When merging, the code reviewer should remove the temporary commit on RMG-Py.


rmgpy/data/kinetics/family.py
```diff
@@ -3868,3 +3872,82 @@ def getObjectiveFunction(kinetics1,kinetics2,obj=informationGain,T=1000.0):
...
+    def load_yaml(self, path, class_dictionary):
+        with open(path, 'r') as f:
+            data = yaml.safe_load(stream=f)
+
+        data = {key:data[key] for key in data if key!='class'}
+        args = recursive_make_object(data, class_dictionary)
+        self.__init__(**args)
+
+
+def loadFilterFits(self):
```
mliu49: My first impression is that this should be a method of the KineticsDatabase class. Let me know if you considered that and whether there's a reason why that wouldn't work.

ajocher: Good idea, loadFilterFits is now a method of the KineticsDatabase class. Where would you recommend to include the unit-test (currently in familyTest.py; do we want to start a databaseTest.py in this folder)?

rmgpy/solver/base.pyx
```diff
@@ -255,11 +269,17 @@ cdef class ReactionSystem(DASx):
         self.networkLeakRates = numpy.zeros((self.numPdepNetworks), numpy.float64)
         self.maxNetworkLeakRateRatios = numpy.zeros((len(self.prunableNetworks)), numpy.float64)
         self.sensitivityCoefficients = numpy.zeros((self.numCoreSpecies, self.numCoreReactions), numpy.float64)
-        self.unimolecularThreshold = numpy.zeros((self.numCoreSpecies), bool)
-        self.bimolecularThreshold = numpy.zeros((self.numCoreSpecies, self.numCoreSpecies), bool)
+        self.unimolecularThreshold = numpy.zeros((self.numCoreSpecies, len_families), bool)
+
+        # Value number 3 is the number of reaction families in RMG
```
amarkpayne: `# len_families is the number of reaction families in RMG`

rmgpy/solver/base.pyx
```diff
@@ -448,24 +468,30 @@ cdef class ReactionSystem(DASx):
...
+
+        # Number of families in RMG
+        len_families = len(getDB('kinetics').families.keys())
+
         # Set unimolecular and bimolecular thresholds as true for any concentrations greater than 0
         numCoreSpecies = len(self.coreSpeciesConcentrations)
         for i in xrange(numCoreSpecies):
-            if self.coreSpeciesConcentrations[i] > 0:
-                self.unimolecularThreshold[i] = True
+            for k in xrange(len_families):
```
amarkpayne: If we want to vectorize this I think that one of these should work:

```
    if self.coreSpeciesConcentrations[i] > 0:
        self.unimolecularThreshold[i,:] = True
```

or

```
self.unimolecularThreshold = np.tile(self.coreSpeciesConcentrations > 0, (len_families, 1)).T
```

If you choose the second one double check that I didn't accidently swap rows and colums--it depends on if self.coreSpeciesConcentrations is a row vector or column vector

If you choose to do this then do the same for bimolecular and trimolecular

rmgpy/solver/base.pyx
```diff
@@ -979,32 +1008,41 @@ cdef class ReactionSystem(DASx):
...
-                (unimolecularThresholdRateConstant,
-                 bimolecularThresholdRateConstant,
-                 trimolecularThresholdRateConstant) = self.get_threshold_rate_constants(modelSettings)
-                unimolecularThresholdVal = toleranceMoveToCore * charRate / unimolecularThresholdRateConstant
-                bimolecularThresholdVal = toleranceMoveToCore * charRate / bimolecularThresholdRateConstant
-                trimolecularThresholdVal = toleranceMoveToCore * charRate / trimolecularThresholdRateConstant
+                (kvals_uni, kvals_bi) = self.get_threshold_rate_constants()
+
+                unimolecularThresholdVal = numpy.zeros(len(kvals_uni))
+                for i in xrange(len(kvals_uni)):
```
amarkpayne: If `kvals_uni` is a numpy array then I think you can get away with `unimolecularThresholdVal = toleranceMoveToCore * charRate / kvals_uni`, and similar for bi and tri molecular

rmgpy/data/kinetics/family.py
```diff
@@ -3868,3 +3872,82 @@ def getObjectiveFunction(kinetics1,kinetics2,obj=informationGain,T=1000.0):
...
+class ArrheniusRMGObject(RMGObject):
+    """
+    Child class of RMG Object for storing filter Arrhenius fits.
+    """
+
+    def __init__(self, unimol=None, bimol=None):
+        if unimol:
+            self.unimol = unimol
+        else:
+            self.unimol = {}
```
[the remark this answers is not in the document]
mliu49: Both `dict()` and `{}` instantiate a dictionary. To get an empty set you have to use `set()`.

Note that instantiating a non-empty dictionary/set automatically determines the type based on the contents. So `{'foo': 'bar'}` gives a dictionary while `{'foo', 'bar'}` gives a set.

rmgpy/data/kinetics/family.py
```diff
@@ -3868,3 +3872,82 @@ def getObjectiveFunction(kinetics1,kinetics2,obj=informationGain,T=1000.0):
     
     return obj(ks1,ks2), N1 == 0
 
+################################################################################
+
+class ArrheniusRMGObject(RMGObject):
```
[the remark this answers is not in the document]
mliu49: I plan on looking at this in more detail later, but it seems from first glance that this is just a container class for the fitted max rates for each family? If so, then it seems to me like "Arrhenius" isn't really a necessary descriptor, while things like "Filter", "Fits", "Fitted", "Limit", "Family", "Rate", or "Threshold" might be more helpful to put in the name. E.g. "FittedRateLimit", "FilterLimitFits", "FittedFilterThreshold" (FFT :P), "FamilyRateLimit", etc.

rmgpy/data/kinetics/family.py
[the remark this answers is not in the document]
amarkpayne: Ah sorry, I just now realized that I have had this backwards in my head for a long time. Thanks for catching this @mliu49

rmgpy/data/kinetics/family.py
```diff
@@ -3868,3 +3872,82 @@ def getObjectiveFunction(kinetics1,kinetics2,obj=informationGain,T=1000.0):
...
+
+    def __init__(self, unimol=None, bimol=None):
+        if unimol:
+            self.unimol = unimol
+        else:
+            self.unimol = {}
+        if bimol:
+            self.bimol = bimol
+        else:
+            self.bimol = {}
```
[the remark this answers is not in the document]
amarkpayne: ignore this, sorry I was wrong

rmgpy/reduction/reductionTest.py
```diff
@@ -62,9 +64,18 @@ def setUpClass(cls):
         cls.targets = targets
         cls.error = error
 
+        # load kinetic database and forbidden structures
+        rmg.database = RMGDatabase()
+        path = os.path.join(settings['database.directory'])
+
+        # kinetics family loading
+        rmg.database.loadKinetics(os.path.join(path, 'kinetics'),
+                                       kineticsFamilies=['H_Abstraction'],
```
amarkpayne: Do we need to load `H_Abstraction` in particular for these tests? If not can we get away with loading a smaller family?

rmgpy/solver/baseTest.py
```diff
@@ -55,6 +58,21 @@ def setUp(self):
...
+
+        # load kinetic database and forbidden structures
+        self.rmg.database = RMGDatabase()
+        path = os.path.join(settings['database.directory'])
+
+        # forbidden structure loading
+        self.rmg.database.loadForbiddenStructures(os.path.join(path, 'forbiddenStructures.py'))
+        # kinetics family loading
+        self.rmg.database.loadKinetics(os.path.join(path, 'kinetics'),
+                                       kineticsFamilies=['H_Abstraction'],
```
amarkpayne: Here as well

rmgpy/solver/baseTest.py
```diff
@@ -66,15 +84,28 @@ def testSurfaceInitialization(self):
...
+        rtol = 1e-8
+        sensitivity = False
+        sens_atol = 1e-6
+        sens_rtol = 1e-4
+        filterReactions = False
+        unimolecularFilterFit = self.rmg.reactionSystems[0].unimolecularFilterFit
+        bimolecularFilterFit = self.rmg.reactionSystems[0].bimolecularFilterFit
+        conditions = {}
+
+        reactionSystem.initializeModel(coreSpecies, coreReactions, edgeSpecies, edgeReactions, surfaceSpecies,
```
amarkpayne: I think you can use keyword arguments here so that you don't have to add the default values for things like `atol` in the previous lines

rmgpy/solver/baseTest.py
```diff
@@ -98,10 +129,21 @@ def testSurfaceLayeringConstraint(self):
...
```
amarkpayne: Same here

rmgpy/solver/baseTest.py
```diff
@@ -126,10 +168,21 @@ def testAddReactionsToSurface(self):
...
```
amarkpayne: Same here

rmgpy/solver/liquidTest.py
```diff
@@ -40,10 +40,20 @@
...
 from rmgpy.solver.base import TerminationTime
-from rmgpy.rmg.main import RMG
 from rmgpy.rmg.settings import ModelSettings, SimulatorSettings
+from rmgpy.data.rmg import RMGDatabase
+from rmgpy.rmg.main import RMG
+from rmgpy import settings
+from rmgpy.tools.loader import loadRMGPyJob
 
 ################################################################################
+class ConcentrationPrinter(object):
```
amarkpayne: Pycharm complains that there isn't 2 spaces between this and the separator above, but this is very minor

rmgpy/solver/liquidTest.py
```diff
@@ -40,10 +40,20 @@
...
+    def __init__(self):
+        self.species_names = []
+        self.data = []
+
+    def update(self, subject):
+        self.data.append((subject.t , subject.coreSpeciesConcentrations))
```
amarkpayne: extra space here before the comma

rmgpy/solver/liquidTest.py
```diff
@@ -40,10 +40,20 @@
...
 
 class LiquidReactorCheck(unittest.TestCase):
```
amarkpayne: Add a line above this class definition

rmgpy/solver/liquidTest.py
```diff
@@ -119,11 +152,26 @@ def testComputeFlux(self):
...
+        unimolecularFilterFit = self.rmg.reactionSystems[0].unimolecularFilterFit
+        bimolecularFilterFit = self.rmg.reactionSystems[0].bimolecularFilterFit
+        conditions = {}
+
         c0 = {self.C2H5: 0.1, self.CH3: 0.1, self.CH4: 0.4, self.C2H6: 0.4}
 
         rxnSystem = LiquidReactor(self.T, c0, 1, termination=[])
 
-        rxnSystem.initializeModel(coreSpecies, coreReactions, edgeSpecies, edgeReactions)
+        rxnSystem.initializeModel(coreSpecies, coreReactions, edgeSpecies, edgeReactions, surfaceSpecies,
```
amarkpayne: here as well

rmgpy/solver/liquidTest.py
```diff
@@ -315,11 +363,25 @@ def test_compute_derivative(self):
...
+        filterReactions = False
+        unimolecularFilterFit = self.rmg.reactionSystems[0].unimolecularFilterFit
+        bimolecularFilterFit = self.rmg.reactionSystems[0].bimolecularFilterFit
+        conditions = {}
         
         c0 = {self.CH4: 0.2, self.CH3: 0.1, self.C2H6: 0.35, self.C2H5: 0.15, self.H2: 0.2}
 
         rxnSystem0 = LiquidReactor(self.T, c0, 1, termination=[])
-        rxnSystem0.initializeModel(coreSpecies, coreReactions, edgeSpecies, edgeReactions)
+        rxnSystem0.initializeModel(coreSpecies, coreReactions, edgeSpecies, edgeReactions, surfaceSpecies,
```
amarkpayne: here as well

rmgpy/solver/simpleTest.py
```diff
@@ -44,11 +44,50 @@
 import rmgpy.constants as constants
 from rmgpy.chemkin import loadChemkinFile
 from rmgpy.rmg.settings import ModelSettings, SimulatorSettings
+from rmgpy.data.rmg import RMGDatabase
+from rmgpy.rmg.main import RMG
+from rmgpy import settings
+from rmgpy.tools.loader import loadRMGPyJob
 
 ################################################################################
+class ConcentrationPrinter(object):
```
amarkpayne: Same style changes here (two lines above and after, and extra space before comma)

rmgpy/solver/simpleTest.py
```diff
@@ -342,9 +381,26 @@ def testSpecificColliderModel(self):
...
+        sens_rtol = 1e-4
+        filterReactions = False
+        unimolecularFilterFit = [] #self.rmg.reactionSystems[0].unimolecularFilterFit
+        bimolecularFilterFit =[] #self.rmg.reactionSystems[0].bimolecularFilterFit
+        conditions = {}
+
         # Initialize the model
         rxnSystem = SimpleReactor(T,P,initialMoleFractions=initialMoleFractions,nSims=1,termination=None)
-        rxnSystem.initializeModel(speciesList, reactionList, [], [])
+        rxnSystem.initializeModel(speciesList, reactionList, edgeSpecies, edgeReactions, surfaceSpecies,
```
amarkpayne: Here as well

rmgpy/solver/surfaceTest.py
```diff
@@ -41,11 +42,52 @@
...
 from rmgpy.solver.base import TerminationTime, TerminationConversion
 import rmgpy.constants as constants
+from rmgpy.data.rmg import RMGDatabase
+from rmgpy.rmg.main import RMG
+from rmgpy import settings
+from rmgpy.tools.loader import loadRMGPyJob
 
 ################################################################################
 
+class ConcentrationPrinter(object):
```
amarkpayne: style changes here as well

rmgpy/solver/surfaceTest.py
```diff
@@ -100,7 +142,22 @@ def testSolveH2(self):
...
+        rtol = 1e-8
+        sensitivity = False
+        sens_atol = 1e-6
+        sens_rtol = 1e-4
+        filterReactions = False
+        unimolecularFilterFit = self.rmg.reactionSystems[0].unimolecularFilterFit
+        bimolecularFilterFit = self.rmg.reactionSystems[0].bimolecularFilterFit
+        conditions = {}
+
+        rxnSystem.initializeModel(coreSpecies, coreReactions, edgeSpecies, edgeReactions, surfaceSpecies,
```
amarkpayne: Here as well

rmgpy/data/kinetics/familyTest.py
```diff
@@ -639,6 +640,58 @@ def testSaveFamily(self):
...
+        # forbidden structure loading
+        cls.rmg.database.loadForbiddenStructures(os.path.join(path, 'forbiddenStructures.py'))
+        # kinetics family loading
+        cls.rmg.database.loadKinetics(os.path.join(path, 'kinetics'),
+                                       kineticsFamilies=['H_Abstraction'],
+                                       reactionLibraries=[])
+
+        cls.rmg = loadRMGPyJob(inputFile, chemkinFile, spc_dict, generateImages=False, checkDuplicates=False)
+
+    def testLoad(cls):
```
amarkpayne: Use `self` as the first argument of this method. The one above was a class method which (by convention only) uses `cls`.

rmgpy/data/kinetics/familyTest.py
```diff
@@ -639,6 +640,58 @@ def testSaveFamily(self):
...
+        """
+        Test if loading Arrhenius fits of the highest rates in each training reaction family from a file located at
+        `fits_location` on disk works.
+        """
+
+        loadFilterFits(cls.rmg)
+
+        cls.assertIsNotNone(cls.rmg.reactionSystems[0].unimolecularFilterFit)
```
amarkpayne: Can you also check that these objects are `ArrheniusRMGObject` instances?

rmgpy/data/kinetics/familyTest.py
```diff
@@ -639,6 +640,58 @@ def testSaveFamily(self):
...
+        cls.assertIsNotNone(cls.rmg.reactionSystems[0].bimolecularFilterFit)
+
+
+class ConcentrationPrinter(object):
```
amarkpayne: Define this class in rmgpy.rmg.listener.py and import it from there. This will allow you to delete this here, and also where they occur in simpleTest.py etc.

rmgpy/data/kinetics/familyTest.py
```diff
@@ -639,6 +640,58 @@ def testSaveFamily(self):
...
+    def __init__(self):
+        self.species_names = []
```
amarkpayne: Does `self.species_names` ever get used?

rmgpy/rmg/react.py
```diff
@@ -93,7 +93,8 @@ def react_all(core_spc_list, numOldCoreSpecies, unimolecularReact, bimolecularRe
     reactions and splits reaction families per task for improved load balancing in parallel runs.
     """
 
-    families = getDB('kinetics').families.keys()
+    family_names = getDB('kinetics').families.keys()
```
amarkpayne: You can avoid calling `getDB` twice by doing the following:

```
family_names, families = list(getDB('kinetics').families.items())
```

Again, `list` for python3 reasons

rmgpy/data/kinetics/family.py
```diff
@@ -3868,3 +3872,82 @@ def getObjectiveFunction(kinetics1,kinetics2,obj=informationGain,T=1000.0):
...
+
+    def load_yaml(self, path, class_dictionary):
+        with open(path, 'r') as f:
+            data = yaml.safe_load(stream=f)
+
+        data = {key:data[key] for key in data if key!='class'}
```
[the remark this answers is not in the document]
ajocher: def load_yaml() is now located in kinetics/database.py. Thanks, the make_object method works fine!

rmgpy/data/kinetics/family.py
```diff
@@ -3868,3 +3872,82 @@ def getObjectiveFunction(kinetics1,kinetics2,obj=informationGain,T=1000.0):
...
     return obj(ks1,ks2), N1 == 0
 
+################################################################################
+
+class ArrheniusRMGObject(RMGObject):
+    """
+    Child class of RMG Object for storing filter Arrhenius fits.
+    """
+
+    def __init__(self, unimol=None, bimol=None):
```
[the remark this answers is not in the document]
ajocher: Done. Now in kinetics/database.py

rmgpy/data/kinetics/family.py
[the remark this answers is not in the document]
ajocher: Renamed to "FilterLimitFits".

rmgpy/data/kinetics/family.py
```diff
@@ -3868,3 +3872,82 @@ def getObjectiveFunction(kinetics1,kinetics2,obj=informationGain,T=1000.0):
...
+    on disk.
+
+    For reaction filtering, the loaded fits per reaction family are sorted to correspond to the user
+    defined families.
+    """
+
+    from rmgpy.data.rmg import getDB
+    families = getDB('kinetics').families.keys()
+
+    fits_location = os.path.join(settings['database.directory'],'FilterArrheniusFits')
```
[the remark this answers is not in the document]
ajocher: Done.

rmgpy/data/kinetics/family.py
```diff
@@ -3868,3 +3872,82 @@ def getObjectiveFunction(kinetics1,kinetics2,obj=informationGain,T=1000.0):
...
+
+    logging.debug("""Loading Arrhenius fits from {0} based on training reactions
+    to generate custom filter criteria for each reaction family""".format(
+        os.path.join(fits_location)))
```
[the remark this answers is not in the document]
ajocher: Done.

rmgpy/data/kinetics/family.py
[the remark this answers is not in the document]
ajocher: Done. Also updated in the RMG-database ipython notebook and file.

rmgpy/data/kinetics/family.py
```diff
@@ -3868,3 +3872,82 @@ def getObjectiveFunction(kinetics1,kinetics2,obj=informationGain,T=1000.0):
...
+    class_dictionary = {'ScalarQuantity': ScalarQuantity,
+                        'Arrhenius': Arrhenius,
+                        'ArrheniusRMGObject': ArrheniusRMGObject,
+                        }
+    obj = ArrheniusRMGObject()
+    obj.load_yaml(fits_location, class_dictionary)
+
+    # List of unimolecular kinetics
+    filter_unimol_families = []
+    unimol_kinetics_list_tmp  = []
```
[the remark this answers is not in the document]
ajocher: Outdated by the piece of code suggested in the two following comments.

rmgpy/data/kinetics/family.py
```diff
@@ -3868,3 +3872,82 @@ def getObjectiveFunction(kinetics1,kinetics2,obj=informationGain,T=1000.0):
...
+    for filter_family, Arrhenius_fit in obj.unimol.items():
+        filter_unimol_families.append(filter_family)
+        unimol_kinetics_list_tmp.append(Arrhenius_fit)
+
+    # List of bimolecular kinetics
+    filter_bimol_families = []
+    bimol_kinetics_list_tmp = []
+    for filter_family, Arrhenius_fit in obj.bimol.items():
```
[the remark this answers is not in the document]
ajocher: Outdated by the piece of code suggested in the following comment.

rmgpy/data/kinetics/family.py
```diff
@@ -3868,3 +3872,82 @@ def getObjectiveFunction(kinetics1,kinetics2,obj=informationGain,T=1000.0):
...
+        filter_bimol_families.append(filter_family)
+        bimol_kinetics_list_tmp.append(Arrhenius_fit)
+
+    # Sort to match family list in getDB. Add None for surface families.
+    unimol_kinetics_list = []
+    bimol_kinetics_list = []
+    for family in families:
+        for index, filter_family in enumerate(filter_unimol_families):
```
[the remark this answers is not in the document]
ajocher: Thanks for the more elegant piece of code, just fit and filter_family had to be switched.

rmgpy/data/kinetics/family.py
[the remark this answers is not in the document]
amarkpayne: I think this is also in line ~74 of checkFilterFits.py

rmgpy/rmg/main.py
```diff
@@ -1443,74 +1460,84 @@ def updateReactionThresholdAndReactFlags(self,
...
+                                    self.trimolecularReact[i,j,k,l] = True
+                                    self.trimolecularThreshold[i,j,k,l] = True
         else:
             # We are not filtering reactions
             if new_core_species:
                 # React all the new core species unimolecularly
                 for i in xrange(prevNumCoreSpecies, numCoreSpecies):
-                    self.unimolecularReact[i] = True
+                    for k in xrange(len_families):
+                        self.unimolecularReact[i,k] = True
```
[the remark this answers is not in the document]
amarkpayne: Thanks! One thing that I need to check into is how we handle the fact that for bimolecular and trimolecular arrays that tensors are symmetric for a given family. Therefore, is `self.bimolecularReact[:, prevNumCoreSpecies:, :] = True` good enough or do we need to add an additional line of `self.bimolecularReact[prevNumCoreSpecies:, :, :] = True`?

rmgpy/solver/liquidTest.py
amarkpayne: This got moved to listner.py, but there is still an extra space after `subject.t`

rmgpy/data/kinetics/familyTest.py
amarkpayne: Thanks! I think you can use `Arrhenius` instead of `Arrhenius()` i.e. you can hand it the class itself, you don't have to instantiate an Arrhenius object.

rmgpy/data/kinetics/familyTest.py
amarkpayne: Thanks! Add a blank line at the end of listener.py as well

rmgpy/rmg/react.py
amarkpayne: Sorry I realized that I made a mistake here. There is a way to do it in python 3 I think but it might be more trouble than it is worth. Let's just revert this back, you had it right the first time

rmgpy/data/kinetics/familyTest.py
ajocher: Done.

rmgpy/data/kinetics/family.py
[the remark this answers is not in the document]
ajocher: Thanks! Changed in checkFilterFits.py

rmgpy/solver/liquidTest.py
ajocher: Sorry, missed that space. Now it is gone.

rmgpy/data/kinetics/familyTest.py
ajocher: Done.

rmgpy/rmg/react.py
ajocher: Updated and tested to work in python 3.

rmgpy/rmg/main.py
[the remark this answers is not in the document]
amarkpayne: Just to document, we learned that these tensors are upper triangle matrices for each family slice in bimolecular, and that RMG does not use the bottom triangle to prevent reacting pairs twice.

rmgpy/rmg/main.py
```diff
@@ -1443,74 +1460,84 @@ def updateReactionThresholdAndReactFlags(self,
...
-                                          :prevNumCoreSpecies] = self.trimolecularThreshold
+                                          :prevNumCoreSpecies,
+                                          :len_families] = self.trimolecularThreshold
                     self.trimolecularThreshold = trimolecularThreshold
                 
             if skipUpdate:
                 return
             
             # Always update the react and threshold arrays
             for i in xrange(numCoreSpecies):
```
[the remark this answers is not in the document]
amarkpayne: Just to document, we discussed offline that this implementation is equivalent to the previous where as far as the upper triangle is concerned. @ajocher and I have both done testing and confirmed that the output is the same, with the exception that the numpy code also returns the bottom triangle portion which is unused.

I also did some profiling and found that the two methods have pretty much identical timings. So it turns out that running through the full tensors twice in numpy is about as fast as running through half of the tensors in for loops.

Feel free to pick whichever implementation you like. Part of me likes the short numpy code, but after learning that RMG only uses the upper triangle anyways and that the speed is equivalent I think it was probably fine as is. Furthermore, this is not crucial to the PR, and this is not the rate determining part of our code.

testing/checkFilterFits.py
```diff
@@ -0,0 +1,288 @@
...
+        path_notebook = os.path.join(databaseDirectory, '../scripts/generateFilterArrheniusFits.ipynb')
+
+        obj = ArrheniusRMGObject()
+
+        class_dictionary = {'ScalarQuantity': ScalarQuantity,
+                            'Arrhenius': Arrhenius,
+                            'ArrheniusRMGObject': ArrheniusRMGObject,
+                            }
+
+        self.load_yaml(obj, path, class_dictionary)
```
amarkpayne: Change this line to `obj.load_yaml(path, class_dictionary)`. This will allow you to get rid of the extra function in line 169 that is being flagged by codacy.

testing/checkFilterFits.py
```diff
@@ -0,0 +1,288 @@
...
+                units = 's^-1' if molecularity == 1 else 'm^3/(mol*s)'
+
+                arr = Arrhenius().fitToData(Ts, np.array(k_max_list), units)
+                return arr
+
+        else:
+            arr = None
+            return arr
+
+    def is_close(self, a, b, rel_tol=1e-2, abs_tol=0.0):
```
amarkpayne: Make the following change to prevent this from being flagged by codacy:

```
    @staticmethod
    def is_close(a, b, rel_tol=1e-2, abs_tol=0.0):
```

ajocher: Thanks, done.

testing/checkFilterFits.py
ajocher: I removed `load_yaml` and replaced it with the method we already have in `rmgpy/data/kinetics/database.py`. I also did some more clean-up and replaced `ArrheniusRMGObject` by loading this class from `rmgpy/data/kinetics/database.py`.

rmgpy/data/kinetics/database.py
```diff
@@ -799,3 +802,71 @@ def reconstructKineticsFromSource(self, reaction, source, fixBarrierHeight=False
...
+
+        unimol_kinetics_list = []
+        bimol_kinetics_list = []
+        for family in families:
+            for filter_family, fit in obj.unimol.items():
+                if family == filter_family:
+                    unimol_kinetics_list.append(fit)
+                    break
+            else:  # We could not find the filter fits for this family. Raise an exception or else the indices will be off
+                raise ValueError('Unable to find filter fits for family {0}'.format(family))
```
mliu49: Perhaps include in the message what the user should do? e.g. fit filter thresholds for that family

Aside: I'm not super keen on `fits` as a noun. It just sounds weird...

rmgpy/data/kinetics/database.py
```diff
@@ -799,3 +802,71 @@ def reconstructKineticsFromSource(self, reaction, source, fixBarrierHeight=False
...
+                            'FilterLimitFits': FilterLimitFits,
+                            }
+
+        obj = FilterLimitFits()
+        obj.load_yaml(fits_location, class_dictionary)
+
+        unimol_kinetics_list = []
+        bimol_kinetics_list = []
+        for family in families:
+            for filter_family, fit in obj.unimol.items():
```
mliu49: If `obj.unimol` is a dictionary like this implies, why not do this?
```
for family in families:
    try:
        unimol_kinetics_list.append(obj.unimol[family])
    except KeyError:
        raise ValueError
```

rmgpy/data/kinetics/familyTest.py
```diff
@@ -40,14 +40,16 @@
...
 from rmgpy.data.kinetics.database import KineticsDatabase
 from rmgpy.data.kinetics.family import TemplateReaction
+from rmgpy.kinetics.arrhenius import Arrhenius
 from rmgpy.data.rmg import RMGDatabase
 from rmgpy.molecule import Molecule
 from rmgpy.species import Species
-from rmgpy.molecule.group import Group
+from rmgpy.tools.loader import loadRMGPyJob
+from rmgpy.rmg.main import RMG
+from rmgpy.rmg.listener import ConcentrationPrinter
```
mliu49: Imports should be in alphabetical order if possible.

rmgpy/data/kinetics/familyTest.py
```diff
@@ -639,6 +641,54 @@ def testSaveFamily(self):
             shutil.rmtree(os.path.join(settings['test_data.directory'], 'testing_database/kinetics/families/intra_H_copy'))
 
 
+###################################################
+class TestLoadFilterFits(unittest.TestCase):
```
mliu49: I know I said that you could keep it in familyTest.py, but I think this is probably more suited for kineticsTest.py since you're loading the KineticsDatabase functionality.

rmgpy/kinetics/arrhenius.pxd
```diff
@@ -33,7 +33,7 @@ from rmgpy.quantity cimport ScalarQuantity, ArrayQuantity
 ################################################################################
 
 cdef class Arrhenius(KineticsModel):
-    
+
```
mliu49: Could you remove this change?

rmgpy/reduction/reductionTest.py
```diff
@@ -62,9 +64,18 @@ def setUpClass(cls):
...
         cls.error = error
 
+        # load kinetic database and forbidden structures
+        rmg.database = RMGDatabase()
+        path = os.path.join(settings['database.directory'])
+
+        # kinetics family loading
+        rmg.database.loadKinetics(os.path.join(path, 'kinetics'),
+                                       kineticsFamilies=['R_Addition_COm'],
+                                       reactionLibraries=[])
```
mliu49: Could you fix the indentation?

rmgpy/reduction/reductionTest.py
```diff
@@ -62,9 +64,18 @@ def setUpClass(cls):
...
+
         reactionModel = rmg.reactionModel
         initialize(rmg.outputDirectory, reactionModel.core.reactions)
-    
+
```
mliu49: Could you either undo the change or remove this line completely? (single blank line between methods)

rmgpy/rmg/main.py
```diff
@@ -488,6 +489,11 @@ def initialize(self, **kwargs):
             for reactionSystem in self.reactionSystems:
                 reactionSystem.viscosity = solventData.getSolventViscosity(reactionSystem.T.value_si)
 
+        # Read custom reaction filtering YAML file here and store content as lists
+        if self.filterReactions:
+            KineticsDatabase.loadFilterFits(self.database.kinetics, self.databaseDirectory,
+                                            self.reactionSystems)
```
mliu49: Should this be `self.database.kinetics.loadFilterFits(self.databseDirectory, self.reactionSystems)`?

rmgpy/rmg/main.py
```diff
@@ -1415,20 +1427,26 @@ def generateCanteraFiles(self, chemkinFile, **kwargs):
 
     def initializeReactionThresholdAndReactFlags(self):
         numCoreSpecies = len(self.reactionModel.core.species)
+
+        # Number of families in RMG
+        len_families = len(getDB('kinetics').families)
```
mliu49: I might call this `numFamilies` for consistency.

rmgpy/rmg/main.py
```diff
@@ -1415,20 +1427,26 @@ def generateCanteraFiles(self, chemkinFile, **kwargs):
...
+            self.bimolecularReact = np.zeros((numCoreSpecies, numCoreSpecies, len_families),bool)
+            self.unimolecularThreshold = np.zeros((numCoreSpecies, len_families),bool)
+            self.bimolecularThreshold = np.zeros((numCoreSpecies, numCoreSpecies, len_families),bool)
             if self.trimolecular:
-                self.trimolecularReact = np.zeros((numCoreSpecies, numCoreSpecies, numCoreSpecies), bool)
-                self.trimolecularThreshold = np.zeros((numCoreSpecies, numCoreSpecies, numCoreSpecies), bool)
+                self.trimolecularReact = np.zeros((numCoreSpecies, numCoreSpecies, numCoreSpecies, 
+                    len_families), bool)
+                self.trimolecularThreshold = np.zeros((numCoreSpecies, numCoreSpecies, numCoreSpecies, 
+                    len_families), bool)
```
mliu49: It might be better to line this up with the first numCoreSpecies. Not totally sure though.

Or just keep it on one line, like line 1449.

rmgpy/rmg/react.py
```diff
@@ -118,59 +117,70 @@ def react_all(core_spc_list, numOldCoreSpecies, unimolecularReact, bimolecularRe
...
     # Select reactive species that can undergo unimolecular reactions:
-    spc_tuples = [(core_spc_list[i],)
-                  for i in xrange(numOldCoreSpecies) if (unimolecularReact[i] and core_spc_list[i].reactive)]
+    spc_tuples = []
+    for i in xrange(numOldCoreSpecies):
+        fam_leftovers = []
+        for k, family in enumerate(family_names):
+            # Find reactions involving the species that are unimolecular and only generate tuple if family is not
+            # bimolecular in forward and backward direction.
+            if not (len(families[k].forwardTemplate.reactants)==2 and len(families[k].forwardTemplate.products)==2):
```
mliu49: Probably doesn't make a difference performance-wise, but seems more intuitive to check `reactants == 1 or products == 1`.

Similarly, for bimolecular, you could check `reactants == 2 or products == 2`.

rmgpy/rmg/react.py
```diff
@@ -73,12 +74,10 @@ def react(spc_fam_tuples, procnum=1):
 
     return reactions
 
-
```
mliu49: Could you undo the spacing changes? (two empty lines between top-level functions)

https://www.python.org/dev/peps/pep-0008/#blank-lines

rmgpy/solver/base.pyx
```diff
@@ -55,6 +55,7 @@ from rmgpy.quantity import Quantity
 from rmgpy.chemkin import getSpeciesIdentifier
 from rmgpy.reaction import Reaction
 from rmgpy.species import Species
+from rmgpy.data.rmg import getDB
```
mliu49: To reduce the complexity of our imports and the risk of adding more cyclic imports, would it make sense provide this value to the ReactionSystem directly, e.g. from the RMG class?

rmgpy/solver/liquid.pyx
```diff
@@ -166,18 +169,34 @@ cdef class LiquidReactor(ReactionSystem):
                 self.Keq[j] = rxn.getEquilibriumConstant(self.T.value_si)
                 self.kb[j] = self.kf[j] / self.Keq[j]
 
-    def get_threshold_rate_constants(self, modelSettings):
+    def get_threshold_rate_constants(self):
         """
         Get the threshold rate constants for reaction filtering.
 
-        modelSettings is not used here, but is needed so that the method
+        lists are not used here, but are needed so that the method
```
mliu49: What does `lists` refer to?

rmgpy/solver/simple.pyx
```diff
@@ -251,22 +258,35 @@ cdef class SimpleReactor(ReactionSystem):
...
+
+        # Evaluate kinetics at user defined temperature
+        # unimolecular reactions
+        kvals_uni_tmp = []
+        for k, unimol_kinetics in enumerate(self.unimolecularFilterFit):
+            if unimol_kinetics:
+                kvals_uni_tmp.append(unimol_kinetics.getRateCoefficient(self.T.value_si))
+            else:
+                # Set the maximum unimolecular rate to be kB*T/h
+                kvals_uni_tmp.append(2.08366122e10 * self.T.value_si)
```
mliu49: Could you import kB from constants?

rmgpy/rmg/listener.py
```diff
@@ -146,3 +146,13 @@ def update(self, reactionSystem):
             )
             
         SimulationPlot(csvFile=csvFile, numSpecies=10, ylabel='Moles').plot(pngFile)
+
+
+###################################################
+class ConcentrationPrinter(object):
```
mliu49: Is this only used for unit tests?

.travis.yml
```diff
@@ -51,6 +52,22 @@ jobs:
...
+        # Clone RMG-database
+        - git clone -b FilterCritPerReactFamily_database https://github.com/ReactionMechanismGenerator/RMG-database.git
+        - cd RMG-Py
+        - conda env create -q -f environment_linux.yml
+        - source activate rmg_env
+        # Install codecov for coverage report
+        - conda install -y -c conda-forge codecov
+        - conda list
+        # Setup MOPAC license key
+        - yes 'Yes' | $HOME/miniconda/envs/rmg_env/bin/mopac $MOPACKEY > /dev/null
```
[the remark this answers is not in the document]
amarkpayne: This is my fault, I blindly copied without taking a closer look

rmgpy/kinetics/arrhenius.pxd
ajocher: Done.

rmgpy/solver/base.pyx
ajocher: Import now replaced in various files by value from the RMG class.

ajocher:
Is it correct that we can't simulate anymore without loading the database? Is there a way to avoid that? 
- The value is now obtained from the RMG class.

There were a lot of additions to the solver tests, but no new assertions. Were the additions only to make the tests run? Are there new assertions that should be added? 
- The additions were added to pass the solver tests, because of the storage location of the Arrhenius fits for the custom filter criteria per reaction family. After offline discussion we are now storing the Arrhenius fits for filtering in `modelSettings` and the changes to the solver tests are reverted.

rmgpy/solver/base.pyx
```diff
@@ -978,33 +991,33 @@ cdef class ReactionSystem(DASx):
...
+                                unimolecularThreshold[i,k] = True
                 for i in xrange(numCoreSpecies):
                     for j in xrange(i, numCoreSpecies):
-                        if not bimolecularThreshold[i,j]:
-                            if coreSpeciesConcentrations[i]*coreSpeciesConcentrations[j] > bimolecularThresholdVal:
-                                bimolecularThreshold[i,j] = True
+                        for k in xrange(len(bimolecularThresholdVal)):
+                            if not bimolecularThreshold[i,j,k]:
+                                if coreSpeciesConcentrations[i]*coreSpeciesConcentrations[j] > bimolecularThresholdVal[k]:
+                                    bimolecularThreshold[i,j,k] = True
```
mliu49: Would this also be faster as a matrix operation (including uni and trimolecular)? You would end up filling the entire matrix/tensor though.

rmgpy/solver/liquid.pyx
```diff
@@ -173,11 +174,27 @@ cdef class LiquidReactor(ReactionSystem):
...
+
+        # Generate dictionary with reaction families as keys and kinetics as values
+        unimolecular_threshold_rate_constant = {
+            key: value for key, value in zip(families, unimolecular_threshold_rate_constant_tmp)
+        }
+        bimolecular_threshold_rate_constant = {
+            key: value for key, value in zip(families, bimolecular_threshold_rate_constant_tmp)
+        }
+        trimolecular_threshold_rate_constant = {
+            key: value for key, value in zip(families, trimolecular_threshold_rate_constant_tmp)
```
mliu49: Why are these dictionaries?

rmgpy/solver/simple.pyx
```diff
@@ -253,20 +258,34 @@ cdef class SimpleReactor(ReactionSystem):
...
+            if bimol_kinetics:
+                kvals_bi_tmp.append(min(1.0e8, bimol_kinetics.getRateCoefficient(self.T.value_si)))
+            else:
+                # Value estimated, based on the same order of magnitude of hydrogen colliding at 1000 K.
+                kvals_bi_tmp.append(1e8)
+
+        kvals_uni = numpy.asarray(kvals_uni_tmp, dtype=numpy.float64)
+        kvals_bi = numpy.asarray(kvals_bi_tmp, dtype=numpy.float64)
+
+        return (kvals_uni, kvals_bi)
```
mliu49: We should return a value for trimolecular reactions here instead of using kvals_bi directly in base.pyx.

Additionally, the treatment of trimolecular families is a bit confusing. We should discuss.

rmgpy/rmg/listener.py
mliu49: It's not clear to me what this actually does...

rmgpy/solver/liquid.pyx
```diff
@@ -173,11 +174,17 @@ cdef class LiquidReactor(ReactionSystem):
...
+        families = getDB('kinetics').families.keys()
+
         # Set the maximum bi/trimolecular rates based on the Smoluchowski and Stokes-Einstein equations
-        bimolecular_threshold_rate_constant = 22.2 * self.T.value_si / self.viscosity
-        trimolecular_threshold_rate_constant = 0.11 * self.T.value_si / self.viscosity
+        unimolecular_threshold_rate_constant = []
+        bimolecular_threshold_rate_constant = []
+        trimolecular_threshold_rate_constant = []
+        for k in xrange(len(families)):
+            unimolecular_threshold_rate_constant.append(2.08366122e10 * self.T.value_si)
```
mliu49: Couple more suggestions here:
- Keep the comment for unimolecular threshold from before: `# Set the maximum unimolecular rate to be kB*T/h`
- Import constants and use constants.kb/constants.h for the unimolecular threshold
- Make sure to convert these to numpy arrays like in the method simple.pyx, which is now required by the type declaration in base.pyx
- I still don't like importing from the database module here. I can think of two alternatives: 1) since the simulate method now has num_families, you can pass it to this method, 2) if the arrhenius fits are also loaded for liquid simulations, you can check the length of the dictionary from model settings. If you choose 1, make sure to update the function signature for simple.pyx as well.

ajocher: - Keeping the comment for unimolecular threshold from before: `# Set the unimolecular rate to kB*T/h`.
- Using now constants.kB and constants.h for unimolecular threshold.
- Using numpy operations to assign values.
- Chose alternative 2) to avoid database module loading.
- To be consistent and potentially faster I also use numpy operations in simple.pyx `get_threshold_rate_constants`

rmgpy/data/kinetics/familyTest.py
amarkpayne: Just to follow-up, we also break the imports into three sections: modules in the standard library, modules from our dependencies (listed in our conda environment files) and then a third block for our own modules (rmgpy + arkane).

With this, logging should be in the first section, and numpy should be in the second section (leave a blank line between sections). 

Also, I don't think that `from rmgpy.molecule.group import Group` is currently being used

rmgpy/data/kinetics/database.py
```diff
@@ -799,3 +802,81 @@ def reconstructKineticsFromSource(self, reaction, source, fixBarrierHeight=False
...
+        `RMG-database/input/FilterArrheniusFits.yml`.
+
+        For reaction filtering, the loaded fits per reaction family are sorted to correspond to the user
+        defined families.
+        """
+
+        families = self.families.keys()
+
+        path = os.path.join(database_directory, 'FilterArrheniusFits.yml')
+        path_notebook = os.path.join(database_directory, './../scripts/generateFilterArrheniusFits.ipynb')
```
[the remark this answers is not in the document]
amarkpayne: You'll want to use `os.path.join(os.path.dirname(database_directory), 'scripts', 'generateFilterArrheniusFits.ipynb')`, as `/` is not always the directory delimiter. Not that we support Windows anymore, but we should be careful anyways.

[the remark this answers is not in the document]
ajocher: Replaced `'scripts/generateFilterArrheniusFits.ipynb'` with `'scripts', 'generateFilterArrheniusFits.ipynb'`

rmgpy/data/kinetics/kineticsTest.py
```diff
@@ -1091,3 +1094,37 @@ def test_add_atom_labels_for_reaction_3(self):
...
+        Load an RMG job to test loading the YAML file with custom filter criteria.
+        """
+        folder = os.path.join(os.path.dirname(__file__), '../../solver/files/listener/')
+        inputFile = os.path.join(folder, 'input.py')
+        chemkinFile = os.path.join(folder, 'chemkin/chem.inp')
+        spc_dict = os.path.join(folder, 'chemkin/species_dictionary.txt')
+
+        # set-up RMG object
+        cls.rmg = RMG()
+        cls.rmg = loadRMGPyJob(inputFile, chemkinFile, spc_dict, generateImages=False, checkDuplicates=False)
```
mliu49: It seems like this test could be simplified significantly since the RMG instance isn't actually necessary.

rmgpy/rmg/react.py
```diff
@@ -118,59 +120,69 @@ def react_all(core_spc_list, numOldCoreSpecies, unimolecularReact, bimolecularRe
...
+                        # Find reactions involving the species that are trimolecular
+                        if trimolecularReact[i, j, k, l] and core_spc_list[i].reactive and \
+                                core_spc_list[j].reactive and core_spc_list[k].reactive:
+                            if family in major_families and len(core_spc_list[i].molecule[0].atoms) > min_atoms or \
+                                    len(core_spc_list[j].molecule[0].atoms) > min_atoms or \
+                                    len(core_spc_list[k].molecule[0].atoms) > min_atoms:
+                                spc_tuples.append(((core_spc_list[i], core_spc_list[j],
+                                                   core_spc_list[k]), family))
+                            else:
+                                fam_leftovers.append(family)
```
mliu49: I think the trimolecular section is missing an `if fam_leftovers:` part.

ajocher: Yes, apparently the ´if fam_leftovers:´ statement went missing during some rebase. I added it back in. 

Also, there is now a molecularity check for trimolecular families.

rmgpy/data/kinetics/kineticsTest.py
ajocher: Simplified, though not sure if in the most elegant way.

rmgpy/data/kinetics/kineticsTest.py
```diff
@@ -1091,3 +1093,29 @@ def test_add_atom_labels_for_reaction_3(self):
...
+        global database
+        cls.database = database
+
+    def testLoad(self):
+        """
+        Test if loading Arrhenius fits of the highest training reaction rates in each reaction family from
+        `RMG-database/input/FilterArrheniusFits.yml` works.
+        """
+        modelSettingsList = []
+        modelSettingsList.append(ModelSettings)
```
mliu49: Thanks for redoing this test! This was exactly what I was thinking of.

Only fix is that I think you need to do `ModelSettings()` here to create an instance of the class rather than passing the class itself. You could also add a second ModelSettings instance to test that it's saved properly to multiple.
```
modelSettingsList = [ModelSettings(), ModelSettings()]
```

rmgpy/rmg/react.py
```diff
@@ -118,59 +120,73 @@ def react_all(core_spc_list, numOldCoreSpecies, unimolecularReact, bimolecularRe
...
-                    spc_fam_tuples.append((spc_tuple, item))
-            else:
-                spc_fam_tuples.append((spc_tuple, ))
-
-    return react(spc_fam_tuples, procnum), [fam_tuple[0] for fam_tuple in spc_fam_tuples]
+                    fam_leftovers = []
+                    for l, family in enumerate(family_names):
+                        # Find reactions involving the species that are trimolecular
+                        if (len(families[k].forwardTemplate.reactants) == 3 or len(
+                                families[k].forwardTemplate.products) == 3):
```
mliu49: Could you put the `len(` on the second line?

ajocher: ´len(´on second line now.

rmgpy/data/kinetics/kineticsTest.py
ajocher: Thanks, done.

rmgpy/data/kinetics/kineticsTest.py
```diff
@@ -1091,3 +1093,28 @@ def test_add_atom_labels_for_reaction_3(self):
...
+        """
+        global database
+        cls.database = database
+
+    def testLoad(self):
+        """
+        Test if loading Arrhenius fits of the highest training reaction rates in each reaction family from
+        `RMG-database/input/FilterArrheniusFits.yml` works.
+        """
+        modelSettingsList = [ModelSettings(), ModelSettings()]
```
amarkpayne: `model_settings_list`

rmgpy/rmg/reactTest.py
```diff
@@ -116,10 +121,11 @@ def testReactAll(self):
...
 
         n = len(spcs)
-        reaction_list, spc_tuples = react_all(spcs, n, np.ones(n), np.ones([n, n]), np.ones([n, n, n]), procnum)
+        reaction_list, spc_tuples = react_all(spcs, n, np.ones([n, num_families]), np.ones([n, n, num_families]),
+                                              np.ones([n, n, n, num_families]), procnum)
         self.assertIsNotNone(reaction_list)
-        self.assertEqual(len(reaction_list), 34)
-        self.assertEqual(len(spc_tuples), 34)
+        self.assertEqual(len(reaction_list), 27)
+        self.assertEqual(len(spc_tuples), 27)
```
amarkpayne: @mliu49 can you check this before you merge?

mliu49: Confirmed that this is ok.

rmgpy/rmg/main.py
```diff
@@ -1443,74 +1456,66 @@ def updateReactionThresholdAndReactFlags(self,
...
-                for i in xrange(numCoreSpecies):
-                    for j in xrange(i, numCoreSpecies):
-                        for k in xrange(j, numCoreSpecies):
-                            if not self.trimolecularThreshold[i,j,k] and rxnSysTrimolecularThreshold[i,j,k]:
-                                # We've shifted from not reacting to reacting
-                                self.trimolecularReact[i,j,k] = True
-                                self.trimolecularThreshold[i,j,k] = True
+                self.trimolecularReact = np.logical_and(np.logical_not(self.trimolecularThreshold),
+                                                        rxnSysTrimolecularThreshold)
+                self.unimolecularThreshold = np.logical_or(self.trimolecularReact, self.trimolecularThreshold)
```
mliu49: This should be `trimolecularThreshold`

mjohnson541:
Recommendation to deal with the slowdown we're observing in this PR during filter threshold checking in the solver.   

There actually isn't any reason you need to check all of the family thresholds each iteration, because you aren't going to react anything during the simulation.  All you really need is the maximum of each species concentration, each product of species concentration and so on.  

So instead of looping over all of the families calculate all of these products each iteration and save the maximum value for each (A,B,C,A*B,B*C,A*C,A*B*C) and then do the family analysis of which things to react with which families once after the simulation.  

This should remove the family dimension of the loop, which should make filter checking in the solver faster by about a factor of the number of families checked.

mliu49:
I implemented @mjohnson541's suggestion, bringing the filter threshold evaluation outside of the simulation loop. Now we calculate the concentration values as before without the extra family dimension.

In testing with the liquid oxidation test, this change resulted in `simulate` dropping out of the top 25 time-consuming functions. So I think it was clear that this was one bottleneck.

However, the timing for this job dropped from 20 min to 12 min, while it only takes 6 min on master (on my desktop, which is a bit slow). The most time-consuming thing now is deepcopy, which is being called by `react_species`. This PR enables family splitting in serial mode as well, which increases the number of species tuples generated, thereby increasing the number of copy operations. For the liquid oxidation test, deepcopy time increased from 92 s on master to 353 s on this branch.

mjohnson541:
Is there a reason the species part of the family tuples can't all be references of the same species objects?  I guess I don't understand why more deepcopy's should be necessary to encode the same individual sets of react steps we ran before.

mliu49:
For the record, we used to (~2+ years ago) not copy species before reacting. I _think_ the change was related to the first parallelization effort, in order to send independent objects to the child processes.

Since we've implemented a completely different parallelization approach which does not require copying species (again, I think), it would be worth the time to see if we can stop copying.

My worry is that in the meantime, we've made changes which rely (unknowingly) on having a copy of the species rather than the actual object in the core. We should verify extensively that doing so does not introduce bugs.

mliu49:
I skimmed through the reaction generation code again, and I'm fairly sure that it's safe to not copy species before reacting. I added a unit test to confirm that a couple notable Species and Molecule attributes are not affected by the reaction generation process. Let me know if you think there are other attributes we should check.

Based on that, I removed species copying from `react_species`. I also removed a redundant resonance structure generation step in `ensure_independent_atom_ids`. Finally, I restored the prior behavior where family splitting is only employed when running in parallel, which reduces the number species tuples created for reaction generation.

To summarize the runtime differences:

| stage | runtime | change |
| ----- | ----- | ----- |
| master | 442 s | -- |
| custom filter criteria v1 | 1270 s | +828 |
| optimize filter threshold | 780 s | -490 |
| remove species copy | 509 s | -271 |
| remove res gen | 442 s | -67 |
| change family splitting | 339 s | -103 |

Two of the changes I added make custom filter criteria more efficient, while two are unrelated and could also be applied on master. That means that for this particular example, custom filtering still hurts performance. I think that additional profiling would be a good idea, probably with longer jobs.

ajocher:
Thanks for helping, I really appreciate it. My untested suspicion is that some of the Numpy vectorization operations we have added, especially in the .pyx files might not be as fast as the for loops that were there before.

amarkpayne:
> Thanks for helping, I really appreciate it. My untested suspicion is that some of the Numpy vectorization operations we have added, especially in the .pyx files might not be as fast as the for loops that were there before.

I agree with @ajocher . Specifically, I think my recommendations [here](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/78a7a307c98bd66d573d11494941ab96ad82d83a/rmgpy/rmg/main.py#L1495) were bad and should be reverted back to for loops

JacksonBurns:

PR #1592, opened 2019-05-02
Parse CCSD(T) energies (no F12) in Molpro

alongd:
I found out that some jobs don't have the ` Electronic Energy at 0 [K]:                      -768.275662 [H]` line we're currently looking for when parsing energies from a non-F12 CCSD calculation in Molpro. I made a test out of this example, and modified the parser accordingly.

### DIFF
--- a/arkane/molpro.py
+++ b/arkane/molpro.py
@@ -50,7 +50,7 @@ class MolproLog(Log):
     """
 
     def __init__(self, path):
-        self.path = path
+        super(MolproLog, self).__init__(path)
 
     def getNumberOfAtoms(self):
         """
@@ -392,3 +392,9 @@ class MolproLog(Log):
             raise Exception('Unable to find imaginary frequency in Molpro output file {0}'.format(self.path))
         negativefrequency = -float(frequency)
         return negativefrequency
+
+    def loadScanEnergies(self):
+        """
+        Rotor scans are not implemented in Molpro
+        """
+        raise NotImplementedError('Rotor scans not implemented in Molpro')
--- a/arkane/molpro.py
+++ b/arkane/molpro.py
@@ -38,6 +38,7 @@ from rmgpy.statmech import IdealGasTranslation, NonlinearRotor, LinearRotor, Har
 
 from arkane.common import get_element_mass
 from arkane.log import Log
+
 ################################################################################
 
 
@@ -45,10 +46,9 @@ class MolproLog(Log):
     """
     Represents a Molpro log file. The attribute `path` refers to the
     location on disk of the Molpro log file of interest. Methods are provided
-    to extract a variety of information into Arkane classes and/or NumPy
-    arrays. 
+    to extract a variety of information into Arkane classes and/or NumPy arrays.
     """
-    
+
     def __init__(self, path):
         self.path = path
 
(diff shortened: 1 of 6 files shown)


arkane/molpro.py
```diff
@@ -278,17 +278,21 @@ def loadEnergy(self, frequencyScaleFactor=1.):
...
-        E0 = None
+        e0 = None
         with open(self.path, 'r') as f:
             lines = f.readlines()
-            # Determine whether this is f12a or f12b according to the basis set, or whether this is MRCI.
-            f12a, f12b, mrci = False, False, False
+            # Determine whether the sp method is f12,
+            # if so whether we should parse f12a or f12b according to the basis set.
+            # Otherwise, check whether the sp method is MRCI.
+            f12, f12a, f12b, mrci = False, False, False, False
```
cgrambow: Why do you need the `f12` variable? Could you just check that both `f12a` and `f12b` are false?

arkane/molpro.py
```diff
@@ -302,44 +306,45 @@ def loadEnergy(self, frequencyScaleFactor=1.):
...
                     # First search for MRCI+Davidson energy
                     if '(Davidson, relaxed reference)' in line:
-                        E0 = float(line.split()[3])
+                        e0 = float(line.split()[3])
                         logging.debug('Found MRCI+Davidson energy in molpro log file {0}, using this value'.format(
                             self.path))
                         break
-            if E0 is None and mrci:
+                elif not f12:
+                    if 'Electronic Energy at 0' in line:
```
cgrambow: Just curious, do you know when (if ever) this gets used?

arkane/molproTest.py
```diff
@@ -52,58 +52,58 @@ def testLoadDzFromMolproLog_F12(self):
...
     def testLoadQzFromMolproLog_F12(self):
         """
         Uses a Molpro log file for ethylene_qz (C2H4) to test that F12b
         energy can be properly read.
         """
         
-        log=MolproLog(os.path.join(os.path.dirname(__file__),'data','ethylene_f12_qz.out'))
-        E0=log.loadEnergy()
+        log = MolproLog(os.path.join(os.path.dirname(__file__), 'data', 'ethylene_f12_qz.out'))
+        eo = log.loadEnergy()
```
cgrambow: `e0`

arkane/molproTest.py
```diff
@@ -52,58 +52,58 @@ def testLoadDzFromMolproLog_F12(self):
...
         
-        self.assertAlmostEqual(E0 / constants.Na / constants.E_h, -78.472682755635, 5)
+        self.assertAlmostEqual(eo / constants.Na / constants.E_h, -78.472682755635, 5)
```
cgrambow: `e0`

arkane/molpro.py
alongd: We determine f12a and f12b from the basis set, so one of them could be True, but the method won't necessarily be F12

arkane/molpro.py
alongd: Not really. This is actually the original line we were parsing so far for non-F12 sp jobs. The on;y thing "different" about the test I added is that it is QZ (`basis=aug-cc-pvqz`). Maybe that changes something in the output format.

arkane/molpro.py
alongd: They don't have to be on the same line if one uses an F12 method but not an F12 basis set (for which we have model chemistries)

alongd:
I didn't use fixup commits since I had to re-commit stuff to better organize the PR. I think I addressed everything (though I left the f12 flag, let me know if you think of a better way), and force pushed.

PR #1597, opened 2019-05-09
Modify test-database make target to not capture stdout and log

mliu49:
### Motivation or Problem
When database tests fail, you get a million lines of logging statements, making it a pain to find the actual test results.

### Description of Changes
Disable log capturing for `make test-database`.

### Testing
Break the database and try running the tests. A couple easy options: change an index in training reactions to be a duplicate or change a group definition to be more specific than its children.


mliu49:
@mjohnson541, could you please take a look and merge this? It will only take a minute.

mjohnson541:
Can you rebase?

PR #1599, opened 2019-05-13
Tests: Added a Molpro negative frequency test to Arkane

alongd:
Bumping up RMG-Py's coverage by 0.01%!


alongd:
@mjohnson541, this is a minor PR (just added a test). Could you take a look?

alongd:
Sure!

Issue #1598, opened 2019-05-10
Could not determine representative localized structures


AH9316:
### Topic
General area which your question is related to.
 - [ ] Installation of RMG
 - [x] Running an RMG job
 - [ ] Using RMG API
 - [ ] Arkane (formerly CanTherm)
 - [ ] Dependencies
 - [x] An error message 

### Context
With running one RMG simulation, I get an error:
```
Traceback (most recent call last):
  File "/opt/scientific/RMG-Py-stable/rmgpy/scoop_framework/util.py", line 112, in __call__
    return self.myfn(*args, **kwargs)
  File "/opt/scientific/RMG-Py-stable/rmgpy/rmg/react.py", line 75, in reactSpecies
    reactions = getDB('kinetics').generate_reactions_from_families(speciesTuple)
  File "/opt/scientific/RMG-Py-stable/rmgpy/data/kinetics/database.py", line 513, in generate_reactions_from_families
    reaction_list = find_degenerate_reactions(reaction_list, same_reactants, kinetics_database=self)
  File "/opt/scientific/RMG-Py-stable/rmgpy/data/kinetics/common.py", line 279, in find_degenerate_reactions
    rxn0.ensure_species()
  File "rmgpy/reaction.py", line 1108, in rmgpy.reaction.Reaction.ensure_species
    def ensure_species(self, reactant_resonance=False, product_resonance=True):
  File "rmgpy/reaction.py", line 1126, in rmgpy.reaction.Reaction.ensure_species
    ensure_species(self.products, resonance=product_resonance, keep_isomorphic=True)
  File "/opt/scientific/RMG-Py-stable/rmgpy/data/kinetics/common.py", line 171, in ensure_species
    new_item.generate_resonance_structures(keep_isomorphic=keep_isomorphic)
  File "rmgpy/species.py", line 168, in rmgpy.species.Species.generate_resonance_structures
    def generate_resonance_structures(self, keep_isomorphic=True, filter_structures=True):
  File "rmgpy/species.py", line 178, in rmgpy.species.Species.generate_resonance_structures
    self.molecule = self.molecule[0].generate_resonance_structures(keep_isomorphic=keep_isomorphic,
  File "rmgpy/molecule/molecule.py", line 1775, in rmgpy.molecule.molecule.Molecule.generate_resonance_structures
    def generate_resonance_structures(self, keep_isomorphic=False, filter_structures=True):
[...]

Issue #1578, opened 2019-04-17
Unable to determine respresentative structures (resonance)


mefuller:
### Bug Description
During mechanism generation, my RMG run comes to a screeching halt with the following behavior:
```
ResonanceError: Could not determine representative localized structures for species [O-][N+](=O)C1C=CC=CC=1[N]C1=CC=CC=C1

Traceback (most recent call last):
  File "/home/fuller/Documents/CFG Lab/RMG/RMG-Py/rmg.py", line 173, in <module>
    main()
  File "/home/fuller/Documents/CFG Lab/RMG/RMG-Py/rmg.py", line 167, in main
    rmg.execute(**kwargs)
  File "/home/fuller/Dropbox/Documents/CFG Lab/RMG/RMG-Py/rmgpy/rmg/main.py", line 664, in execute
    trimolecularReact=self.trimolecularReact)
  File "/home/fuller/Dropbox/Documents/CFG Lab/RMG/RMG-Py/rmgpy/rmg/model.py", line 679, in enlarge
    unimolecularReact, bimolecularReact, trimolecularReact=trimolecularReact)
  File "/home/fuller/Dropbox/Documents/CFG Lab/RMG/RMG-Py/rmgpy/rmg/react.py", line 143, in reactAll
    rxns = list(react(*spcTuples))
  File "/home/fuller/Dropbox/Documents/CFG Lab/RMG/RMG-Py/rmgpy/rmg/react.py", line 59, in react
    spcTuples)
  File "/home/fuller/Dropbox/Documents/CFG Lab/RMG/RMG-Py/rmgpy/scoop_framework/util.py", line 154, in map_
    return map(WorkerWrapper(args[0]), *args[1:], **kwargs)
  File "/home/fuller/miniconda2/envs/rmg_env/lib/python2.7/site-packages/scoop-0.7.2.0-py2.7.egg/scoop/fallbacks.py", line 49, in wrapper
  File "/home/fuller/Dropbox/Documents/CFG Lab/RMG/RMG-Py/rmgpy/scoop_framework/util.py", line 112, in __call__
    return self.myfn(*args, **kwargs)
  File "/home/fuller/Dropbox/Documents/CFG Lab/RMG/RMG-Py/rmgpy/rmg/react.py", line 75, in reactSpecies
    reactions = getDB('kinetics').generate_reactions_from_families(speciesTuple)
[...]

PR #1600, opened 2019-05-13
Fix resonance structure generation and filtration bugs

mliu49:
### Motivation or Problem
Original issue was `UndeterminableKineticsError` while generating `intra_H_migration` reaction between `C1=C[C]2C=CC3=CC2=C3C1` and `[C]1=CC2=CC3=C2CC=CC13`.

Cause was inconsistent resonance structure generation resulting from minor differences in molecule objects, which led to the inability to identify the correct reverse reaction.

The cause of the resonance structure inconsistency was traced back to an inconsistency in perceived aromatic rings as determined by `Molecule.getAromaticRings()`, which relies on RDKit.

This was compounded by a bug in `generate_optimal_aromatic_resonance_structures` which would only attempt to aromatize the radical delocalization structures with the largest number of aromatic rings.

The combined effect was as follows:
- For one starting structure, RDKit identified 2 potential aromatic rings, which RMG then determined to be not truly aromatic. Then, it did not attempt to aromatize the valid structures with 1 aromatic ring, leading to no aromatic structures.
- For another starting structure, RDKit only identified 1 potential aromatic ring, which RMG found to be valid, therefore successfully generating one aromatic structure.

### Description of Changes
This does not fix the root cause, which is inconsistent aromaticity perception. It does fix the second issue where `generate_optimal_aromatic_resonance_structures` only tries to aromatize the structures with the most aromatic rings.

This modifies the algorithm to instead try all of the delocalized structures, starting with the ones with the most aromatic rings. If that fails, it will move to the structures with the second most aromatic rings, and so on.

This also adds a failing unit test with the bug that was discovered.

With this change, RMG can successfully generate the aforementioned reaction.

### Testing
Try generating reactions for the molecules mentioned above.

### DIFF
--- a/rmgpy/molecule/molecule.py
+++ b/rmgpy/molecule/molecule.py
@@ -1085,9 +1085,17 @@ class Molecule(Graph):
         If `deep` is ``False`` or not specified, a shallow copy is made: the
         original vertices and edges are used in the new graph.
         """
-        other = cython.declare(Molecule)
+        cython.declare(g=Graph, other=Molecule, i=int, v1=Vertex, v2=Vertex)
         g = Graph.copy(self, deep)
         other = Molecule(g.vertices)
+        # Copy connectivity values and sorting labels
+        for i in xrange(len(self.vertices)):
+            v1 = self.vertices[i]
+            v2 = other.vertices[i]
+            v2.connectivity1 = v1.connectivity1
+            v2.connectivity2 = v1.connectivity2
+            v2.connectivity3 = v1.connectivity3
+            v2.sortingLabel = v1.sortingLabel
         other.multiplicity = self.multiplicity
         other.reactive = self.reactive
         return other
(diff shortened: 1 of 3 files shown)


mliu49:
### Additional fix
The cause of #1578 and #1598 was identified to be lack of valid atom sorting labels. Octet and charge filtration both rely on sorting labels to identify atoms. In this case, Kekule structure generation was not copying connectivity values and sorting labels when copying the molecule. The resulting molecule had sorting labels of -1 for all atoms, leading to issues during filtration.

The chosen solution was to move connectivity value and sorting label copying to `Molecule.copy` instead of being done in each individual resonance structure method. This ensures that connectivity values and sorting labels are always transferred when copying a molecule.

alongd:
Looking good! Could you also add a unit test checking the expected resonance structures for cases of combined heteroatom groups and aromatic structures?

mliu49:
I added a unit test. It took me a while to recreate the error message in a unit test, but I figured out that the key was having `keep_isomorphic=True`, which prevents running `isIsomorphic` which would otherwise sort the atoms.

alongd:
I think it looks good. Would you like to rebase?

mliu49:
This should be good to go!

PR #1601, opened 2019-05-14
Determine a species linearity from xyz

alongd:
Arkane uses the species linearity when projecting out rotor frequencies.
This PR automates the determination of the linearity attribute, so now it is optional in input files.
If the problem can be reduced in dimensionality, we first identify that. We then check that all atoms from the 3rd one on form a straight line (either in 2D space or in 3D space) within some precision with the first two atoms.
Examples and documentations were updated. Tests were added.

### DIFF
--- a/arkane/statmech.py
+++ b/arkane/statmech.py
@@ -36,7 +36,7 @@ information for a single species or transition state.
 
 import os.path
 import math
-import numpy
+import numpy as np
 import logging
 
 from rdkit.Chem import GetPeriodicTable
@@ -120,8 +120,8 @@ class ScanLog(object):
                     angles.append(float(tokens[0]) / angleFactor)
                     energies.append(float(tokens[1]) / energyFactor)
 
-        angles = numpy.array(angles)
-        energies = numpy.array(energies)
+        angles = np.array(angles)
+        energies = np.array(energies)
         energies -= energies[0]
 
         return angles, energies
@@ -489,8 +489,8 @@ class StatMechJob(object):
                     fourierRotor = HinderedRotor(inertia=(inertia, "amu*angstrom^2"), symmetry=symmetry)
                     fourierRotor.fitFourierPotentialToData(angle, v_list)
 
-                    Vlist_cosine = numpy.zeros_like(angle)
-                    Vlist_fourier = numpy.zeros_like(angle)
+                    Vlist_cosine = np.zeros_like(angle)
+                    Vlist_fourier = np.zeros_like(angle)
                     for i in range(angle.shape[0]):
                         Vlist_cosine[i] = cosineRotor.getPotential(angle[i])
                         Vlist_fourier[i] = fourierRotor.getPotential(angle[i])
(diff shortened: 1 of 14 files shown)


arkane/statmech.py
```diff
@@ -970,7 +969,131 @@ def determine_qm_software(fullpath):
...
+        return True
+
+    x, y, z = [], [], []
+
+    for coordinate in coordinates:
+        x.append(coordinate[0])
+        y.append(coordinate[1])
+        z.append(coordinate[2])
+
+    # first check whether the problem can be reduced into one or two dimensions
```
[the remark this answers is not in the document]
alongd: I think both methods are equivalent in essence, but your suggestion is significantly more friendly to implement / maintain / debug. I'd happily adopt it.

alongd:
The tolerance is indeed a mystery.
I looked up CCCBDB for some linear molecules at a lousy (B3LYP/6-31G*) and a fancy method (CCSD(T)=FULL/aug-cc-pVQZ), but for CO2 and C2H2 both methods had exactly 180.0 (or 0.0) degrees between all triples. Thinking that they probably used Z-Mats, I ran CO2, C2H2, HCN, NCN(S), NCN(T), C#CC#C, and C#CC#CC#CC#C at wb97xd/6-311+G(d,p) using ARC (no ZMats), but got the same thing: all angles were dead on 180.0 (or 0.0) degrees (I used Gaussian).

So I plan on using 0.2 degrees for the tolerance, just based on my hunch, but I'm open for sugestions.

I added fixup commits and rebased, let me know when to squash and whether you have any other comments

Thanks for helping me improve the algorithm!

arkane/statmech.py
```diff
@@ -969,128 +969,30 @@ def determine_qm_software(fullpath):
...
-        else:
-            return True
-        return False
+    # A tensor containing all distance vectors in the molecule
+    d = -numpy.array([c[:, numpy.newaxis] - c[numpy.newaxis, :] for c in coordinates.T])
+    for i, c in enumerate(coordinates):
+        if i > 1:
+            u1 = d[:, 0, 1] / numpy.linalg.norm(d[:, 0, 1])  # unit vector between atoms 0 and 1
+            u2 = d[:, 1, i] / numpy.linalg.norm(d[:, 1, i])  # unit vector between atoms 1 and i
+            a = math.degrees(numpy.arccos(numpy.clip(numpy.dot(u1, u2), -1.0, 1.0)))  # angle between atoms 0, 1, i
```
cgrambow: Are you using `numpy.clip` to avoid invalid arguments for `numpy.arccos` if there are floating point errors in the calculation of the dot product?

arkane/statmech.py
```diff
@@ -969,128 +969,30 @@ def determine_qm_software(fullpath):
...
-            if ax and az and not nearly_equal(ax, az):
-                break
-            if ay and az and not nearly_equal(ay, az):
-                break
-        else:
-            return True
-        return False
+    # A tensor containing all distance vectors in the molecule
+    d = -numpy.array([c[:, numpy.newaxis] - c[numpy.newaxis, :] for c in coordinates.T])
+    for i, c in enumerate(coordinates):
```
cgrambow: You don't actually need to iterate over `coordinates`, correct? I'd just do
`for i in range(2, len(coordinates))`
and remove the if-statement on the next line.

cgrambow:
That actually really interesting and good to know. 0.2 is probably ok for now, maybe even smaller based on your observations? We can adjust as we learn more. I requested one small change. After that you can squash.

arkane/statmech.py
alongd: Yes, I thought to use cautious although my examples ran fine without it

arkane/statmech.py
alongd: absolutely!

alongd:
Thanks for all the comments, the PR is in a much better shape now. Let me know if there are any other issues, otherwise this should be good to go

alongd:
@cgrambow, I changed the tolerance to 0.1 degrees and added a descriptive comment (squashed)

arkane/statmech.py
```diff
@@ -970,7 +969,35 @@ def determine_qm_software(fullpath):
...
+    number_of_atoms = len(coordinates)
+    if number_of_atoms == 1:
+        return False
+    if number_of_atoms == 2:
+        return True
+
+    # A tensor containing all distance vectors in the molecule
+    d = -numpy.array([c[:, numpy.newaxis] - c[numpy.newaxis, :] for c in coordinates.T])
+    for i in range(2, len(coordinates)):
+        if i > 1:
```
cgrambow: You don't need this if-statement anymore since `i` now starts at 2.

alongd: Thanks, removed

PR #1604, opened 2019-05-22
Added GroupAtom isNitrogen() and isCarbon() methods 

alongd:
Added `isNitrogen()` and `isCarbon()` methods to `GroupAtom`
Also, I happened to take a look at `getPrincipalMomentsOfInertia` in conformer and couldn't resist making some minor changes.

### DIFF
--- a/rmgpy/molecule/group.py
+++ b/rmgpy/molecule/group.py
@@ -442,7 +442,7 @@ class GroupAtom(Vertex):
                         return False
         else:
             if group.lonePairs: return False
-        #Each charge in self must have an equivalent in other
+        # Each charge in self must have an equivalent in other
         if self.charge:
             for charge1 in self.charge:
                 if group.charge:
@@ -462,7 +462,6 @@ class GroupAtom(Vertex):
         # Otherwise self is in fact a specific case of other
         return True
 
-
     def isSurfaceSite(self):
         """
         Return ``True`` if the atom represents a surface site or ``False`` if not.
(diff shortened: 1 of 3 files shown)


rmgpy/molecule/group.py
```diff
@@ -472,22 +471,34 @@ def isSurfaceSite(self):
...
+        allNitrogen = [atomTypes['N']] + atomTypes['N'].specific
+        checkList=[x in allNitrogen for x in self.atomType]
+        return all(checkList)
+
+    def isCarbon(self):
+        """
+        Return ``True`` if the atom represents an sulfur atom or ``False`` if not.
+        """
+        allCarbon = [atomTypes['C']] + atomTypes['C'].specific
+        checkList=[x in allCarbon for x in self.atomType]
```
mliu49: Add spaces around `=`?

rmgpy/molecule/group.py
```diff
@@ -472,22 +471,34 @@ def isSurfaceSite(self):
...
 
+    def isNitrogen(self):
+        """
+        Return ``True`` if the atom represents an sulfur atom or ``False`` if not.
+        """
+        allNitrogen = [atomTypes['N']] + atomTypes['N'].specific
+        checkList=[x in allNitrogen for x in self.atomType]
+        return all(checkList)
+
+    def isCarbon(self):
```
mliu49: This is rather trivial, but you could consider placing these new methods in alphabetical order relative to `isOxygen` and `isSulfur`.

rmgpy/statmech/conformer.pyx
```diff
@@ -338,13 +338,13 @@ cdef class Conformer(RMGObject):
         Finally, the reduced moment of inertia is evaluated from the moment of inertia 
         of each top via the formula (I1*I2)/(I1+I2).  
         
-        Option corresponds to 3 possible ways of calculating the internal reduced moment of inertia
-        as discussed in East and Radom [2]
+        `option` is an integer corresponding to one of three possible ways of calculating the internal reduced moment
```
mliu49: I think double tick marks are used for literals, single ticks are for links.
```
``option``
```

rmgpy/statmech/conformer.pyx
```diff
@@ -442,7 +445,7 @@ cdef class Conformer(RMGObject):
                           
         else:
             
-            raise ValueError("option {0} unimplemented or non-existant".format(option))
+            raise ValueError("option {0} unimplemented or non-existing".format(option))
```
mliu49: Perhaps change this message to `Option {0} is invalid. Should be either 1, 2, or 3.`? Since only three options exist and all are implemented.

alongd:
Thanks! Changes were squashed.

Note that I also deleted `electronpairs` from the comment:
`atomtype, electronpairs, lone pairs, charge, and bond order. Returns`
in GroupAtom.hasWildcards(), https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1604/files#diff-199e8ac7eefa2bc7391991c9df68e6e1L496
I think they meant just lone pairs, which is already specified there

rmgpy/molecule/group.py
```diff
@@ -472,31 +471,42 @@ def isSurfaceSite(self):
...
+        """
+        allCarbon = [atomTypes['C']] + atomTypes['C'].specific
+        checkList = [x in allCarbon for x in self.atomType]
         return all(checkList)
 
     def hasWildcards(self):
         """
         Return ``True`` if the atom has wildcards in any of the attributes:
-        atomtype, electronpairs, lone pairs, charge, and bond order. Returns
+        atomtype, lone pairs, charge, and bond order. Returns
```
mliu49: I think it was probably meant to say radical electrons.

alongd: I agree, changed.
Also, the docstring says "charge", but the method doesn't check for it

mliu49: It seems you squashed the fix into the other commit. So the first commit removes electronpairs and the second commit adds radical electrons. It would be best if you could move it to one commit.

alongd: oops! fixed

PR #1605, opened 2019-05-22
Refactor energy corrections in Arkane and implement new bond additivity correction type

cgrambow:
### Motivation or Problem
Applying energy corrections as part of a statmech job in Arkane takes place in a really messy function and does not allow for easy extension to new bond additivity correction types.

### Description of Changes
This PR refactors energy corrections into separate modules and stores the data separately. Furthermore, it implements a new type of bond additivity correction scheme.

### Testing
Existing unit tests pass.

### Discussion
This PR provides the functionality to select either type of bond additivity correction, but it doesn't yet provide data for the new BAC type. My plan is to completely replace the current atom energies using just the calculated values at the respective levels of theory and to provide newly-derived bond additivity corrections for all levels of theory for both the existing type of BACs and the newly implemented type of BACs once we have selected a reference set of molecules and performed the quantum calculations at all the levels of theory.

This PR can be merged without the new data because it doesn't change current Arkane behavior. Alternatively, we can wait with merging until the new data is available. What do you think? In either case, the current changes are ready for review.


arkane/encorr/mbac.py
```diff
@@ -0,0 +1,141 @@
...
+}
+
+
+def get_bac_correction(model_chemistry, coords, nums, multiplicity=1, mol_corr=0.0):
+    """
+    Given the model chemistry, molecular coordinates, atomic numbers,
+    and dictionaries of BAC parameters, return the total BAC correction
+    (should be SUBTRACTED from energy).
+
+    Note that a molecular correction term other than 0 destroys the size
```
alongd: Could you explain this point? What's a molecular correction term? Is it a general name for BACs and AECs? What are the practical implications of destroying the QC method consistency?

arkane/encorr/mbac.py
```diff
@@ -0,0 +1,141 @@
...
+def geo_to_mol(coords, nums):
+    """
+    Convert molecular geometry specified by atomic coordinates and
+    atomic numbers to RMG molecule.
+
+    Use Open Babel for most cases because it's better at recognizing
+    long bonds. Use RMG for hydrogen because Open Babel can't do it for
+    mysterious reasons.
+    """
+    if len(nums) == 2 and all(n == 1 for n in nums):
```
alongd: Perhaps `if nums == [1, 1]` instead?
Why hard code for H2, or am I not reading it correctly?

arkane/encorr/mbac.py
```diff
@@ -0,0 +1,141 @@
...
+        mol = Molecule()
+        mol.fromXYZ(nums, coords)
+    else:
+        xyz = '{}\n\n'.format(len(nums))
+        xyz += '\n'.join('{0}  {1[0]: .10f}  {1[1]: .10f}  {1[2]: .10f}'.format(s, c) for s, c in zip(coords, nums))
```
alongd: Can you take another look here? Looks like `coords` and `nums` should be switched in the `zip`

documentation/source/users/arkane/input.rst
```diff
@@ -28,6 +28,7 @@ Component                   Description
 ``useHinderedRotors``       ``True`` (by default) if hindered rotors are used, ``False`` if not
 ``useAtomCorrections``      ``True`` (by default) if atom corrections are used, ``False`` if not
 ``useBondCorrections``      ``True`` if bond corrections are used, ``False`` (by default) if not
+``bondCorrectionType``      ``'p'`` for Petersson-type or ``'m'`` for Melius-type bond additivity corrections
```
alongd: Could you specify what the default is?

documentation/source/users/arkane/input.rst
```diff
@@ -188,7 +189,12 @@ to apply atomization energy corrections (AEC) and spin orbit corrections (SOC) f
 (see `Model Chemistry`_). If not interested in accurate thermodynamics (e.g., if only using ``kinetics()``), then
 atom corrections can be turned off by setting ``useAtomCorrections`` to ``False``.
 
-The ``bond`` parameter is used to apply bond corrections (BC) for a given ``modelChemistry``.
+The ``bond`` parameter is used to apply bond corrections (BC) for a given ``modelChemistry`` if using Petersson-type
```
alongd: Can we change `to apply bond corrections (BC)` into `to apply bond additivity corrections (BAC)`, and then just use BAC later on?

documentation/source/users/arkane/input.rst
```diff
@@ -188,7 +189,12 @@ to apply atomization energy corrections (AEC) and spin orbit corrections (SOC) f
...
+bond additivity corrections (BACs) (``bondCorrectionType = 'p'``). When using Melius-type bond additivity corrections
+(``bondCorrectionType = 'm'``), specifying bonds is not required because the molecular connectivity is automatically
```
alongd: Perhaps change `specifying bonds is not required` into `specifying bond orders is not required`

arkane/encorr/pbac.py
```diff
@@ -0,0 +1,73 @@
...
+    except KeyError:
+        raise Exception('Missing BAC parameters for model chemistry {}'.format(model_chemistry))
+
+    # Sum corrections
+    bac = 0.0
+    for symbol, count in bonds.items():
+        if symbol in params:
+            bac += count * params[symbol] * 4184.0
+        elif symbol[::-1] in params:
+            bac += count * params[symbol[::-1]] * 4184.0
```
mliu49: This method will break once we start supporting elements with two letter symbols, right? Although it might be a moot point if we're deprecating this type of correction anyway.

arkane/encorr/corr.py
```diff
@@ -0,0 +1,118 @@
...
+    # basis by subtracting out that model chemistry's atomic energies
+    if atom_energies is None:
+        try:
+            atom_energies = data.atom_energies[model_chemistry]
+        except KeyError:
+            raise Exception('Missing atom energies for model chemistry {}'.format(model_chemistry))
+
+    for symbol, count in atoms.items():
+        if symbol in atom_energies:
+            corr -= count * atom_energies[symbol] * 4.35974394e-18 * constants.Na
```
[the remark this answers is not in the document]
cgrambow: I added comments everywhere but just kept the numerical values in there for now. I'm inclined to not add unit conversions to constants.py, but am open to other suggestions.

arkane/encorr/data.py
```diff
@@ -0,0 +1,405 @@
...
+###############################################################################
+
+"""
+This file provides atomic energy and BAC parameters for several model
+chemistries.
+"""
+
+# Atom energy corrections to reach gas-phase reference state
+# Experimental enthalpy of formation at 0 K, 1 bar for gas phase
+# See Gaussian thermo whitepaper at http://www.gaussian.com/g_whitepap/thermo.htm)
```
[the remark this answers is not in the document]
cgrambow: Turns out the link has actually changed, so I replaced it with a more up-to-date link.

arkane/encorr/data.py
```diff
@@ -0,0 +1,405 @@
...
+# Publishing Corp., New York, 1989. (http://www.science.uwaterloo.ca/~cchieh/cact/tools/thermodata.html)
+atom_hf = {'H': 51.63, 'He': -1.481,
+           'Li': 37.69, 'Be': 76.48, 'B': 136.2, 'C': 169.98, 'N': 112.53, 'O': 58.99, 'F': 18.47, 'Ne': -1.481,
+           'Na': 25.69, 'Mg': 34.87, 'Al': 78.23, 'Si': 106.6, 'P': 75.42, 'S': 65.66, 'Cl': 28.59,
+           'K': 36.841, 'Ca': 41.014, 'Ti': 111.2, 'Cu': 79.16, 'Zn': 29.685, 'Ge': 87.1, 'Br': 25.26,
+           'Kr': -1.481,
+           'Rb': 17.86, 'Ag': 66.61, 'Cd': 25.240, 'Sn': 70.50, 'I': 24.04, 'Xe': -1.481,
+           'Cs': 16.80, 'Hg': 13.19, 'Pb': 15.17}
+
+# Thermal contribution to enthalpy Hss(298 K) - Hss(0 K) reported by Gaussian thermo whitepaper
```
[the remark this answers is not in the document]
cgrambow: These terms just describe the enthalpy changes for the elements going from 0 K to 298 K. I think the Hss notation was left over from a previous version of Arkane, so I just replaced it.

arkane/encorr/data.py
```diff
@@ -0,0 +1,405 @@
...
+# Atom energy corrections to reach gas-phase reference state
+# Experimental enthalpy of formation at 0 K, 1 bar for gas phase
+# See Gaussian thermo whitepaper at http://www.gaussian.com/g_whitepap/thermo.htm)
+# Note: These values are relatively old and some improvement may be possible by using newer values
+# (particularly for carbon).
+# However, care should be taken to ensure that they are compatible with the BAC values (if BACs are used)
+# He, Ne, K, Ca, Ti, Cu, Zn, Ge, Br, Kr, Rb, Ag, Cd, Sn, I, Xe, Cs, Hg, and Pb are taken from CODATA
+# Codata: Cox, J. D., Wagman, D. D., and Medvedev, V. A., CODATA Key Values for Thermodynamics, Hemisphere
+# Publishing Corp., New York, 1989. (http://www.science.uwaterloo.ca/~cchieh/cact/tools/thermodata.html)
+atom_hf = {'H': 51.63, 'He': -1.481,
```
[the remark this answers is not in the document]
cgrambow: PEP 8 does recommend that, but I think I would prefer not writing them in ALL_CAPS because they're basically data structures storing a lot of information rather than single constants. This is also in line with the dictionaries in `common.py`. I'm basically treating `data.py` as a file that contains no logic other than variable assignments, so similar to what a yaml or json file would be. I kept it as a Python file though because the data structure isn't particularly complicated and it allows for easy importing. Let me know if you still think I should change it or if we should consider storing the data in a different format altogether.

arkane/encorr/data.py
```diff
@@ -0,0 +1,405 @@
...
+    # * indicates that the grid size used in the [QChem] electronic
+    # structure calculation utilized 75 radial points and 434 angular points
+    # (i.e,, this is specified in the $rem section of the [qchem] input file as: XC_GRID 000075000434)
+    'm08so/mg3s*': {
+        'H': -0.5017321350 + SOC['H'], 'N': -54.5574039365 + SOC['N'],
+        'O': -75.0382931348 + SOC['O'], 'C': -37.8245648740 + SOC['C'],
+        'P': -341.2444299005 + SOC['P'], 'S': -398.0940312227 + SOC['S']
+    },
+
+    'klip_1': {
```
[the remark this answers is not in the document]
cgrambow: I was thinking to actually just completely replace unused model chemistries once we have the new BAC data. I highly doubt that anyone is using some of the more obscure ones and don't think anyone would miss them.

arkane/encorr/data.py
```diff
@@ -0,0 +1,405 @@
...
+        'H': -0.499946213243 + SOC['H'], 'N': -54.526406291655 + SOC['N'],
+        'O': -74.995458316117 + SOC['O'], 'C': -37.788203485235 + SOC['C']
+    },
+
+    'ccsd(t)-f12/cc-pvdz-f12_h-qz': {
+        'H': -0.499994558325 + SOC['H'], 'N': -54.526406291655 + SOC['N'],
+        'O': -74.995458316117 + SOC['O'], 'C': -37.788203485235 + SOC['C']
+    },
+
+    # We are assuming that SOC is included in the Bond Energy Corrections
```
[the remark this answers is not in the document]
cgrambow: We have two options: Either we add SOCs to all of the atom energies and then fit BACs or we don't add SOCs at all in which case they will be implicitly captured in the BACs. I prefer the second option.

arkane/encorr/data.py
```diff
@@ -0,0 +1,405 @@
...
+
+    # ccsd(t)/aug-cc-pvtz(-pp) atomic energies were fit to a set of 8 small molecules:
+    # CH4, CH3OH, H2S, H2O, SO2, HI, I2, CH3I
+    'ccsd(t)/aug-cc-pvtz(-pp)': {
+        'H': -0.499821176024 + SOC['H'], 'O': -74.96738492 + SOC['O'],
+        'C': -37.77385697 + SOC['C'], 'S': -397.6461604 + SOC['S'],
+        'I': -294.7958443 + SOC['I']
+    },
+
+    # note that all atom corrections but S are fitted, the correction for S is calculated
```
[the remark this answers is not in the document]
cgrambow: My plan was to have just one set of atom energies for both BAC methods and to just use the pure quantum calculations for the atoms at each level of theory as the atomic energies. In that way, all corrections due to systematic errors appear as part of the BACs. As far as I can tell, it's actually not very common in literature to fit both atom and bond corrections when using the conventional style of BACs, so I don't know why Arkane has been doing that thus far. For Melius-type BACs, corrections to the atomic energies are actually built in to the method.

arkane/encorr/data.py
```diff
@@ -0,0 +1,405 @@
...
+        'H': -0.499946213243 + SOC['H'], 'N': -54.586402551258 + SOC['N'],
+        'O': -75.062767632757 + SOC['O'], 'C': -37.842729156944 + SOC['C']
+    },
+
+    'b-ccsd(t)-f12/cc-pcvqz-f12': {
+        'H': -0.49999456 + SOC['H'], 'N': -54.587781507581 + SOC['N'],
+        'O': -75.065397706471 + SOC['O'], 'C': -37.843634971592 + SOC['C']
+    },
+
+
```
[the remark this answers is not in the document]
cgrambow: How did you even catch this

arkane/encorr/mbac.py
```diff
@@ -0,0 +1,141 @@
...
+
+atom_spins = {
+    'H': 0.5, 'C': 1.0, 'N': 1.5, 'O': 1.0, 'F': 0.5, 'Si': 1.0, 'P': 1.5, 'S': 1.0, 'Cl': 0.5, 'Br': 0.5, 'I': 0.5
+}
+
+
+def get_bac_correction(model_chemistry, coords, nums, multiplicity=1, mol_corr=0.0):
+    """
+    Given the model chemistry, molecular coordinates, atomic numbers,
+    and dictionaries of BAC parameters, return the total BAC correction
```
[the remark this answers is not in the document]
cgrambow: I was being inconsistent in other places as well, so I just went ahead and replaced all occurrences of "BAC correction" by "BAC", including in function names.

arkane/encorr/mbac.py
cgrambow: Equation 3 and 4 in the Melius paper describe the molecular correction term. Basically, it is a single additional parameter that corrects for potential systematic errors in the overall electronic structure of the molecule due to the difference in spin quantum number of the molecule and summed spin quantum numbers of the atoms. The paper gives a good example of what it means to destroy size consistency: "For example, the sum of the calculated energies for hydrogen and oxygen atoms is not the same as the energy of the hydroxyl radical when the two atoms are fixed at infinity." Size consistency is a desirable property of quantum chemistry methods, but I don't actually know how much of an issue it would be in practice to have this term. So far, I haven't observed any benefit when fitting the term in addition to the other parameters so right now I'm just passing it as an argument to the function with a value of 0 and don't plan on including it in the `mbac` dictionary in `data.py`.

arkane/encorr/mbac.py
cgrambow: Oops, great catch.

documentation/source/users/arkane/input.rst
cgrambow: It means that the `bonds` dictionary is not required in the input file. To highlight that, I added `` around it so that it gets parsed as plain text in the documentation.

arkane/encorr/pbac.py
cgrambow: Didn't even think of that. I replaced it with a regex which works for all possible scenarios.

arkane/encorr/mbac.py
alongd: Will `if list(nums) == [1, 1]` work?

OK. Could you add a small comment explaining why H2 was hardcoded?

cgrambow: It's mentioned in the docstring of the function. Is that not enough?

cgrambow:
I think I addressed all your comments and made some fixup commits. I just realized that I caused a circular import issue, so I'll fix that, too.

arkane/encorr/mbac.py
alongd: Perfect! (I missed it)

cgrambow:
Added two more fixup commits. I introduced an `exceptions.py` file in the main Arkane directory and added the exceptions there to get rid of the circular imports.

arkane/encorr/corr.py
```diff
@@ -42,20 +42,31 @@
...
                           atom_energies=None, apply_atom_corrections=True,
-                          apply_bacs=False, bac_type='p'):
+                          apply_bac=False, bac_type='p'):
     """
     Calculate a correction to the electronic energy obtained from a
     quantum chemistry calculation at a given model chemistry such that
     it is consistent with the normal gas-phase reference states.
     Optionally, correct the energy using bond additivity corrections.
+
+    model_chemistry: The model chemistry, typically specified as method/basis.
```
mliu49: You can format the arguments to get nice styling when we compile the API documentation. Minimum would be adding the `Args` keyword and indenting:
```
Args:
    model_chemistry: The model chemistry, typically specified as method/basis.
    ...
```

See https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html for even more details.

cgrambow: Good to know!

alongd:
@cgrambow, can you rebase?
What else is needed for this PR?
Let's get it in before the Py3 transision

cgrambow:
Squashed and rebased. Nothing else is needed for this PR, except for BAC data, but we don't have that yet, so we can add it after Py3 transition.

alongd:
Could you take a look at the Codacy comment?

arkane/encorr/corr.py
```diff
@@ -0,0 +1,118 @@
...
+
+    # Step 2: Atom energy corrections to reach gas-phase reference state
+    atom_enthalpy_corrections = {symbol: data.atom_hf[symbol] - data.atom_thermal[symbol] for symbol in data.atom_hf}
+    for symbol, count in atoms.items():
+        if symbol in atom_enthalpy_corrections:
+            corr += count * atom_enthalpy_corrections[symbol] * 4184.0
+        else:
+            raise Exception(
+                'Element "{}" is not yet supported in Arkane.'
+                ' To include it, add its experimental heat of formation'.format(symbol)
```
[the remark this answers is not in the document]
alongd: @cgrambow, was this comment addressed?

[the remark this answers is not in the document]
cgrambow: Now it has been.

cgrambow:
The Codacy comment can be ignored. Apparently Codacy isn't smart enough to figure out that `c` in `coords` can be indexed.

PR #1612, opened 2019-06-02
Derive the frequency scaling factor in Arkane from the frequency level

alongd:
### Motivation or Problem
If the user did not specify the frequency scaling factor, Arkane uses the model chemistry to determine it from a dictionary. However, the model chemistry in Arkane currently respesents the sp level, whereas the freq level should have been used instead.

### Description of Changes
Arkane can now process a model chemistry in an sp//freq form, e.g., `CCSD(T)-F12a/aug-cc-pVTZ//B3LYP/6-311++G(3df,3pd)`, and determines the frequency scaling factor from the **freq** level, if available.
Also, if the frequency scaling factor was not specified and could not be determined from the dictionary, we now raise an error instead of setting it to `1` and warn the user. This way, software like ARC could capture the error and attempt to derive the frequency scaling factor by itself.

### Testing
Added an example which also runs in RMG's tests. Also added a unit test for the new `process_model_chemistry` function.

### DIFF
--- a/arkane/input.py
+++ b/arkane/input.py
@@ -638,14 +638,16 @@ def loadInputFile(path):
             logging.error('The input file {0!r} was invalid:'.format(path))
             raise
 
-    model_chemistry = local_context.get('modelChemistry', '')
-    level_of_theory = local_context.get('levelOfTheory', '')
+    model_chemistry = local_context.get('modelChemistry', '').lower()
+    sp_level, freq_level = process_model_chemistry(model_chemistry)
+
     author = local_context.get('author', '')
-    if 'frequencyScaleFactor' not in local_context:
-        logging.debug('Assigning a frequencyScaleFactor according to the modelChemistry...')
-        frequency_scale_factor = assign_frequency_scale_factor(model_chemistry)
-    else:
+    if 'frequencyScaleFactor' in local_context:
         frequency_scale_factor = local_context.get('frequencyScaleFactor')
+    else:
+        logging.debug('Tying to assign a frequencyScaleFactor according to the frequency '
+                      'level of theory {0}'.format(freq_level))
+        frequency_scale_factor = assign_frequency_scale_factor(freq_level)
     use_hindered_rotors = local_context.get('useHinderedRotors', True)
     use_atom_corrections = local_context.get('useAtomCorrections', True)
     use_bond_corrections = local_context.get('useBondCorrections', False)
@@ -659,7 +661,7 @@ def loadInputFile(path):
     for job in jobList:
         if isinstance(job, StatMechJob):
             job.path = os.path.join(directory, job.path)
-            job.modelChemistry = model_chemistry.lower()
+            job.modelChemistry = sp_level
             job.frequencyScaleFactor = frequency_scale_factor
             job.includeHinderedRotors = use_hindered_rotors
             job.applyAtomEnergyCorrections = use_atom_corrections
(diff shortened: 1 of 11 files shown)


alongd:
I changed the default behaviour if the freq scale factor isn't found back to what it was before this PR: of setting it to 1 and warning the user instead of crushing. I think it makes more because there are cases the user doesn't care about the freq scale factor such as running an RMG network, or running the Explorer tool. We shouldn't make Arkane crush if the user doesn't artificially set the factor to 1 in the input file.

Also, thanks @dranasinghe for spotting this issue to begin with!

I think it should be ready for review now.

alongd:
Added more unit tests to Arkane's inputTest (force pushed)

arkane/input.py
```diff
@@ -659,24 +661,52 @@ def loadInputFile(path):
...
+        if model_chemistry.startswith('cbs-qb3'):
+            # hard code for CBS-QB3-Paraskevas which has the same frequency scaling factor as CBS-QB3
+            freq_level = 'cbs-qb3'
+    elif '//' in model_chemistry:
+        # assume this is an sp//freq format, split
+        sp_level, freq_level = model_chemistry.split('//')
+    else:
+        # assume only the sp level was given in the model chemistry, assign to sp and return None for freq
+        sp_level = model_chemistry
+        freq_level = None
```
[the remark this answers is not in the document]
alongd: I'll change accordingly, but we still need the `/` that differentiates this from a composite method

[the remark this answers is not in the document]
cgrambow: In both cases we set `sp_level` and `freq_level` to the same thing, so we don't have to check composite methods separately.

alongd:
Thanks for the comments! I added fixups, let me know when to squash

arkane/input.py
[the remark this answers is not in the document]
alongd: I agree, added a fixup

arkane/input.py
```diff
@@ -706,7 +706,7 @@ def process_model_chemistry(model_chemistry):
         # assume this is an sp//freq format, split
         sp_level, freq_level = model_chemistry.split('//')
     else:
-        # assume only the sp level was given in the model chemistry, assign to sp and return None for freq
+        # assume the sp freq levels are the same, assign to the model chemistry to both
```
cgrambow: I think some words might have gotten lost here. You probably meant something like `assume the sp and freq levels are the same, assign the model chemistry to both`

alongd:
Thanks!! Rebased and squashed

arkane/input.py
```diff
@@ -659,24 +661,49 @@ def loadInputFile(path):
...
+    elif '//' in model_chemistry:
+        # assume this is an sp//freq format, split
+        sp_level, freq_level = model_chemistry.split('//')
+    else:
+        # assume the sp and freq levels are the same, assign the model chemistry to both
+        # (this could also be a composite method, and we'll expect the same behavior)
+        sp_level = freq_level = model_chemistry
+        if model_chemistry.startswith('cbs-qb3'):
+            # hard code for CBS-QB3-Paraskevas which has the same frequency scaling factor as CBS-QB3
+            freq_level = 'cbs-qb3'
```
mliu49: Sorry about seeing this after merging, but this hard-coding is slightly redundant since there's an entry for `cbs-qb3-paraskevas` in `assign_frequency_scale_factor` (https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/arkane/statmech.py#L1224).

I think we could remove one or the other.

cgrambow: Sorry, didn't check the `assign_frequency_scale_factor` function in detail.

alongd: You're absolutely right, I haven't noticed the redundancy either.

PR #1615, opened 2019-06-04
Don't hard-code for CBS-QB3-Paraskevas in process_model_chemistry

alongd:
It already has a frequency scaling factor under Arkane statmech assign_frequency_scale_factor()

I chose to remove the newly added hard-coding instead of the table entry to leave the code generic, assuming that the scaling factor dictionary is more likely to change soon anyway (e.g., if we fit our own BACs).

Thanks, @mliu49, for spotting this duplicity.


mliu49:
Ah, seems like you also need to update the newly added unit test.

PR #1619, opened 2019-06-07
Scaling zpe

alongd:
### Motivation or Problem
Previously we used the __frequency__ scaling factor to also scale the zero point energy correction in Arkane. Principally, these should be two different factors. Luckily, there's a simple ratio between them, a factor of 1.014 (+/- 0.002, see https://pubs.acs.org/doi/10.1021/ct100326h section 3.1.3) - largely independent of the model chemistry. So only of the factors one should be known, the second can be easily determined. **This erroneous behaviour causes a systematic error of about 1.5 kJ/mol(!) - or 0.35 kcal/mol - to H298** (see below).

Thanks @dranasinghe for helping me understand what I'm doing :)

### Description of Changes
Dividing the freq scaling factor by 1.014 to obtain the ZPEC scaling factor when scaling the __ZPEC__.

Here are some species I ran (purposely selected ones with no torsions) prior and after this change at ccsd(t)-f12/cc-pvtz-f12//wb97xd/6-311++g(d,p), energies are in kJ/mol:

| Species | H298, master | H298, this branch | H298, ATcT |
| ------------- | ------------- | ------------- | ------------- |
| CH4 | -71.95 | -73.56 | -74.525 +/- 0.056 |
| NH3 | -41.81 | -43.06 | -45.558 +/- 0.030 |
| CO2 | -385.07 | -385.49 | -393.475 +/- 0.015 |
| C2H4 | 55.97 | 54.13 | 52.39 +/- 0.12 |
| C2H2 | 232.61 | 231.63 | 228.27 +/- 0.13 |

Before the changes we had an average deviation from ATcT values of 4.5 kJ/mol (for these species at this level), now it is 3.3 kJ/mol (not accounting for CO2, the deviation went down from 3.6 kJ/mol to 2.1 kJ/mol). If these species could be representative, the implication is that **we're improving Arkane's H298 by about 1.2 to 1.5 kJ/mol**, at least for this level.

### DIFF
--- a/arkane/gaussian.py
+++ b/arkane/gaussian.py
@@ -45,7 +45,6 @@ from rmgpy.exceptions import InputError
 from arkane.common import check_conformer_energy, get_element_mass
 from arkane.log import Log
 
-
 ################################################################################
 
 
@@ -65,23 +64,20 @@ class GaussianLog(Log):
         Return the number of atoms in the molecular configuration used in
         the Gaussian log file.
         """
-
         Natoms = 0
-        # Open Gaussian log file for parsing
-        f = open(self.path, 'r')
-        line = f.readline()
-        while line != '' and Natoms == 0:
-            # Automatically determine the number of atoms
-            if 'Input orientation:' in line and Natoms == 0:
-                for i in range(5):
-                    line = f.readline()
-                while '---------------------------------------------------------------------' not in line:
-                    Natoms += 1
-                    line = f.readline()
+
+        with open(self.path, 'r') as f:
             line = f.readline()
-        # Close file when finished
-        f.close()
-        # Return the result
+            while line != '' and Natoms == 0:
+                # Automatically determine the number of atoms
+                if 'Input orientation:' in line and Natoms == 0:
+                    for i in range(5):
+                        line = f.readline()
+                    while '---------------------------------------------------------------------' not in line:
+                        Natoms += 1
+                        line = f.readline()
+                line = f.readline()
+
         return Natoms
 
     def loadForceConstantMatrix(self):
@@ -94,32 +90,29 @@ class GaussianLog(Log):
         are J/m^2. If no force constant matrix can be found in the log file,
         ``None`` is returned.
         """
-
         F = None
 
         Natoms = self.getNumberOfAtoms()
         Nrows = Natoms * 3
 
-        f = open(self.path, 'r')
-        line = f.readline()
-        while line != '':
-            # Read force constant matrix
-            if 'Force constants in Cartesian coordinates:' in line:
-                F = numpy.zeros((Nrows, Nrows), numpy.float64)
-                for i in range(int(math.ceil(Nrows / 5.0))):
-                    # Header row
-                    line = f.readline()
-                    # Matrix element rows
-                    for j in range(i * 5, Nrows):
-                        data = f.readline().split()
-                        for k in range(len(data) - 1):
-                            F[j, i * 5 + k] = float(data[k + 1].replace('D', 'E'))
-                            F[i * 5 + k, j] = F[j, i * 5 + k]
-                # Convert from atomic units (Hartree/Bohr_radius^2) to J/m^2
-                F *= 4.35974417e-18 / 5.291772108e-11 ** 2
+        with open(self.path, 'r') as f:
             line = f.readline()
-        # Close file when finished
-        f.close()
+            while line != '':
+                # Read force constant matrix
+                if 'Force constants in Cartesian coordinates:' in line:
+                    F = numpy.zeros((Nrows, Nrows), numpy.float64)
+                    for i in range(int(math.ceil(Nrows / 5.0))):
+                        # Header row
+                        line = f.readline()
+                        # Matrix element rows
+                        for j in range(i * 5, Nrows):
+                            data = f.readline().split()
+                            for k in range(len(data) - 1):
+                                F[j, i * 5 + k] = float(data[k + 1].replace('D', 'E'))
+                                F[i * 5 + k, j] = F[j, i * 5 + k]
+                    # Convert from atomic units (Hartree/Bohr_radius^2) to J/m^2
+                    F *= 4.35974417e-18 / 5.291772108e-11 ** 2
+                line = f.readline()
 
         return F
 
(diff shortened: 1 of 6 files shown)


alongd:
Thanks for the insightful Q's, it already helped (I re-organized the code a bit)

The terminology for E0 is take from both a conversation with Bill as well as from Arkane's documentation (http://reactionmechanismgenerator.github.io/RMG-Py/users/arkane/input.html#option-2-directly-enter-molecular-properties). Our code wasn't consistent with this terminology. I also modified the documentation a bit more in this PR.

Yes, I agree. I changed ZPEC back to ZPE.

I just went through all scaling factors, we indeed had a few that were actually the ZPE factors. I also added a few more, and now they are all consistent. Note that the CBS-QB3 scaling factor changed as well. Surprisingly, I did *not* observe any changes in H298 before and after changing it (which is "good" but strange). I'll check it a bit more.

alongd:
An interesting observation:

From https://cccbdb.nist.gov/vibscalejust.asp we have the following frequencies scaling factors:
```python
'ccsd(t)/aug-cc-pvdz': 0.963,
'ccsd(t)/aug-cc-pvtz': 0.970,
'ccsd(t)/aug-cc-pvqz': 0.975,
```

But from https://comp.chem.umn.edu/freqscale/version3b2.htm we have:
```python
'ccsd(t)/aug-cc-pvtz': 1.001,
```

In this PR I mixed both, with was probably not a great idea, but the data shows significant inconsistencies... Perhaps eventually we'd like to just refit all the scaling factor after implementing the Truhlar algorithm in ARC (tag @dranasinghe).

dranasinghe:
I looked at CCCBDB, they use frequency scaling instead of ZPE as Truhlar
did. I have my doubts regarding the accuracy of cccbdb results; I find it
strange that ccsd(t)/tz ,qz results are off by ~.0.96-097. I expect them to
be much closer to 0.99. further, if you take a look at the section  "Poor
results"  there are lots of bad frequency calculations.
I agree with the idea that we should refit the frequency scaling factors
with Truhlar's algorithm. Furthermore, we should use ZPE to fit frequency
scaling, if you use frequencies to fit scaling factors we need to worry
about the symmetry of the vibrational mode; calculated vibrational modes
might have different symmetry labels than the experiment. We can avoid all
of it by fitting to ZPE. Question is should we use 15 molecules Truhlar
used or expand the test set? Truhlar uses two test sets one with 9
molecules and other with 15 looks like both give similar results. So I
think we will be fine if you just use the same test set.
I will implement Truhlar's method in ARC this weekend.
Duminda

alongd:
@cgrambow , could you take a look?

arkane/gaussian.py
```diff
@@ -282,48 +280,47 @@ def loadEnergy(self, frequencyScaleFactor=1.):
...
 
                 elif 'E(ZPE)' in line:
-                    scaledZPE = float(line.split()[1]) * constants.E_h * constants.Na
+                    scaled_zpe = float(line.split()[1]) * constants.E_h * constants.Na
                 elif '\\ZeroPoint=' in line:
                     line = line.strip() + f.readline().strip()
                     start = line.find('\\ZeroPoint=') + 11
                     end = line.find('\\', start)
-                    scaledZPE = float(line[start:end]) * constants.E_h * constants.Na * frequencyScaleFactor
+                    scaled_zpe = float(line[start:end]) * constants.E_h * constants.Na * frequencyScaleFactor
```
cgrambow: You're passing your newly defined `zpe_scale_factor` to this method in `statmech.py` but here we're removing the scaled ZPE from the composite E0. Are we sure that should be the ZPE scale factor and not the frequency scale factor?

Or maybe the more important question is, why are we actually doing this? As far as I can tell, `\ZeroPoint` gets printed in the archive string at the end of Gaussian jobs, but if it's a composite method, the last occurrence of ZPE will be from the `E(ZPE)` string instead. Therefore, I am wondering if the ZPE as defined on line 288 ever gets used. Do you know? If it does get used, then why are we multiplying it by the scale factor? Shouldn't it be the same as the statement on line 283?

arkane/statmech.py
```diff
@@ -288,29 +288,29 @@ def load(self, pdep=False):
...
+            energy = Quantity(energy)
+            e_0 = energy.value_si  # in J/mol
         elif isinstance(energy, tuple) and len(energy) == 3:
-            if energy[2] == 'E0':
-                energy_temp = Quantity(energy[:2])
-                E0 = energy_temp.value_si / constants.E_h / constants.Na  # convert J/mol to Hartree
-            elif energy[2] == 'E0-ZPE':
-                energy_temp = Quantity(energy[:2])
-                E0_withZPE = energy_temp.value_si  # in J/mol
+            if energy[2].lower() in ['e_elect', 'e_electronic', 'electronic_energy']:
```
cgrambow: Can we try to make the naming consistent so that there aren't three different variable names that all correspond to the same thing?

arkane/statmech.py
```diff
@@ -401,31 +401,30 @@ def load(self, pdep=False):
...
-            # The E0 that is read from the log file is without the ZPE and corresponds to E_elec
-            if E0 is None:
-                E0 = energyLog.loadEnergy(zpe_scale_factor)  # in J/mol
+        if e0 is None:
+            if e_electronic is None:
+                # The energy read from the log file is without the ZPE
+                e_electronic = energyLog.loadEnergy(zpe_scale_factor)  # in J/mol
             else:
-                E0 = E0 * constants.E_h * constants.Na  # Hartree/particle to J/mol
+                e_electronic = e_electronic * constants.E_h * constants.Na  # convert Hartree/particle into J/mol
```
cgrambow: Can shorten to `e_electronic *= constants.E_h * constants.Na`

arkane/statmech.py
alongd: Sure

arkane/statmech.py
alongd: done

arkane/gaussian.py
cgrambow: Thanks for the explanation. This clears up my confusion.

arkane/statmech.py
```diff
@@ -1220,79 +1220,73 @@ def projectRotors(conformer, F, rotors, linear, is_ts, label):
...
+                 'cbs-qb3-paraskevas': 0.99 * 1.014,  # [5], the 0.99 value is the ZPE scale factor
+                 'ccsd-f12/cc-pvdz-f12': 0.947,  # [2], taken as 'CCSD/cc-pVDZ'
+                 'ccsd(t)/cc-pvdz': 0.979,  # [2]
+                 'ccsd(t)/cc-pvtz': 0.975,  # [2]
+                 'ccsd(t)/cc-pvqz': 0.970,  # [2]
+                 'ccsd(t)/aug-cc-pvdz': 0.963,  # [2]
+                 'ccsd(t)/aug-cc-pvtz': 1.001,  # [3]
+                 'ccsd(t)/aug-cc-pvqz': 0.975,  # [2]
+                 'ccsd(t)/cc-pv(t+d)z': 0.965,  # [2]
+                 'ccsd(t)-f12/cc-pvdz-f12': 0.979,  # [2], taken as 'CCSD(T)/cc-pVDZ'
```
cgrambow: You can update this line and the remaining lines using the newest version of [Truhlar's database](https://comp.chem.umn.edu/freqscale/190107_Database_of_Freq_Scale_Factors_v4.pdf). It has data for CCSD(T)-F12/cc-pVDZ-F12, CCSD(T)-F12/cc-pVTZ-F12 (already have this), and CCSD(T)-F12/cc-pVQZ-F12 (two different versions but they're super close to each other). The CCSD(T)-F12/aug-cc-pVnZ scale factors can be based on the CCSD(T)-F12/cc-pVnZ-F12 factors because the cc-pVnZ-F12 and aug-cc-pVnZ basis sets are very similar to each other.

alongd:
Thanks! Let me know when to squash

alongd:
Squashed. Should be good to go!

alongd:
@cgrambow, rebase again. Should be good to go

PR #1620, opened 2019-06-08
Calculate the translational mode if doesn't exist

alongd:
### Motivation or Problem
Arkane determines poor entropies for atoms. The reason is that it appends a translational partition function to modes only if it finds, e.g., 'Molecular mass:' in a Gaussian input file, but that line appears in a freq calculation, not in an sp calculation log file (atoms only have the latter, which is passed to Arkane as the frequency log file as well).
As a result, we get great H298's, but poor S298's:
image
image

### Description of Changes
Detect if the translational mode was not appended to the conformer modes, and if so determine it from the mass and append it.
This results in improved S298:
image
image
Much, much better, but could still be improved for O and S (I don't know what the source for the deviation is).

### DIFF
--- a/arkane/statmech.py
+++ b/arkane/statmech.py
@@ -365,12 +365,19 @@ class StatMechJob(object):
                                                                     spinMultiplicity=spinMultiplicity,
                                                                     opticalIsomers=opticalIsomers,
                                                                     label=self.species.label)
+        translational_mode_exists = False
         for mode in conformer.modes:
             if isinstance(mode, (LinearRotor, NonlinearRotor)):
                 self.supporting_info.append(mode)
                 break
+            if isinstance(mode, (Translation, IdealGasTranslation)):
+                translational_mode_exists = True
         if unscaled_frequencies:
             self.supporting_info.append(unscaled_frequencies)
+
+        if not translational_mode_exists:
+            # Sometimes the translational mode is not appended to modes for monoatomic species
+            conformer.modes.append(IdealGasTranslation(mass=self.species.molecularWeight))
 
         if conformer.spinMultiplicity == 0:
             raise ValueError("Could not read spin multiplicity from log file {0},\n"


alongd:
@cgrambow, can I pick your brain for why we still see discrepancies for O and S's entropy? Any lead?

cgrambow:
I haven't looked at your code, but I went ahead and calculated HCSON by hand (see attached file). My results are:

| Atom | S (J/mol/K) |
|------|-------------|
| H    | 114.72      |
| C    | 148.99      |
| S    | 161.24      |
| O    | 152.57      |
| N    | 153.30      |

That's the same as ARC/Arkane gets, it seems (based on your graph), so I think ARC is calculating the partition functions correctly.

That calculation assumes that excited states are inaccessible. The full electronic partition function would be
elec_part
but the assumptions we make simplify it to
elec_simpl
My guess would be that excited states contribute significantly to the entropy for oxygen and sulfur. I checked oxygen on the CCCBDB experimental data page [(link)](https://cccbdb.nist.gov/sanyt4.asp), which shows 143.43 J/mol/K for the translational component of the entropy, which is exactly what I calculated in my spreadsheet. However, the electronic contribution is 17.63 J/mol/K, whereas I only calculated 9.13 J/mol/K. If you look at the excited state energies, the first and second excited state of the oxygen atom are extremely close to the ground state (only 158.5 cm-1 and 226.5 cm-1 above the ground state), which explains the difference.

In conclusion, there is no error in Arkane/ARC and we can only do better if we modify the electronic partition function to include excited state energies.

As an aside: @mjohnson541, The `primaryThermoLibrary` has the incorrect entropy for triplet oxygen atom (36.43 cal/mol/K, which corresponds to my calculation and not the true value, 38.49 or 38.46 cal/mol/K (depending on who you ask), that accounts for excited states). And less important, the group additivity value for triplet sulfur atom is wrong and also seems to coincide with the naively calculated value.

[atom_entropies.xlsx](https://github.com/ReactionMechanismGenerator/RMG-Py/files/3287895/atom_entropies.xlsx)

alongd:
Thanks so much for the detailed response, @cgrambow !!
@amarkpayne, now that the theory seems reasonable, could you take a look at the code?

amarkpayne:
Can do! I am taking a look at it now.

mliu49:
Should this be included in the v2.4.0 release?

alongd:
@mliu49, I think the code review should not take too long, and we do fix a bug here, so YES - we should include it in the release. But I wouldn't want the release to just wait for this PR, if it comes to this then it's also OK to release w/o this.

mliu49:
No problem, we can definitely get it in.

It seems that @amarkpayne already approved this. Is it ready to be merged?

alongd:
Yes - just waiting on RMG-Tests... (I restarted since it got stuck)

PR #1629, opened 2019-06-16
Reaction Mechanism Simulator YAML File Generation

mjohnson541:
Adds generation of .rms YAML files for use in ReactionMechanismSimulator.

### DIFF
--- a/rmgpy/yml.py
+++ b/rmgpy/yml.py
@@ -1,0 +1,212 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+###############################################################################
+#                                                                             #
+# RMG - Reaction Mechanism Generator                                          #
+#                                                                             #
+# Copyright (c) 2002-2019 Prof. William H. Green (whgreen@mit.edu),           #
+# Prof. Richard H. West (r.west@neu.edu) and the RMG Team (rmg_dev@mit.edu)   #
+#                                                                             #
+# Permission is hereby granted, free of charge, to any person obtaining a     #
+# copy of this software and associated documentation files (the 'Software'),  #
+# to deal in the Software without restriction, including without limitation   #
+# the rights to use, copy, modify, merge, publish, distribute, sublicense,    #
+# and/or sell copies of the Software, and to permit persons to whom the       #
+# Software is furnished to do so, subject to the following conditions:        #
+#                                                                             #
+# The above copyright notice and this permission notice shall be included in  #
+# all copies or substantial portions of the Software.                         #
+#                                                                             #
+# THE SOFTWARE IS PROVIDED 'AS IS', WITHOUT WARRANTY OF ANY KIND, EXPRESS OR  #
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,    #
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE #
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER      #
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING     #
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER         #
+# DEALINGS IN THE SOFTWARE.                                                   #
+#                                                                             #
+###############################################################################
+"""
+This file defines functions for outputting the RMG generated mechanism to a .rms yaml file, which can be read by the
+Reaction Mechanism Simulator (RMS)
+"""
+
+import os
+import yaml
+
+from rmgpy.chemkin import loadChemkinFile
+from rmgpy.species import Species
+from rmgpy.reaction import Reaction
+from rmgpy.thermo.nasa import NASAPolynomial, NASA
+from rmgpy.thermo.wilhoit import Wilhoit
+from rmgpy.kinetics.arrhenius import Arrhenius, PDepArrhenius, MultiArrhenius, MultiPDepArrhenius
+from rmgpy.kinetics.falloff import Troe, ThirdBody, Lindemann
+from rmgpy.kinetics.chebyshev import Chebyshev
+from rmgpy.data.solvation import SolventData
+from rmgpy.kinetics.surface import StickingCoefficient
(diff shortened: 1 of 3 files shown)


rmgpy/yml.py
```diff
@@ -0,0 +1,168 @@
...
+import os
+from rmgpy.chemkin import loadChemkinFile
+from rmgpy.species import Species
+from rmgpy.reaction import Reaction
+from rmgpy.thermo.nasa import NASAPolynomial, NASA
+from rmgpy.thermo.wilhoit import *
+from rmgpy.kinetics.arrhenius import *
+from rmgpy.kinetics.falloff import *
+from rmgpy.kinetics.chebyshev import *
+from rmgpy.data.solvation import *
```
mliu49: Can you replace all of the * imports with explicit ones?

rmgpy/yml.py
```diff
@@ -0,0 +1,168 @@
...
+from rmgpy.util import makeOutputSubdirectory
+
+def convertchemkin2yml(chemkinpath,spcdictpath=None,output="chem.rms"):
```
mliu49: Can you add spaces to after all commas in this file, in accordance with PEP-8?

rmgpy/yml.py
```diff
@@ -0,0 +1,168 @@
...
+    D["Phases"] = [dict()]
+    D["Phases"][0]["name"] = "phase"
+    D["Phases"][0]["Species"] = [obj2dict(x,spcs) for x in spcs]
+    D["Reactions"] = [obj2dict(x,spcs) for x in rxns]
+    if solventData:
+        D["Solvents"]= [obj2dict(solventData,spcs,label=solvent)]
+    return D
+
+def getradicals(spc):
+    if spc.molecule[0].toSMILES() == "[O][O]":
```
mliu49: What is the purpose of this function? Can you not use Molecule.getRadicalCount?

Also, why does oxygen need to be special cased? Could you add a comment explaining?

rmgpy/yml.py
```diff
@@ -0,0 +1,168 @@
...
+        dsub["A"] = float(obj.A)
+        dsub["B"] = float(obj.B)
+        dsub["C"] = float(obj.C)
+        dsub["D"] = float(obj.D)
+        dsub["E"] = float(obj.E)
+        D["mu"] = dsub
+    elif obj is None:
+        return None
+    else:
+        raise ValueError
```
mliu49: Add a message to this error?

rmgpy/yml.py
```diff
@@ -0,0 +1,168 @@
...
+        D["coefs"] = obj.coeffs.tolist()
+        D["Tmax"] = obj.Tmax.value_si
+        D["Tmin"] = obj.Tmin.value_si
+    elif isinstance(obj,Reaction):
+        D["reactants"] = [x.label for x in obj.reactants]
+        D["products"] = [x.label for x in obj.products]
+        D["kinetics"] = obj2dict(obj.kinetics,spcs)
+        D["type"] = "ElementaryReaction"
+        D["radicalchange"] = sum([getradicals(x) for x in obj.products]) -  sum([getradicals(x) for x in obj.reactants])
+    elif isinstance(obj,Arrhenius):
```
mliu49: Is support for the surface kinetics types necessary? I assume not now, since RMG doesn't have surface reactors?

rmgpy/yml.py
mjohnson541: I think this is faster since the multiplicities are known.  

This is typically used for analyzing radical production in RMS and it's very useful to treat O2 as a stable molecule when you're looking at bulk radical production/loss

rmgpy/yml.py
mjohnson541: done

rmgpy/yml.py
mjohnson541: done

rmgpy/yml.py
mjohnson541: done

rmgpy/yml.py
mjohnson541: I've added them in, RMS can read the catalysis output files properly yet, but it will be able to in the future.

rmgpy/yml.py
```diff
@@ -0,0 +1,175 @@
...
+        spcs,rxns = loadChemkinFile(chemkinPath)
+    writeyml(spcs,rxns,path=output)
+
+def writeyml(spcs, rxns, solvent=None, solventData=None, path="chem.yml"):
+    D = getMechDict(spcs,rxns,solvent=solvent,solventData=solventData)
+    yaml.dump(D,stream=file(path,'w'))
+
+def getMechDict(spcs, rxns, solvent='solvent', solventData=None):
+    D = dict()
+    D["Units"] = dict()
```
amarkpayne: Does ["Units"] ever get updated anywhere else in the code? Does the YAML format for RMS/Cantera assume that an empty dictionary for units defaults to something?

rmgpy/yml.py
```diff
@@ -0,0 +1,175 @@
...
+    D["Reactions"] = []
```
amarkpayne: This line is not needed right? Further down you have `D["Reactions"] = [obj2dict(x, spcs) for x in rxns]`, which will overwrite this line and also return an empty list if there are no reactions.

rmgpy/yml.py
```diff
@@ -0,0 +1,175 @@
...
+        return spc.molecule[0].multiplicity-1
+
+def obj2dict(obj, spcs, label="solvent"):
+    D = dict()
+    if isinstance(obj, Species):
+        D["name"] = obj.label
+        D["type"] = "Species"
+        D["smiles"] = obj.molecule[0].toSMILES()
+        D["thermo"] = obj2dict(obj.thermo, spcs)
+        if D["smiles"]  == "[O][O]":
```
amarkpayne: You defined the `getradicals` function above, so use this to get rid of the if-else statement here.

mjohnson541: done

rmgpy/yml.py
mjohnson541: done

rmgpy/yml.py
mjohnson541: It defaults to SI units for everything, unless units are specified for individual floats

mjohnson541:
Should be ready to merge now

rmgpy/yml.py
amarkpayne: That works then

amarkpayne:
I agree, feel free to merge this as soon as the tests pass (assuming they do).

PR #1630, opened 2019-06-17
Bugfix nasa as dict

goldmanm:
### Motivation or Problem
NASA polynomials would throw an error when calling the `as_dict` method when optional variables like `Tmin` are not set.

### Description of Changes
Added an if statement to the `as_dict` method to only add attributes which are set.

### Testing
RMG tests should be sufficient

### DIFF
--- a/arkane/inputTest.py
+++ b/arkane/inputTest.py
@@ -43,6 +43,8 @@ from rmgpy.statmech.translation import IdealGasTranslation
 from rmgpy.statmech.rotation import NonlinearRotor
 from rmgpy.kinetics.tunneling import Eckart
 from rmgpy.exceptions import InputError
+from rmgpy.thermo.nasa import NASAPolynomial, NASA
+from rmgpy.molecule import Molecule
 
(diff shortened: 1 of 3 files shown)


alongd:
Looking good! Would you like to remove Tmin / Tmax from one of the tests?

amarkpayne:
I am currently working on the yaml file standardization for the isodesmic reaction code, so this PR is very relevant to me. I have looked over the code, and it looks good to me. I like @alongd 's suggestion about adding a test for ensuring that an error is not thrown when Tmin/Tmax is not set, but I don't want this to hold up the PR from being merged. If you have the unit test for this already go ahead and add it. Otherwise, I believe this test will come up naturally in my isodesmic PR (or if not I'll add in a test). Either way @goldmanm feel free to update this branch and merge.

goldmanm:
Added two tests ensuring that `as_dict` saves set attributes and does not save unset attributes.

alongd:
Looking good!

Issue #1634, opened 2019-06-19
Invalid chemkin thermo block generated for an adsorbate with 5 elements


rwest:
### Bug Description
When running a surface chemistry job, this ended up in my `chem_annotated-surface.inp` chemkin file
```
! Gas phase thermo from Thermo library: thermo_DFT_CCSDTF12_BAC. Adsorption correction: + Thermo group additivity estimation: adsorptionPt((NRO)*)
! Binding energy corrected by LSR.
SX(8582)                                    G   100.000  5000.000  948.83      1&
X   1H   3C   1O   2N   1
 8.75901373E+00 7.66223340E-03-2.06362528E-06 4.02513189E-10-3.39045287E-14    2
-2.23373310E+04-3.74620293E+01 3.45624020E+00 3.20307213E-04 5.64909893E-05    3
-7.37256390E-08 2.81889462E-11-1.99942728E+04-5.11341806E+00                   4
```
which then obviously crashed the cantera conversion, leading to the error message
```
Warning: 
ERROR: Unable to parse '/Users/rwest/Dropbox (CoMoChEng)/Northeastern/Papers/2019-02 RMG-Cat Pt database franklin/rmgcat-pt-supporting-info/RMG_CH4catcombust_Pt_coupled/chemkin/chem-surface.inp' near line 195:

Error: Could not generate Cantera files for some reason.
Traceback (most recent call last):
  File "/Users/rwest/Code/Cat/RMG-Py/rmgpy/rmg/main.py", line 923, in execute
    surfaceFile=(os.path.join(self.outputDirectory, 'chemkin', 'chem-surface.inp')))
  File "/Users/rwest/Code/Cat/RMG-Py/rmgpy/rmg/main.py", line 1407, in generateCanteraFiles
    parser.convertMech(chemkinFile, transportFile=transportFile, outName=outName, quiet=True, permissive=True, **kwargs)
  File "/Users/rwest/anaconda/envs/rmg230/lib/python2.7/site-packages/cantera/ck2cti.py", line 2183, in convertMech
    self.loadChemkinFile(surfaceFile, surface=True)
  File "/Users/rwest/anaconda/envs/rmg230/lib/python2.7/site-packages/cantera/ck2cti.py", line 1736, in loadChemkinFile
    label, thermo, comp, note = self.readThermoEntry(thermo, TintDefault)
[...]

PR #1636, opened 2019-06-19
Fix bugs in Chemkin thermo entry element reading/writing

mliu49:
### Motivation or Problem
RMG Chemkin files break Cantera ck2cti for species with 5 or more elements. #1634 

### Description of Changes
- For species with 5 elements, prefer putting the 5th element in columns 74-78 of the first line
- For species with more than 5 elements, fix the line continuation syntax by switching to space delimited entries instead of continuing the `{0!s:<2}{1:>3d}` approach.
- Add read support for thermo entries using the line continuation syntax
- Add unit tests for reading and writing thermo entries

### Testing
Should test with ck2cti and perhaps Chemkin as well. I have not done so yet.

### DIFF
--- a/rmgpy/chemkinTest.py
+++ b/rmgpy/chemkinTest.py
@@ -33,7 +33,8 @@ import os
 
 import rmgpy
 from rmgpy.chemkin import get_species_identifier, load_chemkin_file, load_transport_file, mark_duplicate_reactions, \
-    read_kinetics_entry, read_reaction_comments, read_thermo_entry, save_chemkin_file, save_species_dictionary, save_transport_file
+    read_kinetics_entry, read_reaction_comments, read_thermo_entry, save_chemkin_file, save_species_dictionary, \
+    save_transport_file, write_thermo_entry
 from rmgpy.chemkin import _remove_line_breaks, _process_duplicate_reactions
 from rmgpy.data.kinetics import LibraryReaction
 from rmgpy.exceptions import ChemkinError
@@ -41,7 +42,7 @@ from rmgpy.kinetics.arrhenius import Arrhenius, MultiArrhenius
 from rmgpy.kinetics.chebyshev import Chebyshev
 from rmgpy.reaction import Reaction
 from rmgpy.species import Species
-from rmgpy.thermo import NASA
+from rmgpy.thermo import NASA, NASAPolynomial
 from rmgpy.transport import TransportData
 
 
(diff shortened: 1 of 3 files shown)


rwest:
Ugh. This then led me to dig further and ask is the columns 74-78 extension official and/or in Chemkin or Cantera? 

It seems that it is official, and in Chemkin, but it is not (yet) read by Cantera, although Cantera DOES read the extended line syntax (now that we have it correctly white-space separated)

This is part of Cantera's ck2cti
```python
    def readThermoEntry(self, lines, TintDefault):
        """
        Read a thermodynamics entry for one species in a Chemkin-format file
        (consisting of two 7-coefficient NASA polynomials). Returns the label of
        the species, the thermodynamics model as a :class:`MultiNASA` object, the
        elemental composition of the species, and the comment/note associated with
        the thermo entry.
        """
        identifier = lines[0][0:24].split()
        species = identifier[0].strip()

        if len(identifier) > 1:
            note = ''.join(identifier[1:]).strip()
        else:
            note = ''

        # Normal method for specifying the elemental composition
        composition = self.parseComposition(lines[0][24:44], 4, 5)

        # Chemkin-style extended elemental composition: additional lines
        # indicated by '&' continuation character on preceding lines. Element
        # names and abundances are separated by whitespace (not fixed width)
        if lines[0].rstrip().endswith('&'):
            complines = []
            for i in range(len(lines)-1):
                if lines[i].rstrip().endswith('&'):
                    complines.append(lines[i+1])
                else:
                    break
            lines = [lines[0]] + lines[i+1:]
            comp = ' '.join(line.rstrip('&\n') for line in complines).split()
            composition = {}
            for i in range(0, len(comp), 2):
                composition[comp[i].capitalize()] = int(comp[i+1])

        # Non-standard extended elemental composition data may be located beyond
        # column 80 on the first line of the thermo entry
        if len(lines[0]) > 80:
            elements = lines[0][80:]
            composition2 = self.parseComposition(elements, len(elements)//10, 10)
            composition.update(composition2)

        if not composition:
            raise InputParseError("Error parsing elemental composition for "
                                  "species '{0}'".format(species))
```

The columns 74-78 syntax is in the ckinterp.f source code comments
```
C     OPTIONAL THERMODYNAMIC DATA: (Subroutine CKTHRM)
C     (If this feature is not used, thermodynamic properties are
C     obtained from a CHEMKIN database.)  The format for this option
C     is the word 'THERMO' followed by any number of 4-line data sets:
C
C     Line 1: species name, optional comments, elemental composition,
C             phase, T(low), T(high), T(mid), additional elemental
C             composition, card number (col. 80);
C             format(A10,A14,4(A2,I3),A1,E10.0,E10.0,E8.0,(A2,I3),I1)
```
and the source code 
```
      ICOL = 20
      DO 60 I = 1, 5
         ICOL = ICOL + 5
         IF (I .EQ. 5) ICOL = 74
         ELEM  = LINE(1)(ICOL:ICOL+1)
```
and it's also documented in the 1980 Sandia report of the original CHEMKIN
https://prod-ng.sandia.gov/techlib-noauth/access-control.cgi/1980/808003.pdf (p60)

...so we should open an issue (or pull request) to support it in Cantera.

But until that's merged and released, doing this will make Chemkin files that Cantera can't read.
Should we just do the extended line syntax?

I think I'd rather inconvenience Chemkin-II users than Cantera 2.4.0 users.  I know several of the latter ;-)


Of course, we could also include our own slightly modified ck2cti in RMG. It's a stand-alone Python script that doesn't need all of Cantera.

rwest:
For reference: https://github.com/Cantera/cantera/issues/656

rwest:
I rebased @mliu49's PR onto the new master, converting them to Python 3 and RMG 3 in the process, and added another couple of commits.  Hope you don't mind that I force pushed.
I think this is now ready for a review.

mliu49:
I did some additional refactoring of the thermo writing method and squashed the commits together.

rmgpy/chemkinTest.py
```diff
@@ -428,6 +429,120 @@ def test_mark_duplicate_reactions(self):
...
+
+        # Chemkin entries for testing - note that the values are all the same
+        self.entry1 = """C2H6                    C   2H   6          G   300.000  3000.000  650.73      1
+-3.07954000E-01 2.45269000E-02-1.24130000E-05 3.07724000E-09-3.01467000E-13    2
+-1.06930000E+04 2.26280000E+01 4.03055000E+00-2.14171000E-03 4.90611000E-05    3
+-5.99027000E-08 2.38945000E-11-1.12576000E+04 3.56130000E+00                   4
+"""
+
+        self.entry2 = """CH3NO2X                                     G   300.000  3000.000  650.73      1&
+C 1 H 3 N 1 O 2 X 1
```
rwest: Probably these have to be 5-column wide blocks like `C   1 H   3 N   1 O   2 X   1` now?
Haven't actually run the code or the unit tests to check.

rmgpy/chemkinTest.py
```diff
@@ -428,6 +429,120 @@ def test_mark_duplicate_reactions(self):
...
+-3.07954000E-01 2.45269000E-02-1.24130000E-05 3.07724000E-09-3.01467000E-13    2
+-1.06930000E+04 2.26280000E+01 4.03055000E+00-2.14171000E-03 4.90611000E-05    3
+-5.99027000E-08 2.38945000E-11-1.12576000E+04 3.56130000E+00                   4
+"""
+
+        self.entry3 = """CH3NO2SX                                    G   300.000  3000.000  650.73      1&
+C 1 H 3 N 1 O 2 S 1 X 1
```
rwest: Same for these, I imagine

rmgpy/chemkinTest.py
mliu49: The 5-column wide blocks refers to the original method of placing them in the first line. The extended line is still space delimited. The tests currently pass.

rwest: Ah yes, I misread something. Looks good. Can I "approve"

PR #1639, opened 2019-06-21
Molecular weight growth

mjohnson541:
This adds features that greatly improve RMG's capacity to model molecular weight growth problems:  

-Mass flux based flux algorithm
   -Includes a mass index for adjusting the bias toward weight (molar flux * MW**massIndex) = "flux"

-Average MW termination criterion
   -Most of the old criteria other than time are difficult to use because in many polymerization cases          
             RMG doesn't generate important reactions unless you simulate out much farther than 
             conversion and rate ratio terminations allow you to
   -Time has to be set out very far to make this work, which make it ineffective after initial key    
             reactions have been generated
   -Average MW along with time (as an initial stop) provides good control over how far out the 
             model should be simulated

### DIFF
--- a/rmgpy/solver/base.pxd
+++ b/rmgpy/solver/base.pxd
@@ -63,6 +63,9 @@ cdef class ReactionSystem(DASx):
     cdef public np.ndarray Keq  # equilibrium constants
     cdef public np.ndarray network_leak_coefficients
     cdef public np.ndarray jacobian_matrix
+    cdef public np.ndarray mw
+    cdef public np.ndarray mww #molecular weight based weights of species
+    cdef public np.ndarray mww_network #molecular weight based weights of pdep networks
 
     cdef public np.ndarray core_species_concentrations
     
--- a/rmgpy/solver/base.pyx
+++ b/rmgpy/solver/base.pyx
@@ -1009,6 +1009,7 @@ cdef class ReactionSystem(DASx):
                                  ''.format(len(self.surface_species_indices), len(self.surface_reaction_indices)))
 
             if filter_reactions:
+                mww = self.mww
                 # Calculate thresholds for reactions
                 (unimolecular_threshold_rate_constant,
                  bimolecular_threshold_rate_constant,
@@ -1019,19 +1020,20 @@ cdef class ReactionSystem(DASx):
                 for i in range(num_core_species):
                     if not unimolecular_threshold[i]:
                         # Check if core species concentration has gone above threshold for unimolecular reaction
-                        if core_species_concentrations[i] > unimolecular_threshold_val:
+                        if (flux_basis == 'mass' and core_species_concentrations[i]*mww[i] > unimolecular_threshold_val) or core_species_concentrations[i] > unimolecular_threshold_val:
                             unimolecular_threshold[i] = True
                 for i in range(num_core_species):
                     for j in range(i, num_core_species):
                         if not bimolecular_threshold[i, j]:
-                            if core_species_concentrations[i] * core_species_concentrations[j] > bimolecular_threshold_val:
+                            if (flux_basis == 'mass' and core_species_concentrations[i]*core_species_concentrations[j]*(mww[i]+mww[j]) > bimolecular_threshold_val) or core_species_concentrations[i] * core_species_concentrations[j] > bimolecular_threshold_val:
                                 bimolecular_threshold[i, j] = True
                 if self.trimolecular:
                     for i in range(num_core_species):
                         for j in range(i, num_core_species):
                             for k in range(j, num_core_species):
                                 if not trimolecular_threshold[i, j, k]:
-                                    if (core_species_concentrations[i] *
+                                    if (flux_basis == 'mass' and (core_species_concentrations[i]*core_species_concentrations[j]*core_species_concentrations[k]*
+                                        (mww[i]+mww[j]+mww[k])> trimolecular_threshold_val)) or (core_species_concentrations[i] *
                                             core_species_concentrations[j] *
                                             core_species_concentrations[k]
                                             > trimolecular_threshold_val):
--- a/rmgpy/solver/base.pyx
+++ b/rmgpy/solver/base.pyx
@@ -1258,7 +1258,21 @@ cdef class ReactionSystem(DASx):
                         logging.info('At time {0:10.4e} s, reached target termination RateRatio: '
                                      '{1}'.format(self.t,char_rate/max_char_rate))
                         self.log_conversions(species_index, y0)
-
+                        
+                elif isinstance(term, TerminationNumberAvgMW):
+                    avg_MW = np.dot(self.y[:num_core_species],self.mw[:num_core_species])*constants.Na*1000.0/self.y[:num_core_species].sum()
+                    if avg_MW > term.avg_MW:
+                        terminated = True
+                        logging.info('At time {0:10.4e} s, reached target termination Number Averaged MW: {1}'.format(self.t,avg_MW))
+                        self.logConversions(species_index, y0)
+                        
+                elif isinstance(term, TerminationWeightAvgMW):
+                    avg_MW = np.dot(self.y[:num_core_species],self.mw[:num_core_species]**2)*constants.Na*1000.0/self.y[:num_core_species].dot(self.mw[:num_core_species])
+                    if avg_MW > term.avg_MW:
+                        terminated = True
+                        logging.info('At time {0:10.4e} s, reached target termination Weight Averaged MW: {1}'.format(self.t,avg_MW))
+                        self.logConversions(species_index, y0)
+                        
             # Increment destination step time if necessary
             if self.t >= 0.9999 * step_time:
                 step_time *= 10.0
@@ -1455,3 +1469,21 @@ class TerminationRateRatio:
 
     def __init__(self, ratio=0.01):
         self.ratio = ratio
+        
+class TerminationNumberAvgMW:
+    """
+    Represent an number averaged MW in amu at which a simulation
+    should be terminated. This class has one attribute the average
+    molecular weight.  Useful for molecular growth situations.
+    """
+    def __init__(self, avg_MW=50.0):
+        self.avg_MW = avg_MW
+
+class TerminationWeightAvgMW:
+    """
+    Represent an weight averaged MW in amu at which a simulation
+    should be terminated. This class has one attribute the average
+    molecular weight.  Useful for molecular growth situations.
+    """
+    def __init__(self, avg_MW=50.0):
+        self.avg_MW = avg_MW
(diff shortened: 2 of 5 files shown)


rmgpy/solver/base.pyx
```diff
@@ -594,6 +596,7 @@ cdef class ReactionSystem(DASx):
         cdef int numCoreSpecies, numEdgeSpecies, numPdepNetworks, numCoreReactions
         cdef double stepTime, charRate, maxSpeciesRate, maxNetworkRate, maxEdgeReactionAccum, stdan
         cdef numpy.ndarray[numpy.float64_t, ndim=1] y0 #: Vector containing the number of moles of each species
+        cdef numpy.ndarray[numpy.float64_t, ndim=1] corerates,edgerates,leakrates,MW
```
xiaoruiDong: Please change the nomenclature. Suggestions: coreWeightRates, edgeWeightRates, MWNetworkLeakRates, MW

rmgpy/solver/base.pxd
```diff
@@ -65,8 +65,8 @@ cdef class ReactionSystem(DASx):
     cdef public numpy.ndarray Keq # equilibrium constants
     cdef public numpy.ndarray networkLeakCoefficients
     cdef public numpy.ndarray jacobianMatrix
-    cdef public numpy.ndarray MW #molecular weights of species
-    cdef public numpy.ndarray MWnetwork #molecular weights of pdep networks
+    cdef public numpy.ndarray MWW #molecular weight based weights of species
```
xiaoruiDong: Change nomenclatures to MWWeight and MWWeightNetwork to increase readability?

rmgpy/solver/base.pyx
```diff
@@ -705,8 +707,9 @@ cdef class ReactionSystem(DASx):
         bimolecularThreshold = self.bimolecularThreshold
         trimolecularThreshold = self.trimolecularThreshold
         
-        self.MW = numpy.array([spc.molecularWeight.value_si for spc in coreSpecies+edgeSpecies])
-        self.MWnetwork = numpy.array([sum([sum([x.molecularWeight.value_si for x in r.species]) for r in nwk.reactants]) for nwk in pdepNetworks])
+
```
xiaoruiDong: Remove this extra line and can you add another comment to indicate the content of the block?

rmgpy/solver/base.pyx
```diff
@@ -988,25 +1006,40 @@ cdef class ReactionSystem(DASx):
                 for i in xrange(numCoreSpecies):
                     if not unimolecularThreshold[i]:
                         # Check if core species concentration has gone above threshold for unimolecular reaction
-                        if coreSpeciesConcentrations[i] > unimolecularThresholdVal:
-                            unimolecularThreshold[i] = True
+                        if fluxBasis == 'mass':
+                            if coreSpeciesConcentrations[i]*MWW[i] > unimolecularThresholdVal:
```
mliu49: I think it would look cleaner and reduce code duplication to define a `fluxFactor` which is 1 when `fluxBasis='mole'`.
```
fluxFactor = self.MWW[i] if fluxBasis == 'mass' else 1.0
if coreSpeciesConcentrations[i] * fluxFactor > unimolecularThresholdVal:
    unimolecularThreshold[i] = True
```

And similar setup for bi and trimolecular.

rmgpy/solver/base.pyx
```diff
@@ -978,6 +995,7 @@ cdef class ReactionSystem(DASx):
                     logging.info('Surface now has {0} Species and {1} Reactions'.format(len(self.surfaceSpeciesIndices),len(self.surfaceReactionIndices)))
                     
             if filterReactions:
+                MWW = self.MWW
```
mliu49: This seems unnecessary?

rmgpy/solver/base.pyx
```diff
@@ -1399,4 +1438,13 @@ class TerminationRateRatio:
     
     def __init__(self, ratio=0.01):
         self.ratio = ratio
-        
+
+class TerminationAveMW:
```
mliu49: I would suggest `TerminationAvgMW` instead.

rmgpy/solver/base.pyx
```diff
@@ -1399,4 +1438,13 @@ class TerminationRateRatio:
...
+    """
+    Represent an average MW at which a simulation
+    should be terminated. This class has one attribute the average
+    molecular weight.  Useful for molecular growth situations.
+    """
+
+    def __init__(self, aveMW=50.0):
+        self.aveMW = aveMW
```
mliu49: Also `avgMW` here (and everywhere else it's used).

Could you add a note in the docstring about the units?

rmgpy/solver/base.pyx
```diff
@@ -1212,6 +1245,12 @@ cdef class ReactionSystem(DASx):
                         terminated = True
                         logging.info('At time {0:10.4e} s, reached target termination RateRatio: {1}'.format(self.t,charRate/maxCharRate))
                         self.logConversions(speciesIndex, y0)
+                elif isinstance(term, TerminationAveMW):
+                    aveMW = numpy.dot(self.y[:numCoreSpecies],self.MW[:numCoreSpecies])*6.022e26/self.y[:numCoreSpecies].sum()
```
mliu49: Import Avogadro's constant from `constants`?

rmgpy/solver/base.pyx
```diff
@@ -700,7 +706,10 @@ cdef class ReactionSystem(DASx):
         unimolecularThreshold = self.unimolecularThreshold
         bimolecularThreshold = self.bimolecularThreshold
         trimolecularThreshold = self.trimolecularThreshold
-        
+
+        self.MW = numpy.array([spc.molecularWeight.value_si for spc in coreSpecies+edgeSpecies])
```
mliu49: Iterating over two lists by using `+` is not great because it creates a new list temporarily. Unfortunately I don't know what the optimal way is in cython. Since this happens once per simulation (?), we can probably leave it.

mjohnson541: I've switched to itertools.chain

rmgpy/solver/base.pyx
mjohnson541: done

rmgpy/solver/base.pyx
mjohnson541: done

rmgpy/solver/base.pyx
mjohnson541: done

rmgpy/solver/base.pyx
```diff
@@ -126,7 +126,10 @@ cdef class ReactionSystem(DASx):
         self.Keq = None # equilibrium constants
         self.networkLeakCoefficients = None
         self.jacobianMatrix = None
-        
+
+        self.MW = None
+        self.MWW = None
+        self.MWWnetwork = None
```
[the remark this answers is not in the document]
mjohnson541: done

rmgpy/solver/base.pyx
mjohnson541: I've tried to clean it up and remove most of the if statements, the fluxFactor approach just seems wasteful in the loop to me, although it probably doesn't slow us down that much since I don't think the filter loop is limiting here.

examples/rmg/polybutadiene/input.py
```diff
@@ -0,0 +1,57 @@
...
+)
+
+model(
+    toleranceMoveToCore=1.0,
+    toleranceInterruptSimulation=1.0,
+    maxNumObjsPerIter=1,
+    terminateAtMaxObjects=True,
+    filterReactions=True,
+    fluxBasis='mass',
+    massIndex=2.0,
```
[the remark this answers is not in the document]
mjohnson541: done

mjohnson541:
I've made all of the changes, I'd prefer to keep it out of the formal documentation until I can give better recommendations on what the settings should be for these sorts of runs.

mjohnson541:
@xiaoruiDong The first should be fixed now.  The second thing was just a find-replace mistake that was already fixed, MWave should always be calculated without the massIndex.

rmgpy/solver/base.pyx
```diff
@@ -1229,6 +1229,12 @@ cdef class ReactionSystem(DASx):
                         logging.info('At time {0:10.4e} s, reached target termination RateRatio: {1}'.format(self.t,charRate/maxCharRate))
                         self.logConversions(speciesIndex, y0)
                         
+                elif isinstance(term, TerminationAvgMW):
+                    avgMW = numpy.dot(self.y[:numCoreSpecies],self.mw[:numCoreSpecies])*constants.Na*1000.0/self.y[:numCoreSpecies].sum()
```
xiaoruiDong: What I mean is here, you are just using number average termination criteria. Do you think about giving users options to choose from number average molecular weight sum(y*M)/sum(y) and weight average molecular weight sum(y*M^2)/sum(y*M). Or even add a new termination criterion like the polydispersity index. I've heard they are important parameters but not quite clear about how people use them. I am totally fine if you don't want to add those for now, but I am happy to hear what do you think?

mjohnson541: So in general there are two major considerations for termination criteria:  
1) It needs to trigger at useful times
2) A user needs to have some idea what to set it to, either it needs to have a physical meaning to them or they need to be able to use a default setting

In this case I chose number averaged molecular weight because it is the simplest to calculate, of these.  There are definitely use cases in polymers for Weight averaged molecular weight based termination, so I can add that, although it's bit more complicated for a user to set.  

Polydispersity index however, likely would not make a good termination criterion because it wouldn't trigger at physically meaningful times unless you could guess what the limiting value and general trend were for the polymer before running RMG, if you guess slightly wrong your run could terminate way too early or way too late.

xiaoruiDong: Thanks for your great explanation. I am okay with the current PR. For weight average molecular weight, it may need more specific cases to decide: (1) whether useful, (2) whether used simultaneously with number average molar weight, etc. Anyway, thanks for the updates and efforts.

rmgpy/solver/base.pyx
xiaoruiDong: Considering the readability, @mliu49's suggestion makes sense. The current if statement seems a little bit too long for me.

xiaoruiDong:
> I've made all of the changes, I'd prefer to keep it out of the formal documentation until I can give better recommendations on what the settings should be for these sorts of runs.

@mliu49 Max, I think you need to make a decision on the documentation thing and further merge the branch.

rmgpy/solver/base.pyx
mliu49: I think the revised if statement is ok. Could you add a line break though?

mliu49:
I would still prefer this to be added to the documentation, even if you want to add a disclaimer that it is an experimental feature. Recommended settings can come later.

mjohnson541:
Yes, but by putting it in the documentation we're inherently endorsing it's use.  From a user's perspective we wouldn't add a "experimental" feature unless we intended people to use it.  I don't think we should endorse it until I've finished working out the recommended parameters.

mliu49:
I think that putting it on master is endorsing its use. If you are not ready to endorse it, then we should not merge this PR.

JacksonBurns:

PR #1641, opened 2019-06-24
Enable Restarting RMG Jobs from a Seed Mechanism

amarkpayne:
### Motivation or Problem
There are many instances where it is desirable to restart an RMG job from the last iteration instead of having to start the mechanism generation from the beginning. RMG currently has a restart file, but it has not been maintained and will be deprecated starting in the next release. Furthermore, these restart files are extremely slow.

The agreed upon solution to this problem is to restart RMG jobs from the seed mechanism files that RMG already outputs. These files allow for the complete reconstruction of the model core and edge at the time of the restart. The only missing piece of information missing is the filter tensors, which help RMG decide which species to react, which species have already reacted, and which species not to react because their expected rate is well below what would be included based on the tolerance. This PR saves these filter tensors and allows for them to be loaded so that RMG jobs can be restarted using this technique

### Description of Changes

- Reorganized when seed mechanisms are saved
- Save the filter tensors (and necessary mappings) to a `Filters` sub folder of the seed mechanism
- Add in a block to the RMG input file format for telling RMG to perform a restart job from a seed mechanism
- Automatically generate input files entitled `restart_from_seed.py` that can be run to restart the job
- Updated the documentation to describe how the user can use this restart feature
- Saving the seed mechanism each iteration is now the default behavior

### Testing
I have restarted the superminimal and minimal examples using this technique, and the same mechanism was obtained if the job was run to completion or if the job was restarted from a seed mechanism from an intermediate iteration seed mechanism.

I have not added anything in the way of unit testing yet, but I am happy to add tests for the crucial features of this PR.


rmgpy/rmg/input.py
```diff
@@ -693,6 +693,18 @@ def uncertainty(localAnalysis=False, globalAnalysis=False, uncorrelated=True, co
     }
 
 
+def restart_from_seed(path=None, relative_path=True):
```
mliu49: Could you make this an argument in the other options block instead of it's own block?

I don't think you need to ask about relative_path. You can check whether the path is absolute using `os.path.isabs`, and if not, assume (and say in the documentation) that it is relative to the input file.

amarkpayne: I updated the absolute path issue, thanks for the suggestion!

As for moving this to the other options block I feel that this is not really an option but rather a feature that completely alters the behavior of a normal RMG job. I don't want this to be hidden away in the input file where the user might forget to change this. It is also easier to automatically write restart files if this is its own block, though this can of course be overcome.

I am willing to discuss this further, and change it if you feel strongly about it, but I do feel like it is different from the other options in the block

examples/rmg/commented/input.py
```diff
@@ -1,7 +1,5 @@
 # Uncomment the lines below to restart from a seed mechanism
-# restartFromSeed(path='seed',  # Location of the seed mechanism folder (with Filters subfolder) to load for restarting
-#                 relative_path=True,  # If True (default), the location listed in path will be a relative path from the
-#                                      # input file
+# restartFromSeed(path='seed',  # Location of the seed mechanism (with `Filters` subfolder) to load for restarting
```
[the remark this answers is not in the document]
amarkpayne: Done. As for the documentation, I will add this but it will depend on how we deal with deprecation

rmgpy/rmg/main.py
```diff
@@ -1473,9 +1491,10 @@ def initializeReactionThresholdAndReactFlags(self):
         numCoreSpecies = len(self.reactionModel.core.species)
 
         # Number of families in RMG
+        rmg_families = list(self.database.kinetics.families.keys())
         num_families = len(self.database.kinetics.families)
 
-        if self.filterReactions:
+        if self.filterReactions or self.restart_seed_path:
```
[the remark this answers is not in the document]
mliu49: I think that would mean that previously unreacted things would never be reacted in the restarted simulation, since the flags would never be updated.

Not exactly sure that the optimal behavior is though.

rmgpy/rmg/main.py
```diff
@@ -1356,6 +1356,15 @@ def makeSeedMech(self,firstTime=False):
...
                 yaml.dump(data=family_map, stream=f)
 
+            # Generate a file for restarting from a seed mechanism if this is not a restart job
+            if firstTime and (not self.restart_seed_path):
+                with open(self.inputFile, 'r') as f:
+                    input_text = f.readlines()
+
+                with open(os.path.join(self.outputDirectory, 'restart_from_seed.py'), 'w') as f:
+                    f.write('restartFromSeed(\n    path=\'seed\'\n)\n\n')
+                    f.write(''.join(input_text))
```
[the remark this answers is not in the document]
mliu49: The main difference is not needing to create a variable. I feel like it's a bit cleaner, but it doesn't really matter. You can leave it if you prefer.

rmgpy/rmg/main.py
```diff
@@ -1484,6 +1503,56 @@ def initializeReactionThresholdAndReactFlags(self):
...
+                    raise RuntimeError('Not all of the species specified in the restart mechanism have been included'
+                                       'in the core')
+
+                reordered_core_species = [core_species_map[i] for i in range(len(core_species_map))] + remaining_species
+                self.reactionModel.core.species = reordered_core_species
+
+                # Fill in the newly initialized filter tensors by matching based on family order
+                for i, rmg_family in rmg_family_map.iteritems():
+                    for j, restart_family in restart_family_map.iteritems():
+                        if rmg_family == restart_family:
```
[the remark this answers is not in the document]
mjohnson541: I think that all makes sense.  I can see real cases where Scenario 1 might happen, niche certainly though.  Dr. West has talked about changing families in the middle of a run, but that would probably be better done by an extension of the staged model generation tool.

rmgpy/rmg/main.py
```diff
@@ -1247,7 +1247,7 @@ def makeSeedMech(self,firstTime=False):
                     q += 1
                 self.name = name + str(q)
         
-        seedDir = os.path.join(self.outputDirectory,'seed')
+        seedDir = os.path.join(self.outputDirectory, 'seed')
```
mjohnson541: This change should be in a separate commit

rmgpy/rmg/main.py
```diff
@@ -1356,6 +1356,15 @@ def makeSeedMech(self,firstTime=False):
             with open(os.path.join(filter_dir, 'family_map.yml'), 'w') as f:
                 yaml.dump(data=family_map, stream=f)
 
+            # Generate a file for restarting from a seed mechanism if this is not a restart job
```
mjohnson541: I'm not sure I understand, we don't want to generate a file to restart if the job we're running was ever restarted?

documentation/source/users/rmg/input.rst
```diff
@@ -942,3 +942,29 @@ For example ::
...
+Restarting from a Seed Mechanism
+=================================
+The preferred method for restarting an RMG-Py job is to restart the job from a seed mechanism. There are many scenarios
+when the user might want to do this, including continuing on a job that ran out of time or crashed as the result of a
+now fixed bug. To restart from a seed mechanism, the block below must be added on to the input file.  ::
+
+    restartFromSeed(path='seed')
+
+The ``path`` flag is the path to the seed mechanism folder, which contains three subfolders that must be titled as
+follows: Filters, Seed, Seed_Edge. The path can be a relative path from the input.py file or an absolute path on disk
```
mjohnson541: So I appreciate the simplicity of simply passing the path of the directory in a job you ran, however, in general we should have an interface (at least in addition to the current one) that isn't dependent on the directory structure beneath the input path.  The first approach works great until someone manipulates the files or moves the files around and then the path only works if they rearrange them in the exact right directory structure.  Taking each file as a separate input is much more typical of professional python codes.

rmgpy/rmg/main.py
```diff
@@ -1463,21 +1481,126 @@ def generateCanteraFiles(self, chemkinFile, **kwargs):
...
-                self.trimolecularReact = np.zeros((numCoreSpecies, numCoreSpecies, numCoreSpecies), bool)
-                self.trimolecularThreshold = np.zeros((numCoreSpecies, numCoreSpecies, numCoreSpecies), bool)
-        else:
-            # By default, react everything
-            self.unimolecularReact = np.ones((numCoreSpecies),bool)
-            self.bimolecularReact = np.ones((numCoreSpecies, numCoreSpecies),bool)
-            if self.trimolecular:
-                self.trimolecularReact = np.ones((numCoreSpecies, numCoreSpecies, numCoreSpecies),bool)
-            # No need to initialize reaction threshold arrays in this case
+                self.trimolecularThreshold = np.zeros((numCoreSpecies, numCoreSpecies, numCoreSpecies),bool)
```
mliu49: Spaces after commas.

rmgpy/rmg/main.py
```diff
@@ -1463,21 +1481,126 @@ def generateCanteraFiles(self, chemkinFile, **kwargs):
...
+
+                    except KeyError:  # If we can't find the filters then this is because filtering was not used
+                        logging.warning('No filters were found in file {0}. This is to be expected if the restart '
+                                        'files specified came from an RMG job that did not use filtering. Therefore, '
+                                        'RMG will assume that all of the core species from the restart seed have '
+                                        'already been reacted and will not react them again. Additional species that '
+                                        'were added to the input file but are not in the restart core seed WILL be '
+                                        'reacted with the rest of the core though. If this is not the intended '
+                                        'behavior please stop this job and correct the missing '
+                                        'filters.'.format(self.filtersPath))
```
mliu49: `came from an RMG job that did not use filtering` -> `are from an RMG job without reaction filtering`
`species that were added to the input file but are not` -> `species added to the input file but not`
`core though.` -> `core.`
`restart seed` vs `restart core seed` should probably be consistent?
I think the last sentence is probably not needed. I don't know how many people read the log in real time.

rmgpy/rmg/main.py
```diff
@@ -1463,21 +1481,126 @@ def generateCanteraFiles(self, chemkinFile, **kwargs):
...
+
+                        filters_found = False
+
+                # Reorder the core species to match the indices of the restart filter tensors
+                core_species_map = {}
+
+                for index, spcs in restartSpeciesMap.items():
+                    for j, core_spcs in enumerate(self.reactionModel.core.species):
+                        if core_spcs.isIsomorphic(spcs):
+                            core_species_map[index] = core_spcs
```
mliu49: You could do `core_species_map[index] = self.reactionModel.core.species.pop(j)` instead of doing it after.

Since we break immediately after, it doesn't matter if we shorten the list.

Also, I still think that the restart species map can just be dumped as a list instead of a dictionary with indices as keys.

rmgpy/rmg/main.py
```diff
@@ -1463,21 +1481,126 @@ def generateCanteraFiles(self, chemkinFile, **kwargs):
...
+                            break
+                    else:
+                        raise RuntimeError('Species {0} was defined in the restart file, but was not included in the'
+                                           'core.'.format(spcs))
+                    self.reactionModel.core.species.pop(j)
+
+                remaining_species = self.reactionModel.core.species
+
+                reordered_core_species = [core_species_map[i] for i in range(len(core_species_map))] + remaining_species
+                self.reactionModel.core.species = reordered_core_species
```
mliu49: I think the core species map also doesn't need to be a dictionary.
```
reordered_core_species = []
for spc in restart_species_list:  # if it were a list instead of a dict
    for j, core_spcs in enumerate(self.reactionModel.core.species):
        if core_spcs.isIsomorphic(spc):
            reordered_core_species.append(self.reactionModel.core.species.pop(j)
            break
    else:
        raise RuntimeError

self.reactionModel.core.species = reordered_core_species + self.reactionModel.core.species
```

rmgpy/rmg/main.py
```diff
@@ -1463,21 +1481,126 @@ def generateCanteraFiles(self, chemkinFile, **kwargs):
...
+                    if self.filterReactions:  # Filling in the filter thresholds will suffice
+                        # Fill in the newly initialized filter tensors
+                        self.unimolecularThreshold = unimolecularThreshold
+                        self.bimolecularThreshold = bimolecularThreshold
+                        if self.trimolecular:
+                            self.trimolecularThreshold = trimolecularThreshold
+
+                    else:  # We must set the react flags instead. If it was `True` in the threshold, it should not react
+                        self.unimolecularReact = np.logical_not(unimolecularThreshold)
+                        self.bimolecularReact = np.logical_not(bimolecularThreshold)
```
mliu49: Is `bimolecularThreshold` triangular or the full matrix? Is there any danger in inverting the entire thing if it is triangular?

documentation/source/users/rmg/input.rst
```diff
@@ -944,3 +944,41 @@ For example ::
...
+
+Alternatively, you may also specify the paths to each of the following files/directories: coreSeed (path to seed
+mechanism folder containing files named ``reactions.py`` and ```dictionary.txt``` that will go into the model core),
+edgeSeed (path to edge seed mechanism folder containing files named ``reactions.py`` and ```dictionary.txt``` that will
+go into the model edge), filters (an h5 binary file storing the uni-, bi-, and optionally tri-molecular thresholds
+generated by RMG from previous run (must be RMG version >= 3.0)), speciesMap (a YAML file specifying what species are
```
mliu49: Is the v2.5 cutoff going to be before this PR?

rmgpy/rmg/main.py
```diff
@@ -915,9 +912,13 @@ def execute(self, **kwargs):
...
                 continue
         
-        if not self.generateSeedEachIteration:
+        # Save the final seed mechanism
+        if self.generateSeedEachIteration:
+            self.makeSeedMech()
+        else:
             self.makeSeedMech(firstTime=True)
 
+
```
mliu49: Undo extra line?

examples/rmg/commented/input.py
```diff
@@ -1,3 +1,15 @@
...
+#
+# # Option 1: Specify the path to an RMG (version >= 3.0) generated seed mechanism folder, which contains all of the
+# # required files (core and edge seed, filters and mappings) in their default locations and names in the seed folder.
+# restartFromSeed(path='seed')  # Location of the seed mechanism (with `Filters` subfolder) to load for restarting
+#
+# # Option 2: Specify the paths of each of the required files individually.
+# restartFromSeed(coreSeed = 'seed/Seed'  # Path to core seed folder. Must contain `reactions.py` and `dictionary.txt`
+#                 edgeSeed = 'seed/Seed_edge'  # Path to edge seed folder containing `reactions.py` and `dictionary.txt`
+#                 filters='seed/Filters/filters.h5',
+#                 speciesMap='seed/Filters/species_map.yml')
```
mliu49: Either no spaces or consistent spaces around `=`

rmgpy/rmg/input.py
```diff
@@ -770,6 +770,43 @@ def uncertainty(localAnalysis=False, globalAnalysis=False, uncorrelated=True, co
...
+def restartFromSeed(path=None, coreSeed=None, edgeSeed=None, filters=None, speciesMap=None):
+    parentDir = os.path.dirname(rmg.inputFile)
+    rmg.restart = True
+
+    if path:
+        if any((coreSeed, edgeSeed, filters, speciesMap)):
+            raise InputError('For restarting an RMG job from a seed mechanism, either the path to the RMG generated '
+                             'seed mechanism should be given as `path`, or the path for each of the required files '
+                             'should be explicitly given, but not both. Please take one approach or the other. For '
+                             'further information see the RMG documentation on restarting from a seed mechanism.')
```
mliu49: Link or section?

amarkpayne: Done, thanks!

rmgpy/rmg/main.py
amarkpayne: Thanks for the catch! I forgot that I could change this since I was changing this line anyways

rmgpy/rmg/main.py
```diff
@@ -1463,21 +1481,126 @@ def generateCanteraFiles(self, chemkinFile, **kwargs):
...
+                with h5py.File(self.filtersPath, 'r') as f:
+                    try:
+                        unimolecularThreshold_restart = f.get('unimolecularThreshold').value
+                        bimolecularThreshold_restart = f.get('bimolecularThreshold').value
+                        if self.trimolecular:
+                            trimolecularThreshold_restart = f.get('trimolecularThreshold').value
+
+                        # Expand Thresholds to match number of species in the current model.
+                        unimolecularThreshold = np.zeros(numCoreSpecies, bool)
+                        unimolecularThreshold[:numRestartSpcs] = unimolecularThreshold_restart
```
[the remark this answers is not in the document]
amarkpayne: Good point, thanks!

rmgpy/rmg/main.py
amarkpayne: Thanks!

rmgpy/rmg/main.py
```diff
@@ -1463,21 +1481,126 @@ def generateCanteraFiles(self, chemkinFile, **kwargs):
...
+                                        'reacted with the rest of the core though. If this is not the intended '
+                                        'behavior please stop this job and correct the missing '
+                                        'filters.'.format(self.filtersPath))
+
+                        filters_found = False
+
+                # Reorder the core species to match the indices of the restart filter tensors
+                core_species_map = {}
+
+                for index, spcs in restartSpeciesMap.items():
```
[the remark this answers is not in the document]
amarkpayne: Done, thanks!

rmgpy/rmg/main.py
amarkpayne: It would be triangular because it is saved from RMG. In theory this could cause problems if we ever read from the lower triangle, but this is true regardless of the changes implemented in this PR.

rmgpy/rmg/main.py
amarkpayne: We don't want to generate a restart file if the _current job_ is a restart job because there already exists a file with the restart block defined in it, no need to write the exact same file again.

rmgpy/rmg/main.py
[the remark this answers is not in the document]
amarkpayne: Done

documentation/source/users/rmg/input.rst
amarkpayne: Fixed to allow both options as we discussed offline

documentation/source/users/rmg/input.rst
amarkpayne: Up to you. I'll update this based on whatever we decide. I'll wait to push until I hear back on this.

documentation/source/users/rmg/input.rst
```diff
@@ -944,9 +941,9 @@ For example ::
 
 Restarting from a Seed Mechanism
 =================================
-The preferred method for restarting an RMG-Py job is to restart the job from a seed mechanism. There are many scenarios
-when the user might want to do this, including continuing on a job that ran out of time or crashed as the result of a
-now fixed bug. To restart from a seed mechanism, the block below must be added on to the input file.  ::
+The preferred (and now only) method for restarting an RMG-Py job is to restart the job from a seed mechanism. There are
```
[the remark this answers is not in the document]
amarkpayne: Done

rmgpy/rmg/input.py
```diff
@@ -696,11 +696,15 @@ def pressureDependence(
 def options(name='Seed', generateSeedEachIteration=True, saveSeedToDatabase=False, units='si', saveRestartPeriod=None,
             generateOutputHTML=False, generatePlots=False, saveSimulationProfiles=False, verboseComments=False, 
             saveEdgeSpecies=False, keepIrreversible=False, trimolecularProductReversible=True, wallTime='00:00:00:00'):
+    if saveRestartPeriod:
+        logging.warning("`saveRestartPeriod` flag was set in the input file, but this feature has been removed. Please "
+                        "remove this line from the input file. This will throw an error after RMG-Py 3.1. For "
+                        "restarting an RMG job see the documentation for restarting from a seed mechanism")
```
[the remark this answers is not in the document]
amarkpayne: Done

rmgpy/rmg/main.py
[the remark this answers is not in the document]
amarkpayne: Implemented

rmgpy/rmg/main.py
amarkpayne: Change not included in this PR since I do not change this line. Thanks for the catch!

documentation/source/users/rmg/input.rst
amarkpayne: As @mjohnson541 suggested we can just say >2.4.1

rmgpy/rmg/main.py
amarkpayne: Good point, I'll do this

rmgpy/rmg/main.py
amarkpayne: Done

amarkpayne:
I have force pushed the requested changes and confirmed that the (updated, the species map changed from a dictionary to a list) minimal restart example works.

documentation/source/users/rmg/input.rst
```diff
@@ -944,3 +941,41 @@ For example ::
...
+mechanism folder containing files named ``reactions.py`` and ```dictionary.txt``` that will go into the model core),
+edgeSeed (path to edge seed mechanism folder containing files named ``reactions.py`` and ```dictionary.txt``` that will
+go into the model edge), filters (an h5 binary file storing the uni-, bi-, and optionally tri-molecular thresholds
+generated by RMG from previous run (must be RMG version > 2.4.1)), speciesMap (a YAML file specifying what species are
+at each index in the filters). ::
+
+    restartFromSeed(coreSeed = 'seed/seed'  # Path to core seed folder. Must contain `reactions.py` and `dictionary.txt`
+                   edgeSeed = 'seed/seed_edge'  # Path to edge seed folder containing `reactions.py` and `dictionary.txt`
+                   filters='seed/filters/filters.h5',
+                   speciesMap='seed/filters/species_map.yml')
```
mliu49: Spacing around `=` and indent an extra space

rmgpy/rmg/input.py
```diff
@@ -770,6 +775,44 @@ def uncertainty(localAnalysis=False, globalAnalysis=False, uncorrelated=True, co
...
+def restartFromSeed(path=None, coreSeed=None, edgeSeed=None, filters=None, speciesMap=None):
+    parentDir = os.path.dirname(rmg.inputFile)
+    rmg.restart = True
+
+    if path:
+        if any((coreSeed, edgeSeed, filters, speciesMap)):
+            raise InputError('For restarting an RMG job from a seed mechanism, either the path to the RMG generated '
+                             'seed mechanism should be given as `path`, or the path for each of the required files '
+                             'should be explicitly given, but not both. Please take one approach or the other. For '
+                             'further information see the RMG documentation on restarting from a seed mechanism at'
```
mliu49: Add space at the end

rmgpy/rmg/input.py
```diff
@@ -770,6 +775,44 @@ def uncertainty(localAnalysis=False, globalAnalysis=False, uncorrelated=True, co
...
+
+        # Try to find the paths for all of the required modules
+        rmg.coreSeedPath = os.path.join(path, 'seed')
+        rmg.edgeSeedPath = os.path.join(path, 'seed_edge')
+        rmg.filtersPath = os.path.join(path, 'filters', 'filters.h5')
+        rmg.speciesMapPath = os.path.join(path, 'filters', 'species_map.yml')
+
+        if any([not os.path.exists(file_path) for file_path in [rmg.coreSeedPath, rmg.edgeSeedPath,
+                                                                rmg.filtersPath, rmg.speciesMapPath]]
+               ):
```
mliu49: I think we should do this check after the if statement, in case the use doesn't provide all 4 arguments.

rmgpy/rmg/main.py
```diff
@@ -1417,21 +1468,121 @@ def generateCanteraFiles(self, chemkinFile, **kwargs):
...
+                                        'to the input file but not in the restart core seed WILL be reacted with the '
+                                        'rest of the core.'.format(self.filtersPath))
+
+                        filters_found = False
+
+                # Reorder the core species to match the indices of the restart filter tensors
+                reordered_core_species = []
+                for spc in restartSpeciesList:
+                    for j, oldCoreSpc in enumerate(self.reactionModel.core.species):
+                        if oldCoreSpc.isIsomorphic(spc):
```
mliu49: I think we should use the `strict=False` argument since we don't generate resonance structures for the restart species.

rmgpy/rmg/main.py
```diff
@@ -1417,21 +1468,121 @@ def generateCanteraFiles(self, chemkinFile, **kwargs):
...
+                            self.trimolecularThreshold = trimolecularThreshold
+
+                    else:  # We must set the react flags instead. If it was `True` in the threshold, it should not react
+                        self.unimolecularReact = np.logical_not(unimolecularThreshold)
+                        self.bimolecularReact = np.logical_not(bimolecularThreshold)
+                        if self.trimolecular:
+                            self.trimolecularReact = np.logical_not(trimolecularThreshold)
+
+                else:  # Assume that all species found in the restart core seed have already been reacted
+                    ind = len(restartSpeciesList)  # Index where the additional species begin
```
mliu49: Is this the same as `numRestartSpecies`?

rmgpy/rmg/main.py
```diff
@@ -431,14 +433,9 @@ def initialize(self, **kwargs):
         self.logHeader()
         
         try:
-            restart = kwargs['restart']
+            self.restart_seed_path = kwargs['restart']
```
mliu49: I think this doesn't do anything. It should set the `coreSeedPath`, `edgeSeedPath`, `filtersPath`, and `speciesMapPath` like you do when loading the input file.

Alternatively, if you don't think we should support both methods, you can remove this and set the `-r` flag to simply print a warning that the old restart method has been removed and the flag does not do anything.

rmgpy/rmg/main.py
amarkpayne: Yes, but it had been a while before I had used this, and `ind` here was shorter for this purpose, but did not make sense when I needed it earlier. What would you prefer?

documentation/source/users/rmg/input.rst
amarkpayne: Thanks!

mliu49:
@amarkpayne was correct, and the problem was in fact the extra indentation, which was messing up the family name parsing and causing the reaction to be loaded as a library reaction with `.library` (and `.family`) as `seedMechanism.name` ([here](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/04d4d42e718d3022b937bdd2b8fc94cc812d48db/rmgpy/rmg/model.py#L1419)). We missed the warning, which would have led us to the cause much faster.

I added a few commits to resolve this and some related issues:
- Switch to using the `longDesc` property of the entries instead of the private `_longDesc` attribute in `getLibraryReactions`. Not a bug, but the point of a private attribute is that it shouldn't be used.
- Improve parsing of reaction template and family name in `getLibraryReactions`, including stripping the family name. I discovered while debugging that the reaction template was completely wrong in some cases because the comment strings weren't exactly as expected.
- Remove leading spaces from kinetics comments in `getLibraryReactions`. I wasn't sure the best place to put this, but it seemed to make sense there. This prevents the gradual increase in indentation with each read/write cycle.

rmgpy/rmg/main.py
amarkpayne: Done

rmgpy/rmg/main.py
amarkpayne: Fixed, thanks for the catch!

amarkpayne:
I have also confirmed that with @mliu49 fixes enable unlimited successive restarts. I have also addressed the last remaining comment that I need to fixup, and I have squashed all of the fixups and force pushed. This should be ready for final approval/merging

mjohnson541:
Took a look now, looks fine to me!

PR #1643, opened 2019-06-26
Recursive as dictionary

amarkpayne:
### Motivation or Problem
The methods `as_dict` and `make_object` allow for any object that inherits from RMGObject to be easily saved to a yaml file and reconstructed from a yaml file, respectively. However, as currently written is RMG there is a limited recursion depth to how many objects, dictionaries, and lists can be nested inside of each other.

For example, consider an object that has an attribute that is a dictionary which itself contains another object. The desired affect of `as_dict` would be for the object nested inside of this dictionary to be expanded to a dictionary (i.e. for `as_dict` to be called on this object as well). However, the current implementation will return for this attribute a dictionary that contains the object itself rather than the expanded object.

Furthermore, many objects have to rewrite their `as_dict` and `make_objects` to function properly, even if they inherit from RMGObject. Ideally, newly created objects that inherit from RMGObject should have working `as_dict` and `make_object` methods without needing to overwrite these methods (at least more times than not).

### Description of Changes
Inside the file where we define the RMGObject I have created method for recursively calling `as_dict` for all objects inside of a dictionary or list. I have also created a method for recursively calling `make_object` for all objects inside of a dictionary. Becasue these functions are recursive by calling themselves, the recursion limit for nested objects and iterables should be whatever the recursion stack limit of python is (very large). 

While these functions can be called as standalone functions (potentially useful for scripting), I have used these functions to help re-write the `as_dict` and `make_object` methods of the RMGObject class.

With this, I have also cleaned-up the affected code, and made it so that the `as_dict` and `make_object` methods are only re-written when absolutely necessary (for example, see statmech rotations and torsions), which thanks to the new functions require very little code even when necessary. Note that this is usually necessary whenever passing all attributes to the __init__ method causes issues.

### Testing
I have ran the unit tests, and even successfully recreated an ArkaneSpecies object from a YAML file, but it is likely that there are affected classes that weren't properly unit tested. I am specifically concerned about the Wilhoit class, and the NASA class can probably be cleaned up further.

### Reviewer Note
Because we have a few pull requests open or about to be opened that define new objects that will be saved to a YAML file, these PRs will require these changes, and thus this PR needs to be merged in first. For example, this affects #1577, #1641, and #1561

### DIFF
--- a/arkane/common.py
+++ b/arkane/common.py
@@ -52,7 +52,7 @@ from rmgpy.statmech.translation import IdealGasTranslation
 from rmgpy.statmech.vibration import HarmonicOscillator
 from rmgpy.pdep.collision import SingleExponentialDown
 from rmgpy.transport import TransportData
-from rmgpy.thermo import NASA, Wilhoit, ThermoData
+from rmgpy.thermo import NASA, Wilhoit, ThermoData, NASAPolynomial
 from rmgpy.species import Species, TransitionState
 import rmgpy.constants as constants
 
@@ -231,21 +231,30 @@ class ArkaneSpecies(RMGObject):
             yaml.dump(data=self.as_dict(), stream=f)
         logging.debug('Dumping species {0} data as {1}'.format(self.label, filename))
 
-    def load_yaml(self, path, species, pdep=False):
+    def load_yaml(self, path, label=None, pdep=False):
         """
         Load the all statMech data from the .yml file in `path` into `species`
         `pdep` is a boolean specifying whether or not jobList includes a pressureDependentJob.
         """
-        logging.info('Loading statistical mechanics parameters for {0} from .yml file...'.format(species.label))
+        yml_file = os.path.basename(path)
+        if label:
+            logging.info('Loading statistical mechanics parameters for {0} from {1} file...'.format(label, yml_file))
+        else:
+            logging.info('Loading statistical mechanics parameters from {0} file...'.format(yml_file))
         with open(path, 'r') as f:
             data = yaml.safe_load(stream=f)
-        try:
-            if species.label != data['label']:
-                logging.debug('Found different labels for species: {0} in input file, and {1} in the .yml file. '
-                              'Using the label "{0}" for this species.'.format(species.label, data['label']))
-        except KeyError:
-            # Lacking label in the YAML file is strange, but accepted
-            logging.debug('Did not find label for species {0} in .yml file.'.format(species.label))
+        if label:
+            # First, warn the user if the label doesn't match
+            try:
+                if label != data['label']:
+                    logging.debug('Found different labels for species: {0} in input file, and {1} in the .yml file. '
+                                  'Using the label "{0}" for this species.'.format(label, data['label']))
+            except KeyError:
+                # Lacking label in the YAML file is strange, but accepted
+                logging.debug('Did not find label for species {0} in .yml file.'.format(label))
+
+            # Then, set the ArkaneSpecies label to the user supplied label
+            data['label'] = label
         try:
             class_name = data['class']
         except KeyError:
@@ -270,19 +279,30 @@ class ArkaneSpecies(RMGObject):
                       'NASA': NASA,
                       'NASAPolynomial': NASAPolynomial,
                       'ThermoData': ThermoData,
+                      'np_array': numpy.array,
                       }
         freq_data = None
         if 'imaginary_frequency' in data:
             freq_data = data['imaginary_frequency']
             del data['imaginary_frequency']
+        if not data['is_ts']:
+            if 'smiles' in data:
+                data['species'] = Species(SMILES=data['smiles'])
+            elif 'adjacency_list' in data:
+                data['species'] = Species().fromAdjacencyList(data['adjacency_list'])
+            elif 'inchi' in data:
+                data['species'] = Species(InChI=data['inchi'])
+            else:
+                raise ValueError('Cannot load ArkaneSpecies from YAML file {0}. Either `smiles`, `adjacency_list`, or '
+                                 'InChI must be specified'.format(path))
+            # Finally, set the species label so that the special attributes are updated properly
+            data['species'].label = data['label']
+
         self.make_object(data=data, class_dict=class_dict)
         if freq_data is not None:
             self.imaginary_frequency = ScalarQuantity()
-            self.imaginary_frequency.make_object(data=freq_data, class_dict=dict())
-        self.adjacency_list = data['adjacency_list'] if 'adjacency_list' in data else None
-        self.inchi = data['inchi'] if 'inchi' in data else None
-        self.smiles = data['smiles'] if 'smiles' in data else None
-        self.is_ts = data['is_ts'] if 'is_ts' in data else False
+            self.imaginary_frequency.make_object(data=freq_data, class_dict=class_dict)
+
         if pdep and not self.is_ts and (self.transport_data is None or self.energy_transfer_model is None):
             raise ValueError('Transport data and an energy transfer model must be given if pressure-dependent '
                              'calculations are requested. Check file {0}'.format(path))
--- a/arkane/common.py
+++ b/arkane/common.py
@@ -52,7 +52,7 @@ from rmgpy.statmech.translation import IdealGasTranslation
 from rmgpy.statmech.vibration import HarmonicOscillator
 from rmgpy.pdep.collision import SingleExponentialDown
 from rmgpy.transport import TransportData
-from rmgpy.thermo import NASA, Wilhoit
+from rmgpy.thermo import NASA, Wilhoit, ThermoData
 from rmgpy.species import Species, TransitionState
 import rmgpy.constants as constants
 
@@ -180,16 +180,19 @@ class ArkaneSpecies(RMGObject):
                 self.energy_transfer_model = species.energyTransferModel
             if species.thermo is not None:
                 self.thermo = species.thermo.as_dict()
-                thermo_data = species.getThermoData()
-                h298 = thermo_data.getEnthalpy(298) / 4184.
-                s298 = thermo_data.getEntropy(298) / 4.184
-                cp = dict()
-                for t in [300, 400, 500, 600, 800, 1000, 1500, 2000, 2400]:
-                    temp_str = '{0} K'.format(t)
-                    cp[temp_str] = '{0:.2f}'.format(thermo_data.getHeatCapacity(t) / 4.184)
-                self.thermo_data = {'H298': '{0:.2f} kcal/mol'.format(h298),
-                                    'S298': '{0:.2f} cal/mol*K'.format(s298),
-                                    'Cp (cal/mol*K)': cp}
+                data = species.getThermoData()
+                h298 = data.getEnthalpy(298) / 4184.
+                s298 = data.getEntropy(298) / 4.184
+                temperatures = numpy.array([300, 400, 500, 600, 800, 1000, 1500, 2000, 2400])
+                cp = []
+                for t in temperatures:
+                    cp.append(data.getHeatCapacity(t) / 4.184)
+
+                self.thermo_data = ThermoData(H298=(h298, 'kcal/mol'),
+                                              S298=(s298, 'cal/(mol*K)'),
+                                              Tdata=(temperatures, 'K'),
+                                              Cpdata=(cp, 'cal/(mol*K)'),
+                                              )
 
     def update_xyz_string(self):
         """
@@ -265,6 +268,8 @@ class ArkaneSpecies(RMGObject):
                       'SingleExponentialDown': SingleExponentialDown,
                       'Wilhoit': Wilhoit,
                       'NASA': NASA,
+                      'NASAPolynomial': NASAPolynomial,
+                      'ThermoData': ThermoData,
                       }
         freq_data = None
         if 'imaginary_frequency' in data:
--- a/arkane/commonTest.py
+++ b/arkane/commonTest.py
@@ -40,9 +40,10 @@ import logging
 
 import rmgpy
 import rmgpy.constants as constants
+from rmgpy.pdep.collision import SingleExponentialDown
 from rmgpy.species import Species, TransitionState
 from rmgpy.quantity import ScalarQuantity
-from rmgpy.thermo import NASA
+from rmgpy.thermo import NASA, ThermoData
 
 from arkane import Arkane, input
 from arkane.common import ArkaneSpecies, get_element_mass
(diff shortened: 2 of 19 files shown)


rmgpy/rmgobject.pyx
```diff
@@ -115,3 +115,21 @@ cdef class RMGObject(object):
...
                     pass
         self.__init__(**data)
+
+cpdef expand_to_dictionaries(obj):
+    if isinstance(obj, list):
+        return [expand_to_dictionaries(x) for x in obj]
+
+    elif isinstance(obj, dict):
+        for key, value in obj.iteritems():
+            obj[key] = expand_to_dictionaries(value)
```
alongd: what if the key is an object as well?

rmgpy/rmgobject.pyx
```diff
@@ -115,3 +115,21 @@ cdef class RMGObject(object):
                 except ValueError:
                     pass
         self.__init__(**data)
+
+cpdef expand_to_dictionaries(obj):
```
alongd: Maybe consider changing the name, since the function deals with more than just dicts

alongd: Add a docstring?

rmgpy/rmgobject.pyx
alongd: Are we sure this won't change the original object?

rmgpy/rmgobject.pyx
```diff
@@ -115,3 +115,21 @@ cdef class RMGObject(object):
...
+    if isinstance(obj, list):
+        return [expand_to_dictionaries(x) for x in obj]
+
+    elif isinstance(obj, dict):
+        for key, value in obj.iteritems():
+            obj[key] = expand_to_dictionaries(value)
+        return obj
+
+    elif isinstance(obj, np.ndarray):
+        return obj.tolist()
```
alongd: Do we also know to read a list and generate a numpy array if needed?

rmgpy/rmgobject.pyx
```diff
@@ -133,3 +134,52 @@ cpdef expand_to_dictionaries(obj):
...
+    if isinstance(obj, dict):
+        if 'class' in obj:  # This is a dictionary of an object to be created
+            try:
+                class_to_make = class_dictionary[obj['class']]
+            except KeyError:
+                raise KeyError('Class {0} must be provided in the class_dictionary: {1}'.format(obj['class'],
+                                                                                                class_dictionary))
+
+            for key, val in obj.iteritems():
+                if key == 'class':
```
alongd: maybe change `==` to `!=` and skip the `continue`?

rmgpy/rmgobject.pyx
```diff
@@ -133,3 +134,52 @@ cpdef expand_to_dictionaries(obj):
             return expand_to_dictionaries(obj.as_dict())
         else:
             return obj
+
+cpdef recursive_make_object(obj, class_dictionary, make_final_object=True):
```
alongd: Do we ever call this with `make_final_object=False`?

rmgpy/rmgobject.pyx
```diff
@@ -133,3 +134,52 @@ cpdef expand_to_dictionaries(obj):
...
+                return args
+        else:
+            for key, val in obj.iteritems():
+                obj[key] = recursive_make_object(val, class_dictionary)
+            return obj
+
+    elif isinstance(obj, list):
+        return [recursive_make_object(x, class_dictionary) for x in obj]
+
+    elif isinstance(obj, str):
```
alongd: Can you do `isinstance(obj, (str, unicode))` instead? In Py3 (and Py-3 compatible code like ARC) all strings are unicode by default

rmgpy/rmgobject.pyx
```diff
@@ -133,3 +134,52 @@ cpdef expand_to_dictionaries(obj):
...
+        # Check to see if the string needs to be converted to a float or int
+        if bool(re.search('\d', obj)):
```
alongd: Maybe search for `^\d+$` (start to end only digits) instead of the try except

arkane/common.py
```diff
@@ -252,7 +252,8 @@ def load_yaml(self, path, species, pdep=False):
             raise KeyError("Can only make objects if the `class` attribute in the dictionary is known")
         if class_name != 'ArkaneSpecies':
             raise KeyError("Expected a ArkaneSpecies object, but got {0}".format(class_name))
-        del data['class']
+        data = {key: data[key] for key in data.iterkeys() if key != 'class'}
```
alongd: What's the difference between this and the deleted line?

arkane/common.py
```diff
@@ -268,19 +269,19 @@ def load_yaml(self, path, species, pdep=False):
...
                       'Wilhoit': Wilhoit,
                       'NASA': NASA,
+                      'NASAPolynomial': NASAPolynomial,
                       }
         freq_data = None
         if 'imaginary_frequency' in data:
             freq_data = data['imaginary_frequency']
-            del data['imaginary_frequency']
+            data = {key: data[key] for key in data.iterkeys() if key != 'imaginary_frequency'}
+        data['species'] = Species().fromSMILES(data['smiles'])
```
alongd: sometimes an adjlist or InChI are given, and sometimes non of them

arkane/common.py
```diff
@@ -268,19 +269,19 @@ def load_yaml(self, path, species, pdep=False):
...
+        data['species'].label = data['label']
         self.make_object(data=data, class_dict=class_dict)
         if freq_data is not None:
             self.imaginary_frequency = ScalarQuantity()
             self.imaginary_frequency.make_object(data=freq_data, class_dict=dict())
-        self.adjacency_list = data['adjacency_list'] if 'adjacency_list' in data else None
-        self.inchi = data['inchi'] if 'inchi' in data else None
-        self.smiles = data['smiles'] if 'smiles' in data else None
-        self.is_ts = data['is_ts'] if 'is_ts' in data else False
```
alongd: Are these no longer necessary?

rwest:
Not had chance to review the code, but from the pull request description this sounds like a great improvement. 👏🏻

rmgpy/rmgobject.pyx
amarkpayne: Yes, this gets called with `make_final_object=False` in the RMGObject `make_object` method. This is because the `make_object` method passes the dictionary as keyword arguments into the __init__ method. By doing this, it initializes the object in place.

If you call the function with `make_final_object=True` then the function returns a new instance of the object (i.e. not in place). This mean that if you read in a dictionary from a yaml file, you can simply call this function on the dictionary and it will return the object for you (you do not need to create an "empty" instance of the object first.

However, if you already have an empty instance or want to reinitialize the object, you just need to call the `make_object` method of the object.

rmgpy/rmgobject.pyx
amarkpayne: Thanks for the catch!

rmgpy/rmgobject.pyx
amarkpayne: Done, thanks for the suggestion!

rmgpy/rmgobject.pyx
amarkpayne: I don't think this works for floats (try `bool(re.search('^\d+$', '12.45'))` in python 2). I'll keep looking for a better method though.

arkane/common.py
amarkpayne: The deleted line deletes class from the original object, which changes data. In this file I realize now that this isn't an issue, and I can revert this back (it might be faster to use `del data['class']`).

I had used code like this in other files, where changing `data` could cause problems under certain circumstances. For example, if you were scripting and had defined `data` externally but passed it to functions like `recursive_make_dictionaries` multiple times.

Let me know what your preference on this is, I am happy to leave it in or revert this line.

arkane/common.py
amarkpayne: I think this is something I have been struggling with a little bit. It is hard because we need an rmgpy Species to create an ArkaneSpecies, but because we don't have a species attribute on ArkaneSpecies the species object does not get saved to the yaml file. How should this work ideally? How can we reconstruct an ArkaneSpecies from a yaml file if the file does not have smiles or InChI?

arkane/common.py
amarkpayne: I don't believe these are necessary, but I could be wrong. I tested generating ArkaneSpecies yaml files, and everything looked okay (but I'll check again).

I believe this is the case because if these attributes were in `data` then these attributes would be passed to ArkaneSpecies.__init__() by `make_object`, and all 4 of these can be sent to __init__() and will be set there.

If there is a more subtle reason why these things were there though it is possible that they may still be needed.

amarkpayne:
Thanks @alongd for the timely review and great comments! I have addressed your comments and added fixup commits to add changes where I could. Let me know what you think. In the meantime I think it might be worthwhile if I add some additional unit testing to catch some of the points that you addressed.

rmgpy/rmgobject.pyx
alongd: will `^[\d.]+$` work?

amarkpayne: Closer, but it doesn't catch floats in scientific notation like `'-1.20e-05'` and negative numbers

amarkpayne:
@alongd I have applied a few of the changes that we discussed offline and also added some unit tests. I also further refactored quantity, NASA/NASAPolynomial, and Wilhoit. With this I have added everything that I wanted to add for this PR, so this should be ready for one last round of review (and final changes if necessary).

amarkpayne:
As we discussed offline, the fixup commits that I squashed are on the branch `rad_w_fixup` on official

arkane/common.py
alongd: I think the original like is simpler. If it's not problematic any more, lets revert the change.
Same for `imaginary_frequency` below?

arkane/common.py
```diff
@@ -274,15 +275,23 @@ def load_yaml(self, path, species, pdep=False):
         freq_data = None
         if 'imaginary_frequency' in data:
             freq_data = data['imaginary_frequency']
-            del data['imaginary_frequency']
+            data = {key: data[key] for key in data.iterkeys() if key != 'imaginary_frequency'}
+        if not data['is_ts']:
+            if 'smiles' in data:
+                data['species'] = Species().fromSMILES(data['smiles'])
+            elif 'adjacency_list' in data:
+                data['species'] = Species().fromAdjacencyList(data['adjacency_list'])
```
alongd: would you like to add inchi as well? Maybe using something like:
`Species(molecule=[Molecule().fromInChI().generate_resonance_structures()])`
or is there a simpler method?

arkane/common.py
alongd: I think these are here just to report information in the YAML file, since Species doesn't have a .smiles .adjlist, .inchi, and .is_ts attributes. They shouldn't be processed when reading a YAML file, though. I think it's nice and convenient to have adjList / SMILES in the YAML file

rmgpy/quantity.py
```diff
@@ -227,19 +228,6 @@ def as_dict(self):
             output_dict['uncertainty'] = self.uncertainty
             output_dict['uncertaintyType'] = self.uncertaintyType
         return output_dict
-
-    def make_object(self, data, class_dict):
```
alongd: Great!

rmgpy/statmech/rotation.pyx
```diff
@@ -83,6 +84,12 @@ cdef class Rotation(Mode):
         """
         return (Rotation, (self.symmetry, self.quantum))
 
+    def make_object(self, data, class_dict):
```
alongd: Add in rotation.pxd as well, or is the definition already there?

(Same for torsion)

arkane/common.py
mliu49: You can now do `Species(InChI='whatever')` and `Species(SMILES='whatever')` for instantiation.

amarkpayne: Thanks @mliu49 ! I'll implement this then

rmgpy/rmgobject.pyx
```diff
@@ -115,3 +115,35 @@ cdef class RMGObject(object):
...
                     pass
         self.__init__(**data)
+
+cpdef expand_to_dictionaries(obj):
+    """
+    Takes an object of any type (list, dict, str, int, float, RMGObject, etc.) and returns a dictionary representation 
+    of the object (useful for __repr__ methods and outputting to YAML). The function works recursively to all objects
+    nested in the original object.
+    
+    :param obj: Object to be represented by a dictionary (any type)
```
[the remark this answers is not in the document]
mliu49: Yeah, I like the Google style docstring a bit more. I don't think we have an "official" stance though.

rmgpy/statmech/rotation.pyx
amarkpayne: Good question, a thought the same when I first did this. Because `Rotation` inherits from `RMGObject`, this method is already present in this class (and was declared in rmgobject.pxd) and does not need to be declared in the header. All we are doing here is overwriting the parent method.

I am pretty sure I tried adding this to the header file initially, but I got an error. I'll check this one more time though.

rmgpy/rmgobject.pyx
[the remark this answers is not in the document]
amarkpayne: I'll switch to this then. I also figured out how to make pycharm default to this style here: https://www.jetbrains.com/help/pycharm/settings-tools-python-integrated-tools.html

arkane/common.py
amarkpayne: I am not sure I follow completely--let's talk offline about this

rmgpy/rmgobject.pyx
```diff
@@ -147,3 +148,69 @@ cpdef expand_to_dictionaries(obj):
...
+            try:
+                class_to_make = class_dictionary[obj['class']]
+            except KeyError:
+                raise KeyError('Class {0} must be provided in the class_dictionary: {1}'.format(obj['class'],
+                                                                                                class_dictionary))
+
+            args = {key:new_obj[key] for key in new_obj if key != 'class'}
+            if make_final_object:
+                if hasattr(class_to_make, 'make_object'):
+                    created_obj = class_to_make.__new__(class_to_make)
```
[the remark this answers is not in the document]
amarkpayne: Nevermind, this causes more problems than it is worth, i'll stick with `if hasattr(class_to_make, 'make_object'):`

rmgpy/rmgobject.pyx
mliu49: Idk if this is a good solution, but could you just try casting to float, then int if that fails, then leave as string if that fails?

rmgpy/rmgobject.pyx
mliu49: I like shorter names as long as they're still descriptive. Maybe just "expand_to_dict" or "repr_dict"?

mliu49: Another question, would there be any benefit to making this and `recursive_make_object` class methods of RMGObject?

amarkpayne: I'll refactor to `expand_to_dict`.

amarkpayne: As @mliu49 and I discussed offline, there would be benefit to making `expand_to_dict` and `recursive_make_object` static (as opposed to class) method of RMGObject, as we could call these functions from any class that inherits from RMGObject (meaning we would not have to import these functions explicitly most of the time). The downside of this is that these functions work on (hopefully) all objects, not just those that subclass RMGObject. Therefore, to avoid this confusion we will keep this as is

rmgpy/rmgobject.pyx
amarkpayne: Is there an instance where something could be cast as `int` but not `float`? Also, would this convert all `int`s to `float`s?

rmgpy/rmgobject.pyx
mliu49: In Py3, there is no unicode class, rather the default str class is unicode. So when we transition, I think we would need to remove the check for unicode? Is this correct?

rmgpy/rmgobject.pyx
mliu49: Oh yeah, that's true. Is there a benefit to the first search for digits? Could you skip directly to the check for a decimal point?

rmgpy/rmgobject.pyx
```diff
@@ -53,65 +58,127 @@ cdef class RMGObject(object):
...
+        return [expand_to_dictionaries(x) for x in obj]
+
+    elif isinstance(obj, dict):
+        # Create a new dictionary to store the expanded objects
+        new_obj = dict()
+
+        for key, value in obj.iteritems():
+            new_key = expand_to_dictionaries(key)
+            new_value = expand_to_dictionaries(value)
+            new_obj[new_key] = new_value
```
mliu49: Can you use dictionaries as dictionary keys?

amarkpayne:
I have added the last of the changes that @mliu49 and I discussed here and offline. @alongd let me know if there are any other changes. Otherwise if you approve I will update the branch and squash the fixup commits in preparation for merging.

amarkpayne:
I added your formatting changes and squashed all of the fixup commits. Thanks again @mliu49 !

amarkpayne:
The builds are getting errors related to cairo. It looks like this package was updated on the RMG channel today. @mliu49 any idea if something went wrong with the new build?

mliu49:
Sorry, I was playing around with updating a couple dependencies yesterday. The newest noarch cairocffi build by conda-forge apparently works on mac but not linux. The linux-64 build works though, so I just removed the noarch build.

arkane/common.py
amarkpayne: @alongd and I discussed this offline. We ran an example Arkane job to confirm that these attributes are indeed stored on the YAML file even without these lines as expected.

rmgpy/rmgobject.pyx
amarkpayne: Just to document this here, we figured out that dictionaries themselves are not hashable, and thus cannot serve as keys for other dictionaries. Therefore, we can not simply expand objects to dictionaries if they are being used as keys for a dictionary. 

Max and I had some ideas for how we might tackle this, but this is a subject for a later PR due to its complexity and the fact that no RMGObjects currently have attributes that are dictionaries with objects as keys (that is until we make Molecule an RMGObject)

arkane/common.py
amarkpayne: Settled by allowing a species to be created from SMILES, InChI, or adjacency list

arkane/common.py
```diff
@@ -231,21 +231,26 @@ def save_yaml(self, path):
...
         with open(path, 'r') as f:
             data = yaml.safe_load(stream=f)
-        try:
-            if species.label != data['label']:
-                logging.debug('Found different labels for species: {0} in input file, and {1} in the .yml file. '
-                              'Using the label "{0}" for this species.'.format(species.label, data['label']))
-        except KeyError:
-            # Lacking label in the YAML file is strange, but accepted
-            logging.debug('Did not find label for species {0} in .yml file.'.format(species.label))
+        if label:
```
alongd: You can move it to the `if label` condition above

arkane/common.py
```diff
@@ -231,21 +231,26 @@ def save_yaml(self, path):
...
 
-    def load_yaml(self, path, species, pdep=False):
+    def load_yaml(self, path, label=None, pdep=False):
         """
         Load the all statMech data from the .yml file in `path` into `species`
         `pdep` is a boolean specifying whether or not jobList includes a pressureDependentJob.
         """
-        logging.info('Loading statistical mechanics parameters for {0} from .yml file...'.format(species.label))
+        yml_file = os.path.basename(path)
+        if label:
```
alongd: thanks!!

arkane/common.py
```diff
@@ -270,19 +275,33 @@ def load_yaml(self, path, species, pdep=False):
...
+                data['species'] = Species(SMILES=data['smiles'])
+            elif 'adjacency_list' in data:
+                data['species'] = Species().fromAdjacencyList(data['adjacency_list'])
+            elif 'inchi' in data:
+                data['species'] = Species(InChI=data['inchi'])
+            else:
+                raise ValueError('Cannot load ArkaneSpecies from YAML file {0}. Either `smiles`, `adjacency_list`, or '
+                                 'InChI must be specified'.format(path))
+            # Finally, set the species label so that the special attributes are updated properly
+            if label:  # By default use the label passed by the user
```
alongd: for clarity, maybe relocate the label setting near the `if label != data['label']:` block?

rmgpy/rmgobject.pyx
```diff
@@ -42,6 +41,10 @@ cdef class RMGObject(object):
     cpdef dict as_dict(self):
         """
         A helper function for dumping objects as dictionaries for YAML files
+        
+        Returns: 
+            dict: A dictionary representation of the object
+
```
alongd: remove extra line break?

rmgpy/rmgobject.pyx
```diff
@@ -53,68 +56,24 @@ cdef class RMGObject(object):
...
         """
         A helper function for constructing objects from a dictionary (used when loading YAML files)
+        
+        Args:
+            data (dict): The dictionary representation of the object
+            class_dict (dict): A mapping of class names to the classes themselves
+
+        Returns: 
+            None
+
```
alongd: remove extra line break?

arkane/commonTest.py
```diff
@@ -334,25 +335,67 @@ def test_dump_yaml(self):
...
+        # Create YAML file by running Arkane
+        jobList = self.arkane.loadInputFile(self.dump_input_path)
         for job in jobList:
-            job.execute(outputFile=self.load_output_file)
-        arkane_spc = jobList[0].arkane_species
+            job.execute(outputFile=self.dump_output_file)
+
+        # Load in newly created YAML file
+        arkane_spc_old = jobList[0].arkane_species
+        arkane_spc = ArkaneSpecies.__new__(ArkaneSpecies)
```
alongd: What's the difference between this and `arkane_spc = ArkaneSpecies()`?

amarkpayne: I believe that `arkane_spc = ArkaneSpecies()` raises an error, because ArkaneSpecies objects need to be instantiated either with an rmgpy species object or a conformer object. We need an object in order to call the `load_yaml` method, but we don't want to try and instantiate the object until we call `load_yaml`

rmgpy/rmgobject.pyx
amarkpayne: Done

rmgpy/rmgobject.pyx
amarkpayne: Done

arkane/common.py
amarkpayne: I thought about that, but then the "Loading ..." message will have to come after we load `data = yaml.safe_load(stream=f)`. Which would you prefer?

arkane/common.py
amarkpayne: Done

arkane/commonTest.py
amarkpayne: Confirmed by the following:

```
>>> from arkane.common import ArkaneSpecies
>>> arkane_spc = ArkaneSpecies()
  File "<stdin>", line 1, in <module>
  File "RMG-Py/arkane/common.py", line 95, in __init__
    raise ValueError('No species (or TS) or conformer was passed to the ArkaneSpecies object')
ValueError: No species (or TS) or conformer was passed to the ArkaneSpecies object
```

amarkpayne:
The last round of fixes/additions has been rebased/autosquashed. Let me know if this is ready to merge or if there are any other changes we missed.

amarkpayne:
Thanks @alongd and @mliu49 for all of the hard work you put into this PR!

PR #1644, opened 2019-06-27
Fix Travis builds by adding in missing miniconda dependencies

amarkpayne:
### Motivation or Problem
Lately, every Travis build has been failing during the stage where the conda environment is created, with the following error message:

```
ResolvePackageNotFound: 
  - dde -> pydot-ng
  - dde -> theano=0.9.0 -> pygpu[version='>=0.6.5,<0.7']
  - textgenrnn -> theano=0.9.0 -> pygpu[version='>=0.6.5,<0.7']

```

Interestingly, the conda environment builds just fine from a clean install of anaconda on the current master branch. However, after looking at how we have setup our Travis builds, I saw that we use miniconda on Travis. It appears that miniconda needs the dependencies of pydot-ng and pygpu to be explicitly given.

### Description of Changes
Added pydot-ng and pygpu to the conda environment files.

### Testing
On a clean Ubuntu machine, I first tried creating a fresh anaconda install, and creating the conda environment did not cause any problems. After cleaning this anaconda install, I then followed the travis.yml file instructions as closely as I could and used miniconda. This replicated the error on Travis. Finally, I updated the environment files, and again tried to install the environment from miniconda. This seemed to solve the issue

### Reviewer Note
We will see for certain in a few minutes, but I believe this commit will solve the current Travis issue. Whether this is the best way to solve the issue or whether adding pydot-ng or pygpu will cause issues is something I am not sure about currently.

### DIFF
--- a/environment_linux.yml
+++ b/environment_linux.yml
@@ -4,6 +4,7 @@ channels:
   - rmg
   - rdkit
   - cantera
+  - anaconda
 dependencies:
   - cairo
   - cairocffi
@@ -31,7 +32,9 @@ dependencies:
   - psutil
   - pydas >=1.0.1
   - pydot ==1.2.2
+  - pydot-ng
   - pydqed >=1.0.0
+  - pygpu
   - pymongo
   - pyparsing
   - pyrdl
--- a/environment_mac.yml
+++ b/environment_mac.yml
@@ -4,6 +4,7 @@ channels:
   - rmg
   - rdkit
   - cantera
+  - anaconda
 dependencies:
   - cairo
   - cairocffi
@@ -31,7 +32,9 @@ dependencies:
   - psutil
   - pydas >=1.0.1
   - pydot ==1.2.2
+  - pydot-ng
   - pydqed >=1.0.0
+  - pygpu
   - pymongo
   - pyparsing
   - pyrdl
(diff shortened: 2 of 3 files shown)


alongd:
The changes look reasonable and good. Do we know what suddenly triggered this problem?

amarkpayne:
I am not sure, but perhaps if dde and textgenrnn updated their dependencies?

alongd:
Both are from the rmg conda channel, modified by @cgrambow [2](https://anaconda.org/RMG/textgenrnn/files) and [8](https://anaconda.org/RMG/dde/files) months ago. But perhaps a different dependency of theirs got updated?

amarkpayne:
We have decided to merge this PR to fix the issue, but we want to note that we are not quite sure what changed to cause this issue. One possible explanation is that the latest version of miniconda uses conda version 4.7.5, while (at the time of this PR) the version of anaconda you can download uses conda version 4.6.11 . Perhaps something changed in the newest version of conda that causes conda to be unable to resolve the package dependencies? Otherwise we are not sure what is so special about these two dependencies that need to be added to our environment files.

amarkpayne:
Theory confirmed!! This does appear to be an issue with conda 4.7.5 that conda 4.6.11 does not have. I did a clean install of anaconda, and then upgraded conda to 4.7.5 and got the exact same error message (whereas in 4.6.11 I did not). I'll see if we can report the issue somewhere.

amarkpayne:
I left an issue on conda's Github, let's see what they say. (Issue 8844)

PR #1645, opened 2019-06-28
Delete simulation files after unit testing

amarkpayne:
### Motivation or Problem
Whenever the unit test are run, the simulation logs in `rmgpy.tools.simulateTest.py` are created but never deleted (whereas everything else with this test is deleted).

### Description of Changes
Add statements in `rmgpy.tools.simulateTest.py` to remove the log files after the tests have run.

### Testing
Ran the unit tests and everything passed and the logs were deleted


amarkpayne:
This should pass all of the tests as soon as #1644 is merged

mliu49:
Thanks for catching this. The simulate logs were newly added in #1593.

PR #1646, opened 2019-07-01
Run database tests in Travis after testing RMG-Py code

amarkpayne:
### Motivation or Problem
We are starting to run out of time to complete the database checks if the filter thresholds are to be tested every time as in #1577 . This draft PR explores the idea of running the database test in a separate stage after the unit and functional tests have been run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ env:
 
 stages:
   - test
+  - check
   - name: documentation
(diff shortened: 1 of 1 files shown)


amarkpayne:
This appears to work as expected, so I'll add it on to the PR.

mliu49:
I would suggest running the database tests in parallel with the unit tests instead.

PR #1647, opened 2019-07-09
Automatic Network Generation 3

mjohnson541:
This adds a number of miscellaneous bug fixes and improvements for automatic network generation.  Forbidden structure checking is now done on the edge for automatic network generation, flux reduction is now done on netReactions instead of pathReactions and has an additional very robust fallback algorithm just using rate coefficients.  Fixed a bug related to reduction when an isomer is removed.

### DIFF
--- a/arkane/explorer.py
+++ b/arkane/explorer.py
@@ -282,6 +282,7 @@ class ExplorerJob(object):
             if self.energy_tol != np.inf or self.flux_tol != 0.0:
 
                 rxnSet = None
+                productSet = None
 
                 for T in Tlist:
                     if self.energy_tol != np.inf:
@@ -293,17 +294,26 @@ class ExplorerJob(object):
 
                     for P in Plist:
                         if self.flux_tol != 0.0:
-                            rxns = network.get_rate_filtered_reactions(T, P, self.flux_tol)
-                            if rxnSet is not None:
-                                rxnSet &= set(rxns)
+                            products = network.get_rate_filtered_products(T, P, self.flux_tol)
+                            products = [tuple(x) for x in products]
+                            if productSet is not None:
+                                productSet &= set(products)
                             else:
-                                rxnSet = set(rxns)
+                                productSet = set(products)
 
-                logging.info('removing reactions during reduction:')
-                for rxn in rxnSet:
-                    logging.info(rxn)
 
-                network.remove_reactions(reaction_model, list(rxnSet))
+                if rxnSet:
+                    logging.info('removing reactions during reduction:')
+                    for rxn in rxnSet:
+                        logging.info(rxn)
+                    rxnSet = list(rxnSet)
+                if productSet:
+                    logging.info('removing products during reduction:')
+                    for prod in productSet:
+                        logging.info([x.label for x in prod])
+                    productSet = list(productSet)
+
+                network.remove_reactions(reaction_model, rxns=rxnSet, prods=productSet)
 
                 for rxn in jobRxns:
                     if rxn not in network.pathReactions:
--- a/arkane/explorer.py
+++ b/arkane/explorer.py
@@ -206,6 +206,7 @@ class ExplorerJob(object):
 
         forbiddenStructures = getDB('forbidden')
         incomplete = True
+        checkedSpecies = []
 
         while incomplete:
             incomplete = False
@@ -214,6 +215,15 @@ class ExplorerJob(object):
                     for network in self.networks:
                         # compute the characteristic rate coefficient by summing all rate coefficients
                         # from the reactant channel
+                        for spc in reaction_model.edge.species:
+                            if spc in checkedSpecies:
+                                continue
+                            if forbiddenStructures.isMoleculeForbidden(spc.molecule[0]):
+                                reaction_model.removeSpeciesFromEdge(reaction_model.reactionSystems, spc)
+                                reaction_model.removeEmptyPdepNetworks()
+                            else:
+                                checkedSpecies.append(spc)
+
                         kchar = 0.0
                         for rxn in network.netReactions:  # reaction_model.core.reactions+reaction_model.edge.reactions:
                             if (set(rxn.reactants) == set(self.source)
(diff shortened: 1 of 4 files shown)


rmgpy/rmg/pdep.py
```diff
@@ -472,22 +472,49 @@ def remove_disconnected_reactions(self):
                 logging.info('Removing rxn: {}'.format(rxn))
                 self.pathReactions.remove(rxn)
                 
+        nrxns = []
+        for nrxn in self.netReactions:
+           prod = nrxn.products
+           if nrxn.products not in keptProducts or nrxn.reactants not in keptProducts:
```
alongd: `nrxn.products` is a list of length 1 to 3, right? Can an `in` or `not in` check be performed that way? Shouldn't we iterate through the products in `nrxn.products` before the check?

rmgpy/rmg/pdep.py
```diff
@@ -429,20 +429,30 @@ def solve_SS_network(self,T,P):
...
-        c = mp.qr_solve(Amp,bmp)
-        
-        c = np.array(list(c[0]))
-         
-        if any(c<=0.0):
-            c, rnorm = opt.nnls(A,b)
+
+        con = np.linalg.cond(A)
+
+        try:
```
alongd: I think the error message and comment are a bit confusing: So if it is ill-conditioned we raise an error, but if it is "very ill-conditioned" we troubleshoot?

rmgpy/rmg/pdep.py
```diff
@@ -354,32 +354,42 @@ def get_energy_filtered_reactions(self,T,tol):
                 
         return filtered_rxns
 
-    def get_rate_filtered_reactions(self,T,P,tol):
+    def get_rate_filtered_products(self,T,P,tol):
```
alongd: Add space between comma and argument

arkane/explorer.py
```diff
@@ -294,17 +295,29 @@ def execute(self, outputFile, plot, format='pdf', print_summary=True, speciesLis
...
+                if rxnSet:
+                    logging.info('removing reactions during reduction:')
+                    for rxn in rxnSet:
+                        logging.info(rxn)
+                if productSet:
+                    logging.info('removing products during reduction:')
+                    for prod in productSet:
+                        logging.info([x.label for x in prod])
+
+                if rxnSet:
```
alongd: maybe combine these conditions into the ones above?

rmgpy/rmg/pdep.py
```diff
@@ -536,16 +536,45 @@ def remove_disconnected_reactions(self):
         self.Nreac = len(self.reactants)
         self.Nprod = len(self.products)
 
-    def remove_reactions(self,reactionModel,rxns):
+    def remove_reactions(self,reactionModel,rxns=None,prods=None):
```
alongd: add spaces?

rmgpy/rmg/pdep.py
```diff
@@ -536,16 +536,45 @@ def remove_disconnected_reactions(self):
...
         """
         removes a list of reactions from the network and all reactions/products
         left disconnected by removing those reactions
         """
-        for rxn in rxns:
-            self.pathReactions.remove(rxn)
-            
+        if rxns:
```
alongd: shouldn't this be `if rxns is not None`? Or is there no practical difference?

rmgpy/rmg/pdepTest.py
```diff
@@ -162,19 +162,14 @@ def setUp(self):
         self.pdepnetwork.index = 1
         self.pdepnetwork.explored = []
 
-    
-    def test_SS_solver(self):
```
alongd: But pdep.py still had the method, doesn't it? (or was the method itself removed as well?)
If we're still using the original method somewhere, lets keep/relocate the test for it?

arkane/explorer.py
```diff
@@ -214,6 +215,16 @@ def execute(self, outputFile, plot, format='pdf', print_summary=True, speciesLis
                     for network in self.networks:
                         # compute the characteristic rate coefficient by summing all rate coefficients
                         # from the reactant channel
+                        for spc in reaction_model.edge.species:
+                            if spc in checkedSpecies:
+                                continue
+                            if forbiddenStructures.isMoleculeForbidden(spc.molecule[0]):
+                                reaction_model.removeSpeciesFromEdge(reaction_model.reactionSystems, spc)
+                                reaction_model.removeEmptyPdepNetworks()
+                                logging.error(spc.label)
```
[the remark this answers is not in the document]
mjohnson541: removed

rmgpy/rmg/pdep.py
mjohnson541: keptProducts is constructed a list of list of species (from a pdep network perspective this makes sense since each represents a configuration) so this should work fine.

rmgpy/rmg/pdep.py
```diff
@@ -472,22 +472,49 @@ def remove_disconnected_reactions(self):
                 logging.info('Removing rxn: {}'.format(rxn))
                 self.pathReactions.remove(rxn)
                 
+        nrxns = []
+        for nrxn in self.netReactions:
+           prod = nrxn.products
```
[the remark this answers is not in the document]
mjohnson541: fixed

rmgpy/rmg/pdep.py
mjohnson541: There's no practical difference it won't trigger if rxns = None or rxns = [].

arkane/explorer.py
mjohnson541: done

rmgpy/rmg/pdep.py
mjohnson541: fixed

rmgpy/rmg/pdep.py
mjohnson541: I've adjusted this to use an if statement rather than the first try except, which makes it a bit clearer.  In general if we expect we can handle the matrix we do it normally, otherwise we fall back to arbitrary precision arithmetic, and lastly if we can't manage that we just assume all the isomers that aren't reactant channels have zero concentration which makes the calculation trivial.

rmgpy/rmg/pdepTest.py
mjohnson541: So the method exists, but the problem is that pdepTest doesn't have appropriate objects to test it directly after changes made in solve_SS_network. test_flux_filter does test it though, so it's still under coverage.
I invested sometime trying to figure out how to make it work, but given it's already under reasonable coverage (get_rate_filtered_products isn't a complicated function), it didn't seem worth it.

rmgpy/rmg/pdep.py
mjohnson541: done

arkane/explorer.py
```diff
@@ -227,21 +238,14 @@ def execute(self, outputFile, plot, format='pdf', print_summary=True, speciesLis
...
-                                logging.error(spc.label)
-                            else:
-                                logging.info('adding new isomer {0} to network'.format(spc))
-                                flags = np.array([s.molecule[0].getFormula() == form
-                                                  for s in reaction_model.core.species])
-                                reaction_model.enlarge((network, spc), reactEdge=False, unimolecularReact=flags,
+                            logging.info('adding new isomer {0} to network'.format(spc))
+                            flags = np.array([s.molecule[0].getFormula() == form for s in reaction_model.core.species])
+                            reaction_model.enlarge((network, spc), reactEdge=False, unimolecularReact=flags,
                                                        bimolecularReact=np.zeros((len(reaction_model.core.species),
```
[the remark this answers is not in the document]
mjohnson541: done

mjohnson541:
@alongd

rmgpy/rmg/pdep.py
```diff
@@ -429,20 +439,27 @@ def solve_SS_network(self,T,P):
...
+            c = np.linalg.solve(A,b)
+        else:
+            logging.warn("Matrix Ill-conditioned, attempting to use Arbitrary Precision Arithmetic")
+            mp.dps = 30+int(np.log10(con))
+            Amp = mp.matrix(A.tolist())
+            bmp = mp.matrix(b.tolist())
             
-        c = c.astype(np.float64)
-        
+            try:
```
alongd: Maybe only leave the critical part (`qr_solve`?) inside the `try: except` block?

mjohnson541: So the problem is that nnls (non-negative least squares) is critical as well when required and may fail as well (although given the qr_solve has to succeed to try nnls it shouldn't fail that often, but it should still fall back properly if nnls is called and fails).

alongd:
OK, I think we're all set. Can you rebase?

PR #1649, opened 2019-07-11
Merge model improvements

goldmanm:
### Motivation or Problem
merge_models has some unexpected behavior, like throwing divide by zero errors when empty models are given and renaming the species in a chemkin model

### Description of Changes
This PR cleans up some of the workings of the merge_models tool

1. refactored the `execute` method to call two separate methods that perform different tasks, which allows for better unittesting
2. added a unittest to ensure proper behavior
3. prevents print statements from causing divide by zero errors for models without any reactions :)
4. minimize reindexing to only those species in conflicts.

### Testing
Haha. I ran the new unittest. :) I also ran merge model on minimal and superminimal examples.

### Reviewer Tips
Run merge models on two of your favorite RMG generated models :)


rmgpy/tools/merge_models.py
```diff
@@ -160,4 +161,17 @@ def combine_models(models):
                   '#{0:d}.'.format(i+1, Nrxn - Nrxn0, len(model.reactions)))
     print('The merged model has {0:d} species and {1:d} reactions'
           ''.format(len(final_model.species), len(final_model.reactions)))
+
+    # ensure no species with same name and index
+    label_index_dict = {}
+    for s in final_model.species:
+        if s.label not in label_index_dict.keys():
```
mliu49: This can just be `if s.label not in label_index_dict`. No need to call `keys()`.

rmgpy/tools/merge_models.py
```diff
@@ -160,4 +161,17 @@ def combine_models(models):
...
+            label_index_dict[s.label] = [s.index]
+        else:
+            if s.index in label_index_dict[s.label]:
+                # obtained a duplicate
+                s.index = max(label_index_dict[s.label]) + 1
```
mliu49: Do we want to have unique indices? In other words, do we want to reindex starting from the largest index in the entire model + 1?

rmgpy/rmg/model.py
```diff
@@ -130,11 +130,11 @@ def merge(self, other):
...
-        speciesIndex = 0
-        for spec in finalModel.species:
-            if spec.label not in ['Ar','N2','Ne','He']:
-                spec.index = speciesIndex + 1
-                speciesIndex += 1
+        #speciesIndex = 0
+        #for spec in finalModel.species:
+        #    if spec.label not in ['Ar','N2','Ne','He']:
+        #        spec.index = speciesIndex + 1
+        #        speciesIndex += 1
```
mliu49: Since you implemented a different indexing system, I would just remove this section instead of commenting it out.

rmgpy/tools/merge_models.py
goldmanm: This code isn't guaranteeing unique indexes at all. To do that, we would pretty much have to renumber everything except the first model, which would make this PR much less useful.

From my point of view, if the user starts with an un-indexed mechanism, like aramcomech, the output should also not have indexes.

rmgpy/rmg/model.py
goldmanm: will do.

rmgpy/tools/merge_models.py
goldmanm: Is there a benefit to unique indices that I've overlooked? My main goal for this PR was to not have any CHEMKIN or cantera strings conflict, while keeping those names as similar as possible to the input names.

goldmanm: To give an example of the functioning in this PR: if model 1 has `CH3(15)` and model 2 has `H2O(15)`, then neither of their names would be changed, since both the name and index don't match.

I would think for many users, avoiding chemkin or cantera names changing would be preferable to ensuring every species has a unique number.

goldmanm: Let me know your thoughts on this

rmgpy/tools/merge_models.py
goldmanm: good to know! Just made the code simpler!

rmgpy/tools/merge_models.py
mliu49: Ah, I forgot that we would also need to check for species which already have the same indices.

I think it would only matter if you were merging RMG models and wanted to load the merged model back into RMG, which would parse the indices from the labels. I'm not sure exactly what effect duplicate indices would have in that scenario though. It might not actually matter.

I'm ok with the behavior you implemented. I think the key desired outcome is that the resulting merged mechanism file is directly simulate-able in Chemkin or Cantera.

PR #1651, opened 2019-07-12
Element count from conformer

alongd:
### Motivation or Problem
Previously thermo blocks in Chemkin files outputted by Arkane did not have an element count if the molecular structure wasn't specified by the user.

### Description of Changes
Now we use the xyz information Arkane has for each species (specifically the species.conformer.number attribute) to determine the element count.

Additional minor modifications:
- replaced file `open()` `close()` with `with open()`
- don't append a species adjList dictionary to the species dictionary file if the species is already there

### Testing
Added a thermoTest for the new element_count_from_conformer() method.

### Reviewer Tips
Try running a ThermoJob without specifying a structure and look at the Chemkin output

### DIFF
--- a/arkane/thermo.py
+++ b/arkane/thermo.py
@@ -53,7 +53,7 @@ from rmgpy.molecule import Molecule
 from rmgpy.molecule.util import retrieveElementCount
 
 from arkane.output import prettify
-from arkane.common import ArkaneSpecies
+from arkane.common import ArkaneSpecies, symbol_by_number
 
 
 ################################################################################
@@ -183,15 +183,15 @@ class ThermoJob(object):
         with open(os.path.join(os.path.dirname(outputFile), 'chem.inp'), 'a') as f:
             if isinstance(species, Species):
                 if species.molecule and isinstance(species.molecule[0], Molecule):
-                    element_count = retrieveElementCount(species.molecule[0])
+                    element_counts = retrieveElementCount(species.molecule[0])
                 else:
                     try:
-                        element_count = species.props['element_count']
+                        element_counts = species.props['element_counts']
                     except KeyError:
-                        element_count = {'C': 0, 'H': 0}
+                        element_counts = self.element_count_from_conformer()
             else:
-                element_count = {'C': 0, 'H': 0}
-            chemkin_thermo_string = writeThermoEntry(species, element_count=element_count, verbose=True)
+                element_counts = {'C': 0, 'H': 0}
+            chemkin_thermo_string = writeThermoEntry(species, elementCounts=element_counts, verbose=True)
             f.write('{0}\n'.format(chemkin_thermo_string))
 
         # write species dictionary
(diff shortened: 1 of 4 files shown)


arkane/thermo.py
```diff
@@ -155,54 +155,61 @@ def save(self, outputFile):
...
         if isinstance(species, Species):
             if species.molecule and isinstance(species.molecule[0], Molecule):
-                with open(os.path.join(os.path.dirname(outputFile), 'species_dictionary.txt'), 'a') as f:
-                    f.write(species.molecule[0].toAdjacencyList(removeH=False, label=species.label))
-                    f.write('\n')
+                spec_dict_path = os.path.join(os.path.dirname(outputFile), 'species_dictionary.txt')
+                is_species_in_dict = False
+                if os.path.isfile(spec_dict_path):
+                    with open(spec_dict_path, 'r') as f:
+                        # check whether the species dictionary contains this species, in which case do not re-append
```
goldmanm: nice check

arkane/thermo.py
```diff
@@ -212,6 +212,23 @@ def save(self, outputFile):
...
+        """
+        element_counts = dict()
+        for number in self.species.conformer.number.value_si:
+            symbol = symbol_by_number[number]
+            if symbol in element_counts:
+                element_counts[symbol] += 1
+            else:
+                element_counts[symbol] = 1
+        return element_counts
+
```
goldmanm: It feels like this method might do better as a part of the Conformer class.

arkane/thermoTest.py
```diff
@@ -0,0 +1,73 @@
...
+        element_count = self.thermo_job.element_count_from_conformer()
+        self.assertEqual(element_count, {'H': 4, 'C': 2})
+
+
+
+################################################################################
+
+
+if __name__ == '__main__':
+    unittest.main(testRunner=unittest.TextTestRunner(verbosity=2))
```
goldmanm: Even if we decide to move the conformer counting method to Conformer.pyx, having this unittest would still be super useful (so no need to eliminate the whole thing).

arkane/thermo.py
alongd: I looked at it, but this method uses `symbol_by_number()` from Arkane common.py, I wouldn't like to make Arkane a dependency of conformer, not to add the `symbol_by_number()` function to conformer. I say let's leave it as is. Would you agree?

goldmanm: When looking at the usages of `symbol_by_number` and the corresponding `get_element_mass`, they all relate in someway to the conformer object's attributes (some of the uses involve converting C to 12 in the Log classes, which is then just converted back in arkane). Ideally, I would think almost any method manipulating the atom's XYZ coordinates and their corresponding masses should be within the Conformer class, since this complies with the principle of encapsulation in object oriented programming. 

Though honestly, it would require some restructuring, and I am fine merging it in, we already have many other poor uses of object oriented programing in the code already.

alongd:
@goldmanm, I added another (minor) commit for consistent naming of `element_counts` in Species.props
It's set in Arkane statmech.py and isn't used in RMG
With this, the PR should now be good to go

PR #1656, opened 2019-07-17
Fluorine fixes

rwest:
A couple of small atomType fixes for Fluorine, that were causing segfaults when testing the database https://github.com/ReactionMechanismGenerator/RMG-database/pull/317

### DIFF
--- a/rmgpy/molecule/atomtype.py
+++ b/rmgpy/molecule/atomtype.py
@@ -665,7 +665,7 @@ atomTypes['F'  ].setActions(incrementBond=[],               decrementBond=[],   
 atomTypes['F1s'].setActions(incrementBond=[],               decrementBond=[],               formBond=['F1s'],      breakBond=['F1s'],      incrementRadical=['F1s'], decrementRadical=['F1s'], incrementLonePair=[],      decrementLonePair=[])
 
 #these are ordered on priority of picking if we encounter a more general atomType for make
-allElements=['H', 'C', 'O', 'N', 'S', 'Si', 'Cl', 'Ne', 'Ar', 'He', 'X']
+allElements=['H', 'C', 'O', 'N', 'S', 'Si', 'Cl', 'F', 'Ne', 'Ar', 'He', 'X']
 #list of elements that do not have more specific atomTypes
 nonSpecifics=['H', 'He', 'Ne', 'Ar',]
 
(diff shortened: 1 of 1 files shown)


rmgpy/molecule/atomtype.py
```diff
@@ -665,7 +665,7 @@ def getFeatures(self):
 atomTypes['F1s'].setActions(incrementBond=[],               decrementBond=[],               formBond=['F1s'],      breakBond=['F1s'],      incrementRadical=['F1s'], decrementRadical=['F1s'], incrementLonePair=[],      decrementLonePair=[])
 
 #these are ordered on priority of picking if we encounter a more general atomType for make
-allElements=['H', 'C', 'O', 'N', 'S', 'Si', 'Cl', 'Ne', 'Ar', 'He', 'X']
+allElements=['H', 'C', 'O', 'N', 'S', 'Si', 'Cl', 'F', 'Ne', 'Ar', 'He', 'X']
```
alongd: While at it, and although beyond the scope of the PR title, perhaps add `I` as well?
BTW, if Ar is here, should we also consider He and Ne?

rmgpy/molecule/atomtype.py
```diff
@@ -548,7 +548,7 @@ def getFeatures(self):
 
 atomTypes['F'  ] = AtomType('F',   generic=['R','R!H','Val7'],      specific=['F1s'])
 atomTypes['F1s'] = AtomType('F1s', generic=['R','R!H','F','Val7'],  specific=[],
-                             single=[0,1], allDouble=[0], rDouble=[], oDouble=[], sDouble=[], triple=[0], benzene=[0], lonePairs=[3], charge=[0])
+                             single=[0,1], allDouble=[0], rDouble=[], oDouble=[], sDouble=[], triple=[0], quadruple=[0], benzene=[0], lonePairs=[3], charge=[0])
```
alongd: The default value for `quadruple` is an empty list. Shouldn't that be enough (I guess not from your experience with F, but I wonder why)?
Perhaps a quicker fix for all elements would be to change the default behaviour of AtomType.
I suggest editing [line 118](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/molecule/atomtype.py#L118) to:
`self.quadruple = quadruple or [0]`
so zero quadruple bonds becomes the default.

rwest: Actually, we don't know that the default was causing a problem, we just thought we'd make Fluorine the same as Chlorine. I think our actual problem was solved by the other change.

rmgpy/molecule/atomtype.py
rwest: Good point with `I`. Added in subsequent commit. 
He and Ne are already listed.

Issue #1655, opened 2019-07-17
frequencyScaleFactor in Arkane examples might not be used except in input.py


amarkpayne:
### Bug Description
`frequencyScaleFactor` is defined in the arkane example file `RMG-Py/examples/arkane/species/C2H6/C2H6.py`. However, this factor is being overwritten by the scale factor for the model chemistry. Interestingly though, defining `frequencyScaleFactor` in  `RMG-Py/examples/arkane/species/C2H6/input.py` does overwrite the model chemistry as expected.

Given this, it is possible that defining `frequencyScaleFactor` outside of the input.py file has no effect. At the very least though, we should add a comment to these example files outlining what happens to this parameter when defined in the species files instead of the input file

### How To Reproduce
Run this Arkane example, then look at the output YAML file.

### Additional Context
I am not certain that this is a bug, or that we need to do anything at all yet, but I am documenting this here so that I follow-up on this. After I do a little more digging (including a closer look at the documentation, perhaps I missed a section that clearly explains this) I'll push any necessary changes.

amarkpayne:
Line 334-335 in arkane.statmech.py warns the user that the frequency scale factor is always ignored. This seems fine to me, but I think we should not have this specified in an example species file if this is the case. I'll open a quick PR with this change.

PR #1657, opened 2019-07-17
Fix frequencyScaleFactor in Arkane examples

amarkpayne:
### Motivation or Problem
Setting `frequencyScaleFactor` in an species file instead of an Arkane input file does nothing. Therefore, we should not define this in our example files. See #1655 

### Reviewer Note
~~Must be merged after #1643~~ (Update: This PR can now be reviewed and merged when ready)

### DIFF
--- a/examples/arkane/species/C2H6/input.py
+++ b/examples/arkane/species/C2H6/input.py
@@ -11,6 +11,10 @@ modelChemistry = "CBS-QB3"
 useHinderedRotors = True
 useBondCorrections = False
 author = 'I.B. Modeling'
+# Example of how to manually set the frequencyScaleFactor
+# Factor of 0.99 for CBS-QB3 ZPE scaling (dx.doi.org/10.1063/1.477924)
+# Factor of 1.014 for converting between ZPE and frequency scaling factors (dx.doi.org/10.1021/ct100326h)
+frequencyScaleFactor = 0.99*1.014
 
 species('C2H6', 'C2H6.py',
         structure=SMILES('CC'))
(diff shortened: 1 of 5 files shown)


examples/arkane/species/C2H6/input.py
```diff
@@ -11,6 +11,7 @@
 useHinderedRotors = True
 useBondCorrections = False
 author = 'I.B. Modeling'
+frequencyScaleFactor = 0.99
```
alongd: This is a very confusing point:
The **ZPE** scaling factor for CBS-QB3 is indeed 0.99, but the **frequency** scaling factor (which Arkane requires) for CBS-QB3 is 0.99 * 1.014 =~ 1.004. I know we've all been using 0.99, we only standardized this entire scaling factors  thing recently. The good news is that it's automated, so we don't need to worry about it too much. I suggest either not adding this line, or assign `0.99 * 1.014` (the latter is more transparent).

(The 1.014 is a factor between ZPE scaling and freq. scaling that is universal, according to Truhlar - almost independent of the level of theory)

amarkpayne: I think I'll go with your second suggestion and modify this line and the one above it to:
'''
# Example of how to manually set the frequencyScaleFactor
frequencyScaleFactor = 0.99*1.014  # 0.99 for CBS-QB3, 1.014 for scaling between ZPE and freq. according to Truhlar
'''

The reason I wanted to keep this line was because it is the only species example where we manually set the frequencyScaleFactor. Does this seem reasonable? If so I'll make this change and force push it

alongd: The ref for CBS-QB3's **ZPE** scaling factor is [dx.doi.org/10.1063/1.477924](https://dx.doi.org/10.1063/1.477924), and the ref for the 1.014 factor is [dx.doi.org/10.1021/ct100326h](https://dx.doi.org/10.1021/ct100326h)

amarkpayne: Thanks! I'll add this and push the changes

amarkpayne:
Looks like the Travis builds are working again--I'll restart the others. @alongd is it okay if I go ahead and merge this?

Issue #797, opened 2016-10-26
generateReverseRateCoefficient cannot handle PDepArrhenius with MultiArrhenius rates


mliu49:
This was discovered while using the Plot Kinetics tool on the website.

The input included pressure dependent rates in PLOG format, but with two sets of Arrhenius parameters at each pressure. RMG parsed this as a PDepArrhenius object with a MultiArrhenius rate at each pressure. However, the `generateReverseRateCoefficient` function assumes that PDepArrhenius objects only have Arrhenius rates for each pressure, and errors while trying to calculate the reverse rate.

Questions:
- Is this the correct way for RMG to be interpreting the rate?
- If so, is it ok to modify `generateReverseRateCoefficient` to recognize such a form?

rwest:
See PR https://github.com/ReactionMechanismGenerator/RMG-Py/pull/329 which fixed issue https://github.com/ReactionMechanismGenerator/RMG-Py/issues/147 which has a detailed write-up.  I *think* after Connie's fix the answer to you first bullet point is yes, and thus it'd be helpful if we could reverse such rates, somehow.

mliu49:
Fixed by #1659

PR #1659, opened 2019-07-21
Support reversing PDepArrhenius objects containing MultiArrhenius rates

mliu49:
### Motivation or Problem
This really should have been fixed a long time ago, but better late than never.

#329 implemented the capability for the PDepArrhenius class to hold MultiArrhenius rates instead of just Arrhenius rates. This is commonly seen in literature rates, especially from MESS calculations, e.g.
```
C2H3+O2=CH2CHOO                                    4.07E+27       -4.67    5222.0
  PLOG/1.000E-02    1.55E+24       -5.45    9662.0/
  PLOG/1.000E-02    1.78E-09        4.15    -4707.0/                            ! fit btw. 400 and 1250 K with MAE of 0.4%, 1.2%
  PLOG/1.000E-01    3.48E+56      -15.01    19160.0/
  PLOG/1.000E-01    2.36E+22       -4.52    2839.0/                             ! fit btw. 400 and 1350 K with MAE of 0.2%, 0.5%
  PLOG/3.160E-01    1.25E+64      -16.97    21290.0/
  PLOG/3.160E-01    2.00E+26       -5.43    2725.0/                             ! fit btw. 400 and 1450 K with MAE of 0.2%, 0.6%
  PLOG/1.000E+00    3.34E+61      -15.79    20150.0/
  PLOG/1.000E+00    6.13E+28       -5.89    3154.0/                             ! fit btw. 400 and 1550 K with MAE of 0.2%, 1.1%
  PLOG/3.160E+00    7.34E+53      -13.11    17300.0/
  PLOG/3.160E+00    2.14E+29       -5.80    3520.0/                             ! fit btw. 400 and 1650 K with MAE of 0.3%, 1.5%
  PLOG/1.000E+01    4.16E+48      -11.21    16000.0/
  PLOG/1.000E+01    3.48E+28       -5.37    3636.0/                             ! fit btw. 400 and 1750 K with MAE of 0.4%, 1.9%
  PLOG/3.160E+01    2.33E+43       -9.38    14810.0/
  PLOG/3.160E+01    3.32E+27       -4.95    3610.0/                             ! fit btw. 400 and 1900 K with MAE of 0.6%, 2.5%
  PLOG/1.000E+02    3.41E+39       -8.04    14360.0/
  PLOG/1.000E+02    1.03E+27       -4.72    3680.0/                             ! fit btw. 400 and 2100 K with MAE of 0.9%, 3.1%
```

As noted in #797, RMG could not reverse these rates and would print the following error:
```
AttributeError: 'rmgpy.kinetics.arrhenius.MultiArrhenius' object has no attribute 'T0'
```

### Description of Changes
This PR adds the ability to reverse such rates along with a unit test. Since RMG can reverse MultiArrhenius rates, this simply reverses the MultiArrhenius rate associated with each pressure.

This also adds `rcond=None` to all calls to `numpy.linalg.lstsq`. In numpy 1.14, the default behavior for the `rcond` parameter changed. Passing `rcond=None` silences the warnings printed by numpy and accepts the new default behavior.

https://docs.scipy.org/doc/numpy/reference/generated/numpy.linalg.lstsq.html

### Testing
Check that tests pass. Jobs containing such rates (e.g. from the Klippenstein_Glarborg2016 kinetics library) will crash during the check collision limit violators step, which is another way to test.

### DIFF
--- a/rmgpy/reactionTest.py
+++ b/rmgpy/reactionTest.py
@@ -32,25 +32,26 @@
 This module contains unit tests of the rmgpy.reaction module.
 """
 
-import numpy
 import unittest
+
+import cantera as ct
+import numpy
 from external.wip import work_in_progress
 
-from rmgpy.quantity import Quantity
-from rmgpy.species import Species, TransitionState
-from rmgpy.molecule import Molecule
-from rmgpy.reaction import Reaction
-from rmgpy.quantity import Quantity
-from rmgpy.statmech.translation import Translation, IdealGasTranslation
-from rmgpy.statmech.rotation import Rotation, LinearRotor, NonlinearRotor, KRotor, SphericalTopRotor
-from rmgpy.statmech.vibration import Vibration, HarmonicOscillator
-from rmgpy.statmech.torsion import Torsion, HinderedRotor
-from rmgpy.statmech.conformer import Conformer
-from rmgpy.kinetics import Arrhenius, ArrheniusEP, MultiArrhenius, \
-                           PDepArrhenius, MultiPDepArrhenius, \
-                           SurfaceArrhenius, StickingCoefficient
-from rmgpy.thermo import Wilhoit, ThermoData, NASA, NASAPolynomial
 import rmgpy.constants as constants
+from rmgpy.kinetics import Arrhenius, ArrheniusEP, MultiArrhenius, PDepArrhenius, MultiPDepArrhenius, \
+                           ThirdBody, Troe, Lindemann, Chebyshev, SurfaceArrhenius, StickingCoefficient
+from rmgpy.molecule import Molecule
+from rmgpy.quantity import Quantity
+from rmgpy.reaction import Reaction
+from rmgpy.species import Species, TransitionState
+from rmgpy.statmech.conformer import Conformer
+from rmgpy.statmech.rotation import NonlinearRotor
+from rmgpy.statmech.torsion import HinderedRotor
+from rmgpy.statmech.translation import IdealGasTranslation
+from rmgpy.statmech.vibration import HarmonicOscillator
+from rmgpy.thermo import Wilhoit, ThermoData, NASA, NASAPolynomial
+
 
 ################################################################################
 
@@ -961,8 +962,6 @@ class TestReaction(unittest.TestCase):
         Test the Reaction.generateReverseRateCoefficient() method works for the ThirdBody format.
         """
 
-        from rmgpy.kinetics import ThirdBody
-
         arrheniusLow = Arrhenius(
             A = (2.62e+33,"cm^6/(mol^2*s)"), 
             n = -4.76, 
@@ -1008,8 +1007,6 @@ class TestReaction(unittest.TestCase):
         """
         Test the Reaction.generateReverseRateCoefficient() method works for the Lindemann format.
         """
-
-        from rmgpy.kinetics import Lindemann
 
         arrheniusHigh = Arrhenius(
             A = (1.39e+16,"cm^3/(mol*s)"), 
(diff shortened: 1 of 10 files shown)


rmgpy/reactionTest.py
```diff
@@ -46,7 +46,9 @@
 from rmgpy.statmech.vibration import Vibration, HarmonicOscillator
 from rmgpy.statmech.torsion import Torsion, HinderedRotor
 from rmgpy.statmech.conformer import Conformer
-from rmgpy.kinetics import Arrhenius, ArrheniusEP, SurfaceArrhenius, StickingCoefficient
+from rmgpy.kinetics import Arrhenius, ArrheniusEP, MultiArrhenius, \
```
amarkpayne: Can you also remove some of the unused imports in this file? For example, from `rmgpy.quantity import Quantity` is stated twice.

rmgpy/reactionTest.py
```diff
@@ -46,7 +46,9 @@
...
+                           PDepArrhenius, MultiPDepArrhenius, \
+                           SurfaceArrhenius, StickingCoefficient
```
amarkpayne: This can fit on two lines instead of three if you want. Is there a reason you went with three lines?

rmgpy/reactionTest.py
```diff
@@ -689,7 +691,6 @@ def testGenerateReverseRateCoefficientPDepArrhenius(self):
         """
         Test the Reaction.generateReverseRateCoefficient() method works for the PDepArrhenius format.
         """
-        from rmgpy.kinetics import PDepArrhenius
```
amarkpayne: There are some similar imports on line 1241, can some of these also be removed?

rmgpy/reactionTest.py
```diff
@@ -746,12 +747,66 @@ def testGenerateReverseRateCoefficientPDepArrhenius(self):
...
+
+        arrhenius1 = MultiArrhenius(
+            arrhenius=[
+                Arrhenius(A=(1.0e12, "s^-1"), n=1.0, Ea=(10.0, "kJ/mol"), T0=(300.0, "K"), Tmin=(300.0, "K"), Tmax=(2000.0, "K")),
+                Arrhenius(A=(1.0e12, "s^-1"), n=1.0, Ea=(10.0, "kJ/mol"), T0=(300.0, "K"), Tmin=(300.0, "K"), Tmax=(2000.0, "K")),
+            ],
+            comment="""This data is completely made up""",
+        )
+
+        pressures = numpy.array([0.1, 10.0])
```
amarkpayne: Is there a reason why you defined these as separate variable here before creating the PDepArrhenius object?

rmgpy/reactionTest.py
mliu49: I added line breaks by category, e.g. high-P, pdep, surface. I can put them on two lines if preferred.

rmgpy/reactionTest.py
mliu49: I copied an existing test. I thought about condensing it, but thought might be clearer this way.

rmgpy/reactionTest.py
mliu49: Done

rmgpy/reactionTest.py
mliu49: Done

amarkpayne:
> The code itself looks good, thanks for the PR! I have a couple of quick comments but this looks mostly ready to go. Before you rebase, please swap the order of the commits so that the updated code that allows PDepArrhenius to be reversed comes before the unit test.

Just to document it, @mliu49 and I discussed the order of commits offline. It makes more sense as is, as having the test first shows that we have a bug in the old implementation, and then the next commit finally fixes the bug.

rmgpy/reactionTest.py
amarkpayne: I see it now--this is fine then

PR #1660, opened 2019-07-22
Revert Fix Travis builds by adding in missing miniconda dependencies

amarkpayne:
Packages requiring libgcc have been rebuilt to require libgcc-ng instead. With this the conda solver can resolve the environment without needing pydot-ng and pygpu to be given explicitly

### Motivation or Problem
In #1644 we added some packages explicitly to the environment files to fix failing Travis builds. It has since been determined that all that was necessary was to add in the anaconda channel so that `pydot-ng`, `pygpu`, and `libgpuarray` could be found (these packages were previously taken from the Free channel, but this was removed as a default channel in conda 4.7). The only packages that need these dependencies are DDE and anything that depends on theano, which we are about to update/get rid of. When this happens, we should get rid of these dependencies from our environment files, as they will no longer be needed.

### Description of Changes

1. Removed the anaconda channel
2. Removed pydot-ng
3. Removed pygpu

### Testing
Note that the Travis builds for this PR will fail until DDE/Theano are dealt with, which must happen before the transition to python 3.

### DIFF
--- a/environment_linux.yml
+++ b/environment_linux.yml
@@ -4,7 +4,6 @@ channels:
   - rmg
   - rdkit
   - cantera
-  - anaconda
 dependencies:
   - cairo
   - cairocffi
@@ -32,9 +31,7 @@ dependencies:
   - psutil
   - pydas >=1.0.1
   - pydot ==1.2.2
-  - pydot-ng
   - pydqed >=1.0.0
-  - pygpu
   - pymongo
   - pyparsing
   - pyrdl
--- a/environment_mac.yml
+++ b/environment_mac.yml
@@ -4,7 +4,6 @@ channels:
   - rmg
   - rdkit
   - cantera
-  - anaconda
 dependencies:
   - cairo
   - cairocffi
@@ -32,9 +31,7 @@ dependencies:
   - psutil
   - pydas >=1.0.1
   - pydot ==1.2.2
-  - pydot-ng
   - pydqed >=1.0.0
-  - pygpu
   - pymongo
   - pyparsing
   - pyrdl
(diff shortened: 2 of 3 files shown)


mliu49:
Any idea why the Travis builds are showing a `boost` conflict on this branch but not master?

amarkpayne:
> Any idea why the Travis builds are showing a `boost` conflict on this branch but not master?

Good question--I saw this on my local build as well. I tried doing some debugging with conda-tree on this but I didn't get very far. My going theory was that without being able to find pydot-ng, pygpu, and/or libgpuarray that the solver tried solving the environment by constraining the version of some other package(s) that ultimately lead to an error in boost. I do know though that in the current environment that the version of `boost` is low enough to satisfy MUQ but the version of `libboost` is high enough to satisfy RDKit

amarkpayne:
Sorry, I misunderstood the plan for updating DDE. There is no python 2 version of its replacement, so it will be updated after the upgrade to python 3. Therefore, this PR will wait until after python 3

cgrambow:
Should I just include the changes in this PR in the DDE/Chemprop update PR after the switch to Py3?

amarkpayne:
> Should I just include the changes in this PR in the DDE/Chemprop update PR after the switch to Py3?

That is fine with me, whichever you prefer. I need to make one more change to this as soon as @mliu49 updates some of the conda build files (we need to remove the anaconda channel from here as well) but as soon as that is done I'll add the changes here and you can just cherry-pick the commit.

cgrambow:
Is this ready for cherry-picking?

amarkpayne:
> Is this ready for cherry-picking?

Yes, feel free to cherry-pick this and close this PR

PR #1661, opened 2019-07-22
Test that we can make a sample molecule for every atom type.

rwest:
### Motivation or Problem
This unit test at the start of this PR was inspired by efforts to debug something in some changes to the database (https://github.com/ReactionMechanismGenerator/RMG-database/pull/317) , which caused a segfault when running the database Tests, but was in fact caused by an undetected bug in RMG (https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1656) not a problem in the database.  I felt there ought to have been a unit test to prevent the bug.

### Description of Changes
The goal is to make a group definition out of each atom type, and then make a sample molecule for each group definition. This should test the group definition and the `make_sample_molecule` method.

At first, the test failed, and that was either because of 
1) bug(s) in the way the test is currently written, or 
2) bug(s) in our atom type definitions 
3) bug(s) in our `make_sample_molecule` method.

I then fixed several of the bugs - mostly in the algorithm to make aromatic rings for sample molecules.
The remaining bugs I split into a "work in progress" unit test - they involve aromatic O and S atoms (['O4b', 'S4b'] ) which seem to be poorly supported currently.

I then made a more stringent test  - that not only do we generate sample molecules without crashing, but that they're correct. This test is also currently marked a "work in progress". The list of problems is Xo, Cq, Sib, Sibf, Siq. Fixing those can be a separate issue.

### DIFF
--- a/rmgpy/molecule/atomtype.py
+++ b/rmgpy/molecule/atomtype.py
@@ -404,7 +404,7 @@ ATOMTYPES['N3t'] = AtomType('N3t', generic=['R', 'R!H', 'R!H!Val7', 'N', 'Val5']
 # examples for N3t: N2, N#C, N#[C], N#CC
 ATOMTYPES['N3b'] = AtomType('N3b', generic=['R', 'R!H', 'R!H!Val7', 'N', 'Val5'], specific=[],
                             single=[0], all_double=[0], r_double=[0], o_double=[0], s_double=[0], triple=[0], quadruple=[], benzene=[2], lone_pairs=[1], charge=[0])
-# examples for N3b: Oxazole, Pyradine, Pyrazine, 1,3,5-Triazine, Benzimidazole, Purine
+# examples for N3b: Oxazole, Pyridine, Pyrazine, 1,3,5-Triazine, Benzimidazole, Purine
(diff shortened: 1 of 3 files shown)


rwest:
Just updated to RMG 3.

JacksonBurns:
This PR is abandoned but up-to-date and simple enough that it could be merged. I am going to change the base branch to `main`, update it, and see what happens.

JacksonBurns:
Rebased and CI is now running.

rwest:
Couldn't make sample molecules for types ['Ctc', 'N5bd', 'O4b', 'P3b', 'P5b', 'P5bd', 'S4b']

rwest:
Now down to ['N5bd', 'O4b', 'P3b', 'P5b', 'P5bd', 'S4b']. 
Noting that these are all benzene atom types, I am hoping they all have the same problem.

rwest:
Got several more working.
There is still some (other) problem with O4b and S4b, but this gets them closer.
Note that Sib and Sibf atom types do not crash, but do not create a correct sample molecule at this time.
(The unit test just checks for exceptions, not that the sample molecules are right, so these errors are undetected)

JacksonBurns (approved):
LGTM. Thanks for bringing this one back from the dead. Approving now (tests should pass w/o problems) and then feel free to merge.

PR #1665, opened 2019-07-23
Debug travis

amarkpayne:
### Motivation or Problem
A bug (well, at least we hope it is a bug and not a new feature) in Travis Worker 6.2.1 means that these builds only run with 1 CPU instead of the expected two. This made us realize that we have a bug in one of our unit tests that does not check the number of CPUs properly (fixed by this PR). We also want to display CPU and memory information in the Travis build log so that we can debug errors like this in the future.

While this PR will not fix Travis builds completely for 6.2.1 (there is no work around for insufficient memory) it fixes everything that we can right now, and we have opened up issues with Travis.

### Description of Changes

- Reset module level rmgpy.rmg.main.maxproc back to the default of 1 at the end of individual tests
- Print CPU and memory info in the Travis logs.

### Testing
I have already tested that our bug fixes work in Travis Worker 6.2.1, so we just need this to build in Travis Worker 6.2.0 so that we can merge this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,9 @@ stages:
 
 before_install:
   - cd ..
+  # Display CPU and memory information on the Linux builds
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then lscpu; fi
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then free -m; fi
   # Install miniconda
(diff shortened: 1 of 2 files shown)


amarkpayne:
See https://github.com/travis-ci/worker/issues/604 for the issue we created with Travis on these problems

mliu49:
You can rebase and I'll merge this.

PR #1668, opened 2019-07-24
Do not deploy RMG-tests for master and stable branches

mliu49:
Simplify logic to deploy only from feature branches
Changes behavior to not deploy from master or stable branches

### Motivation or Problem
Running RMG-tests is relatively expensive since it consumes at most 3 Travis builds and takes over an hour to complete. Since the goal of RMG-tests is to compare models between feature branches and master, it seems unnecessary to test on the master and stable branches.

### Description of Changes
Effective change is to prevent builds on master and stable. Also cleaned up the logic a bit in the deploy script.

### Testing
It's a bit hard to test, but we should see an RMG-tests build triggered by the push build for this branch. We should confirm that the correct message is displayed for the PR build. However, we'll only be able to confirm that it doesn't build on master after merging.

### Reviewer Tips
Double-check that the bash code is correct.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -2,31 +2,24 @@
 
 set -e # exit with nonzero exit code if anything fails
 
-echo 'Travis Build Dir: '$TRAVIS_BUILD_DIR
-
 if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
-  DEPLOY_BRANCH=$TRAVIS_PULL_REQUEST
+  echo "RMG-tests should not deploy from pull requests"
+  exit 0
+elif [ "$TRAVIS_BRANCH" == "master" ]; then
+  echo "RMG-tests should not deploy from the master branch"
+  exit 0
+elif [ "$TRAVIS_BRANCH" == "stable" ]; then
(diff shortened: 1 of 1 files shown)


deploy.sh
```diff
@@ -2,32 +2,25 @@
 
 set -e # exit with nonzero exit code if anything fails
 
-echo 'Travis Build Dir: '$TRAVIS_BUILD_DIR
-
 if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
```
amarkpayne: Any reason why this is `!= "false" ` instead of just `== "true" ` ?

mliu49: This environment variable will contain the pull request number if it is a pull request and false if not.

amarkpayne:
This was in the build after merging the PR. Looks like we are good to go!

```
[0K$ bash ./deploy.sh
RMG-tests should not deploy from the master branch
travis_time:end:01e43cd8:start=1564056530518451346,finish=1564056530525338338,duration=6886992
[0Ktravis_fold:end:after_success.2
[0K
Done. Your build exited with 0.
```

PR #1669, opened 2019-07-24
Created a new reactor type MBSampledReactor for modeling laser lab Mass-Spec experiments

jimchu10:
### Motivation or Problem
To model the sampling delay in mass-spec of Green group laser apparatus, it is important to generate a new reactor type.

### Description of Changes
New reactor type added, and can be used in simulate.py

### Testing
The MBSampledReactor added works same as what it was before rebase on master. Same time-dependent concentration for species by giving same input files


mliu49:
I think it would make sense for MBSampledReactor to be a subclass of SimpleReactor. I compared the two, and it seemed like only the `residual` method was changed. Doing it this way would make it much easier to maintain.

I added a commit making this change, but have not tested it yet beyond checking that it compiled corrected. Could you run a simulation to see if this works and gives the same result as before?

mliu49:
Update regarding my previous commit. Turns out that it won't work because SimpleReactor defines a `jacobian` method which necessarily cannot be included for MBSampledReactor.

rmgpy/tools/simulateTest.py
```diff
@@ -74,6 +76,23 @@ def test_liquid(self):
         shutil.rmtree(os.path.join(folder, 'solver'))
         os.remove(os.path.join(folder, 'simulate.log'))
 
+    def test_mbSampled(self):
+        """Test that we can simulate an MBSampledReactor"""
+        folder = os.path.join(os.path.dirname(rmgpy.__file__), 'tools/data/sim/mbSampled')
```
amarkpayne: Need to join tools, data, sim, and mbSampled. Can't use `/`

amarkpayne:
When I run the test locally, this is the plot that I get. Is this what the simulation is supposed to do?

simulation_1_30

rmgpy/tools/simulateTest.py
mliu49: Thanks for keeping up with Windows support 😉

mliu49:
Yeah, the concentrations are all extremely low, so they don't show up on the plot.

rmgpy/tools/simulateTest.py
amarkpayne: RMG-Py: potentially Windows compatible since 2019 :)

PR #1670, opened 2019-07-25
Check numpy version before passing rcond to numpy.linalg.lstsq

mliu49:
### Motivation or Problem
#1659 added the `rcond` argument to calls to `numpy.linalg.lstsq`. A side-effect is that we broke compatibility with numpy versions prior to 1.14.

Original plan was to simply increase our numpy version requirement, but after thinking about it a bit more, this change does not seem like sufficient reason to do so.

### Description of Changes
This PR adds automatic determination of the appropriate value for rcond depending on the numpy version.

It also updates and futurizes `utilities.py`.

### Testing
Will test in an environment with an older version of numpy. Travis should be sufficient to test that it works in newer versions of numpy.


mliu49:
I've confirmed that this works with numpy 1.11

Issue #1671, opened 2019-07-27
Unexpected behavior in Arkane kinetic fitting


goldmanm:
### Bug Description
If a user of arkane does not input temperature ranges for high-pressure kinetics, Arkane will space the values between 298 and 2500 K using equal spacing in 1/T (located in generateKinetics of kinetics.py)

If the user specifies Tmin, Tmax, and Tcount, then Arkane will space the points out equally in T space (located in __init__ in kinetics.py).

The difference in spacing also goes against what the original commit message says (beaafb5e31fccafcd2f4e8827b096). I would think that to be consistent, we should either use T or 1/T spacing for both. My personal preference is 1/T spacing. Unless people have opposition, I cam implement this change.

alongd:
I agree we should use consistent 1/T spacing

amarkpayne:
I think this issue has been resolved by the merged PR. If I am wrong please feel free to reopen this.

PR #1672, opened 2019-07-28
Fix kinetics fitting with Tmin & Tmax in Arakne

goldmanm:
### Motivation or Problem

Previously, there were two spots where a list of temperatures
was created for fitting high pressure kinetics in Arkane.
One fit using inverse temperature spacing and the other one fit
with linear temperature spacing. See #1671

### Description of Changes

This PR consolidates creating a listing of temperatures
to the __init__ method, so there is only one spot where temperatures
are initialized. The default min and max in `generateKinetics` was
moved into the __init__ method. Three unittests were written for
obtaining the proper temperatures in the __init__ method.

`generateKinetics` previously accepted an arguement Tlist, which
is already an attribute of KineticsJob, creating redundancy.
The Tlist argument was removed and self.Tlist was used
in this method instead, simplifying the code.

Calls to quantity.Quantity in the __init__ method were unnecessary
before setting them with quantity.Temperature, so the __init__
code was simplified by setting attributes with (value, 'K') instead.

Documentation was updated to better describe the three methods
of specifying temperatures when calling `kinetics` in Arkane

### Testing
did `make test` and ran the kinetics examples in RMG. also did `make html` to check for errors in the documentation.

### Reviewer Tips
Read over code to make sure it makes sense.

### DIFF
--- a/arkane/kinetics.py
+++ b/arkane/kinetics.py
@@ -70,35 +70,19 @@ class KineticsJob(object):
                  Tcount=0,
                  sensitivity_conditions=None):
         self.usedTST = False
-        if Tmin is not None:
-            self.Tmin = quantity.Quantity(Tmin)
-        else:
-            self.Tmin = None
-
-        if Tmax is not None:
-            self.Tmax = quantity.Quantity(Tmax)
-        else:
-            self.Tmax = None
-
-        self.Tcount = Tcount
+        self.Tmin = Tmin if Tmin is not None else (298, 'K')
+        self.Tmax = Tmax if Tmax is not None else (2500, 'K')
+        self.Tcount = Tcount if Tcount > 3 else 50
 
         if Tlist is not None:
-            self.Tlist = quantity.Quantity(Tlist)
-            self.Tmin = quantity.Quantity(numpy.min(self.Tlist.value_si), "K")
-            self.Tmax = quantity.Quantity(numpy.max(self.Tlist.value_si), "K")
+            self.Tlist = Tlist
+            self.Tmin = (min(self.Tlist.value_si), 'K')
+            self.Tmax = (max(self.Tlist.value_si), 'K')
             self.Tcount = len(self.Tlist.value_si)
         else:
-            if Tmin and Tmax is not None:
-
-                if self.Tcount <= 3.:
-                    self.Tcount = 50
-
-                stepsize = (self.Tmax.value_si - self.Tmin.value_si) / self.Tcount
-
-                self.Tlist = quantity.Quantity(numpy.arange(self.Tmin.value_si,
-                                                            self.Tmax.value_si + stepsize, stepsize), 'K')
-            else:
-                self.Tlist = None
+            self.Tlist = (1 / numpy.linspace(1 / self.Tmax.value_si,
+                                            1 / self.Tmin.value_si,
+                                            self.Tcount), 'K')
 
         self.reaction = reaction
         self.kunits = None
@@ -145,10 +129,7 @@ class KineticsJob(object):
         If `plot` is True, then plots of the raw and fitted values for the kinetics
         will be saved.
         """
-        if self.Tlist is not None:
-            self.generateKinetics(self.Tlist.value_si)
-        else:
-            self.generateKinetics()
+        self.generateKinetics()
         if output_directory is not None:
             try:
                 self.write_output(output_directory)
@@ -176,7 +157,7 @@ class KineticsJob(object):
         logging.debug('Finished kinetics job for reaction {0}.'.format(self.reaction))
         logging.debug(repr(self.reaction))
 
-    def generateKinetics(self, Tlist=None):
+    def generateKinetics(self):
         """
         Generate the kinetics data for the reaction and fit it to a modified Arrhenius model.
         """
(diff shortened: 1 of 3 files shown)


arkane/kinetics.py
```diff
@@ -71,34 +71,28 @@ def __init__(self, reaction,
                  sensitivity_conditions=None):
         self.usedTST = False
         if Tmin is not None:
-            self.Tmin = quantity.Quantity(Tmin)
+            self.Tmin = Tmin
         else:
-            self.Tmin = None
+            self.Tmin = (298, 'K')
```
alongd: Could you simplify these assignments into:
`self.Tmin = Tmin if Tmin is not None else (298, 'K')`?
Same for Tmax?

We could also simplify Tcount:
`self.Tcount = Tcount if Tcount > 3. else 50`

arkane/kinetics.py
```diff
@@ -71,34 +71,28 @@ def __init__(self, reaction,
...
 
+        if Tcount <= 3.:
+            Tcount = 50
         self.Tcount = Tcount
 
         if Tlist is not None:
-            self.Tlist = quantity.Quantity(Tlist)
-            self.Tmin = quantity.Quantity(numpy.min(self.Tlist.value_si), "K")
-            self.Tmax = quantity.Quantity(numpy.max(self.Tlist.value_si), "K")
+            self.Tlist = Tlist
```
alongd: maybe skip this assignment, and just use Tlist?

arkane/kinetics.py
```diff
@@ -71,34 +71,28 @@ def __init__(self, reaction,
...
+            self.Tmin = (min(self.Tlist.value_si), 'K')
+            self.Tmax = (max(self.Tlist.value_si), 'K')
             self.Tcount = len(self.Tlist.value_si)
         else:
-            if Tmin and Tmax is not None:
```
alongd: nice cleanup

arkane/kineticsTest.py
```diff
@@ -0,0 +1,94 @@
...
+        self.assertEqual(len(Tlist), kjob.Tcount)
+
+    def test_give_Trange_for_kineticsjob(self):
+        """
+        Ensures that Tlist is set when a range of temperatures is specified
+        """
+        rxn = Reaction(transitionState=TransitionState())
+        kjob = KineticsJob(rxn, Tmin=(50, 'K'), Tmax=(4000, 'K'), Tcount=5)
+        self.assertEqual(5, len(kjob.Tlist.value_si))
+        self.assertAlmostEqual(50, min(kjob.Tlist.value_si))
```
alongd: why `assertAlmostEqual`?

arkane/kineticsTest.py
```diff
@@ -0,0 +1,94 @@
...
+        self.assertAlmostEqual(4000, max(kjob.Tlist.value_si))
+        inverse_tlist = 1 / kjob.Tlist.value_si
+        self.assertAlmostEqual(inverse_tlist[1] - inverse_tlist[0],
```
alongd: nice test!

documentation/source/users/arkane/input.rst
```diff
@@ -818,24 +818,28 @@ High Pressure Limit Kinetics
...
 Use a ``kinetics()`` function to make Arkane execute the high-pressure limit kinetic
 parameters computation for a reaction. The ``'label'`` string must correspond to that of
-a defined ``reaction()`` function. If desired, define a temperature range and number
-of temperatures at which the high-pressure rate coefficient will be tabulated and saved to
-the outupt file. The 3-parameter modified Arrhenius coefficients will automatically be fit
-to the computed rate coefficients. The quantum tunneling factor will also be displayed.
+a defined ``reaction()`` function.
 
-Below is a typical ``kinetics()`` function::
+You have three options for specifying the temperature which a modified Arrhenius
```
alongd: Should it be `to which a modified...` or `will be fit to.`? (add `to`)

documentation/source/users/arkane/input.rst
```diff
@@ -818,24 +818,28 @@ High Pressure Limit Kinetics
...
+expression will be fit.
+
+Give an explicit list of temperatures to fit::
 
     kinetics(
     label = 'H + C2H4 <=> C2H5',
-    Tmin = (400,'K'), Tmax = (1200,'K'), Tcount = 6,
+    Tlist = ([400,500,700,900,1100,1200],'K'),
```
alongd: This is misleading since we only take Tmin and Tmax and the count from this list, we don't actually use the intermediate values. Perhaps we should (what do you think?). Either way, the documentation should reflect what we do.

arkane/kinetics.py
goldmanm: Good idea. I just added that.

arkane/kinetics.py
goldmanm: I would need to assign something to `self.Tlist` at some point, since it is used in `generateKinetics`.

I recognized the `self.Tlist.value_si` looks a little clunky, but since the Tlist passed to this method is of the format `([100,200,500,800,1200], 'K'), I would need to use `Tlist[0]` to get the array, which is not very clear to someone just looking at this method.

arkane/kineticsTest.py
goldmanm: good catch. changed to assertEqual for the Tcount variables

documentation/source/users/arkane/input.rst
goldmanm: added a `to` before `which`

documentation/source/users/arkane/input.rst
goldmanm: I definitely think we should use all the values, which fortunately is what the current implementation does. In `generateKinetics`, we calculate the TST rate at each temperature, shown below.

```
        for i, t in enumerate(self.Tlist.value_si):
            klist[i] = self.reaction.calculateTSTRateCoefficient(t)
```

PR #1676, opened 2019-08-08
fix bug with duplicate library reactions

mjohnson541:
This sets the rxn.family attribute to the name of the library it was loaded from rather than the library it was originally from this prevents reactions in a seed mechanism that were originally from a reaction library from being duplicated if that reaction library is used as a reaction library or seed mechanism.


alongd:
I guess you guys already discussed it, but seems like we're misusing the family attribute. Also, won't that end in a weird comment in the mechanism file, such as `From reaction family: KnownLibrary`?

mjohnson541:
The comments will be the same.  It isn't exactly a misuse, LibraryReaction already had a family attribute that was identical to the original library, now the family attribute is the library it was loaded from.  That doesn't exactly make it a clear solution to the problem, but it does make sense in context.

mliu49:
For documentation purposes, the second commit is to address issues I've had with loading reaction libraries with the comment `Originally from reaction library: Unclassified`, which is appended whenever one imports a Chemkin file in as a reaction library.

I think it's safer to only do the additional parsing for autoGenerated libraries.

mjohnson541:
@alongd does this solve your issue?

rmgpy/data/kinetics/library.py
```diff
@@ -245,6 +245,7 @@ def getLibraryReactions(self):
                                       duplicate=entry.item.duplicate, reversible=entry.item.reversible,
                                       allow_pdep_route=entry.item.allow_pdep_route,
                                       elementary_high_p=entry.item.elementary_high_p)
+                rxn.family = self.label #the library the reaction was loaded from (opposed to originally from)
```
alongd: This works because we also check the family attribute before adding a reaction as "duplicate"?

mliu49: When we check for existing reactions, we use the family attribute to identify the "source" of a reaction, regardless of whether it's a LibraryReaction or TemplateReaction. The library attribute is not used for this check. When saving Chemkin files, we do use the library attribute to write the source library in the comments.

So this fix does something which was not originally intended, but happens to do exactly what we want, by using the family attribute to store the actual source (the seed library) and the library attribute to store the original source (the original library).

PR #1677, opened 2019-08-08
Reduce species copying and resonance structure generation when reacting

mliu49:
### Motivation or Problem
Two performance related changes, cherry-picked from #1577.

### Description of Changes
1. Do not make deep copies of species in `react_species` ([reasoning](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1577#issuecomment-519315774))
2. Remove a redundant resonance structure generation step

### Testing
Tested with the liquid oxidation test from RMG-tests, but on the FilterCritPerReactFamily branch.

Should confirm that RMG-tests show no differences in models except for runtime.

### DIFF
--- a/rmgpy/data/kinetics/kineticsTest.py
+++ b/rmgpy/data/kinetics/kineticsTest.py
@@ -1091,3 +1091,103 @@ class TestKinetics(unittest.TestCase):
         self.assertIn('*2',found_labels)
         self.assertIn('*3',found_labels)
 
+    def test_species_preserved_after_generate_reactions(self):
+        """
+        Test that Species objects do not retain changes after generating reactions
+
+        This tests a case involving identical reactants
+        """
+        reactant1 = Species(index=1, label='ethyl', SMILES='C[CH2]')
+        reactant1_copy = reactant1.copy(deep=True)  # These copies record the state of the original attributes
+        expected_product_1 = Species(SMILES='CC')
+        expected_product_2 = Species(SMILES='C=C')
+
+        reaction_list = self.database.kinetics.generate_reactions_from_families(
+            [reactant1, reactant1], only_families=['Disproportionation'], resonance=True
+        )
+
+        # First confirm that we get the expected reaction
+        self.assertEqual(len(reaction_list), 1)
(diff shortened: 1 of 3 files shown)


rmgpy/data/kinetics/kineticsTest.py
```diff
@@ -1091,3 +1091,97 @@ def test_add_atom_labels_for_reaction_3(self):
         self.assertIn('*2',found_labels)
         self.assertIn('*3',found_labels)
 
+    def test_species_preserved_after_generate_reactions(self):
+        """Test that Species objects do not retain changes after generating reactions"""
+        r1 = Species(index=1, label='ethyl', SMILES='C[CH2]')
```
mjohnson541: Can you use something more descriptive than `r1`?  We more commonly use `r` for reaction than reactant, which can result in some confusion here.

rmgpy/data/kinetics/kineticsTest.py
```diff
@@ -1091,3 +1091,97 @@ def test_add_atom_labels_for_reaction_3(self):
...
+        self.assertTrue(case_1 or case_2)
+
+        r1_out, r2_out = reaction.reactants
+
+        # The species in the output reaction should be new objects because we made them from Molecule objects
+        self.assertIsNot(r1, r1_out)
+        self.assertIsNot(r1, r2_out)
+
+        # The molecule objects should be the same references for the first output species
+        self.assertIsNot(r1.molecule[0], r1_out.molecule[0])
```
mjohnson541: Why is one of them the molecule objects the same reference and the other isn't?  Why can't all identical molecule objects be the same reference?

rmgpy/data/kinetics/kineticsTest.py
mliu49: I feel like it should be clear from context, but I changed them anyway.

rmgpy/data/kinetics/kineticsTest.py
mliu49: If they were the same object then you couldn't assign different labels to the atoms.

PR #1683, opened 2019-08-12
Chemprop

cgrambow:
### Motivation or Problem
DDE does not work with Python 3.

### Description of Changes
Switch the ML thermo estimator to chemprop. New models have been added in database PR ReactionMechanismGenerator/RMG-database#351.

### Testing
Once Python 3 transition has occurred, check that tests pass and ML example can be run.

### Question for Reviewer
I don't understand how I should be editing the different environment files. I added chemprop to the py3 environment file. Should I also add it to the other ones?

### DIFF
--- a/environment_py3.yml
+++ b/environment_py3.yml
@@ -4,12 +4,14 @@ channels:
   - rmg
   - rdkit
   - cantera
-  - anaconda
+  - pytorch
+  - conda-forge
 dependencies:
   - cairo
   - cairocffi
   - cantera >=2.3.0
   - cclib
+  - chemprop
   - coolprop
   - coverage
   - cython >=0.25.2
--- a/rmgpy/ml/__init__.py
+++ b/rmgpy/ml/__init__.py
@@ -1,6 +1,3 @@
-#!/usr/bin/env python
-# -*- coding: utf-8 -*-
-
 ###############################################################################
 #                                                                             #
 # RMG - Reaction Mechanism Generator                                          #
--- a/rmgpy/ml/estimator.py
+++ b/rmgpy/ml/estimator.py
@@ -1,6 +1,3 @@
-#!/usr/bin/env python
-# -*- coding: utf-8 -*-
-
 ###############################################################################
 #                                                                             #
 # RMG - Reaction Mechanism Generator                                          #
@@ -28,17 +25,24 @@
 #                                                                             #
 ###############################################################################
 
+import contextlib
 import os
+from argparse import Namespace
+from typing import Callable, Union
+
+try:
+    import chemprop
+except ImportError as e:
+    chemprop = None
+    chemprop_exception = e
 import numpy as np
 
-os.environ['KERAS_BACKEND'] = 'theano'
-from dde.predictor import Predictor
-
+from rmgpy.molecule import Molecule
+from rmgpy.species import Species
 from rmgpy.thermo import ThermoData
 
 
-class MLEstimator():
-
+class MLEstimator:
     """
     A machine learning based estimator for thermochemistry prediction.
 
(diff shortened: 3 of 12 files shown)


mliu49:
Regarding environment files, my thought is that by the end of this, we will go back to having a single environment file for both mac and linux (and dropping the windows one) which will be `environment_py3.yml`, though we'll probably rename it to `environment.yml`.

I assume this breaks the Python 2 build? If so, then this will not be merged until the last stage of the transition.

rmgpy/ml/__init__.py
```diff
@@ -1,6 +1,3 @@
-#!/usr/bin/env python
-# -*- coding: utf-8 -*-
```
mliu49: Did you remove all of the shebangs intentionally? Are they not necessary in Python 3?

rmgpy/ml/estimator.py
```diff
@@ -28,17 +25,24 @@
 #                                                                             #
 ###############################################################################
 
+from argparse import Namespace
```
mliu49: Put from imports below imports

environment_py3.yml
```diff
@@ -1,15 +1,16 @@
 name: rmg_env
 channels:
+  - pytorch
```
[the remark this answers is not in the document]
cgrambow: It has to be before the defaults channel because the pytorch version in defaults is relatively old. If we were too add pytorch to the RMG channel, then RMG would have to be before defaults, not sure what the preferred option is.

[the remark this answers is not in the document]
cgrambow: Oh actually, I just checked pytorch in default Anaconda and it seems they updated to the current version 5 days ago. I can test if the defaults build works correctly with chemprop and then could remove the pytorch channel completely.

rmgpy/ml/__init__.py
cgrambow: The coding specification is not necessary in Python 3 because that is now the default.

I removed the first line as well because the modules aren't meant to be run as scripts and Python doesn't need that line if you're importing a module. That line is only useful if you plan to run a Python file without telling the command line explicitly that you're using Python. Most RMG files have that first line but unless they're scripts, none of those files need it. We could consider removing it everywhere. Or would you rather I keep it here to be consistent with the rest of RMG?

rmgpy/ml/estimator.py
cgrambow: Is that a PEP-8 rule or an RMG rule? I thought imports were supposed to be alphabetical, but I might be wrong.

mliu49: Great question! I thought it was in PEP-8 but apparently not. It seems that it's just the default PyCharm style.

I like grouping from and direct imports separately, but I would want to get everyone's opinions before making it an RMG rule.

rmgpy/ml/__init__.py
mliu49: In that case, I think we can remove them everywhere. That can be a separate PR though.

cgrambow: Should I keep it here then?

environment_py3.yml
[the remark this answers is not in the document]
cgrambow: Removed pytorch channel and force pushed.

rmgpy/ml/estimator.py
cgrambow: So in any case we want the following order for each module:

1. standard library imports
2. related third party imports
3. local application/library specific imports

And then I think I agree with you that it makes sense to first have normal imports and then from imports and imports should be alphabetical within each group.

cgrambow:
So it seems that even the pytorch builds on the official anaconda channel are not reliable. The only error-free builds are available on the pytorch channel. This is what was causing your issue, @mliu49. So I actually went ahead and added the pytorch channel back in. I also upgraded the chemprop build on the RMG channel to use pytorch-cpu, which is actually a lot smaller as well and doesn't install CUDA dependencies. Obviously this means that chemprop cannot be used with GPU but that should be fine for RMG users.

Tested locally, on Mithrim, on RMG, and on Pharos.

Added a bunch of fixup commits to address the issues.

mliu49:
Thanks! I think the fixes look good. What led to the addition of hyperopt and tensorboardx as dependencies compared to before?

cgrambow:
They already were dependencies of chemprop before but when I built the environment locally I had conda-forge already in my .condarc so it didn't complain that it couldn't resolve those dependencies.

mliu49:
Incorporated into #1713 and #1724.

PR #1686, opened 2019-08-12
Upgrade Modules to Python 3

amarkpayne:
### Changes

- PEP-8 reformatting
- Renaming local variables following PEP-8
- Futurize
- Fix relative imports
- Fix iterators


amarkpayne:
Not complete yet, but I want to see what happens with the tests and Codacy checks

rwest:
Looks like a great start! Please can you post somewhere a list of steps taken to do similar on other unmerged pull requests, other branches, and other projects that build on or use RMG?

amarkpayne:
> Looks like a great start! Please can you post somewhere a list of steps taken to do similar on other unmerged pull requests, other branches, and other projects that build on or use RMG?

Certainly! First off, I recommend taking a look at the google doc that @mliu49 sent out in the shared Slack channel, there are a lot of good details in there. The general workflow I am taking so far is as follows (@mliu49, @mjohnson541, @alongd, and @xiaoruiDong feel free to add anything or correct any mistakes I make):

1. **PEP-8 reformatting** (optional but highly encouraged. It has made a lot of the files easier to parse through already). As Max showed me I highly recommend using PyCharm's Reformat Code tool (Code > Reformat Code) and then going through the file to fix up any mistakes or changes that still need to be made. This should focus just on whitespace and formatting issues.

2. **Renaming local variables following PEP-8** (optional, but I think it is worthwhile). Again, PyCharm's refactor feature is great for this. DO NOT rename any public variables (functions/methods, classes, etc.)

3. **Futurize** (required) Get the futurize script `conda install future` and then call it on a single file with `futurize -0 -wn file_name`. The zero flag performs both stage 1 and stage 2 improvements to the code. The `w` flag actually writes changes to the file, and the `n` flag prevents creating a backup file (this is what git is for of course). I haven't run into as many of the issues with this as Max and Matt, but one issue that comes up is that if the code has any division statements it will try to import old style division and place it everywhere in your code. If this happens I recommend reverting the file back, placing a `from __future__ import division` statement at the top of the import block, and running the futurize script again. Thanks to Max for figuring out that this prevents the script from using old style division (make sure to unit test that old style division wasn't needed anywhere).

4. **Replace relative import statements with absolute import statements** (required). The futurize scripts will fix the relative import statements, but we want to get rid of them completely. For example, no `from molecule import Molecule`, replace this with `from rmgpy.molecule import Molecule`.

5. **Replace iter[keys, values, items] with [keys, values, items]** (required). Searching the files for `iter` is helpful here. If an iterator won't work and a list is need be sure to convert with `list()`

6. **Unit Test one more time** (required)

7. **Squash commits so that there is one commit for each of steps 1-5.** (required)

8. **Open PR to py3 branch** or master (if this is after the initial transition or an existing PR)

mjohnson541:
It should be noted that you shouldn't rename local variables because if you do so it will make rebasing on RMG an absolute nightmare.

amarkpayne:
I have altered the author dates on these commits so that they show up in the easiest order to review them in. Note that in the "Futurize" commit I introduce the absolute import feature, but this is always removed in the next commit "replace relative imports".

amarkpayne:
Please note that all of the steps (PEP-8 reformatting, renaming local variables, futurize, replace relative imports) have been done for all files list here and with my name on them in the google doc. If you don't see a commit about it, it just means that the step was unnecessary for that file.

I have also separated the commits out by groups of files to make it easier to review. @alongd feel free to do what you want with the Arkane commits at the end of this PR.

alongd:
Thanks! The Arkane commits were cherry-picked and rebased on top of the Arkane Py3 PR

mliu49:
@alongd, can you confirm that it's safe to remove the last 5 commits from this branch?

alongd:
I confirm. They were implemented appropriately in the Arkane branch, split by content and squashed into existing commits.

rmgpy/rmg/outputTest.py
```diff
@@ -28,13 +28,12 @@
...
 
-from __future__ import absolute_import
 import os
 import shutil
 import unittest
 
-from .model import CoreEdgeReactionModel, ReactionModel
-from .output import *
+from rmgpy.rmg.model import CoreEdgeReactionModel, ReactionModel
+from rmgpy.rmg.output import *
```
mliu49: I think we would also like to replace star imports if possible.

rmgpy/__init__.py
```diff
@@ -116,8 +115,7 @@ def load(self, path=None):
             elif os.path.exists(os.path.join(working_dir, 'rmgrc')):
                 self.filename = os.path.join(working_dir, 'rmgrc')
             else:
-                return  # fail silently, instead of raising the following error:
-                raise SettingsError('Could not find an RMG settings file to load!')
+                return  # fail silently, instead of raising an error
```
mliu49: Do we want to fail silently? If someone were to delete the `rmgrc` file, would anything work?

environment_linux.yml
```diff
@@ -18,6 +18,7 @@ dependencies:
   - ffmpeg
   - gprof2dot
   - graphviz
+  - h5py
```
mliu49: Could you add it to `environment_py3.yml` as well?

rmg.py
```diff
@@ -51,12 +51,11 @@ def parse_command_line_arguments(command_line_args=None):
...
-    parser = argparse.ArgumentParser(description="""
-                                                 Reaction Mechanism Generator (RMG) is an automatic chemical reaction
-                                                 mechanism generator that constructs kinetic models composed of
-                                                 elementary chemical reaction steps using a general understanding of
-                                                 how molecules react.
-                                                 """)
+    parser = argparse.ArgumentParser(description='Reaction Mechanism Generator (RMG) is an automatic chemical reaction '
+                                                 'mechanism generator that constructs kinetic models composed of '
+                                                 'elementary chemical reaction steps using a general understanding of '
+                                                 'how molecules react.')
```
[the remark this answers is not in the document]
amarkpayne: Interesting--so I checked that the message looked the same from the command line (by running python rmg.py -h) before and after the change and I thought it was displaying the same way. If not though, we will have to get clever because apparently the `ArgumentParser` function removes this from the output. For example, check out the message before and after changing the description to the following:

```
    parser = argparse.ArgumentParser(description='Reaction Mechanism Generator (RMG) is an automatic chemical reaction\n\n\n\n\n'
                                                 'mechanism generator that constructs kinetic models composed of\n'
                                                 'elementary chemical reaction steps using a general understanding of\n'
                                                 'how molecules react.')
```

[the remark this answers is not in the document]
mliu49: Oh, that's interesting! Maybe argparse automatically formats the description so it looks nice. I didn't check the behavior before commenting.

In that case, we can leave it.

rmgpy/__init__.py
amarkpayne: So the comment associated with the last change is the following:

```
Quietly give up if can't find a settings file.  

The default one is empty anyway ;-) We may find a 
smarter way to do this, and indeed a better place to 
store settings, within anaconda apps.
```

The comment is correct--there is nothing but comments in the current rmgrc file, so by default we do not use this file to assign the correct path to the database directory and test_data directory. I have also confirmed that all of the unit tests pass if we delete this file.

How should we proceed? I think that we should either allow this to fail silently, or make it such that the user would have to specify the path to this file, as it is a rather niche feature, and we wouldn't want a settings file like this to be loaded unintentionally.

environment_linux.yml
amarkpayne: Done, thanks!

rmgpy/__init__.py
mliu49: I see. I think we should fail silently for now and rethink exactly what we want out of the config file in the future.

amarkpayne:
Thanks for all of the comments @mliu49! You were right about removing the * imports, we were getting imports in very strange ways. I have rebased this branch on the current py3 branch, let me know what you think.

rmgpy/rmg/inputTest.py
```diff
@@ -28,10 +28,11 @@
 #                                                                             #
 ###############################################################################
 
-import unittest 
+import unittest
 
-from rmgpy.rmg.main import RMG
 from rmgpy.rmg import input as inp
```
mliu49: What would be the difference between this and `import rmgpy.rmg.input as inp`?

rmgpy/rmg/main.py
```diff
@@ -228,13 +231,13 @@ def clear(self):
...
         self.execTime = []
-    
+
     def loadInput(self, path=None):
         """
         Load an RMG job from the input file located at `inputFile`, or
         from the `inputFile` attribute if not given as a parameter.
         """
-        from input import readInputFile
+        from .input import readInputFile
```
mliu49: Replace with absolute import.

rmgpy/rmg/main.py
```diff
@@ -253,24 +256,25 @@ def loadInput(self, path=None):
...
-            
+
     def loadThermoInput(self, path=None):
         """
         Load an Thermo Estimation job from a thermo input file located at `inputFile`, or
         from the `inputFile` attribute if not given as a parameter.
         """
-        from input import readThermoInputFile
-        if path is None: path = self.inputFile
+        from .input import readThermoInputFile
```
mliu49: Here also

rmgpy/rmg/main.py
```diff
@@ -302,114 +305,127 @@ def checkLibraries(self):
...
+                                                self.database.kinetics.libraries[libIter].name))
+
     def saveInput(self, path=None):
         """
         Save an RMG job to the input file located at `path`, or
         from the `outputFile` attribute if not given as a parameter.
         """
-        from input import saveInputFile
-        if path is None: path = self.outputFile
+        from .input import saveInputFile
```
mliu49: And here

rmgpy/rmg/model.py
```diff
@@ -673,29 +681,32 @@ def addNewSurfaceObjects(self,obj,newSurfaceSpecies,newSurfaceReactions,reaction
...
+        lost_surface_spcs = (surf_spcs - newSurfaceSpecies) | added_spcs
+        added_surface_spcs = newSurfaceSpecies - surf_spcs
+
+        self.newSurfaceSpcsAdd = self.newSurfaceSpcsAdd | added_surface_spcs
+        self.newSurfaceRxnsAdd = self.newSurfaceRxnsAdd | added_surface_rxns
+        self.newSurfaceSpcsLoss = self.newSurfaceSpcsLoss | lost_surface_spcs
+        self.newSurfaceRxnsLoss = self.newSurfaceRxnsLoss | lost_surface_rxns
+
+        return not (
+                    self.newSurfaceRxnsAdd != set() or self.newSurfaceRxnsLoss != set() or self.newSurfaceSpcsLoss != set() or self.newSurfaceSpcsAdd != set())
```
mliu49: I think this should either be one line or split after each `or`.

rmgpy/rmg/inputTest.py
amarkpayne: After doing some searching and some playing around in python, there is no difference between these two statements, and it is purely up to us as to which one we prefer. It is worth noting that if it wasn't for the `as inp` statements in each that there would be a slight difference, as `import rmgpy.rmg.input` would put rmgpy.rmg into the namespace, whereas `from rmgpy.rmg import input` would not. There is an even more interesting but subtle effect if we were trying to import a module level variable (see the second from the top response [here](https://stackoverflow.com/questions/710551/use-import-module-or-from-module-import?noredirect=1&lq=1)), but this does not apply in this case.

Given this, which one do we prefer?

rmgpy/rmg/main.py
amarkpayne: Thanks for the catch! I forgot to check import statements nested in the code. I did some further searching using regex in PyCharm and I didn't find any further instances in the files covered by this PR, so I think we have it all now.

rmgpy/rmg/main.py
amarkpayne: Done

rmgpy/rmg/main.py
amarkpayne: Done

rmgpy/rmg/model.py
amarkpayne: Done, I agree this is much better

amarkpayne:
Final note: the commit `Add missing InputError import in arkane/util.py` is no longer needed and will be dropped.

rmgpy/rmg/inputTest.py
mliu49: Good to know!

I have a slight preference for `import rmgpy.rmg.input as inp` since it's cleaner and seems a bit more straightforward. I always figured that from imports were mainly if you wanted something from within a file. Up to you whether or not to change it though.

mliu49:
I'm still a bit undecided on the best way to handle commits for this process. Perhaps we can discuss tomorrow.

amarkpayne:
I have squashed the commits and rebased on the current py3 branch

Issue #1703, opened 2019-08-23
Cannot convert str to rmgpy.kinetics.uncertainties.RateUncertainty when loading surface kinetics rules


rwest:
### Bug Description
Running a job with heterogeneous catalysis (surface) reaction families causes a crash when loading the database:

```
oup database from /Users/rwest/Code/Cat/RMG-database/input/statmech/groups...
Loading Platts additivity group database from /Users/rwest/Code/Cat/RMG-database/input/solvation/groups...
Trimolecular reactions are turned on
Adding rate rules from training set in kinetics families...
Traceback (most recent call last):
  File "rmg.py", line 185, in <module>
    main()
  File "rmg.py", line 179, in main
    rmg.execute(**kwargs)
  File "/Users/rwest/Code/Cat/RMG-Py/rmgpy/rmg/main.py", line 631, in execute
    self.initialize(**kwargs)
  File "/Users/rwest/Code/Cat/RMG-Py/rmgpy/rmg/main.py", line 472, in initialize
    self.loadDatabase()
  File "/Users/rwest/Code/Cat/RMG-Py/rmgpy/rmg/main.py", line 398, in loadDatabase
    family.addKineticsRulesFromTrainingSet(thermoDatabase=self.database.thermo)
  File "/Users/rwest/Code/Cat/RMG-Py/rmgpy/data/kinetics/family.py", line 1138, in addKineticsRulesFromTrainingSet
    data = deepcopy(entry.data)
  File "/Users/rwest/anaconda/envs/rmg241/lib/python2.7/copy.py", line 190, in deepcopy
    y = _reconstruct(x, rv, 1, memo)
  File "/Users/rwest/anaconda/envs/rmg241/lib/python2.7/copy.py", line 329, in _reconstruct
    y = callable(*args)
  File "rmgpy/kinetics/arrhenius.pyx", line 66, in rmgpy.kinetics.arrhenius.Arrhenius.__init__
    KineticsModel.__init__(self, Tmin=Tmin, Tmax=Tmax, Pmin=Pmin, Pmax=Pmax, uncertainty=uncertainty, comment=comment)
  File "rmgpy/kinetics/model.pyx", line 115, in rmgpy.kinetics.model.KineticsModel.__init__
    self.uncertainty = uncertainty
TypeError: Cannot convert str to rmgpy.kinetics.uncertainties.RateUncertainty
```

### How To Reproduce
Steps to reproduce the behavior.
```
[...]

PR #1705, opened 2019-08-23
Fix a few bugs and add tests: KineticsModel and uncertainty handling.

rwest:
### Motivation or Problem
Issue #1703 describes a problem whereby `SurfaceArrhenius` objects can't be copied with `deepcopy` (meaning they can't be loaded in a kinetics family, or pretty much used for anything).
It was due to the new `uncertainty` attribute being left out of the `__repr__` or `__reduce__` methods.
While fixing and debugging, and adding unit tests, I caught some other bugs.
The `KineticsModel` had some errors, also in the `__repr__` method, and the logic behind the isIdenticalTo() was broken and it'd return False when it should be True (which was causing the new unit tests I wrote for the surface kinetics module to fail).

### Description of Changes
Added some unit tests that initially failed, and then fixed the bugs that they demonstrated.

### Testing
It was test-driven development. A test I wrote to isolate one bug in fact led to finding two more.
Tested locally. Running Travis now. Test coverage around these modules is still pretty low, and even when lines of code are executed, the logic is not always being tested that the lines of code do the right thing. But improving that situation should wait until after Py3!

### DIFF
--- a/rmgpy/kinetics/model.pyx
+++ b/rmgpy/kinetics/model.pyx
@@ -227,15 +227,15 @@ cdef class KineticsModel:
         Returns ``True`` if Tmin, Tmax for both objects match.
         Otherwise returns ``False``
         """
-        if self.Tmin is not None and otherKinetics.Tmin is not None and not self.Tmin.equals(otherKinetics.Tmin):
-            return False
+        if self.Tmin is not None and otherKinetics.Tmin is not None and self.Tmin.equals(otherKinetics.Tmin):
+            pass
         elif self.Tmin is None and otherKinetics.Tmin is None:
             pass
         else:
             return False
 
-        if self.Tmax is not None and otherKinetics.Tmax is not None and not self.Tmax.equals(otherKinetics.Tmax):
-            return False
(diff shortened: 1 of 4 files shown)


mliu49:
I went ahead and rebased this on top of master. I also incorporated futurization and RMG 3 changes into the relevant commits. I think this looks good, but perhaps we should have a second reviewer?

rwest:
Thanks for rebasing, Max.  A quick glance looks good to me (I have no time for a thorough review).

xiaoruiDong:
Thanks for the contribution, Dr. West. Thanks for the rebasing, Max. I went through the codes and they looked good to me. I ran the RMG-test and all tests related to this PR passed. I also implemented PEP-8 formatting to this PR to be consistent with other RMG code.

xiaoruiDong:
Corrections for `namespace` has been made. Sorry for my ignorant.

mliu49:
@xiaoruiDong, could you take a final look and approve? I added another commit fixing an issue with `Chebyshev.is_identical_to` which was uncovered by the fix to `KineticsModel.is_identical_to`.

PR #1714, opened 2019-08-29
Modified the ArkaneSpecies.update_xyz_string() method

alongd:
Now outputs [a standardized form of the xyz](https://en.wikipedia.org/wiki/XYZ_file_format)
Test extended accordingly

Addresses issue 1 reported in #1713 

I think this PR should be merged into the py3 branch before #1713, since it doesn't break Py2 compatibility.

### DIFF
--- a/arkane/common.py
+++ b/arkane/common.py
@@ -42,7 +42,7 @@ import yaml
 
 import rmgpy.constants as constants
 from rmgpy import __version__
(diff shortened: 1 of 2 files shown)


alongd:
Something's messed up with the coverage report. I don't think this small PR bumped the coverage by 3%...

PR #1721, opened 2019-09-13
Reduce the amount of logging messages that get too frequently printed during debug

alongd:
### Motivation or Problem
When debugging RMG or a software that depends on RMG the amount of debugging statements that get printed jam the buffer, not allowing the developer to conveniently see the actual original error massage.

### Description of Changes
Some logging statements, most of which are debugging level, were commented out. I only marked those which get printed too frequently in my experience. The purpose here was to flag them in the code for a discussion of how to best deal with them - leave commented out (so can be conveniently brought back to life if needed) or completely deleting them.
I think we can start the discussion even though we won't merge until the Py3 transition, since the same logging statements will be relevant after transitioning.

### DIFF
--- a/rmgpy/constraints.py
+++ b/rmgpy/constraints.py
@@ -90,7 +90,7 @@ def fails_species_constraints(species):
 
     max_radicals = species_constraints.get('maximumRadicalElectrons', -1)
     if max_radicals != -1:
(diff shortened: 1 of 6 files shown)


alongd:
@amarkpayne, this PR should be ready for review now

Issue #785, opened 2016-10-13
transitioning to python 3


goldmanm:
Python3 is likely where new developing for python is occurring.  After some discussion of different ideas and issues we thought there are kind of two parts to the transition.
1. writing test cases and other changes that improve current RMG functionality (which also aid in transitioning to python 3)
2. applying other changes to RMG that do not provide direct benefit to current RMG functioning, like `2to3` or upgrading dependencies. 

@rspeth, we were wondering how much effort was involved in upgrading Cantera to python3 (and maintaining a python2 version) and if you have any other wisdom.

goldmanm:
## Here is information from our discussion:

### many people think python 3 'is' the future of python

more intuitive feel

* `1 / 2 = 0.5` not `1 / 2 = 0`
* Unicode support is default
* `print('this is now a method')`
* input/output classes streamlined 
* generally more coherant programing experience
	
no more bug fixes provided for 2.X after 2020

any new updates applied only to 3.X since 2013

### converting is difficult

version 3 not backwards compatible

* some programs exist to 'help' but still take training
	
need all dependencies compatible

**no direct research value**

### the conversion process has many helpful tools

1. ensure good test coverage of code with `coverage.py`
1. ensure dependency compatibility with `caniusepython3`
1. rewrite [ambiguous parts of code](http://python3porting.com/preparing.html). `__future__` imports help 
1. use porting converters like `2to3` for the rest
1. touch up failing test cases

**much easier than rewriting from java**



### RMG-Py currently has 44% of its code tested

You can run `coverage.py` with command line `make test`. Then open `testing/coverage/*.html` to see methods aren't tested. 

`coverage` looks at how many lines of executable code are called when all the test methods are run.

Recent master branch has 44% coverage. ['good' coverage is around 80%](https://docs.python.org/2.7/howto/pyporting.html#have-good-test-coverage).

The amount of code called when running minimal is 21%,


**Warning**: Website has little to no tests.


### areas of RMG with lower testing

* uncertainty analysis ~ 10%
* pressure dependence ~ 10%
* non-canterm stat mech ~ 15%
* quantum mechanics ~ 20%
* diffusion limited kinetics ~ 25%
* kinetics groups and libraries ~ 20%
* the main RMG class ~ 30%
* the main RMG database class ~ 40%

### examples of critical, untested methods

* rmg.main.RMG
	* initializeLog()
	* execute()
	* register_listeners()
	* saveInput()
	* checkInput()
[...]

PR #1724, opened 2019-09-23
RMG futurization: Python 3 compatibility and PEP-8 compliance

mliu49:
### Motivation or Problem
Due to the imminent end of Python 2 support, it is important that RMG is updated to be compatible with Python 3.

### Description of Changes
This PR is the culmination of almost six week's work by me, @alongd, @amarkpayne, @mjohnson541, and @xiaoruiDong in futurizing RMG-Py and Arkane.

Main outcomes:
- Python 3 compatibility (and loss of Python 2 compatibility)
- PEP-8 name changes (camelCase to lowercase_with_underscores)
- PEP-8 code style changes

This PR also includes the removal of all remaining scoop functionality, given it has been non-functional since parallel processing has been refactored to use Python multiprocessing, and the reduction module, which was heavily tied into the scoop framework and therefore also non-functional.

### Merging
The standard test suite and RMG-tests should test most of the key functionality. Since the individual components of this branch have been previously reviewed, I plan on merging this directly once tests pass.

### Next Steps
More detailed testing is definitely needed, but my proposed plan is to do that after merging this into master. The main motivation behind that is so we can switch back to the standard contributor workflow as soon as possible, and allow everyone to begin updating existing PRs and more easily make new PRs for fixing any remaining bugs.


mliu49:
I added a commit implementing a check for the Python version to make it obvious to users if they need to update their environment. Otherwise, I think compiling or running RMG would lead to a non-obvious SyntaxError. Could someone take a look?

mliu49:
Closes #785 🎉

Makefile
```diff
@@ -4,9 +4,9 @@
 #
 ################################################################################
 
-.PHONY : all minimal main solver check cantherm clean install decython documentation mopac_travis
+.PHONY : all minimal main solver check pycheck arkane clean install decython documentation mopac_travis
```
amarkpayne: (Outside the scope of this PR) Are our .PHONY targets up to date? Should things like `test` and `eg0` be in there?

PR #1729, opened 2019-09-26
Remove unnecessary zip folder unpacking in arkane/mainTest.py

mliu49:
### Motivation or Problem
CH2CHOOHscans was previously being unpacked into the current working directory and was not being cleaned up.

### Description of Changes
- Add some variables to improve readability of path generation
- Specify the directory to which CH2CHOOHscans should be unpacked
- Check whether q2dtor is installed before unpacking scans and running the q2dtor example

### Testing
Should be sufficient for Travis to pass.

### Reviewer Question
While I was testing this, I discovered that this test doesn't seem to actually run q2dtor. My guess is that it's because the `r0.pes` file exists (since that's the only difference between this example and the test in ndTorsionsTest).

Was this intentional? If so, I will remove the changes that check that q2dtor is installed. If not, we should fix the example.


mliu49:
This has been completely changed. We concluded that the scans were not needed at all, so this PR now simply removes the part where we extract the zip file.

PR #1730, opened 2019-09-26
Remove future imports and py2/py3 compatibility imports

mliu49:
### Motivation or Problem
Since we are already incompatible with Python 2, we should remove future imports.

### Description of Changes
Remove all `__future__` imports and some try/except imports for renamed modules.

### Testing
Check that tests pass.


xiaoruiDong:
@mliu49 Thanks Max! The PR looks great. I searched the keyword 'future' in this PR, and realized the `future` package is remained in the `environment.yml`. Do you agree we should also remove that as well?

mliu49:
Thanks for pointing that out! I don't think we need it anymore, as we already removed the imports which we were using from it. I added another commit removing it.

PR #1731, opened 2019-09-26
Improve group additivity ranking for aromatic species

mliu49:
### Motivation or Problem
Estimating thermo for aromatic species using group additivity can sometimes give weird results because we will get an estimate for every resonance structure and then take the one with the lowest H298. In some cases, this can lead to severe underprediction of enthalpy if there is an inappropriate group which results in a lower enthalpy than a more appropriate group.

For example:

image

This species has resonance structures where the radical can delocalize into the aromatic ring. One of those structures gives the estimate with the lowest H298 of 61.4 kcal/mol:
`Thermo group additivity estimation: group(Cs-(Cds-Cds)CbHH) + group(Cb-Cs) + group(Cb-(Cds-Cds)) + group(Cds-Cds(Cds-Cds)Cs) + group(Cb-H) + group(Cds-CdsCbH) + group(Cb-H) + group(Cds-Cds(Cds-Cds)H) + group(Cb-H) + group(Cb-H) + group(Cds-CdsHH) + group(Cdd-CdsCds) + polycyclic(s2_6_6_ben_ene_1) + radical(Benzyl_S_dihydronaphthalene) `

However the calculated H298 of the molecule is 101.7 kcal/mol. If we were to use GAV to estimate the thermo of resonance structure depicted above, then we would get 107.3 kcal/mol with the following groups:
`Thermo group additivity estimation: group(Cbf-CbCbCbf) + group(Cbf-CbCbCbf) + group(Cb-(Cds-Cds)) + group(Cb-H) + group(Cb-H) + group(Cb-H) + group(Cb-H) + group(Cb-H) + group(Cb-H) + group(Cb-H) + group(Cds-CdsCbH) + group(Cds-CdsHH) + polycyclic(s2_6_6_naphthalene) + radical(Cds_S)`

### Description of Changes
This PR adds an additional parameter, number of aromatic rings, by which to rank GAV estimates, in addition to H298.

The number of aromatic rings is determined by the actual bond orders in the resonance structure, rather than true aromaticity detection.

The result is that estimates for resonance structures with more aromatic rings is prioritized over those with fewer, even if it results in a higher H298.

### Testing
Besides testing with the species mentioned above, I ran comparisons against the SABIC_aromatics_1dHR library (not currently on master), which contains ~350 aromatic/near-aromatic species calculated using CBS-QB3 with 1D hindered rotors.

The H298 MAE on master is 5.263 kcal/mol, while the MAE on this branch is 5.228 kcal/mol. This is not a substantial difference, mainly because this change does not affect many species. However, it does suggest that this change does not have noticeable negative effects.

### DIFF
--- a/rmgpy/molecule/molecule.pxd
+++ b/rmgpy/molecule/molecule.pxd
@@ -248,6 +248,8 @@ cdef class Molecule(Graph):
 
     cpdef identify_ring_membership(self)
 
+    cpdef int count_aromatic_rings(self)
+
     cpdef tuple get_aromatic_rings(self, list rings=?)
 
     cpdef list get_deterministic_sssr(self)
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -1797,17 +1797,17 @@ class ThermoDatabase(object):
                     # Group values are generally fitted to the most aromatic resonance structure
                     num_arom_rings = species.molecule[i].count_aromatic_rings()
 
-                    entries.append((thermo, sum_rank, -num_arom_rings))
+                    entries.append((i, thermo, sum_rank, -num_arom_rings))
 
                 # Sort first by number of aromatic rings, then rank, then by enthalpy at 298 K
-                entries = sorted(entries, key=lambda entry: (entry[2], entry[1], entry[0].get_enthalpy(298.)))
-                indices = [thermo_data_list.index(entry[0]) for entry in entries]
+                entries.sort(key=lambda entry: (entry[3], entry[2], entry[1].get_enthalpy(298.)))
+                indices = [entry[0] for entry in entries]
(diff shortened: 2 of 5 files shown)


rmgpy/molecule/molecule.pxd
```diff
@@ -248,6 +248,8 @@ cdef class Molecule(Graph):
 
     cpdef identify_ring_membership(self)
 
+    cpdef int count_aromatic_rings(self)
```
alongd: Could you change the commit message to lower_case_with_underscores to be consistent with the code?
(same comment for the `Minor code improvements to ThermoDatabase.prioritizeThermo` commit)

mliu49: Done!

PR #1732, opened 2019-09-27
Limit aromaticity perception for strained polycyclics

mliu49:
### Motivation or Problem
This implements an additional heuristic in aromaticity perception which limits the types of molecules which are considered aromatic. Generally, aromaticity requires that a molecule be planar, in order to enable the sharing of pi-electrons across the molecule. (Some amount of curvature is acceptable, such as with corannulene or bucky-balls.)

If there is substantial strain in a molecule, it can negate the potential effects of aromaticity. For example,

image

is quite unstable despite having the correct number of pi-electrons to suggest aromaticity.

### Description of Changes
The new heuristic checks for sharing of more than 2 atoms between two rings, which will always break planarity. In such cases, RMG will no longer consider either ring to be aromatic, even if Huckel's rule is met.

The practical effect of doing so is that those rings will be represented using single/double bonds and aliphatic carbon atom types, which will affect thermo and kinetic estimation.

For the example above, the previous best resonance structure was

image

with a GAV estimated H298 of 359.4 kJ/mol.

With these changes, the new best resonance structure is

image

with a GAV estimated H298 of 397.6 kJ/mol.

Unfortunately, this is still quite far from the quantum result of 744.6 kJ/mol using M06-2X/cc-pVTZ, but is at least in the right direction.

Previously, a major concern with this change was the possibility of having duplicate species due to not having a "canonical" resonance structure. For aromatics, the fully aromatic structure is the most reliable for identifying identical species because it is always consistent., while structures with double bonds may have multiple possible double bond arrangements. However, that is no longer an issue since the implementation of the resonance-agnostic isomorphism method.

Now, I see this mainly as improving the accuracy of resonance structure representations, with a mild effect of shifting thermo and kinetics estimates for these strained species towards aliphatic behavior.

### Testing
I have been running jobs with this change for a very long time. It doesn't substantially affect models, but does seem to slightly reduce the appearance of weird aromatic species.

### Reviewing
Not sure who to assign, so if anyone would like to volunteer, that would be appreciated :)

### DIFF
--- a/rmgpy/molecule/molecule.py
+++ b/rmgpy/molecule/molecule.py
@@ -2437,7 +2437,35 @@ class Molecule(Graph):
 
         if rings is None:
             rings = self.get_relevant_cycles()
-            rings = [ring for ring in rings if len(ring) == 6]
+
+        def filter_fused_rings(_rings):
+            """
+            Given a list of rings, remove ones which share more than 2 atoms.
+            """
+            cython.declare(toRemove=set, i=cython.int, j=cython.int, toRemoveSorted=list)
+
+            if len(_rings) < 2:
+                return _rings
+
+            to_remove = set()
+            for i, j in itertools.combinations(range(len(_rings)), 2):
+                if len(set(_rings[i]) & set(_rings[j])) > 2:
+                    to_remove.add(i)
+                    to_remove.add(j)
+
+            to_remove_sorted = sorted(to_remove, reverse=True)
+
+            for i in to_remove_sorted:
+                del _rings[i]
+
+            return _rings
+
+        # Remove rings that share more than 3 atoms, since they cannot be planar
+        rings = filter_fused_rings(rings)
+
+        # Only keep rings with exactly 6 atoms, since RMG can only handle aromatic benzene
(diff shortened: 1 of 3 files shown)


JacksonBurns:
Changing base and rebasing.

@mliu49 is this PR still relevant?

mliu49:
Thanks for updating the branch! I think this is still relevant and worth merging.

It was waiting for review and then I was supposed to rebase the branch, but I got busy with things and forgot...

JacksonBurns:
> Thanks for updating the branch! I think this is still relevant and worth merging.
> 
> It was waiting for review and then I was supposed to rebase the branch, but I got busy with things and forgot...

Sure thing! In that case let's get a review on this (assuming the tests pass) and get it merged.

JacksonBurns:
@mliu49 I think this is pretty close to ready - the failing unit tests seem like they are only failing because this PR is an improvement. Could you update this branch, check those tests, and update them if that's the case?

JacksonBurns:
@rwest @hwpang this looks ready to go AFAIK - need input on the scientific side.

rwest (approved):
Looks good to me

PR #1752, opened 2019-10-04
Fix two calls to Exception.message, which no longer exists

mliu49:
### Motivation or Problem
Exceptions no longer have a message attribute. We fixed a bunch of cases already, but missed some.

### Description of Changes
Replace calls to Exception.message with str(Exception).

### Testing
These changes fixed the issues I was having. I don't know if it's necessary to do additional testing.


ajocher:
Thanks for fixing. This looks fine to me and ready to be merged.

mliu49:
Thanks!

PR #1754, opened 2019-10-08
Update file headers and shebangs

mliu49:
### Motivation or Problem
Now that we're using Python 3, the recommended shebang has changed. Since these files are not Python 2 compatible, it is recommended to explicitly refer to `python3` in the shebang. Additionally, utf-8 encoding is now default in Python 3, so the encoding line is no longer necessary.

While the shebang is technically unnecessary if the file is not intended to be executable, I think it's still useful to keep around to indicate that these are Python 3 files.

### Description of Changes
Update the `update_headers` function in `utilities.py` and update all file headers.


alongd:
Perhaps we'd like to do the same in the -db repo? Current .py files there say `#!/usr/bin/env python`

mliu49:
Yes, we should also do this for the database repo. It can be included in the (currently non-existent) PR updating scripts to work with Python 3.

PR #1756, opened 2019-10-09
Fix conda build issues

amarkpayne:
### Motivation or Problem
In order to build the conda package for RMG-Py 3.0, there are a couple of dependencies that needed to be updated in the conda recipe, as well as a bug or two that needed fixing. While this PR does not update the version numbers, it is possible to build a binary from this branch with these changes **so long as a python 3 version of the database exists as well**.

### Description of Changes

- Update dependencies in the conda recipe
- Fix a bug in setup.py that prevented `python setup.py install` from working (by calling the solver check that has now been moved to utilities.py)
- Prevent throwing an error if utilities.py is not available, which is the case for the binary builds.

### Testing
I built the RMG 3.0.0b successfully from this branch, so everything should be in order.

### DIFF
--- a/rmg.py
+++ b/rmg.py
@@ -35,8 +35,14 @@ import os.path
 import logging
 
 # Before importing any RMG modules, check Python version
-import utilities
-utilities.check_python()
+try:
+    import utilities
+except ImportError:  # This is likely the binary install version, which does not include utilities
+    # If this is in fact the binary version, conda has ensured that the python version is correct.
+    # It is okay to skip this test here
+    utilities = None
+else:
+    utilities.check_python()
 
 import rmgpy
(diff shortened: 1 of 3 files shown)


rmg.py
```diff
@@ -35,8 +35,14 @@
...
-import utilities
-utilities.check_python()
+try:
+    import utilities
+    utilities.check_python()
+
+except ImportError:  # This is likely the binary install version, which does not include utilities
+    # If this is in fact the binary version, conda has ensured that the python version is correct.
+    # It is okay to skip this test here
+    pass
```
mliu49: Perhaps we should put `utilities.check_python()` in the else clause?

amarkpayne: I agree, this is safer. I have made the change, do you want me to push the change now?

PR #1735, opened 2019-09-30
Update IPython notebooks

mliu49:
### Motivation or Problem
This PR updates the notebooks located in `RMG-Py/ipython` to work with Python 3 and RMG API changes.

The only exception is `globalUncertainty.ipynb`, which will be updated along with the MUQ changes in a separate PR.

### Description of Changes
In the process of updating the notebooks, a few bugs were also discovered and fixed.

The parse_source example model also had to be updated to reflect RMG-database changes (in particular, the replacement of the R_Recombination tree with auto-generated groups).

### Testing
I have run all of the notebooks locally. To confirm, the reviewer can do the same. All of the notebooks can be executed in their entirety without changing any user inputs, and should run without any errors.

PR #1760, opened 2019-10-11
BugFix: file_format in ReactionDrawer

alongd:
These are leftovers from the PEP8 conversions where the reserved keyword `format` was replaced with `file_format` in draw.py.
This fixed was verified to function correctly (within ARC).


mliu49:
This fix is already part of #1735.

alongd:
Great! I'll go ahead and close this PR.

Issue #1762, opened 2019-10-11
RMG parallel instructions for job schedulers


cgrambow:
This is not a super big issue and you can decide if it warrants changing or not. The [RMG parallel instructions for SGE](https://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/running.html?highlight=parallel) explicitly list `-pe singlenode 24`. However, `singlenode` is a configuration that exists solely on Pharos and is not specific to SGE, so these instructions won't work for anyone not in our group and should probably be removed.

In fact, I would propose not mentioning SLURM and SGE at all because everyone will have a different set-up and running RMG in parallel does not require a job scheduler (correct me if I'm wrong).

PR #1763, opened 2019-10-11
Removing RMG job scheduler instructions and updating RMG-parallel run…

ajocher:
… instructions.

### Motivation and Changes
This pull request is in response to #1762. Keeping in mind that people outside of the group will not benefit from group specific job scheduler instructions they were removed completely as suggested by @cgrambow. Instructions on how to run an RMG job in parallel have been added to '6. Running a Job' instead.  

### Testing
I tested the updated instructions for an RMG parallel run with the RMG minimal example and 3 processes.

### Reviewer Tips
Confirm that everything that was deleted should be deleted. In addition, you could also run a parallel run with the updated instructions.

### DIFF
--- a/documentation/source/users/rmg/running.rst
+++ b/documentation/source/users/rmg/running.rst
@@ -20,83 +20,17 @@ Run with CPU profiling::
 
     python rmg.py input.py -p
 
-We recommend you make a job-specific directory for each RMG simulation. Some jobs can take quite a while to complete, so we also recommend using a job scheduler if working in an linux environment. 
+Run with multiprocessing for reaction generation and QMTP::
 
-The instructions below describe special cases for running an RMG job.
+    python rmg.py -n <Max number of processes allowed> input.py 
 
-Running RMG in parallel with SLURM
-----------------------------------
-
-RMG has the option to use multiple processes on one node for reaction generation and on-the-fly Quantum Mechanics Thermodynamic Property (QMTP) calculation. Here is an example submission script for an RMG-Py job with a SLURM scheduler.
-
-The job reserves 24 tasks on a single node, but uses only 12 processes in parallel during
-the RMG-Py simulation.
-
-Make sure that: 
-
-- the queue named ``debug`` exists on your SLURM scheduler. 
-- you modify the path to the parent folder of the RMG-Py installation folder.
-- you have an anaconda environment named ``rmg_env`` that contains RMG-Py's dependencies.
-- the working directory from which you launched the job contains the RMG-Py input file ``input.py``
-
-.. code:: bash
-
-    #!/bin/bash
-
-    #SBATCH -p debug
-    #SBATCH -J jobname
-    #SBATCH -n 24
-
-    Processes=12
-    RMG_WS=/path/to/RMG/parent/folder
-    export PYTHONPATH=$PYTHONPATH:$RMG_WS/RMG-Py/
-
-    source activate rmg_env
-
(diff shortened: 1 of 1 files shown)


documentation/source/users/rmg/running.rst
```diff
@@ -20,83 +20,17 @@ Run with CPU profiling::
...
-Make sure that: 
-
-- the queue named ``debug`` exists on your SLURM scheduler. 
-- you modify the path to the parent folder of the RMG-Py installation folder.
-- you have an anaconda environment named ``rmg_env`` that contains RMG-Py's dependencies.
-- the working directory from which you launched the job contains the RMG-Py input file ``input.py``
+Run with multiprocessing for reaction generation and QMTP::
 
-.. code:: bash
+    python rmg.py input.py -n <Max number of processes allowed>
```
cgrambow: Maybe this is too nitpicky, but I prefer having the optional argument in front of the positional one. I.e., `python rmg.py -n nproc input.py`.

documentation/source/users/rmg/running.rst
```diff
@@ -20,83 +20,17 @@ Run with CPU profiling::
...
-    Processes=12
-    RMG_WS=/path/to/RMG/parent/folder
-    export PYTHONPATH=$PYTHONPATH:$RMG_WS/RMG-Py/
-
-    source activate rmg_env
-
-    python $RMG_WS/RMG-Py/rmg.py -n $Processes input.py
-
-    source deactivate
+We recommend you make a job-specific directory for each RMG simulation. Some jobs can take quite a while to complete, so we also recommend using a job scheduler if working in an linux environment.
```
cgrambow: an -> a, linux -> Linux

documentation/source/users/rmg/running.rst
```diff
@@ -20,83 +20,17 @@ Run with CPU profiling::
...
 
 Currently, multiprocessing is implemented for reaction generation and the generation of QMfiles when using the QMTP option to compute thermodynamic properties of species. The processes are spawned and closed within each function. The number of processes is determined based on the ratio of currently available RAM and currently used RAM. The user can input the maximum number of allowed processes from the command line. For each reaction generation or QMTP call the number of processes will be the minimum value of either the number of allowed processes due to user input or the value obtained by the RAM ratio. The RAM limitation is employed, because multiprocessing is forking the base process and the memory limit (SWAP + RAM) might be exceeded when using too many processors for a base process large in memory.
 
 In python 3.4 new forking contexts 'spawn' and 'forkserver' are available. These methods will create new processes which share nothing or limited state with the parent and all memory passing is explicit. Once RMG is transferred to python 3 it is recommended to use the spawn or forkserver forking context to potentially allow for an increased number of processes.
```
cgrambow: Now that we're on Python 3, is this paragraph still relevant? Or should it be updated to say this will be implemented in the future?

ajocher: @kspieks and I were planning to work on this today afternoon. By then we should have a better idea if it works. I would like to keep it as is until we know more.

documentation/source/users/rmg/running.rst
ajocher: Will do

documentation/source/users/rmg/running.rst
ajocher: done

cgrambow:
Looks good. Before I merge, you can squash.

Issue #1768, opened 2019-10-16
Arkane's output.py file should be restructured


alongd:
We all agree that the Arkane output.py file should be restructured. One possibility for a format change, as suggested by @amarkpayne, could be outputting a nice HTML instead.

We plan to output RMG libraries in Arkane (thermo + kinetics), which makes the thermo and kinetics blocks outputted to this file unnecessary, and also confusing.

Also, the current `conformer` block is generated using Abstract Syntax Trees, which is hard to maintain, perhaps we could replace it with a simpler code.

PR #1769, opened 2019-10-17
Output RMG libraries in Arkane

alongd:
### Motivation or Problem
The current Arkane output is not in a format that can be directly pasted into RMG libraries. See #1768 

### Description of Changes
Now Arkane will output thermo and kinetic libraries into an RMG_libraries folder. Note that only species with structure will be considered for the thermo library, and only reactions with structures for all their reactants and products will be considered for the kinetics libarry.

### Testing
Added a unit test for an intermediate newly added function, could add functional tests as well.

### Reviewer Tips
Run a thermo job, specifying structure for a species, and see the resulting thermo library.
Run kinetics and pdep jobs, specifying structures for all species, and see the resulting kinetics library.

### DIFF
--- a/arkane/input.py
+++ b/arkane/input.py
@@ -645,7 +645,7 @@ def load_input_file(path):
             if atom_energies is not None:
                 job.arkane_species.atom_energies = atom_energies
 
-    return job_list, reaction_dict, species_dict, transition_state_dict, network_dict
+    return job_list, reaction_dict, species_dict, transition_state_dict, network_dict, model_chemistry
 
 
 def process_model_chemistry(model_chemistry):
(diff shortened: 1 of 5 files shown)


arkane/statmech.py
```diff
@@ -1103,6 +1103,7 @@ def assign_frequency_scale_factor(freq_level):
                  'ccsd(t)-f12/aug-cc-pvtz': 0.998,  # [3], taken as CCSD(T)-F12a/cc-pVTZ-F12
                  'ccsd(t)-f12/aug-cc-pvqz': 0.998,  # [3], taken as 'CCSD(T)-F12b/VQZF12//CCSD(T)-F12a/TZF'
                  }
+    freq_level.replace('**', '(d,p)')
```
mjohnson541: I don't think I understand, why do we want to use BAC from one basis set and freq scaling factor from a different basis?

alongd:
I removed the `b3lyp/6-31**` commit, I think I haven't explained myself clearly. I'll make a new PR for that. Rebased.

PR #1773, opened 2019-10-18
Fix solvaion radical bug

### LINKED ISSUE #1743: Abraham GAV for estimating dG for radicals is malfunctioning
### Bug Description
RMG applies heuristics to estimate deltaG of solvation for radicals based on Abraham parameters of their saturated counterparts. If working properly, the Abraham parameter A for species CCO and CC[O] should be different. However, RMG gives the same set of parameters for these two species, suggesting a possible bug. 

### How To Reproduce
Run Abraham GAV estimate for species CCO and CC[O], and compare results. 

### Expected Behavior
Different Abraham A parameter for the two species mentioned.

### Installation Information
 - OS (include version if known): (e.g. Windows 10, Ubuntu 18, macOS 10.13, etc.)
 - Installation method: (e.g. installation from binary or source, with or without anaconda)
 - RMG version information:
   
   - RMG-Py: master
   - RMG-database: master

### Additional Context
I started looking at resolving the bug, but need time and help to resolve the issue.

yunsiechung:
### Motivation or Problem
Solvation correction for radical groups were not applied for radical species even if their radical groups were found in the solvation radical database.

### Description of Changes
Before, when the radical solvation correction would apply, it applied the correction for each individual radical electron found by using such for loop:
`for electron in range(1, atom.radical_electrons):`
But this was wrong, because the range should start from 0, not 1. If the number of radical electron is 1, then the range is from (1,1), which doesn't really start the for loop at all.
I changed this for loop to:
`for electron in range(atom.radical_electrons):`
such that if the number of electron is 1, the range will be from [0, 1], making the correction to be applied once. Since the number of radical electrons are always 0 or positive integer, this change should be appropriate.

### Testing
I also added a unit test in solvationTest.py, which checks whether the solvation radical group is found for the radical speices '[OH]' when calculating its solute data from group additivity method.

### Reviewer Tips
fixes #1743

### DIFF
--- a/rmgpy/data/solvationTest.py
+++ b/rmgpy/data/solvationTest.py
@@ -193,6 +193,12 @@ class TestSoluteDatabase(TestCase):
         solute_data = self.database.get_solute_data_from_groups(species)
         self.assertIsNotNone(solute_data)
 
+    def test_radical_solute_group(self):
+        """Test that the existing radical group is found for the radical species when using group additivity"""
+        species = Species(molecule=[Molecule(smiles='[OH]')])
+        solute_data = self.database.get_solute_data_from_groups(species)
+        self.assertTrue('radical' in solute_data.comment)
+
     def test_correction_generation(self):
         """Test we can estimate solvation thermochemistry."""
         self.testCases = [
--- a/rmgpy/data/solvationTest.py
+++ b/rmgpy/data/solvationTest.py
@@ -195,9 +195,19 @@ class TestSoluteDatabase(TestCase):
 
     def test_radical_solute_group(self):
         """Test that the existing radical group is found for the radical species when using group additivity"""
-        species = Species(molecule=[Molecule(smiles='[OH]')])
-        solute_data = self.database.get_solute_data_from_groups(species)
-        self.assertTrue('radical' in solute_data.comment)
+        # First check whether the radical group is found for the radical species
+        rad_species = Species(smiles='[OH]')
+        rad_solute_data = self.database.get_solute_data_from_groups(rad_species)
+        self.assertTrue('radical' in rad_solute_data.comment)
+        # Then check that the radical and its saturated species give different solvation free energies
+        saturated_struct = rad_species.molecule[0].copy(deep=True)
+        saturated_struct.saturate_radicals()
+        sat_species = Species(molecule=[saturated_struct])
+        sat_solute_data = self.database.get_solute_data_from_groups(sat_species)
+        solvent_data = self.database.get_solvent_data('water')
+        rad_solvation_correction = self.database.get_solvation_correction(rad_solute_data, solvent_data)
+        sat_solvation_correction = self.database.get_solvation_correction(sat_solute_data, solvent_data)
+        self.assertNotAlmostEqual(rad_solvation_correction.gibbs / 1000, sat_solvation_correction.gibbs / 1000)
 
     def test_correction_generation(self):
         """Test we can estimate solvation thermochemistry."""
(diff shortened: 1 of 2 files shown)


rmgpy/data/solvationTest.py
```diff
@@ -193,6 +193,12 @@ def test_radical_and_lone_pair_generation(self):
         solute_data = self.database.get_solute_data_from_groups(species)
         self.assertIsNotNone(solute_data)
 
+    def test_radical_solute_group(self):
+        """Test that the existing radical group is found for the radical species when using group additivity"""
+        species = Species(molecule=[Molecule(smiles='[OH]')])
+        solute_data = self.database.get_solute_data_from_groups(species)
+        self.assertTrue('radical' in solute_data.comment)
+
```
oscarwumit: Just a thought, should we add a test to check if we get the same (we shouldn't) Abraham solute parameters for OH radical and water? If we get the same values then we know something is wrong.

oscarwumit:
Thanks for the PR. I add a minor comment.

rmgpy/data/solvationTest.py
yunsiechung: I could add that test. I will just add an additional test for checking whether their A parameters are different

mliu49:
What's the status on this PR?

yunsiechung:
@mliu49 I'll make the change and force-push this PR again!!

yunsiechung:
@oscarwumit I added another unittest that you suggested before. Could you review it and approve it if it's okay?

rmgpy/data/solvationTest.py
```diff
@@ -193,6 +193,22 @@ def test_radical_and_lone_pair_generation(self):
...
+        self.assertTrue('radical' in rad_solute_data.comment)
+        # Then check that the radical and its saturated species give different solvation free energies
+        saturated_struct = rad_species.molecule[0].copy(deep=True)
+        saturated_struct.saturate_radicals()
+        sat_species = Species(molecule=[saturated_struct])
+        sat_solute_data = self.database.get_solute_data_from_groups(sat_species)
+        solvent_data = self.database.get_solvent_data('water')
+        rad_solvation_correction = self.database.get_solvation_correction(rad_solute_data, solvent_data)
+        sat_solvation_correction = self.database.get_solvation_correction(sat_solute_data, solvent_data)
+        self.assertNotEqual(rad_solvation_correction.gibbs / 1000, sat_solvation_correction.gibbs / 1000)
```
mliu49: You may want to use `assertNotAlmostEqual` here since you're comparing (I assume) floats which have gone through some calculations.

rmgpy/data/solvationTest.py
```diff
@@ -193,6 +193,22 @@ def test_radical_and_lone_pair_generation(self):
         solute_data = self.database.get_solute_data_from_groups(species)
         self.assertIsNotNone(solute_data)
 
+    def test_radical_solute_group(self):
+        """Test that the existing radical group is found for the radical species when using group additivity"""
+        # First check whether the radical group is found for the radical species
+        rad_species = Species(molecule=[Molecule(smiles='[OH]')])
```
mliu49: You can simplify this species creation to `Species(smiles='[OH]')`.

rmgpy/data/solvationTest.py
yunsiechung: Thanks for the comment! I applied this change

rmgpy/data/solvationTest.py
yunsiechung: I applied this change!

mliu49:
Thanks! Could you rebase?

yunsiechung:
@mliu49 I just rebased and pushed again!

PR #1778, opened 2019-10-22
Improve Emax determination when solving pdep networks

mliu49:
### Motivation or Problem
Previously, Emax was set to a fixed value above the highest energy isomer. However, for large molecules at high temperatures, the hard-coded value may not be sufficient to capture the full equilibrium distribution.

The result was that Arkane/RMG would not be able to successfully evaluate rate coefficients in these networks.

### Description of Changes
This PR implements a more complex approach of evaluating the equilibrium distribution and then adjusting Emax until the population at Emax is a specified fraction of the most populated energy.

It starts with the previous default of 40*RT above the highest energy isomer, fits a curve to the equilibrium distribution, and extrapolates to identify the energy at which the population drops to 0.1% of the most populated energy.

### Testing
This was tested a while ago with some pdep networks with C12 species, where the change allowed Arkane to evaluate rate coefficients over a wider temperature range than before. It should be tested again though.

One concern I have is the performance of this in RMG pdep jobs, which should be evaluated before merging this.

### Reviewing
Although I'm opening this as a draft PR, I would like to get other opinions on whether or not the approach makes sense.

### DIFF
--- a/rmgpy/pdep/network.py
+++ b/rmgpy/pdep/network.py
@@ -35,6 +35,7 @@ import logging
 import math
 
 import numpy as np
+from scipy.optimize import curve_fit, brentq
 
 import rmgpy.constants as constants
 from rmgpy.exceptions import NetworkError, InvalidMicrocanonicalRateError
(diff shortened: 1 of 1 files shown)


mliu49:
After running some RMG jobs with this change, I found that it didn't work at all. While writing this comment, I just realized that there is a bug in the implementation, where I'm not using the highest energy isomer in line 513. That would result in likely underestimating e_max, which would then make it even more likely to error, which was what I had observed.

I will try to test this again after fixing that.

JacksonBurns:

PR #1788, opened 2019-10-27
Terachem

alongd:
### Motivation or Problem
We'd like to expand the electronic structure software Arkane supports

### Description of Changes
Added full support for [TeraChem](http://www.petachem.com/products.html).

I couldn't figure out how to parse the moments of inertia out of a TeraChem log file (I think they might not be reported, although they are certainly calculated and used internally), so some general functionalities for calculating them from xyz were added to common.py. Note that neither RDKit nor OB were used to have the capability to calculate moments of inertia for TSs as well. These functions are used in statmech if no rotational mode exists in conformer.modes. Could be useful in the future for additional ESS. I verified that the results are in agreement with those Gaussian would have calculated for a few species, and added tests.

Sine we have additional future plans to expand ESS support in Arkane (Orca, Psi4), we'll end up with many ESS log parsing and respective tests files under the `arkane` folder. For this to not get out of hand, a `logs` subdir was created for all ESS log files and their tests, including the `Log` parent class.

### Testing
Plenty of tests were added.

### Reviewer Tips
See that the code makes sense, verify that the tests are comprehensive and run them.

I'm tagging @dranasinghe as the reviewer and @mjohnson541 as the merger.


arkane/common.py
```diff
@@ -590,3 +591,94 @@ def get_element_mass(input_element, isotope=None):
...
+        tensor[0, 0] += mass * (cm_coord[1] * cm_coord[1] + cm_coord[2] * cm_coord[2])
+        tensor[1, 1] += mass * (cm_coord[0] * cm_coord[0] + cm_coord[2] * cm_coord[2])
+        tensor[2, 2] += mass * (cm_coord[0] * cm_coord[0] + cm_coord[1] * cm_coord[1])
+        tensor[0, 1] -= mass * cm_coord[0] * cm_coord[1]
+        tensor[0, 2] -= mass * cm_coord[0] * cm_coord[2]
+        tensor[1, 2] -= mass * cm_coord[1] * cm_coord[2]
+    tensor[1, 0] = tensor[0, 1]
+    tensor[2, 0] = tensor[0, 2]
+    tensor[2, 1] = tensor[1, 2]
+    return tensor
```
dranasinghe: I could n't recreate the Gaussian output for CH4 manually. I used excel to do the maths.

arkane/logs/terachem.py
```diff
@@ -0,0 +1,327 @@
...
+            line = f.readline()
+            while line != '':
+                # Read force constant matrix
+                if '*** Hessian Matrix (Hartree/Bohr^2) ***' in line:
+                    force = np.zeros((n_rows, n_rows), np.float64)
+                    for i in range(int(math.ceil(n_rows / 6.0))):
+                        # Matrix element rows
+                        for j in range(n_rows):  # for j in range(i*6, Nrows):
+                            line = f.readline()
+                            while len(line.split()) != 7:
```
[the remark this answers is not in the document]
alongd: good catch! test added

alongd:
Remaining things to do:

- [ ] parse E from either freq (first) or opt (last), appropriately
- [ ] in scans, use opt E
- [ ] check number of scan points according to the input
- [ ] check termination
- [ ] test failed job

alongd:
@dranasinghe, I addressed all the comments we discussed (online and offline), please take a look when you get a chance.

alongd:
@goldmanm, I implemented in TeraChem the rotor scan checks we discussed, please [take a look](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/terachem/arkane/logs/terachem.py#L271) and let me know what you think. Feel free to replicate for Gaussian and/or make suggestions.

goldmanm:
The part about rotor scans looks good. I had one comment about it, listed below.

I've implemented the gaussian changes, and I'm just waiting on the more substantial bugfix pr #1810  to go through to reduce merge conflicts.

arkane/logs/terachem.py
```diff
@@ -260,32 +275,49 @@ def load_scan_energies(self):
...
+                    elif len(v_list) < v_index - 1:
+                        # seems like points in this scan are missing... add None's instead,
+                        # later they'll be removed along with the corresponding angles
+                        v_list.extend([None] * (v_index - 1 - len(v_list)))
+                    else:
+                        # we added more points that we should have, something is wrong with the log file or this method
+                        raise LogError(f'Could not parse scan energies from {self.path}')
         logging.info('   Assuming {0} is the output from a TeraChem PES scan...'.format(os.path.basename(self.path)))
 
+        v_list.append(v_list[0])  # TeraChem does not repeat the first scan point (unlike Gaussian), append it
```
goldmanm: Is there a reason why we need to repeat the data at the end? I think adding data should be avoided if possible. The only reason I can think of is how we define 'angles', will have a point at the end.

A workaround might be to not add the extra datapoint and to replace the angle definition on line 310 so that it does not have the 2$\pi$ angle.

``angles = np.linspace(0, 2 * math.pi * (len(v_list) -1) / len(v_list), len(v_list), np.float64)``

goldmanm: for example, if there are 4 data points (90 degree spacing), the angles output will be 0, 0.5$\pi$, $\pi$, and 1.5$\pi$.

alongd: I found a way to repeat the last point in TeraChem instead, I'll change these lines. We do it as a quality assurance method of checking that a 360 scan gives almost the same energy as the initial point (checked in ARC)

alongd:
@mjohnson541, I just rebased, this PR was reviewed by @dranasinghe, could you let me know if you have additional comments and be the merger?

arkane/statmech.py
```diff
@@ -428,6 +427,36 @@ def load(self, pdep=False, plot=False):
         logging.debug('    Reading optimized geometry...')
         coordinates, number, mass = geom_log.load_geometry()
 
+        if self.species.conformer is not None and len(self.species.conformer.modes):
```
mjohnson541: Why is this commit necessary?  Why are these modes missing?

arkane/commonTest.py
```diff
@@ -50,8 +50,6 @@
 from arkane.input import job_list
 from arkane.statmech import InputError, StatMechJob
 
-################################################################################
```
mjohnson541: Why did you make these modifications?

arkane/statmech.py
alongd: I couldn't figure out how to parse the moments of inertia out of a TeraChem log file (I think they might not be reported, although they are certainly calculated and used internally). TeraChem's manual did not mention anything about it, and I couldn't find it by Googling either. Duminda reported the same thing for Orca, so this functionality could be used when parsing Orca as well.

arkane/commonTest.py
alongd: I think we're over using these #### separators, and it looks cleaner w/o them. I could revert this minor change if you'd like.

arkane/statmech.py
mjohnson541: Okay, so normally Arkane builds these modes, but when that data isn't there they just don't get added to the Conformer object?

arkane/commonTest.py
mjohnson541: I'm not really opposed to the change, but it's probably worth having a larger discussion first.  I think we should remove this commit from the PR for the time being.

alongd:
@mjohnson541, thanks for the comments, I reverted the style changes (rebased and squashed). That commit was not removed, though, it instead applies a consistent style to all of Arkan'es files (now keeping the #### lines). The only exception is common.py where the #### lines are deleted, but added in a subsequent commit. Sorry for that, it just created an annoying conflict otherwise.
I also added a minor commit, correcting some docstring in RMG Molecule, since it's very minor I thought it's OK to just append it here instead of creating a PR for it. No other changes other than that.
I think this PR should be good to go, let me know if you have any additional comments.

arkane/statmech.py
alongd: Correct, Arkane will still parse these modes as usual if it can, otherwise we append them here

alongd:
@mjohnson541, I think all comments were addressed. This branch is now rebased. Please feel free to merge unless there are additional concerns.

PR #1789, opened 2019-10-28
Updates to RMG profiling

### LINKED ISSUE #1173: MakeProfileGraph does not work on remote servers (profiling issue)
I saw that others were doing profiling runs on remote servers, and I realized that I forgot to post an issue about this. When you run profiling on a remote server with the `-p` flag, the job gets hung up after model generation completion. I believe this is due to MakeProfileGraph needing to access the display in order to generate the profiling pdf. You can see this is the case by doing a profile run on the head node while forward your Xserver by `ssh -Y`.

To avoid this issue, you can cherry-pick two commits from my branch `profile_tools`. The first commit comments out the call to MakeProfileGraph (the profile stats will still be saved to the `RMG.profile` file). The second commit also prints the local time for each function in seconds in the profile graph pdf.

To make this work, you will need to copy the RMG.profile file to your local PC, and then run `python $RMG input.py -P`. A capital `-P` flag does not run the RMG job again but rather just generates the pdf.

### LINKED ISSUE #1783: Profile information not saved if job killed by SLURM scheduler
### Motivation or Problem
When RMG is killed suddenly by a SLURM scheduler for running over the allocated wallclock time, it doesn't get a chance to save the profiling information (when running with the `-p` option), which means you can't tell why it was taking so long in the first place.

### Desired Solution
I wonder if we could call `dump_stats` at the end of each generation, without much overhead.
https://docs.python.org/2/library/profile.html#pstats.Stats.dump_stats
I have not looked into this very far.

amarkpayne:
### Motivation or Problem
fixes #1783 
fixes #1173 

There are a couple of improvements that needed to be made to make profiling RMG runs easier, especially when this is being done on larger models that are run on remote servers. This PR fixes a few bugs encountered frequently when profiling, and enhances the profiling graph to include self time in seconds.

### Description of Changes

- Output the self time (in seconds) spent on a specific function call. This number is displayed in parenthesis at the bottom of each cell. It was possible to calculate this before from the self time percentage and the total time, but having the number there is helpful

- Prevents RMG from trying to create the profile graph it cannot access a display, which would otherwise cause this job to crash or stall. The necessary information to build the graph later is still saved though.

- Add a flag that only allows RMG to run a specific amount of iterations at most. This is useful to gather profiling information during different stages of RMG model generation (i.e. restart an RMG job containing 500 species and use the next 10 iterations to profile how RMG performs when the core is fairly large).

- Save the RMG profiling information after each iteration just in case the job crashes

### Testing
I still need to test out these features, particularly to make sure that saving the profiling information works as expected (e.g. not saving only the information from the last iteration). Until I can test this out further I will leave this PR as a draft.


amarkpayne:
@rwest and @mazeau some of the changes to this PR might be relevant to your work. If so please let me know what you think, though I still need to test these changes out a bit more. @kspieks feel free to add commits to this PR

amarkpayne:
TODO before merging:

- Update docstring on RMG class for new __init__ kwargs and new attributes
- Updated commented input file

rwest:
Were you able to figure out what requires the display? We frequently run on remote linux clusters and I've not noticed this being a problem. I thought it just output it to pdf. Does it depend on the version of gprof2dot or something?

amarkpayne:
> Were you able to figure out what requires the display? We frequently run on remote linux clusters and I've not noticed this being a problem. I thought it just output it to pdf. Does it depend on the version of gprof2dot or something?

I believe it has something to do with whichever code they are calling to generate the pdf. I tried looking into gprof2dot to see which code they call/where they do this but I didn't get very far. I can try again though. I agree though that it doesn't make sense that it needs a display to make the pdf even if it isn't trying to open and display the pdf.

amarkpayne:
After testing, @kspieks and I discovered that my initial solution for saving the profile information did not function as desired, and only the profile information for the first iteration was saved (and overwritten repeatedly). The issue is that dumping the stats disables the profiling.

I then tried solutions involving `Profile.enable()` after dumping the stats, and even passing the stats to a `pstats.Stats` object with `Stats(rmg_profiler)`. Both of these attempts did not work as some of the profiling information gets lost in this process (the number of calls to certain processes is off, and certain processes are missing).

Given this, I am fairly certain that if `Profile.disable()` is ever called then we will always have incorrect profiling information if we try to continue profiling after.

My solution to this was to reproduce all of the calls involved with `Profile.dump_stats()` but excluding the call to `Profile.disable()`. This seems to work surprisingly, and after testing with the minimal example I found that the profile information was essentially the same with and without this change to save the profiling information each iteration.

I also tested what would happen if the job were to crash. You can load in the saved information, but it is worth noting that any functions that have not completed by the time of the crash (most notably `RMG.execute()`) will not be reported. This really only affects the top most functions, and I verified that the calls to functions like `RMG.initialize()` and "model.enlarge()" which would have completed at least once showed the correct amount of time and number of calls.

With this, I think this PR is ready for review, though if I find some time I still want to get to the bottom of the pdf display issue.

mliu49:
Just wondering, what are your thoughts about the per-iteration profiling stuff which I was playing around with a while ago? It's on the [`improved_profiling`](https://github.com/ReactionMechanismGenerator/RMG-Py/commits/improved_profiling) branch. It's not necessarily related to these changes, but this reminded me that it existed.

rmgpy/rmg/main.py
```diff
@@ -99,6 +100,8 @@ class RMG(util.Subject):
     Attribute                           Description
     =================================== ================================================
     `input_file`                        The path to the input file
+    `profiler`                          A cProfile.Profile object for time profiling RMG
+    `stats_file`                        File path to store the profile stats, usually RMG.profile
```
mliu49: Do you think it's necessary to add this as an attribute? We don't allow user specification of the filename, so it seems like we should just construct it using the `output_directory` each time.

rmgpy/rmg/main.py
```diff
@@ -1741,6 +1746,11 @@ def save_everything(self):
         # Notify registered listeners:
         self.notify()
 
+        if self.profiler:  # Save the profile information in case the job crashes
+            with open(self.stats_file, 'wb') as f:
+                self.profiler.snapshot_stats()
+                marshal.dump(self.profiler.stats, f)
```
mliu49: It seems like profile saving time will be included in the profile. Is this worth trying to avoid?

rmgpy/rmg/main.py
amarkpayne: Fair point, I only added `stats_file` as an attribute to pass it down to this module from rmg.py. I can redefine it though from knowledge of the output directory here though, I'll make this change

rmgpy/rmg/main.py
amarkpayne: Absolutely, if possible this would be ideal. However, I think this would only be possible if we call `Profile.disable()` or equivalent, which seemed to cause other errors.

I don't remember seeing saving the profile information in the profile graph, I'll look to see how much time it is taken. presumably this is negligible.

rmgpy/rmg/main.py
```diff
@@ -1374,6 +1405,15 @@ def make_seed_mech(self, first_time=False):
                     with open(self.input_file, 'r') as input_file:
                         f.write(''.join(input_file.readlines()))
 
+            # Also, save the seed to the previous_seeds directory on specified iterations
+            if self.save_seed_modulus != -1:
+                if np.mod(self.reaction_model.iteration_num, self.save_seed_modulus) == 0:
```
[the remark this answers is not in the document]
kspieks: Will do; I'll combine those. The documentation says np.mod(x1, x2) is equivalent to x1 % x2. I can change to % since it's shorter

rmgpy/rmg/main.py
```diff
@@ -1374,6 +1405,15 @@ def make_seed_mech(self, first_time=False):
...
+                    dst = os.path.join(previous_seeds_dir,
+                                       'iteration_number_{0}'.format(self.reaction_model.iteration_num))
+                    logging.info('Copying seed from seed directory to '
+                                 'previous_seeds/iteration_number_{0}'.format(self.reaction_model.iteration_num))
```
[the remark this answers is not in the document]
kspieks: Ah that was sloppy. Good catch thanks!

rmgpy/rmg/main.py
amarkpayne: Quick follow-up: Saving the profile information every iteration does show up on the graph. On the average this takes 0.1 seconds per iteration based on running the minimal example. For the minimal example this amounted to 2% of the total run time, but I would expect for actual profiling jobs this will be negligible.

mliu49: I was actually wondering if it will become more substantial for longer jobs as the amount of data accumulates.

amarkpayne: I can test this out, but my hunch is that it will not. Unlike the chemkin file, which accumulates species and reactions each iteration, the profile file should not accumulate new functions (all of the relevant functions should be called at least once after the first iteration), it should just be updating the values of int/floats used to store the times and number of calls for each function. This at least is my naive belief about how the code should work, but it is worth testing

kspieks: I made a separate function for saving the time profiling information so cProfile could explicitly time that. For superminimal, saving the time profiling info took 6.12% of the total job time at ~0.06s per iteration. For minimal, saving the time profiling info took just 2.18% of the total job time at ~0.08s per iteration. Since the percentage of total time spent saving the info decreases as the job size increases, I agree with @amarkpayne that it should scale well with large jobs. @mliu49 I think this branch is ready to be merged 👍

kspieks:
@amarkpayne can you review this PR when you get a chance?

rmgpy/rmg/mainTest.py
```diff
@@ -255,6 +257,56 @@ def tearDownClass(cls):
...
+        num_dir_actual = sum(os.path.isdir(os.path.join(path, i)) for i in os.listdir(path))
+        num_dir_expected = self.max_iter//2 + 1   # +1 is for saving iteration 0
+        self.assertEqual(num_dir_actual, num_dir_expected)
+
+    def test_max_iter(self):
+        """
+        Test the command line argument of -i
+        """
+        df = pd.read_excel(os.path.join(self.outputDir, 'statistics.xls'))
+        num_rows, num_col = df.shape
```
amarkpayne, line 293: Change to `num_rows = df.shape[0]`

rmgpy/rmg/mainTest.py
```diff
@@ -35,6 +35,8 @@
 
 from nose.plugins.attrib import attr
 
+import pandas as pd
```
amarkpayne, line 38: Let's add pandas to the requirements file, environment file, and the conda meta recipe file so that this dependency is explicit

kspieks:
Thanks for the catching those! I change, squashed, and pushed :)

examples/rmg/commented/input.py
```diff
@@ -204,6 +204,9 @@
     keepIrreversible=False,
     # Allows families with three products to react in the diverse direction (default).
     trimolecularProductReversible=True,
+    # Allows a seed to be saved every n iterations.
+    # The default of -1 causes the iteration to only be saved at the end of the RMG job
+    saveSeedModulus=-1
```
rwest, line 209: Is this `saveSeedModulus` in the documentation?

rmgpy/util.py
```diff
@@ -176,6 +176,9 @@ def parse_command_line_arguments(command_line_args=None):
     parser.add_argument('-t', '--walltime', type=str, nargs=1, default='00:00:00:00',
                         metavar='DD:HH:MM:SS', help='set the maximum execution time')
 
+    parser.add_argument('-i', '--maxiter', type=int, nargs=1, default=None,
```
rwest, line 179: Is this `--maxiter` or `-i` flag in the documentation?

amarkpayne: It shows up in the command line help documentation by typing `python rmg.py -h`. I don't think we document these flags anywhere else. If we were to add it to the documentation, I am trying to think of where it should go. Maybe a small new section in the RMG documentation outlining a few tips for profiling RMG for developers? Maybe include it instead on the GitHub wiki? Let me know what you think

examples/rmg/commented/input.py
amarkpayne: You are right, we forgot to add this to the main documentation. It made it into the commented input file, but it needs to be added to `RMG-Py/users/rmg/input.rst` as well.

rmgpy/util.py
rwest: Maybe they deserve a new section, but I think the command line options should be in the main documentation.

examples/rmg/commented/input.py
kspieks: Thanks for catching this. I working on an explanation of how to use this new feature and will submit the PR today :)

rmgpy/util.py
kspieks: Sounds good. I'll add a new section underneath `1.6 Running a Job` to the main documentation that demonstrates the various input flags. I'll also create a new section with tips on both time and memory profiling RMG jobs.

rwest: That would be great. Thank you!

kspieks:
Please see the new PR #1934 that updates the documentation. Thanks!

PR #1793, opened 2019-10-29
Include Arkane in code coverage report

### LINKED ISSUE #1792: Arkane not included in codecov
### Bug Description
Currently, Arkane is not included in the code coverage report. I believe this is because the arkane folder and Arkane.py are not included as sources in .coveragerc.

It is possible that this is intentional (for example perhaps we want the coverage to reflect RMG-Py itself), but in case it is not I have created a branch called `arkane_coverage` which I believe fixes this issue. If we feel that this change is what we want I will open the PR.

amarkpayne:
### Motivation or Problem
fixes #1792 

Adds the arkane folder and Arkane.py to code coverage.

### Description of Changes
Updated the sources in the .coveragerc file 

### Testing
Checking that Arkane files show up in the coverage report for this PR should suffice

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,9 +3,14 @@
 plugins = Cython.Coverage
 branch = True
 source =
+    arkane
+    Arkane.py
     rmgpy
     rmg.py
-omit = *Test.py
+omit =
+    *Test.py
+    */test_data/*
+    arkane/data/*
 
 [report]
 show_missing = False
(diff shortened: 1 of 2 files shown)


mjohnson541:
Seems perfectly reasonable to me to add Arkane to coverage.  After talking with @mliu49 it seems like this was really just an unintentional side effect of moving cantherm out of rmgpy.

alongd:
Will we only be able to see this effect in future PRs after merging this one? The codecov report has `?`s since it has nothing to compare to for some of the files

.coveragerc
```diff
@@ -3,9 +3,13 @@
...
 branch = True
 source =
+    arkane
+    Arkane.py
     rmgpy
     rmg.py
-omit = *Test.py
+omit =
+    *Test.py
+    */test_data/*
```
alongd: Would you like to add the data subfolder in Arkane?

amarkpayne: Good idea, I'll force push with this change

alongd:
Could you rebase?

amarkpayne:
Done

PR #1794, opened 2019-10-29
Rename submodules in rmgpy.tools

mliu49:
### Motivation or Problem
The submodules in the `tools` module are a bit of a mess in terms of naming. The 3.0 release seems like a good opportunity to rename them.

Note that RMG-website imports will need to be updated, as will any other dependencies which use the tools module.

### Description of Changes
Rename submodules in `rmgpy.tools` and update imports in RMG-Py.

- Generally, changed to short, all lowercase names without underscores.
- Test files are identified with capital `Test`.

### Reviewing
Any comments or suggestions about the new names are welcome.


amarkpayne:
This is correct according to PEP-8 (https://www.python.org/dev/peps/pep-0008/#package-and-module-names).

I think all of this looks good to go, but should we do the same for the scripts folder? The modules in the script folder are all consistent, but use a different convention.

Rebase and I'll approve and merge

mliu49:
Should be good to go once tests pass.

amarkpayne:
Thanks for the PR @mliu49 !

PR #1795, opened 2019-10-29
Disable warning when importing chemprop

mliu49:
### Motivation or Problem
When importing chemprop, the following warning is logged:
```
WARNING:root:No kyotocabinet available
```

This can be confusing to users and cause them to worry that something has gone wrong with the RMG simulation.

### Description of Changes
This is a somewhat hacky fix which simply disables logging before importing chemprop and re-enables it after.

### Testing
This has been confirmed to suppress the warning.

### Reviewer Tips
Test eg0 or importing the `ml.estimator` module. There should be no more warning.

Suggestions for alternative approaches are also welcome.

### DIFF
--- a/rmgpy/ml/estimator.py
+++ b/rmgpy/ml/estimator.py
@@ -28,15 +28,18 @@
 ###############################################################################
 
 import contextlib
+import logging
 import os
 from argparse import Namespace
 from typing import Callable, Union
 
+logging.disable()
 try:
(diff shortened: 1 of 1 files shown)


cgrambow:
Remake your RMG environment and try again.

mliu49:
Thanks!

Fixed by updated `descriptastorus` conda package.

PR #1796, opened 2019-10-30
Fix WSL permisson bug for renaming directories

### LINKED ISSUE #1786: WSL permission denied when renaming seed folder
### Bug Description
```
Making seed mechanism...
  File "/home/mingyu/RMG/RMG-Py/rmg.py", line 106, in <module>
    main()
  File "/home/mingyu/RMG/RMG-Py/rmg.py", line 100, in main
    rmg.execute(**kwargs)
  File "/home/mingyu/RMG/RMG-Py/rmgpy/rmg/main.py", line 749, in execute
    self.make_seed_mech()
  File "/home/mingyu/RMG/RMG-Py/rmgpy/rmg/main.py", line 1272, in make_seed_mech
    os.rename(seed_dir, os.path.join(temp_seed_dir))
PermissionError: [Errno 13] Permission denied: '/home/mingyu/species/tms_syngas_in_silicon/seed' -> '/home/mingyu/species/tms_syngas_in_silicon/seed_tmp'
```
I met this error message, which seems to be similiar with [#1123](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1123)
### How To Reproduce
The odd thing is that my model won't be disturbed with this error for the first run, but crashed like this when I ran it again. Even if I ran my model on a branch or create a new folder to store the input file (Since Alon mentioned cleaning the existing file might be helpful), this error would occur at any time, which is also the another thing that makes me confused. That is, for a same input file, this `rename seed error` is not fixed in the same place, it's possible that this error message took more or less time to occur for one run.

### Installation Information
 - Ubuntu 18.04
 - Installation method:  installation from source with anaconda)
 - RMG version information:
   
   - RMG-Py: latest
   - RMG-database: latest

amarkpayne:
### Motivation or Problem
fixes #1786 

As noted in the issue this fixes, the Windows Subsystem for Linux sometimes throws a permission error 
when trying to renaming non-empty directories, even when the user has the permissions set properly on the directories and files to do so. This is problematic, because we rename the seed directory folder temporarily every time we save a new seed mechanism so that the previous seed mechanism is not lost in case we encounter an error when saving the new one. Renaming the directory does not require copying large amounts of data, so this is the preferred solution in general, but as noted can cause problems for WSL users.

There appear to be several issues opened over a few years complaining about this issue if I am not mistaken, so it appears to be a reoccurring problem. Given this, it is best to handle this case ourselves.

### Description of Changes
A try/except statement was added that avoids renaming the directory in the event of a permission error and instead copies the directory to a new location, which does not throw a permission error as far as I can tell in WSL. Note that this solution is much slower than renaming, so it should only be done as a last resort, which is exactly what we have implemented here.

### Testing
See the issue. We have tested this branch to confirm that at the very least the WSL bug goes away. It might be worth double checking that the files are copied over and then later deleted as desired.

### DIFF
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -615,6 +615,8 @@ class RMG(util.Subject):
         self.initialize_reaction_threshold_and_react_flags()
         self.reaction_model.initialize_index_species_dict()
 
+        self.initialize_seed_mech()
+
     def register_listeners(self):
         """
         Attaches listener classes depending on the options 
@@ -726,7 +728,7 @@ class RMG(util.Subject):
         self.save_everything()
 
         if self.generate_seed_each_iteration:
-            self.make_seed_mech(first_time=True)
+            self.make_seed_mech()
 
         max_num_spcs_hit = False  # default
 
@@ -798,7 +800,7 @@ class RMG(util.Subject):
                                 from arkane.output import prettify
                                 logging.error(prettify(repr(self.reaction_model.core.reactions)))
                             if not self.generate_seed_each_iteration:  # Then we haven't saved the seed mechanism yet
-                                self.make_seed_mech(first_time=True)  # Just in case the user wants to restart from this
+                                self.make_seed_mech()  # Just in case the user wants to restart from this
                             raise
 
                         self.rmg_memories[index].add_t_conv_N(t, x, len(obj))
@@ -952,7 +954,7 @@ class RMG(util.Subject):
         if self.generate_seed_each_iteration:
             self.make_seed_mech()
         else:
-            self.make_seed_mech(first_time=True)
+            self.make_seed_mech()
 
         self.run_model_analysis()
 
@@ -1238,23 +1240,31 @@ class RMG(util.Subject):
         else:
             logging.info("No collision rate violators found.")
 
-    def make_seed_mech(self, first_time=False):
+    def initialize_seed_mech(self):
         """
-        causes RMG to make a seed mechanism out of the current chem_annotated.inp and species_dictionary.txt
-        this seed mechanism is outputted in a seed folder within the run directory and automatically
-        added to as the (or replaces the current) 'Seed' thermo and kinetics libraries in database
-        
-        if run with first_time=True it will change self.name to be unique within the thermo/kinetics libraries
-        by adding integers to the end of the name to prevent overwritting
+        Initialize the process of saving the seed mechanism by performing the following:
 
-        This also writes the filter tensors to the `filters` sub-folder for restarting an RMG job from a seed mechanism
+        1. Create the initial seed mechanism folder (the seed from a previous iterations will be deleted)
+        2. Save the restart-from-seed file (unless the current job is itself a restart job)
+        3. Ensure that we don't overwrite existing libraries in the database that have the same name as this job
         """
+        # Make the initial seed mechanism folder
+        seed_dir = os.path.join(self.output_directory, 'seed')
+        if os.path.exists(seed_dir):  # This is a seed from a previous RMG run. Delete it
+            shutil.rmtree(seed_dir)
+        os.mkdir(seed_dir)
 
-        logging.info('Making seed mechanism...')
+        # Generate a file for restarting from a seed mechanism if this is not a restart job
+        if not self.restart:
+            with open(os.path.join(self.output_directory, 'restart_from_seed.py'), 'w') as f:
+                f.write('restartFromSeed(path=\'seed\')\n\n')
+                with open(self.input_file, 'r') as input_file:
+                    f.write(input_file.read())
 
+        # Change self.name to be unique within the thermo/kinetics libraries by adding integers to the end of the
+        # name to prevent overwriting
         name = self.name
-
-        if self.save_seed_to_database and first_time:  # make sure don't overwrite current libraries
+        if self.save_seed_to_database:  # make sure we don't overwrite current libraries
             thermo_names = list(self.database.thermo.libraries.keys())
             kinetics_names = list(self.database.kinetics.libraries.keys())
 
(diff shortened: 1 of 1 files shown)


amarkpayne:
@ajocher do you have time to be the reviewer/merger for this PR? I wanted someone who was familiar with the filter tensor/seed mech stuff. @alongd I am tagging you because I think you have run into similar issues in Windows before.

rmgpy/rmg/main.py
```diff
@@ -1268,7 +1268,16 @@ def make_seed_mech(self, first_time=False):
             if os.path.exists(seed_dir):  # This is a seed from a previous RMG run. Delete it
```
ajocher: I wonder why you delete the seed from a previous RMG run while you move the seed from a previous iteration to a temporary directory in case you run into errors. Probably I am not fully aware how restart and seed works, but what happens if I want to use a seed from a previous RMG run and do not copy it out of the seed folder? Will it be deleted at this point and my seed from the previous RMG run is gone for good? In this case wouldn't it be good to also just move the seed to a temporary directory?

rmgpy/rmg/main.py
```diff
@@ -1268,7 +1268,16 @@ def make_seed_mech(self, first_time=False):
...
                 shutil.rmtree(seed_dir)
         else:  # This is a seed from the previous iteration. Move it to a temporary directory in case we run into errors
-            os.rename(seed_dir, os.path.join(temp_seed_dir))
+            try:
+                os.rename(seed_dir, temp_seed_dir)
```
ajocher: Have you considered the tempdirectory option? I am curious if it would have advantages in this case. https://docs.python.org/3/library/tempfile.html#tempfile.tempdir

rmgpy/rmg/main.py
amarkpayne: Good question! Currently, if the user restarts from the seed in this folder, we make a copy in a new folder called `previous_restart` so that the seed doesn't get lost. See below.

https://github.com/ReactionMechanismGenerator/RMG-Py/blob/6483418a348ea471b017c0d17b3ec68e00e36127/rmgpy/rmg/main.py#L488-L499


If the user instead were to run a new RMG job from this folder the seed from the previous job would be lost. This was the behavior before I implemented the restart stuff, but I am willing to consider other options if you think this is worth while.

rmgpy/rmg/main.py
amarkpayne: I had not, thanks for the suggestion! It looks interesting, but I am not sure if I can move existing files inside of a temporary directory. Since we already have the previous seed files in a non-temporary place, it should be much faster to rename/move the files (which does not actually move the data on the disk) rather than copying. So if we can move these files into the temporary directory without making a copy then it might be interesting to implement. The only added advantage I think is just the automatic cleanup, but it would be cleaner code.

amarkpayne:
I have refactored how we handle the first_time code in saving the seed mechanism. I need to verify still that this change doesn't break functionality but I think it is at least ready for a first review. @mliu49 do you also have time to take a quick look?

rmgpy/rmg/main.py
```diff
@@ -1268,7 +1268,16 @@ def make_seed_mech(self, first_time=False):
             if os.path.exists(seed_dir):  # This is a seed from a previous RMG run. Delete it
                 shutil.rmtree(seed_dir)
         else:  # This is a seed from the previous iteration. Move it to a temporary directory in case we run into errors
-            os.rename(seed_dir, os.path.join(temp_seed_dir))
+            try:
+                os.rename(seed_dir, os.path.join(temp_seed_dir))
+            except PermissionError:  # The Windows Subsystem for Linux (WSL) can have problems with renaming
```
mliu49: Do you know if the permissions error is with overwriting or renaming? For example, could the issue potentially be resolved by deleting the temp_seed_dir before renaming?

rmgpy/rmg/main.py
```diff
@@ -1234,23 +1236,31 @@ def check_model(self):
...
+            shutil.rmtree(seed_dir)
+        os.mkdir(seed_dir)
 
-        logging.info('Making seed mechanism...')
+        # Generate a file for restarting from a seed mechanism if this is not a restart job
+        if not self.restart:
+            with open(os.path.join(self.output_directory, 'restart_from_seed.py'), 'w') as f:
+                f.write('restartFromSeed(path=\'seed\')\n\n')
+                with open(self.input_file, 'r') as input_file:
+                    f.write(''.join(input_file.readlines()))
```
mliu49: Could you do `f.write(input_file.read())`?

rmgpy/rmg/main.py
```diff
@@ -1260,26 +1270,38 @@ def make_seed_mech(self, first_time=False):
...
 
+    def make_seed_mech(self):
+        """
+        causes RMG to make a seed mechanism out of the current chem_annotated.inp and species_dictionary.txt
+        this seed mechanism is outputted in a seed folder within the run directory and automatically
+        added to as the (or replaces the current) 'Seed' thermo and kinetics libraries in database
+
+        `initialize_seed_mech` should be called once before this function is ever called.
+
+        This also writes the filter tensors to the `filters` sub-folder for restarting an RMG job from a seed mechanism
```
mliu49: Could you reword this docstring? It's written a bit weirdly and isn't really accurate.

I think it should start with "This method saves a seed mechanism..." or "Save a seed mechanism...". It is also not generating it from the chem_annotated.inp and species_dictionary.txt but rather the actual reaction model.

rmgpy/rmg/main.py
amarkpayne: Done, thanks for the catch!

rmgpy/rmg/main.py
amarkpayne: Updated! I should have updated this the last time I modified this function in a PR

rmgpy/rmg/main.py
amarkpayne: From what I have seen I believe this is an issue with `os.rename` itself in the WSL, and occurs regardless of if we are trying to overwrite an existing directory. In fact, in this function the renaming is not being used to overwrite an existing directory. If I can find the links I based this on I will post them here.

amarkpayne: I can't find the original issues I was looking at, but these links seem relevant:
```
https://github.com/microsoft/WSL/issues/1529
https://github.com/microsoft/WSL/issues/3395
https://github.com/microsoft/vscode-remote-release/issues/208
https://github.com/Microsoft/WSL/issues/3738
https://code.visualstudio.com/docs/remote/wsl#_i-see-eaccess-permission-denied-error-trying-to-rename-a-folder-in-the-open-workspace
https://github.com/Microsoft/WSL/issues/2097
```
I am putting these in a code block so that this PR doesn't get linked to all of these issues

mliu49:
I'm not sure if this is related to the changes in this PR, but I just tested with the superminimal example, and running the restart job leads to the following error:
```
  File "../../rmg.py", line 111, in <module>
    main()
  File "../../rmg.py", line 105, in main
    rmg.execute(**kwargs)
  File "/home/mjliu/Code/RMG-Py/rmgpy/rmg/main.py", line 651, in execute
    self.initialize(**kwargs)
  File "/home/mjliu/Code/RMG-Py/rmgpy/rmg/main.py", line 615, in initialize
    self.initialize_reaction_threshold_and_react_flags()
  File "/home/mjliu/Code/RMG-Py/rmgpy/rmg/main.py", line 1562, in initialize_reaction_threshold_and_react_flags
    unimolecular_threshold_restart = f.get('unimolecular_threshold').value
AttributeError: 'NoneType' object has no attribute 'value'
```

amarkpayne:
> I'm not sure if this is related to the changes in this PR, but I just tested with the superminimal example, and running the restart job leads to the following error:
> 
> ```
> Traceback (most recent call last):
>   File "../../rmg.py", line 111, in <module>
>     main()
>   File "../../rmg.py", line 105, in main
>     rmg.execute(**kwargs)
>   File "/home/mjliu/Code/RMG-Py/rmgpy/rmg/main.py", line 651, in execute
>     self.initialize(**kwargs)
>   File "/home/mjliu/Code/RMG-Py/rmgpy/rmg/main.py", line 615, in initialize
>     self.initialize_reaction_threshold_and_react_flags()
>   File "/home/mjliu/Code/RMG-Py/rmgpy/rmg/main.py", line 1562, in initialize_reaction_threshold_and_react_flags
>     unimolecular_threshold_restart = f.get('unimolecular_threshold').value
> AttributeError: 'NoneType' object has no attribute 'value'
> ```

This probably means that the filters were never saved. Which iteration was the restart file from (i.e. when you ran the super-minimal job the first time, how many iterations did you let it go through before you stopped the job?)

I had tested this on the minimal example and it seemed to work, so I am surprised that the super-minimal job is having this error. I might need to re-think how I test the restart functionality

amarkpayne:
Quick update: I am able to replicate the issue on my end. I ran the job to completion and then restarted the job, which should work. I am looking into what is going on now.

amarkpayne:
I think I have figured it out. The super-minimal job does not use filtering, so the filters are never saved. This should be handled by the try/except block in 1561-1593 of main.py, but the error thrown is an AttributeError instead of a key error. I am pretty sure that this would have been tested when we did the original restart PR, but perhaps it was overlooked. Additionally, it is possible that the behavior of h5py has changed so that it throws a different error.

Either way, this issue is not a result of this PR, but it should be an easy fix. @mliu49 would you prefer me to include the necessary changes in this PR or a separate one? I am planning on adding a commit with the fix, along with some functional tests for restarting the minimal job and the super-minimal job

mliu49:
Please open a separate PR. If you've confirmed that this PR does not change behavior, then I think we can go ahead and merge this.

amarkpayne:
I just checked current master and can confirm that this issue also affects it, so this error is not a result of the changes in this PR. I'll open up an issue and a separate PR to fix this. Thanks for the catch @mliu49 !

PR #1799, opened 2019-10-30
Test and fix Molecule.remove_van_der_waals_bonds

rwest:
### Motivation or Problem
There was a bug where `remove_van_der_waals_bonds` would modify the dictionary of bonds while iterating over it. I guess Python 3 detects this and crashes, wheras Python 2 didn't.

### Description of Changes

This pull request adds a failing unit test, then a patch that makes it pass.

### Reviewer Tips
For testing, you could run the tests after the first commit (and see them fail) then run them again after the second commit (and see them pass).

Note the test coverage is misleading because this code is cythonized. Coverage should go up, as I added a test.

### DIFF
--- a/rmgpy/molecule/group.py
+++ b/rmgpy/molecule/group.py
@@ -1130,7 +1130,7 @@ class Group(Graph):
         """
         Remove all bonds that are definitely only van der Waals bonds.
         """
-        cython.declare(atom=GroupAtom, bond=GroupBond)
+        cython.declare(bond=GroupBond)
         for bond in self.get_all_edges():
             if bond.is_van_der_waals(wildcards=False):
                 self.remove_bond(bond)
--- a/rmgpy/molecule/molecule.py
+++ b/rmgpy/molecule/molecule.py
@@ -718,7 +718,7 @@ class Bond(Edge):
         Return ``True`` if the bond represents a van der Waals bond or 
         ``False`` if not.
         """
-        return self.is_order(0) or self.order == 'vdW'  # todo: remove 'vdW'
+        return self.is_order(0)
 
     def is_order(self, other_order):
         """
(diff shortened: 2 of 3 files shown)


ChrisBNEU:
This worked for me. I was getting the following error before while running a reaction with a Pt catalyst: 
RuntimeError: dictionary changed size during iteration (see attached file)
After merging I am no longer getting that error. 

[RMGCat_HANModelError.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/3821444/RMGCat_HANModelError.txt)

amarkpayne:
I am assigning myself as the merger for now to give an opportunity for one of the newer developers to take a look at it if they want the experience. If there is no movement on this by tomorrow evening though I'll go ahead and review/merge this in.

amarkpayne:
I think this fix is fine, but @mliu49 had implemented something very similar [here](https://github.com/ReactionMechanismGenerator/RMG-Py/commit/6c8004efe09e7869c60958d1efd814d12cdcb032#diff-199e8ac7eefa2bc7391991c9df68e6e1R1137). Ideally we wouldn't have duplicated code like this, but I don't know if adding this method to the parent class `Graph` makes sense either.

I think the next best thing would be to make these functions consistent, but this isn't a big deal either. @rwest do you mind updating this branch when you get a chance? If you want feel free to make the code consistent between these functions, or if there is a better solution to get rid of duplicating code I am all for that as well. Either way I think this PR is good to go and I'll merge it as soon as it is updated.

rwest:
Thanks for spotting the code duplication. It also annoys me, but I think adding it to "Graph" doesn't make sense, as a Graph should know nothing about whether its 'edges' are van der Waals or not. So I did what you called the "next best thing" and made my fix the same as Max's.

While updating the code I spotted a couple of tiny tweaks. Just rebased and pushed.

PR #1802, opened 2019-11-01
Fix PES figure generation in png format

xiaoruiDong:
### Motivation or Problem
One of the arguments of rmgpy.pdep.draw is renamed from `format` to `file_format` in a previous commit, but the corresponding codes are not fully refactorized, causing failures in generating PES figures in png format. This will further hinder the PES from being generated and displayed on the RMG-website.

### Description of Changes
rename `format` to `file_format` .

### Testing
By calling the draw function with `file_format="png"`, it can generate PES figure in .png format


xiaoruiDong:
Thanks @alongd. Docstring changed as well.

PR #1805, opened 2019-11-04
Update sensitivity example for py3

oscarwumit:
### Motivation or Problem
Update sensitivity example as part of py3 merge task.

### Description of Changes
1. correct the format of chemkin input file in the example, before it resulted in a parser error
"Thermo file has no default temperature ranges
(The line it would be on is '    300.000  1000.000  5000.000\n' but that is not formatted as such)
"
2. removed legacy exclusion mark for kinetics families

### Testing
Tested the example locally. The sensitivity analysis finished successfully.


oscarwumit:
@amarkpayne could you please merge this PR? I do not have merge access.Thanks.

PR #1808, opened 2019-11-05
Arkane pdep py3 fixes

goldmanm:
### Motivation or Problem
I ran a pdep job on Python 3 and encountered some small issues. This commit fixed these bugs.
### Description of Changes
Figure.close no longer works, so this commit changes how to close the figure. It also renames a variable for consistency.

### Testing
Ran an RMG pdep job and got the expected result.

### Reviewer Tips
Make sure the code looks adequate

### DIFF
--- a/arkane/statmech.py
+++ b/arkane/statmech.py
@@ -742,7 +742,7 @@ class StatMechJob(object):
             ['$0$', '$\pi/4$', '$\pi/2$', '$3\pi/4$', '$\pi$', '$5\pi/4$', '$3\pi/2$', '$7\pi/4$', '$2\pi$'])
 
         self.hindered_rotor_plots.append((fig, rotor_index))
-        fig.close()
+        plt.close(fig)
 
     def save_hindered_rotor_figures(self, directory):
         """


arkane/statmech.py
```diff
@@ -740,7 +740,7 @@ def create_hindered_rotor_figure(self, angle, v_list, cosine_rotor, fourier_roto
             ['$0$', '$\pi/4$', '$\pi/2$', '$3\pi/4$', '$\pi$', '$5\pi/4$', '$3\pi/2$', '$7\pi/4$', '$2\pi$'])
 
         self.hindered_rotor_plots.append((fig, rotor_index))
-        fig.close()
+        plt.close(fig)
```
alongd: We use `fig.close()` in the kinetic_library_to_training_tools.py script as well (L208, L283). Do you think these should be replaced as well? If so, let add them to this PR

rmgpy/pdep/network.py
```diff
@@ -327,7 +327,7 @@ def set_conditions(self, T, P, ymB=None):
 
         n_isom = self.n_isom
         n_reac = self.n_reac
-        n_proc = self.n_prod
+        n_prod = self.n_prod
```
alongd: This is indeed a typo, but `n_prod` (or `n_proc`...) isn't used in this function. Should we remove this line instead?

arkane/statmech.py
mliu49: Thanks for checking for other uses! That case is actually not related, since fig is a BytesIO object rather than a matplotlib figure.

rmgpy/pdep/network.py
goldmanm: We could remove it in this PR, though I do use it in #1810. If it is removed in this PR, I'll just add it back in the other one.

Issue #1750, opened 2019-10-03
Reactors no longer normalize mole fractions


mliu49:
### Bug Description
RMG previously would normalize initial mole fractions specified in input files. This functionality was removed in #1331. It's not clear whether or not it was intentional.

I'm not sure whether this actually affects reactor simulations, since they are done with moles and concentrations. It seems that this effectively changes the initial basis from 1 mole total to whatever the actual numbers are.

### How To Reproduce
Take any example, e.g. superminimal, and change the mole fractions to random things, e.g.
```
    initialMoleFractions={
        'H2': 200,
        'O2': 100,
    },
```

Run the job, and check out the simulation profiles, which will show 200 moles of H2 initially.

### Expected Behavior
I'm not sure exactly what the expected behavior would be for a new user. The input key is called initialMoleFractions, so it seems unintuitive for the values to be effectively interpreted as initialMoles.

I might be somewhat biased since I'm used to the old behavior of normalizing mole fractions. It was very convenient for using easy numbers to specify reactant ratios.

hwpang:
It looks like this issue is solved by  #1809.

PR #1809, opened 2019-11-05
Allow mole fractions to be normalized if they do not sum to one

kspieks:
Added some code in rmgpy/rmg/input.py that normalizes the mole fractions if necessary. This resolves issue #1750 

### Motivation or Problem
RMG previously would normalize initial mole fractions specified in input files. This functionality was removed in #1331.

### Description of Changes
Added some code in rmgpy/rmg/input.py that normalizes the mole fractions if necessary

### Testing
I tested this on the superminimal example with two conditions:
```
initialMoleFractions={
        'H2': 0.67, 'O2': 0.33,
    },
```
and 
```
initialMoleFractions={
        'H2': 67, 'O2': 33,
    },
```
The feature seems to work as expected

### Reviewer Tips
Minor question: I used `for spec, molfrac in initialMoleFractions.items():` to be explicit about what we were looping over. The code immediately above this addition in line 195 uses a more general `for key, value in initialMoleFractions.item():`. I assume these should be consistent. What do you recommend? Thanks!

### DIFF
--- a/rmgpy/rmg/input.py
+++ b/rmgpy/rmg/input.py
@@ -226,6 +226,23 @@ def simple_reactor(temperature,
             [not isinstance(x, list) for x in initialMoleFractions.values()]):
         nSims = 1
 
+    # normalize mole fractions if not using a mole fraction range
+    if all([not isinstance(x, list) for x in initialMoleFractions.values()]):
+        total_initial_moles = sum(initialMoleFractions.values())
+        if total_initial_moles != 1:
+            logging.warning('Initial mole fractions do not sum to one; normalizing.')
+            logging.info('')
+            logging.info('Original composition:')
+            for spec, molfrac in initialMoleFractions.items():
+                logging.info('{0} = {1}'.format(spec, molfrac))
+            for spec in initialMoleFractions:
+                initialMoleFractions[spec] /= total_initial_moles
+            logging.info('')
+            logging.info('Normalized mole fractions:')
+            for spec, molfrac in initialMoleFractions.items():
+                logging.info('{0} = {1}'.format(spec, molfrac))
+            logging.info('')
+
     termination = []
     if terminationConversion is not None:
         for spec, conv in terminationConversion.items():
--- a/rmgpy/rmg/inputTest.py
+++ b/rmgpy/rmg/inputTest.py
@@ -28,6 +28,7 @@
 ###############################################################################
 
 import unittest
+from unittest.mock import patch
 
 import rmgpy.rmg.input as inp
 from rmgpy.rmg.main import RMG
@@ -35,7 +36,7 @@ from rmgpy.rmg.main import RMG
 
 ###################################################
 
-def setUpModule(self):
+def setUpModule():
     """
     A method that is run before the class.
     """
@@ -46,7 +47,7 @@ def setUpModule(self):
     inp.set_global_rmg(rmg)
 
 
-def tearDownModule(self):
+def tearDownModule():
     # remove RMG object
     global rmg
     rmg = None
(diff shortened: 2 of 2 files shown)


rmgpy/rmg/input.py
```diff
@@ -206,6 +206,20 @@ def simple_reactor(temperature,
             elif value[1] < value[0]:
                 raise InputError('Initial mole fraction range out of order: {0}'.format(key))
 
+    # normalize mole fractions if applicable
+    totalInitialMoles = sum(initialMoleFractions.values())
```
mliu49: This should be `snake_case`. Also, this doesn't seem to account for mole fraction ranges.

mliu49:
It seems that ranged reactors do their own normalization [here](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/rmg/main.py#L2200).

So we can probably just do this for fixed initial mole fractions.

kspieks:
> It seems that ranged reactors do their own normalization [here](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/rmg/main.py#L2200).
> 
> So we can probably just do this for fixed initial mole fractions.
Er I'm confused. ~line 2205 in the`generate_cond()` method seems to normalize even if the value in the `initial_mole_fractions` dictionary is not a list? So when the method is called in ~line 674 in main.py, is it already normalizing, thus making this branch redundant? 
But making the value 100 vs 1.0 produces a different plot. I think I'm now confused where the issue is


> It seems that ranged reactors do their own normalization [here](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/rmg/main.py#L2200).
> 
> So we can probably just do this for fixed initial mole fractions.

As a fix, I moved the normalization code to underneath the 
`if not isinstance(temperature, list) and not isinstance(pressure, list) and all(
            [not isinstance(x, list) for x in initialMoleFractions.values()]):
        nSims = 1`
for both the simple and liquid reactor. This successfully normalizes when the mole fractions are given as scalars, not lists. Upon testing the superminimal example with both `initialMoleFractions={ 'H2': 0.67, 'O2': 0.33, }` and `initialMoleFractions={ 'H2': 67, 'O2': 33, }`, I observe the expected normalization behavior since the final plot is always scaled between 0 and 0.7.

_However_, if I use a mole fraction range, such as `initialMoleFractions={ 'H2': [67, 77], 'O2': 33, }`, the final plot is scaled from 0 to 70, while using `initialMoleFractions={ 'H2': [0.67, 0.77], 'O2': 0.33, }` scales from 0 to 0.7. _It seems like the range reactor is not normalizing?_ I think I'm confused on the behavior of the range reactor

kspieks:
> It seems that ranged reactors do their own normalization [here](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/rmg/main.py#L2200).
> 
> So we can probably just do this for fixed initial mole fractions.

As a fix, I moved the normalization code to underneath the
if not isinstance(temperature, list) and not isinstance(pressure, list) and all( [not isinstance(x, list) for x in initialMoleFractions.values()]): nSims = 1
for both the simple and liquid reactor. This successfully normalizes when the mole fractions are given as scalars, not lists. Upon testing the superminimal example with both initialMoleFractions={ 'H2': 0.67, 'O2': 0.33, } and initialMoleFractions={ 'H2': 67, 'O2': 33, }, I observe the expected normalization behavior since the final plot is always scaled between 0 and 0.7.

However, if I use a mole fraction range, such as initialMoleFractions={ 'H2': [67, 77], 'O2': 33, }, the final plot is scaled from 0 to 70, while using initialMoleFractions={ 'H2': [0.67, 0.77], 'O2': 0.33, } scales from 0 to 0.7. It seems like the range reactor is not normalizing? I think I'm confused on the behavior of the range reactor

mliu49:
@kspieks, what's the status of this PR?

kspieks:
@mliu49 I believe this PR is now ready to be merged based on the tests I ran.

I removed the feature from the liquidReactor because running the liquid_phase example revealed significant changes to the behavior of the liquidReactor by causing it to perform many more iterations than running the same liquid_phase example on the master branch commit #9a469234f48d8968a0be8e4d823bcff534ff7722 aka updated this afternoon. I had to kill the example because the plot in the solver directory was different, and it had not converged after twice as many examples so the statistics file was larger and had additional species and reactions in the core and edge. Presumably this is because the liquidReactor is expecting initialConcentrations, thus giving it mole fractions changes the behavior. I propose leaving the liquidReactor as it was on master.

In contrast, simpleReactor is expecting initialMoleFractions so the normalization features works as expected. The superminimal example was tested under 3 cases. The results from running on this branch vs master commit #9a469234f48d8968a0be8e4d823bcff534ff7722 aka updated this afternoon are compared:

1) initialMoleFractions={ 'H2': 0.67, 'O2': 0.33, }
The number of species and reactions in the core and edge from the statistics files are identical, the final plot in the solver directory looks identical, and diffing the final simulation csv also proves they are identical. 

2) initialMoleFractions={ 'H2':  67,  'O2': 33, }
The number of species and reactions in the core and edge from the statistics files are identical, and the final plot in the solver directory looks identical. The final simulation csv is different due to H2 being 67 in one and 0.67 in another. This also changes the numbers in all columns by a factor of 100. Strangely, the ODE solver seemed to take slightly different steps making it difficult to compare. However, recursively diffing the seed directory proves that the files in the filters, seed, and seed_edge directory are identical.

3) initialMoleFractions={ 'H2’:[0.50, 0.70], 'O2': 0.33, }
The number of species and reactions in the core and edge from the statistics files are identical, the final plot in the solver directory looks identical, and diffing the final simulation csv also proves they are identical

mliu49:
You're right, liquid reactor does not need normalization since it uses concentrations. I wasn't thinking properly. Sorry for the extra work you had to do to try that.

kspieks:
No worries at all! It didn't take long to test the liquid reactor

rmgpy/rmg/input.py
```diff
@@ -225,6 +225,19 @@ def simple_reactor(temperature,
     if not isinstance(temperature, list) and not isinstance(pressure, list) and all(
             [not isinstance(x, list) for x in initialMoleFractions.values()]):
         nSims = 1
+        # normalize mole fractions if applicable
```
mliu49: I think we may want this even if T and P are ranges, since those are independent parameters.

I would add a second if statement here with only `all([not isinstance(x, list) for x in initialMoleFractions.values()])`. Also, the comment could be updated to say `if not using a mole fraction range`.

rmgpy/rmg/input.py
```diff
@@ -225,6 +225,19 @@ def simple_reactor(temperature,
...
+            logging.warning('Initial mole fractions do not sum to one; normalizing.')
+            logging.info('Original composition:')
+            for spec, molfrac in initialMoleFractions.items():
+                logging.info('{0} = {1}'.format(spec, molfrac))
+            for spec in initialMoleFractions:
+                initialMoleFractions[spec] /= total_initial_moles
+            logging.info('Normalized mole fractions:')
+            for spec, molfrac in initialMoleFractions.items():
+                logging.info('{0} = {1}'.format(spec, molfrac))
+            logging.info('')
```
mliu49: This is really minor, but I think it would be nice to put the blank lines before `Original composition` and `Normalized mole fractions` like [before](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/5a07cc9aa932c0d2d031d81254381caa48f9bd83/rmgpy/rmg/input.py#L151).

kspieks:
Thanks for the feedback! I pushed the changes that addressed your comments (though I still have to look up how to undo whitespace in a pushed commit). Any idea when #1842 will be merged?

rmgpy/rmg/input.py
```diff
@@ -225,19 +225,22 @@ def simple_reactor(temperature,
...
-            for spec, molfrac in initialMoleFractions.items():
-                logging.info('{0} = {1}'.format(spec, molfrac))
-            for spec in initialMoleFractions:
-                initialMoleFractions[spec] /= total_initial_moles
-            logging.info('Normalized mole fractions:')
-            for spec, molfrac in initialMoleFractions.items():
-                logging.info('{0} = {1}'.format(spec, molfrac))
-            logging.info('')
+        # normalize mole fractions if not using a mole fraction range
+        if all([not isinstance(x, list) for x in initialMoleFractions.values()]):
```
mliu49: This should be indented to the same level as the above if statement:
```
if T,P,X are not ranges:
    nSims=1
if X are not ranges:
    normalize mole fractions
```

mliu49:
You can add another commit to remove the added whitespace. I would suggest squashing all of these commits together since it's a relatively small change. You can do that when you rebase onto master.

#1842 will be merged once all tests pass, maybe within the next hour or so.

rmgpy/rmg/input.py
kspieks: My apologies, I was rushing to get to the mini TG 😬 The changes should be correct now. All commits have been squashed and rebased onto master. Thanks again for catching my mistakes!

mliu49:
I think this is ready to be merged. Could someone else take a look at the unit tests that I added? @kspieks @mjohnson541

kspieks:
> I think this is ready to be merged. Could someone else take a look at the unit tests that I added? @kspieks @mjohnson541

I thought the unit tests looked good. I should've thought to add those, so thanks for doing that!

PR #1811, opened 2019-11-05
Fix: numeric energy input for ARKANE

jeehyunatrmg:
### Motivation or Problem
I wanted to manually put energy in Arkane input file since Arkane doesn't contain model chemistry that I used. However, Arkane assumed float type as 'path' and giving out the error.

### Description of Changes
added lines that can recognize the float type before it assumes as the path

### Testing
it worked in my laptop..?

### Reviewer Tips
Actually, Jerry did this fix...

### DIFF
--- a/arkane/statmech.py
+++ b/arkane/statmech.py
@@ -316,7 +316,9 @@ class StatMechJob(object):
             except KeyError:
                 raise InputError('Model chemistry {0!r} not found in from dictionary of energy values in species file '
                                  '{1!r}.'.format(self.modelChemistry, path))
-        if not os.path.isfile(energy.path):
+        if not hasattr(energy, 'path'):
+            pass	
+        elif not os.path.isfile(energy.path):
             modified_energy_path = os.path.join(directory, energy.path)
             if not os.path.isfile(modified_energy_path):
                 raise InputError('Could not find single point energy file for species {0} '


arkane/statmech.py
```diff
@@ -316,7 +316,9 @@ def load(self, pdep=False, plot=False):
             except KeyError:
                 raise InputError('Model chemistry {0!r} not found in from dictionary of energy values in species file '
                                  '{1!r}.'.format(self.modelChemistry, path))
-        if not os.path.isfile(energy.path):
+        if not hasattr(energy, 'path'):
+            pass	
+        elif not os.path.isfile(energy.path):
```
mliu49: I would suggest putting the check in the same statement: `if hasattr(energy, 'path') and not os.path.isfile(energy.path)`

Also, I'm not sure if Alon has opinions on use of `hasattr` vs checking `isinstance(energy, Log)`.

goldmanm:
I also worked on a similar issue, which is in the PR Arkane bugfixes. You can see the change at this commit:
cba4f3f6aa507d633c3bf9982e5

arkane/statmech.py
alongd: I think `isinstance(energy, Log)` is cleaner, and it should definitely be in the same statement.
@jeehyunatrmg, we can do it together tomorrow if you'd like, or simply go-ahead and make these fixes.

alongd:
We can either use this PR or @goldmanm's commit. This PR could also be a good practice. Either way we should also add a test.

amarkpayne:
What is the current status of this PR? Are we using the change here or are we using the commit from @goldmanm ?

goldmanm:
9509839a16b2d12ac448e5b9f29894904a fixed this issue.

PR #1814, opened 2019-11-07
Made a jupyter notebook to discover low T pathways.

rwest:
It has been suggested many times that one might "help" RMG discover low temperature pathways (that might be hard to find by just following flux) by adding the intermediate species that you would expect to be important.

The goal of this notebook is to allow you to generate all the species that would occur by systematically applying known low-temperature pathways to a starting fuel molecule.
It then creates an RMG input file block at the end of the notebook.

Hopefully if you put these species into the model core at the start, RMG will generate the appropriate reactions.

General suggestions and comments welcome (as well as careful code review, of course)

The pathways are based on this diagram from https://doi.org/10.1016/j.combustflame.2015.07.005
image2
(although that top loopy blue arrow is not included. Anyone know how important it is?)

Although it's designed to look at the .ipynb, I've included a corresponding .py script for review because it's easier to annotate lines of code there.

It's a bit hard to find in the PR interface, so here's a direct link to the notebook:
https://github.com/rwest/RMG-Py/blob/lowTnotebook/ipython/LowTPathways.ipynb


alongd:
Perhaps we'd like to add this script to RMG, and ask RMG to find and add all these species to the core (if the users declare they care about low T using some input file flag)?

rwest:
Yes, if it works (as intended), then adding it as a core RMG feature is definitely worth considering. Would help many novice users. But I feel someone should test it helps first.

mjohnson541:
I actually have a notebook for the same purpose, but I took a bit different approach.  Although I think I just counted on RMG to handle the last couple decomposition steps at the end.  

Rather than using RMG to directly generate reactions I reacted the species manually (mostly without isomorphism searches and with very limited use of isomorphism checks). This has the advantage of allowing me to simply tell it to skip pathways that involve unnaturally strained intra_H_migrations, which can be important for very large fuel molecules where you otherwise might get way too many pathways.  It also may be a lot faster especially for large molecules, which could potentially matter a lot.

rwest:
Interesting. 

I see two steps that involve intra_H_migration:
ROO --> QOOH
O2QOOH  --> HOOQjOOH   (--> OH + ketohydroperoxide)

Would it help to limit both?
What would you call "unnaturally strained"? 
Would having RMG estimate the rate and excluding the slower ones be a way to handle this? Or make a more specific reaction template that only matches certain TS ring sizes?

So many questions :)

mjohnson541:
I think it's especially important when you consider the fact that you're doing it twice (so if you get twice as many pathways as you need the first time you'll probably get maybe 4 times as many pathways as you want after the second migration (when they branch).  The big rule based mechanisms certainly avoid those pathways.  Although I left the atom distances up to the user's choice:  
getAllMigrations(mols,distmin=5,distmax=7)

This was more of a rules based approach using RMG.  It's a bit more complicated to write than just reacting everything and filtering it, but I think it's more practical if we want to start integrating this into RMG.  Something I've been meaning to do is have reaction families that represent a series of reactions that you could simply apply and get all of these pathways and doing that at large scale (where we have lots of those kinds of families) I think the manual approach would probably be significantly faster.  

No idea if that matters just here in the notebook though.

JacksonBurns:
Changing base to main and rebasing (since the latest changes allow CI to run on forks). This would be easy to merge and nice to have (assuming it is still functional/correct).

PR #1815, opened 2019-11-07
Gaussian energies

dranasinghe:
### Motivation or Problem
Arkane parser for Gaussian does not parse MP2, double hybrid DFT, CCSD, CCSD(T) energies. Previous to this change, HF energy was appended as e0 or energies should be given manually in the Arkane input file.

### Description of Changes
CBS-QB3 calculation contains CCSD(T), MP2, and CCSD energies, therefore, new if statements were added before CBS-QB3. The elif statements were arranged in the order of appearance in the output file

### Testing
unit test added 

### Reviewer Tips
run a Gaussian calculation and check e0

### DIFF
--- a/arkane/gaussian.py
+++ b/arkane/gaussian.py
@@ -45,6 +45,7 @@ from arkane.common import check_conformer_energy, get_element_mass
 from arkane.exceptions import LogError
 from arkane.log import Log
 
+
 ################################################################################
 
 
@@ -212,7 +213,7 @@ class GaussianLog(Log):
                         elif 'Rotational constant (GHZ):' in line:
                             inertia = [float(line.split()[3])]
                             inertia[0] = constants.h / (8 * constants.pi * constants.pi * inertia[0] * 1e9) \
-                                * constants.Na * 1e23
+                                         * constants.Na * 1e23
                             rotation = LinearRotor(inertia=(inertia[0], "amu*angstrom^2"), symmetry=symmetry)
                             modes.append(rotation)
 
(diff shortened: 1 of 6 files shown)


goldmanm:
@dranasinghe, I would merge the first two commits, since they relate to the same effect. 

The tests are showing an error when reading `examples/arkane/species/C2H5/ethyl_cbsqb3.log`

I don't know the CBS always appears in `E2(CBS)= `. If it does you could include the entire string. If it does not you can check for both `' E2('' and `')= ' (with the spaces)`to ensure it doesn't pick up the other dataline.

arkane/gaussian.py
```diff
@@ -274,7 +274,7 @@ def load_energy(self, zpe_scale_factor=1.):
                 if 'SCF Done:' in line:
                     e_elect = float(line.split()[4]) * constants.E_h * constants.Na
                     elect_energy_source = 'SCF'
-                elif ' E2(' in line:
+                elif ' E2(' in line and ' E(' in line:
```
goldmanm: I am probably missing something. What line in ethyl_cbsqb3.log should this be parsing?

When I type `grep ' E2(' ethyl_cbsqb3.log | grep ' E('`, I don't find any matching lines.

The unit tests are passing, though I don't understand why.

dranasinghe:
Dear @goldmanm Mark, i made the changes you suggested. Further, i added the file previously gave the error as a test.

arkane/gaussian.py
dranasinghe: Yes, that how it should be. In my previous commit, I only put " E2("  and it found a match at the end of the summary gaussian print. That's why the unit test failed.  " E2("  and "E(" does not appear together in the cbs-qb3 calculation and it will correctly parse the " cbs-qb3 energy" , but it will appear in the double-hybrid calculations.

PR #1827, opened 2019-11-22
Reverse recipes should be in reverse order

mazeau:
I found that when generating reverse templates, the action list would go in the same order as the forward, and this would cause issues when making and breaking multiple bonds.

```    ['FORM_BOND', '*1', 1, '*2'],
    ['BREAK_BOND', '*2', 1, '*3'],
    ['FORM_BOND', '*3', 1, '*4'],
    ['CHANGE_BOND', '*3', 1, '*4']
```
This change fixes this template's reverse recipe.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1139,10 +1139,7 @@ class KineticsFamily(Database):
             data = deepcopy(entry.data)
             data.change_t0(1)
 
-            if type(data) is Arrhenius:
-                # more specific than isinstance(data,Arrhenius) because we want to exclude inherited subclasses!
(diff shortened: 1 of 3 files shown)


rwest:
Please can you make a PR from a dedicated topic branch so it just contains the changes described in the PR (in this case your commit "reverses the recipe actions for reverse templates" fb8ade7). Otherwise everything you do on your `master` ends up being added to this PR.
I think perhaps the only way to fix is to close this pull request and open a new one.

Ideally a topic branch pushed to the official fork (if you don't have permission to create branches there we can probably grant it), else one pushed to your fork but check the box to "Allow edits from maintainers")

Issue #1682, opened 2019-08-10
Arkane Pdep Functional Test Fails Often


amarkpayne:
### Bug Description
The latest build of #1561 failed with the following error:

```
======================================================================

FAIL: A general test for a PDep job in Arkane

----------------------------------------------------------------------

Traceback (most recent call last):

  File "/home/travis/build/ReactionMechanismGenerator/RMG-Py/arkane/pdepTest.py", line 104, in testPDepJob

    self.assertAlmostEquals(rxn.kinetics.getRateCoefficient(1000.0, 1.0), 88.88253229631246)

AssertionError: 88.8819863725482 != 88.88253229631246 within 7 places
```

This PR does not change any part of Arkane or RMG's pdep code though, and in fact this error only showed when I recently rebased this branch on current master. I notice that this test fails often when I run the functional tests locally, and it appears that the number for the rate coefficient appears to drift slightly for reasons unknown (it appears to happen whenever we add a new Arkane example or functional test, for example I believe this was why the most recent change was made to this unit test [here](https://github.com/ReactionMechanismGenerator/RMG-Py/commit/66a04b7ed53eee4ed23c83e9ae24245929bd6c8e).)

I also wonder if it is somehow related to the error with the most recent Chlorine DB branch build [here](https://github.com/ReactionMechanismGenerator/RMG-database/pull/317), where the order of the reactants has changed when the functional tests are run in this build.

A going theory that I have is that this is related to Arkane's use of global and module variables, which might cause problems when we run all of the Arkane examples at once, but I am not sure.

In the meantime, we might want to consider re-writing this test. Thoughts?

### How To Reproduce
Run the RMG functional tests locally. Check out the `errorCancelingReactions` branch if need be.

mjohnson541:
That's incorrect, the [change](https://github.com/ReactionMechanismGenerator/RMG-Py/commit/66a04b7ed53eee4ed23c83e9ae24245929bd6c8e) in the multi-D torsions PR was made because we switched from calculating 1D HRs semi-classically to quantum mechanically.  This impacts pressure dependent rate calculations because they are dependent through TST and thermochemistry on the 1D HR partition functions.

amarkpayne:
[...]

PR #1829, opened 2019-11-23
Reverse recipes should be in reverse order

mazeau:
This is a duplicate of PR #1827, but only contains the relevant commits.

I found that when generating reverse templates, the action list would go in the same order as the forward, and this would cause issues when making and breaking multiple bonds.

```    ['FORM_BOND', '*1', 1, '*2'],
    ['BREAK_BOND', '*2', 1, '*3'],
    ['FORM_BOND', '*3', 1, '*4'],
    ['CHANGE_BOND', '*3', 1, '*4']
```
This change fixes this template's reverse recipe.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -243,7 +243,7 @@ class ReactionRecipe(object):
         of the reaction that this is the recipe for.
         """
         other = ReactionRecipe()
-        for action in self.actions:
+        for action in reversed(self.actions):  # Play the reverse recipe in the reverse order
             if action[0] == 'CHANGE_BOND':
                 other.add_action(['CHANGE_BOND', action[1], str(-int(action[2])), action[3]])
             elif action[0] == 'FORM_BOND':


rmgpy/data/kinetics/family.py
```diff
@@ -258,6 +258,7 @@ def get_reverse(self):
                 other.add_action(['GAIN_PAIR', action[1], action[2]])
             elif action[0] == 'GAIN_PAIR':
                 other.add_action(['LOSE_PAIR', action[1], action[2]])
+        other.actions = reversed(other.actions) # Play the reverse recipe in the reverse order
```
mliu49: `reversed` returns an iterator. Do we expect any issues with that?

mazeau: I didn't think that would cause any issues.  I'm able to run it perfectly fine.  Maybe @rwest feels differently?

rwest: Hard to anticipate, but might as well try `other.actions = list(reversed(other.actions))`.

Then should (rebase and) look into why the unit tests failed.

mliu49: Actually, I have a better suggestion: use reversed directly on the for loop, `for action in reversed(self.actions)`

mliu49:
The failing unit tests are all related to reaction generation:
```
ERROR: test_reaction_degeneracy_independent_of_generatereactions_direction
  File "/home/travis/build/ReactionMechanismGenerator/RMG-Py/rmgpy/data/kinetics/kineticsTest.py", line 565, in test_reaction_degeneracy_independent_of_generatereactions_direction
    self.assertEqual(forward_reactions[0].degeneracy, reverse_reactions[0].degeneracy,
IndexError: list index out of range
FAIL: Test that the ``react_all`` function works in serial
  File "/home/travis/build/ReactionMechanismGenerator/RMG-Py/rmgpy/rmg/reactTest.py", line 124, in test_react_all
    self.assertEqual(len(flat_rxn_list), 44)
AssertionError: 25 != 44
FAIL: Test that the ``react_all`` function works in parallel using Python multiprocessing
  File "/home/travis/build/ReactionMechanismGenerator/RMG-Py/rmgpy/rmg/reactTest.py", line 149, in test_react_all_parallel
    self.assertEqual(len(flat_rxn_list), 44)
AssertionError: 25 != 44
```

I don't have any intuition for why this change would have these effects, so you would need to do some debugging.

rwest:
I can't imagine why this change would cause a
```
5124FAIL: A general test for a PDep job in Arkane
5125----------------------------------------------------------------------
5126Traceback (most recent call last):
5127  File "/home/travis/build/ReactionMechanismGenerator/RMG-Py/arkane/pdepTest.py", line 107, in test_pdep_job
5128    self.assertAlmostEquals(rxn.kinetics.get_rate_coefficient(1000.0, 1.0), 88.88253229631246)
5129AssertionError: 88.9961408998604 != 88.88253229631246 within 7 places (0.11360860354794511 difference)
```
that was showing up in Travis, so I've triggered a new travis build, in case it's something weird or random.

mliu49:
It seems that a lot of new builds are failing due to that. It's discussed somewhat in #1682, but we were not able to reach a conclusion. It is possible that this new wave of failures is caused by a new version of some dependency, but the idea that math could change (by 0.1!) is quite odd.

PR #1832, opened 2019-11-26
fixed the solvation GAV heuristic

yunsiechung:
### Motivation or Problem
Solvation group additivity was estimating solute data by averaging over all molecules in species.molecule list (so it was averaging over all resonance structure). However, when the solvation solute groups were fitted from the experimental values, they were fitted using the SMILES of the most stable resonance structure in gas phase. Therefore, averaging over all resonance structures is not a correct way. Instead, it should be estimating solute data from only the most stable structure found by gas-phase. 

Extra Note:
Once might be concerned that the most stable resonance structure might change in liquid phase. But the change in the resonance structures when a species is solvated is already reflected in the experimental solvation energies, which are measured for a certain species in general, not for its particular resonance structures. Since we fitted the solute group to these experimental solvation energies using the SMILES of the most stable gas phase resonance structures, we should be consistent with using the same SMILES when we estimate the solute data.

### Description of Changes
In `def get_solute_data_from_groups`, instead of iterating over all molecules and averaging, it only calculates the solute data for the first item in the molecule list,  `species.molecule[0]` as it is the SMILES of the most stable gas phase resonance structure found.

### Testing
I did not add any test since the unittest for `get_solute_data_from_groups` already exists.
I can add more unittests if necessary.

### Reviewer Tips

### DIFF
--- a/rmgpy/data/solvation.py
+++ b/rmgpy/data/solvation.py
@@ -702,32 +702,14 @@ class SolvationDatabase(object):
         additivity method. If no group additivity values are loaded, a
         :class:`DatabaseError` is raised.
         
-        It averages (linearly) over the desciptors for each Molecule (resonance isomer)
-        in the Species.
+        It estimates the solute data for the first item in the species's
+        molecule list because it is the most stable resonance structure found
+        by gas-phase thermo estimate.
         """
-        solute_data = SoluteData(0.0, 0.0, 0.0, 0.0, 0.0)
-        count = 0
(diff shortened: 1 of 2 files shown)


yunsiechung:
@mliu49 The existing unit test checks the values, but the one checked does not have any resonance structure, so the value did not change with this code. 
When I compared the previous code with this code by comparing their estimated solvation values, they do change but the changes are not substantial in general.

Issue #1833, opened 2019-11-30
numpy.linalg.linalg.LinAlgError: SVD did not converge during rmgpy.reaction.Reaction.reverse_arrhenius_rate


rwest:
### Bug Description
During calls to `rmgpy.reaction.Reaction.reverse_arrhenius_rate` I am now quite often getting `numpy.linalg.linalg.LinAlgError: SVD did not converge in Linear Least Squares` errors, that cause RMG to immediately crash.

One context was @ajuluc and I working on a tool to check collision limit violations.
Another was our chemkin mechanism importer script. 

It might be that there are NAN values creeping in, but I don't see why that would happen in these contexts, and for the importer script it never used to be a problem, and it wasn't a problem when @chrisbNEU ran the same input on his ubuntu machine on Wednesday, IIRC. So I suspect it might be an OS or compiler or hardware specific problem.
Based on some discussion at https://github.com/numpy/numpy/issues/12941  etc.

Partly just sharing this here as a place for others to comment if they run into similar issues.

Here's some stack trace:
```
  File "rmgpy/reaction.py", line 791, in rmgpy.reaction.Reaction.generate_reverse_rate_coefficient
  File "rmgpy/reaction.py", line 872, in rmgpy.reaction.Reaction.generate_reverse_rate_coefficient
  File "rmgpy/reaction.py", line 854, in rmgpy.reaction.Reaction.generate_reverse_rate_coefficient
  File "rmgpy/reaction.py", line 829, in rmgpy.reaction.Reaction.generate_reverse_rate_coefficient
  File "rmgpy/reaction.py", line 788, in rmgpy.reaction.Reaction.reverse_arrhenius_rate
  File "rmgpy/kinetics/arrhenius.pyx", line 173, in rmgpy.kinetics.arrhenius.Arrhenius.fit_to_data
  File "/Users/rwest/anaconda/envs/rmg3/lib/python3.7/site-packages/numpy/linalg/linalg.py", line 2156, in lstsq
    x, resids, rank, s = gufunc(a, b, rcond, signature=signature, extobj=extobj)
  File "/Users/rwest/anaconda/envs/rmg3/lib/python3.7/site-packages/numpy/linalg/linalg.py", line 101, in _raise_linalgerror_lstsq
    raise LinAlgError("SVD did not converge in Linear Least Squares")
numpy.linalg.linalg.LinAlgError: SVD did not converge in Linear Least Squares
```

And here's my numpy config
```
[...]

PR #1834, opened 2019-12-01
Enable fitting and reversing of negative reaction rates.

rwest:
### Motivation or Problem
Sometimes people represent a complex rate expression as the sum of Arrhenius, or sum of PDepArrhenius terms, one or more of which is negative. Although an overall negative rate expression is not allowed, one part of the sum may be. 
When generating reverse rate expressions, these multiple sum type expressions are reversed one part at a time. The function to generate a reverse rate expression for an individual Arrhenius expression does so by evaluating the rate at some array of temperatures, calculating the reverses (from the equilibrium constants) then  doing least squares regression to fit a new Arrhenius. But that last step involves taking logarithms of the rates, which leads to `nan` if they're negative, which then messes up the least squares regression.

This was the underlying cause of #1833 (not problems in my LAPACK implementation, as originally suspected)

### Description of Changes
If trying to fit a negative rate, negate it then fit it then negate the A factor.
Also ensure that all k(T) have the same sign, and are all finite.
Also added a couple of unit tests, that would fail without this fix.

### Testing
Unit tests have been added.
(The second one doesn't really test the result, just tries not to crash. But it used to crash)

### DIFF
--- a/rmgpy/kinetics/arrhenius.pyx
+++ b/rmgpy/kinetics/arrhenius.pyx
@@ -38,6 +38,8 @@ from rmgpy.exceptions import KineticsError
 from rmgpy.kinetics.uncertainties import rank_accuracy_map
 from rmgpy.molecule.molecule import Bond
 
+# Prior to numpy 1.14, `numpy.linalg.lstsq` does not accept None as a value
+RCOND = -1 if int(np.__version__.split('.')[1]) < 14 else None
 ################################################################################
 
 cdef class Arrhenius(KineticsModel):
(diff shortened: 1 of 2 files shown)


rwest:
The Codacy complaints are about a few spaces missing after commas in some code generated by our own `__repr__` calls.

rmgpy/kinetics/arrhenius.pyx
```diff
@@ -152,6 +152,13 @@ cdef class Arrhenius(KineticsModel):
         data.
         """
         import scipy.stats
+        assert all(np.isfinite(klist)), "Rates must all be finite, not inf or NaN"
+        if any(klist<0):
+            assert all(klist<0), "Rates must all be positive or all be negative."
```
mliu49: Could you convert these assertions to `raise ValueError` instead?

rwest: Done. as a fixup, when rebasing to latest master.

Issue #1726, opened 2019-09-23
Python 3 finalization tasks


mliu49:
Once #1724 is merged, the bulk of the Python 3 transition will be completed. However, there are final testing and clean up tasks which will still need to be done. The purpose of this issue is to keep track of these tasks and make sure that they are done before releasing RMG 3.0.

To do:
- [x] Remove all `__future__` imports #1730 
- [x] Update all IPython notebooks #1735
- [ ] Update RMG-database scripts and notebooks https://github.com/ReactionMechanismGenerator/RMG-database/pull/364
- [x] Remove dde models from RMG-database ReactionMechanismGenerator/RMG-database#358
- [x] Update documentation as needed @hwpang  (particularly regarding installation @mliu49  and @amarkpayne ) #1767
- [ ] Update global uncertainty analysis (transition to MUQ 2) #1738 

Testing:
- [x] Run a pdep job @alongd
- [x] Run a job with ml thermo @yunsiechung 
- [ ] Run a job with QMTP @agnes (debug conformer issue @dranasinghe and @alongd )
- [x] Run a catalysis job @maforsuelo 
- [x] Run a parallel job - natural gas combustion case for 3d 20h no error but see for warning messages below
- [x] Run a liquid job @hwpang 
- [x] Run a large pdep network in Arkane @alongd
- [x] Run an isotopes job @goldmanm 
- [x] Run sensitivity and local uncertainty analysis @oscarwumit 
- [x] Test all scripts @yunsiechung 
- [x] Test all IPython notebooks

Note that some of jobs are run by RMG-tests, but the point here is to run a longer job and ideally a case that is important to you.

Please comment (or edit) if you have suggestions for other items to add to either list.

ajocher:
Running the examples/rmg/1,3-hexadiene case without pdep: after 6:52 min run time this error happens

  File "/home/ajocher/codes/RMG/RMG3/RMG-Py/rmg.py", line 107, in <module>
    main()
  File "/home/ajocher/codes/RMG/RMG3/RMG-Py/rmg.py", line 101, in main
    rmg.execute(**kwargs)
  File "/home/ajocher/codes/RMG/RMG3/RMG-Py/rmgpy/rmg/main.py", line 651, in execute
    self.initialize(**kwargs)
  File "/home/ajocher/codes/RMG/RMG3/RMG-Py/rmgpy/rmg/main.py", line 486, in initialize
    self.load_database()
[...]

PR #1840, opened 2019-12-03
Python3 Script check

yunsiechung:
### Motivation or Problem
I tested all externally-used scripts in python 3 version RMG. This is related to the issue #1726.

* One thing I could not check is running semi-empirical QM job using `thermoEstimator.py` in `examples/scripts/thermoEstimator/run_QM.sh` because I don't have the mopac license. Since it only throws an error that I don't have the mopac license, I don't expect `thermoEstimator.py` to have any error.

The internally-used scripts like `generateTree.py`, `machineWriteDatabase.py`,  and `rmg2to3.py` are not checked in this test.

### Description of Changes
- All the scripts ran fine in python 3, but some were missing examples or had minor error with example files, so I added / fixed the examples.
- `isotopes.py` does not have an example because running this example takes too much time. This script seems to run fine when I checked it.
- `convertFAME.py` is deleted because it is no longer used.
- `reduction` example folder is deleted because it uses the script that is no longer available.
- One major change is that all the examples relevant to scripts are moved under `examples/scripts`.

### Testing
I ran all the script example bash scripts and everything ran successfully.

### Reviewer Tips
You can run bash scripts in `examples/scripts` to check that everything works fine


yunsiechung:
travis-ci/push failed on the following:
```
A general test for a PDep job in Arkane ... FAIL

FAIL: A general test for a PDep job in Arkane

  File "/home/travis/build/ReactionMechanismGenerator/RMG-Py/arkane/pdepTest.py", line 107, in test_pdep_job

    self.assertAlmostEquals(rxn.kinetics.get_rate_coefficient(1000.0, 1.0), 88.88253229631246)

AssertionError: 88.9961408998604 != 88.88253229631246 within 7 places (0.11360860354794511 difference)
````

This does not seem to be related to any of my code changes, and I'm not sure what's causing this

mliu49:
I found the related issue: #1682 

It seems like we never figured out the root cause, but it went away on its own in each case.

yunsiechung:
I force-pushed this again after rebasing to official/master and got the same error as before with travis-ci/push

PR #1842, opened 2019-12-04
Collection of small fixes

mliu49:
### Motivation or Problem
This PR addresses a number of minor issues in preparation for the 3.0 release.

### Description of Changes
Please see the individual commit messages for the changes.

### Testing
I think passing unit tests should be sufficient since these are all relatively minor.

### Reviewer Tips
Thoughts on the approaches chosen here would be nice.

I think the increased tolerance for the pdep test may be worth additional discussion. There are still some ongoing efforts to try and identify the root cause of that issue.

### DIFF
--- a/arkane/pdepTest.py
+++ b/arkane/pdepTest.py
@@ -104,7 +104,9 @@ class ArkaneTest(unittest.TestCase):
             reaction_list = read_reactions_block(chem, dictionary)
         rxn = reaction_list[0]
         self.assertIsInstance(rxn.kinetics, Chebyshev)
-        self.assertAlmostEquals(rxn.kinetics.get_rate_coefficient(1000.0, 1.0), 88.88253229631246)
+        # Accept a delta of 0.2, which could result from numerical discrepancies
+        # See RMG-Py #1682 on GitHub for discussion
+        self.assertAlmostEquals(rxn.kinetics.get_rate_coefficient(1000.0, 1.0), 88.88253229631246, delta=0.2)
 
         files = [f for f in os.listdir(os.path.join(self.directory, 'sensitivity', ''))
                  if os.path.isfile(os.path.join(self.directory, 'sensitivity', f))]
@@ -116,7 +118,7 @@ class ArkaneTest(unittest.TestCase):
                 if '1000.0' in line:
                     break
         sa_coeff = line.split()[-2]
-        self.assertEquals(float(sa_coeff), -8.23e-6)
+        self.assertAlmostEquals(float(sa_coeff), -8.23e-6, delta=0.02e-6)
 
     @classmethod
     def tearDown(cls):
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -2758,63 +2758,6 @@ class KineticsFamily(Database):
             for species in reaction.reactants + reaction.products:
                 species.generate_resonance_structures()
 
-    def get_w0(self, rxn):
-        """
-        calculates the w0 for Blower Masel kinetics by calculating wf (total bond energy of bonds formed)
-        and wb (total bond energy of bonds broken) with w0 = (wf+wb)/2
-        """
-        mol = None
-        a_dict = {}
-        for r in rxn.reactants:
-            m = r.molecule[0]
-            a_dict.update(m.get_all_labeled_atoms())
-            if mol:
-                mol = mol.merge(m)
-            else:
-                mol = m.copy(deep=True)
-
-        recipe = self.forward_recipe.actions
-
-        wb = 0.0
-        wf = 0.0
-        for act in recipe:
-
-            if act[0] == 'BREAK_BOND':
-                bd = mol.get_bond(a_dict[act[1]], a_dict[act[3]])
-                wb += bd.get_bde()
-            elif act[0] == 'FORM_BOND':
-                bd = Bond(a_dict[act[1]], a_dict[act[3]], act[2])
-                wf += bd.get_bde()
-            elif act[0] == 'CHANGE_BOND':
-                bd1 = mol.get_bond(a_dict[act[1]], a_dict[act[3]])
-
-                if act[2] + bd1.order == 0.5:
-                    mol2 = None
-                    for r in rxn.products:
-                        m = r.molecule[0]
-                        if mol2:
-                            mol2 = mol2.merge(m)
-                        else:
-                            mol2 = m.copy(deep=True)
-                    bd2 = mol2.get_bond(a_dict[act[1]], a_dict[act[3]])
-                else:
-                    bd2 = Bond(a_dict[act[1]], a_dict[act[3]], bd1.order + act[2])
-
-                if bd2.order == 0:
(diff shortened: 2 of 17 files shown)


mliu49:
Changing the tolerance for the pdep test worked for that assertion, but the numerical differences also carry through to the sensitivity analysis, which is causing the test to fail.

rmgpy/molecule/translator.py
```diff
@@ -531,6 +531,8 @@ def _write(mol, identifier_type, backend):
             logging.debug('Backend {0} is not able to generate {1} for this molecule:\n'
                           '{2}'.format(option, identifier_type, mol.to_adjacency_list()))
 
+    logging.error('Unable to generate identifier for this molecule:\n{0}'.format(mol.to_adjacency_list()))
```
alongd: Maybe instead add the adjlist to the error message below?

mliu49: I thought about doing that originally. One consideration is that with the current setup, logging goes into RMG.log, while the error goes into stderr, which is only printed to the terminal (or slurm/sge output files). I can move it though.

alongd: Fair enough, then let's keep it. I'm OK with all other commits too.

rmgpy/data/kinetics/family.py
```diff
@@ -2758,63 +2758,6 @@ def add_atom_labels_for_reaction(self, reaction, output_with_resonance=True):
             for species in reaction.reactants + reaction.products:
                 species.generate_resonance_structures()
 
-    def get_w0(self, rxn):
-        """
-        calculates the w0 for Blower Masel kinetics by calculating wf (total bond energy of bonds formed)
-        and wb (total bond energy of bonds broken) with w0 = (wf+wb)/2
-        """
-        mol = None
-        a_dict = {}
```
amarkpayne: I checked and these functions are essentially the same, but the one we are keeping in arrhenius.pyx has a few additional PEP-8 changes. Can you make these changes?

- `ADict` --> `a_dict`
- Opitionally, things like `bd2bde` --> `bd2_bde`. There are a couple of these variables that I different, but it is PEP-8 as is, so only change if you prefer it

rmgpy/data/kinetics/family.py
```diff
@@ -2758,63 +2758,6 @@ def add_atom_labels_for_reaction(self, reaction, output_with_resonance=True):
             for species in reaction.reactants + reaction.products:
                 species.generate_resonance_structures()
 
-    def get_w0(self, rxn):
```
amarkpayne: These functions are also listed under rmgpy.data.kinetics.family in rmg2to3.py. I am not sure if this needs to be updated, as what you really want is for these functions to instead call Arrhenius. But I doubt this will come up anyways

rmgpy/thermo/thermoengine.py
```diff
@@ -98,14 +98,15 @@ def process_thermo_data(spc, thermo0, thermo_class=NASA, solvent_name=''):
...
 
-    if thermo.__class__ != thermo0.__class__:
-        # Compute RMS error of overall transformation
-        Tlist = np.array([300.0, 400.0, 500.0, 600.0, 800.0, 1000.0, 1500.0], np.float64)
-        err = 0.0
-        for T in Tlist:
-            err += (thermo.get_heat_capacity(T) - thermo0.get_heat_capacity(T)) ** 2
-        err = math.sqrt(err / len(Tlist)) / constants.R
-        # logging.log(logging.WARNING if err > 0.2 else 0, 'Average RMS error in heat capacity fit to {0} = {1:g}*R'.format(spc, err))
+    # if thermo.__class__ != thermo0.__class__:
```
amarkpayne: I know it is better to comment out so that it is easy to find in case anyone else ever needs it, but calculating the error in the fit is fairly straight forward, so if this is something we decide to implement in the future it won't take long to write from scratch, and the resulting implementation might be better for it. I think we can delete this code altogether, but if you prefer it can stay commented out

arkane/pdepTest.py
```diff
@@ -104,7 +104,9 @@ def test_pdep_job(self):
             reaction_list = read_reactions_block(chem, dictionary)
         rxn = reaction_list[0]
         self.assertIsInstance(rxn.kinetics, Chebyshev)
-        self.assertAlmostEquals(rxn.kinetics.get_rate_coefficient(1000.0, 1.0), 88.88253229631246)
+        # Accept a delta of 0.2, which could result from numerical discrepancies
+        # See RMG-Py #1682 on GitHub for discussion
+        self.assertAlmostEquals(rxn.kinetics.get_rate_coefficient(1000.0, 1.0), 88.88253229631246, delta=0.2)
```
amarkpayne: What do you think about making this a percentage test. 

```
self.assertLess((abs(rxn.kinetics.get_rate_coefficient(1000.0, 1.0) - 88.88253229631246)) / 88.88253229631246,
                        0.01)
```

rmgpy/molecule/translator.py
rwest: Might not be relevant here but the discussion reminded me:  I am a fan of using `logging.exception()` in an exception handler to add context and  make sure the original stack trace ends up in the log.  Can then `raise()` the original exception as if you hadn’t caught it

rwest: Might not be relevant here but the discussion reminded me:  I am a fan of using `logging.exception()` in an exception handler to add context and  make sure the original stack trace ends up in the log.  Can then `raise()` the original exception as if you hadn’t caught it

arkane/pdepTest.py
mliu49: I was disappointed that `assertAlmostEquals` did not support a percentage delta, but I think I would prefer the simplicity of using it with a fixed delta over calculating the percentage. I could change the delta to 0.01*88.88 if you prefer.

amarkpayne: No that is okay, we can leave it as is then. I do like the simplicity of just using delta

rmgpy/thermo/thermoengine.py
mliu49: I removed the code. You're right that it would be easy to rewrite if ever needed.

rmgpy/data/kinetics/family.py
mliu49: I don't think the function has existed long enough for anyone to have scripts using it (except Matt). I added a change to remove them from rmg2to3.py.

rmgpy/data/kinetics/family.py
mliu49: Done

rmgpy/molecule/translator.py
mliu49: Yes, I was actually hoping that this would be a good opportunity to use logging.exception(), but that has to be called from an exception handler. It seemed easier to log this message once in this function rather than logging the exception in each function which calls it.

I do think there are a lot of instances where we use `logging.error` when catching exceptions where `logging.exception` could be used instead.

mliu49:
There is one new commit replacing the `mock` external library with `unittest.mock`.

Issue #1843, opened 2019-12-05
scipy.linalg.eig_banded appears to give unreliable results inside solve_schrodinger_equation


amarkpayne:
### Bug Description
This was discovered in the course of debugging #1682. I am opening this up as a separate issue to refocus our efforts, and because it is possible that this is not the only issue affecting the failing pdep test.

Inside `HinderedRotor.solve_schrodinger_equation` (rmgpy.statmech.torsion.pyx) we construct a complex-valued Hamiltonian matrix banded in lower triangular form (i.e. the full Hamiltonian is a 401x401 matrix, but here we only store the diagonals that are non-zero that appear on and below the main diagonal. Therefore, `H` is only 6x401, but represents a full 401x401 Hamiltonian). This banded representation then gets passed to `scipy.linalg.eig_banded`, which takes as input a complex-valued Hamiltonian represented in this way and returns the eigenvalues.

Here is where the problem occurs. When running two different versions of RMG-Py (see below for more details), the exact same Hamiltonian input into `scipy.linalg.eig_banded` yields completely different eigenvalues, where some of them don't even match to 1 significant figure. As far as we can tell the Hamiltonian in question (which I have attached by outputting as a list to a YAML file) is not poorly conditioned. It is worth noting that this can be hard to reproduce, because it appears to be system dependent. I am seeing this issue based on current master, but others aren't necessarily.

What is reproducible though is that the eigenvalues are different if you request `scipy.linalg.eig_banded` to return the eigenvectors as well, which as noted in the documentation for the underlying FORTRAN LAPACK function `chbevd.f` at http://www.netlib.org/lapack/explore-html/d5/db8/chbevd_8f_source.html will cause a different algorithm to be used. It is not surprising that the eigenvalues will be different if a different algorithm is used, but it is surprising if they are so different that they don't even agree to 1 significant digit. The eigenvalues were different when the eigenvectors were requested in both RMG-Py versions, and worse they were still different across the different RMG-Py versions.

This is potentially a major bug, because we use these eigenvalues (energies) to calculate the partition function of the hindered rotor, which of course is the most important quantity in all of statmech which we derive everything else from. The resulting partition functions can differ in as much as the second significant digit, so it is important that we figure out what is going on here.

### How To Reproduce
Finding different versions of RMG-Py where the output of `scipy.linalg.eig_banded` differs can be difficult to find. It appears to be system dependent, and we really don't have a good understanding of why at this point.

[...]

PR #1844, opened 2019-12-05
Create the dense Hamiltonian and use np.linalg.eigh to find eigenvalues

amarkpayne:
### Motivation or Problem
A potential solution for #1843. In initial testing, we have found that using np.linalg.eigh to calculate the eigenvalues from the dense Hamiltonian appears to give consistent results across different platforms. Further testing is needed to confirm this. If this is the case then this PR will close the issue. 

### Description of Changes

- Convert banded Hamiltonian into dense Hamiltonian. Only the lower triangle and main diagonal elements are constructed, which is all np.linalg.eigh needs.

- Use replace scipy.linalg.eig_banded with np.linalg.eigh

### Testing
Verify that this passes on the newest Travis worker version

### DIFF
--- a/arkane/pdepTest.py
+++ b/arkane/pdepTest.py
@@ -104,7 +104,7 @@ class ArkaneTest(unittest.TestCase):
             reaction_list = read_reactions_block(chem, dictionary)
         rxn = reaction_list[0]
         self.assertIsInstance(rxn.kinetics, Chebyshev)
-        self.assertAlmostEquals(rxn.kinetics.get_rate_coefficient(1000.0, 1.0), 88.88253229631246)
+        self.assertAlmostEquals(rxn.kinetics.get_rate_coefficient(1000.0, 1.0), 88.9003516559412)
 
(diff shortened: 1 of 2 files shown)


mliu49:
FYI, `scipy.linalg.eigh` gives slightly different results than `numpy.linalg.eigh`.

amarkpayne:
> FYI, `scipy.linalg.eigh` gives slightly different results than `numpy.linalg.eigh`.

Of course it is haha. Is it the same slightly different results as anything else we have tried?

mliu49:
It seems to me like what this change implies is that the lapack algorithm for finding eigenvalues of banded sparse matrices is less consistent (or more sensitive to floating point error) than the one for dense matrices.

If we are confident in that statement, then using a dense matrix is an acceptable solution. However, I think we should directly construct the dense matrix instead of converting it from the sparse representation.

If we are not confident in that statement, I would be more inclined to increase the tolerance for passing the pdep test, which would mean that we accept some level of numerical variation in the results.

mjohnson541:
Ultimately, loosened tolerances for test.

Issue #1153, opened 2017-08-17
API not updated or accurate


goldmanm:
Our API is only somewhat automatic. It is automatic below the class level, but requires manual addition for new classes and modules.

The API seems to be years behind in adding new methods.

Also note that the Wiki doesn't describe the API updating.

It seems like we could do 3+ things:

1. move to an entirely automated api (we will lose the pretty formatting but everything will be accurate and docstrings will be utilized more effectively)
1. manually update the api pages that require manual updating (anything not stored in a Class structure)
1. deprecate the api (source code for the win!)
1. keep as status quo (confusion is a source of strength!)

amarkpayne:
I think that having an accurate and up to date API is more important than having an out of date API that looks pretty. What would we use to do this? Can we try this out just to see what it would look like?

mliu49:
#1273 updates some portions of the API. I mainly fixed updated things that I noticed while doing other things, so it isn't a thorough update.

We can definitely consider a fully automated API, although we would need to look into how to do that and exactly what that would look like.

PR #1845, opened 2019-12-10
API documentation updates

mliu49:
### Motivation or Problem
The API documentation is substantially out-of-sync with the current code structure. #1153 is definitely still relevant.

### Description of Changes
This PR adds a bunch of un-documented modules to the API documentation and also fixes various issues with docstrings so that they can be displayed properly.

### Testing
The documentation has been compiled to verify that the changes look correct.

### To Do
This PR does not exhaustively update the API documentation, i.e. there are still portions of the code which are not included. While adding documentation pages is easy, some decision-making is required in choosing how to organize and format the documentation.

Here's a partial list of modules which are not documented:

- [x] arkane.common
- [x] arkane.explorer
- [x] arkane.util
- [ ] rmgpy.kinetics.arrhenius.ArrheniusEP and ArrheniusBM
- [ ] rmgpy.kinetics.surface
- [ ] rmgpy.kinetics.uncertainties
- [ ] rmgpy.statmech.ndTorsions
- [ ] rmgpy.thermo.thermoengine
- [ ] rmgpy.tools

I'm not sure how many of these could/should be documented before the release. Contributions from others would be necessary if we want them to be documented.

### DIFF
--- a/arkane/__init__.py
+++ b/arkane/__init__.py
@@ -32,6 +32,7 @@ initialize imports
 """
 
 from arkane.common import ArkaneSpecies
+from arkane.explorer import ExplorerJob
 from arkane.kinetics import KineticsJob
 from arkane.main import Arkane
 from arkane.pdep import PressureDependenceJob
--- a/arkane/ess/log.py
+++ b/arkane/ess/log.py
@@ -52,88 +52,78 @@ class Log(object):
 
     def get_number_of_atoms(self):
         """
-        Return the number of atoms in the molecular configuration used in
-        the MolPro log file.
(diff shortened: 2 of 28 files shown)


alongd:
Looking good!
Would we like to automate the API documentation build & deployment?

mliu49:
Build and deployment are automated as part of the overall documentation. Keeping it up to date is only partially automated due to how we have it set up currently.

Long term, it would definitely be advisable to refactor the API to be completely auto-generated (and completely dependent on docstrings), but it would take a bit of effort to do so.

Issue #1846, opened 2019-12-10
Restarting from a seed mechanism does not handle empty filters properly


amarkpayne:
### Bug Description
Trying to restart from a seed mechanism throws the following error if the RMG job used to generate the seed mechanism did not use filtering:

```
Traceback (most recent call last):
  File "rmg.py", line 111, in <module>
    main()
  File "rmg.py", line 105, in main
    rmg.execute(**kwargs)
  File "rmgpy/rmg/main.py", line 649, in execute
    self.initialize(**kwargs)
  File "rmgpy/rmg/main.py", line 615, in initialize
    self.initialize_reaction_threshold_and_react_flags()
  File "rmgpy/rmg/main.py", line 1538, in initialize_reaction_threshold_and_react_flags
    unimolecular_threshold_restart = f.get('unimolecular_threshold').value
AttributeError: 'NoneType' object has no attribute 'value'
```

When designing the restart feature, we decided to handle this situation in a Try/Except block, where if the filters were missing we would assume the previous job did not use filtering, and set the current filters accordingly (see the code [here](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/cafae8d56818a1fcba88033e69a71acc62d71a15/rmgpy/rmg/main.py#L1561-L1593)). However, the Except block expects a KeyError to be thrown instead of an AttributeError.

### How To Reproduce
Run the superminimal model to completion. Then, try restarting from the generated seed mechanism.

### Additional Context
I should have this fixed up soon, along with some functional tests to help us catch errors like this.

PR #1847, opened 2019-12-10
Restart attribute error

amarkpayne:
### Motivation or Problem
See #1846 

### Description of Changes
This PR address the issue identified in #1846, as well as dealing with a deprecation warnings we were getting from h5py. Finally, this PR adds functional tests that run restart jobs (both with and without filters) to try to catch errors like this in the future.

### Testing
The functional tests added should catch the error this PR addresses

### Additional note
The functional tests simply check that the restart jobs can complete. A more robust test would be ensure that the same model is achieved, but this would require running a full RMG job, and then restarting from an intermediate seed. For the sake of time in running our tests, this was not implemented.

On my computer, the whole suite of tests in mainTest.py including the new functional tests can run in less than 150 seconds, so I don't think this will drastically increase our testing time.


rmgpy/rmg/mainTest.py
```diff
@@ -183,6 +184,78 @@ def test_make_cantera_input_file(self):
...
+        cls.outputDir = os.path.join(cls.testDir, 'output_no_filters')
+        cls.databaseDirectory = settings['database.directory']
+
+        os.mkdir(cls.outputDir)
+        initialize_log(logging.INFO, os.path.join(cls.outputDir, 'RMG.log'))
+
+        cls.rmg = RMG(input_file=os.path.join(cls.testDir, 'restart_no_filters.py'),
+                      output_directory=os.path.join(cls.outputDir))
+
+        cls.rmg.execute()
```
mliu49: I feel like executing the job should be part of the test method rather than in setup.

amarkpayne: I agree. Sorry, I copied from the tests above, but I agree this makes more sense.

PR #1848, opened 2019-12-12
Tree generation improvements

mjohnson541:
Generate resonance structures before calculating thermo for tree generation training sets and add SABIC_aromatics to the thermo libraries used during tree generation.


rmgpy/data/kinetics/family.py
```diff
@@ -3798,11 +3798,15 @@ def fix_labels_mol(mol, root_labels):
             if estimate_thermo:
                 for j, react in enumerate(r.item.reactants):
                     if rxns[i].reactants[j].thermo is None:
-                        rxns[i].reactants[j].thermo = tdb.get_thermo_data(react)
+                        therm_spc = deepcopy(rxns[i].reactants[j])
```
mliu49: Does it matter if you use `react` or `rxns[i].reactants[j]`? If not, could you just loop over `rxns[i].reactants`?

mliu49:
In RMG-tests, there are a lot of reactions which were estimated by rate rules on master but estimated by exact match to training reactions on this branch. Do you have any idea why that would be?

PR #1849, opened 2019-12-12
ND Hindered Rotor Improvements

mjohnson541:
This has a number of numerical improvements for the calculation of ND classical and semiclassical hindered rotor partition functions.  I've removed use of SmoothBivarateSpline interpolation as this resulted in some negative fitted internal reduced moment of inertia values and I've fixed issues with the integration routine.

Also includes miscellaneous Arkane improvements: lets Arkane deal with RateUncertainty objects, makes the semiclassicalND rotors example consistent with the other Toulene examples, and adds a error trigger in internal reduced moment of inertia calculations that prevents use of 0-indexed atom numbers that result in all kinds of weird outputs from the function.

### DIFF
--- a/rmgpy/statmech/conformer.pyx
+++ b/rmgpy/statmech/conformer.pyx
@@ -384,7 +384,8 @@ cdef class Conformer(RMGObject):
         elif pivots[0] in top1 and pivots[1] in top1:
             raise ValueError('Both pivot atoms included in top; you must specify only one pivot atom that belongs'
                              ' with the specified top.')
-
+        elif 0 in top1:
+            raise ValueError('Top must be one indexed, top1: {}'.format(top1))
         # Enumerate atoms in other top
         top2 = [i + 1 for i in range(n_atoms) if i + 1 not in top1]
 
--- a/rmgpy/statmech/ndTorsions.py
+++ b/rmgpy/statmech/ndTorsions.py
@@ -31,6 +31,7 @@ import logging
 import os
 import os.path
 import subprocess
+import itertools
 
 import numdifftools as nd
 import numpy as np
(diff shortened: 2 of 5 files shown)


rmgpy/statmech/conformer.pyx
```diff
@@ -384,7 +384,8 @@ cdef class Conformer(RMGObject):
         elif pivots[0] in top1 and pivots[1] in top1:
             raise ValueError('Both pivot atoms included in top; you must specify only one pivot atom that belongs'
                              ' with the specified top.')
-
+        elif 0 in top1:
```
alongd: a very minor comment: we don't need `elif` if the above condition raises an error
a more important comment: maybe repeat the check below for `top2` as well which could have the `0` index instead of `top1`

rmgpy/statmech/ndTorsions.py
```diff
@@ -597,12 +598,22 @@ def calc_partition_function(self, T):
         """
         calculate the classical/semiclassical partition function at a given temperature
         """
-        rngs = [(0.0, 2.0 * np.pi) for x in range(len(self.pivots))]
 
         def f(*phis):
-            return self.rootD(*phis) * np.exp(-self.V(*phis) / (constants.R * T))
-
-        intg = inte.nquad(f, ranges=rngs)[0]
+            return (self.rootD(*phis)) * np.exp(-self.V(*phis) / (constants.R * T))
```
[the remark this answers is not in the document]
mjohnson541: done

rmgpy/statmech/ndTorsions.py
```diff
@@ -597,12 +598,22 @@ def calc_partition_function(self, T):
...
+        
+        rphis = np.linspace(0, 2.0*np.pi, 30)
```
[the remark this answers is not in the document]
mjohnson541: I never seen anyone write *args operators that way and I think the spacing would make it more easy to confuse with multiplication with a missing first argument. Since it's an operation on the variable itself I think it makes the most sense to not have a space.

rmgpy/statmech/conformer.pyx
mjohnson541: elif just keeps it in one if block, which I think makes it a little bit easier to read (all the error checks separated in one block), although it makes it a bit less obvious that the checks are distinct.  

You can't run the check for top2 because top2 is inferred from top1. Essentially we only know if the input was 0 indexed if we are given a value we know we shouldn't see.

rmgpy/statmech/conformer.pyx
```diff
@@ -384,7 +384,8 @@ cdef class Conformer(RMGObject):
...
+            raise ValueError('Top must be zero indexed.')
```
alongd: Should the message say "should be one-indexed" instead?

mjohnson541: XD...yes done

Issue #1850, opened 2019-12-13
Dependency issue causes significant memory leak on RMG Py3


kspieks:
### Bug Description
Significant memory leaks have been observed when using an rmg_env created around early Oct 2019. Memory leaks are not observed when using a recently created rmg_env, whose various dependencies are at an updated version.

Below is the output from diffing the output from conda list from the two rmg_env (this output is tab spaced and easier to read that exporting the environment and diffing the yml files). The older rmg_env is listed first, hence the older version number for each dependency. 

It is unclear which of these dependencies is responsible for the memory leak, but no memory leak is observed with the newer dependencies. **The recommendation is to remove and recreate the rmg_env if you created an rmg_env for Python 3.7 around early Oct 2019 and you think you are experiencing memory leaks.** Using the exact dependencies in this file resolves the issue (just change the extension to .yml since GitHub only lets me upload .txt files, not .yml files), though just using `conda env create -f environment.yml` and letting conda pull the latest dependencies should also work well.
[Functional_rmg_env.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/3961523/Functional_rmg_env.txt)

```
diff Original_RMG_Server_rmg_env.txt Erebor_rmg_env.txt 
6c6
< absl-py                   0.8.0                    py37_0  
> absl-py                   0.8.1                    py37_0  
9c9
< attrs                     19.2.0                     py_0  
> attrs                     19.3.0                     py_0  
22c22
< cffi                      1.12.3           py37h2e261b9_0  
> cffi                      1.13.2           py37h2e261b9_0  
24a25
> cloudpickle               1.2.2                      py_0  
30,31c31,32
< dbus                      1.13.6               h746ee38_0  
< decorator                 4.4.0                    py37_1  
> dbus                      1.13.12              h746ee38_0  
> decorator                 4.4.1                      py_0  
33c34
< descriptastorus           2.0.0                      py_0    rmg
> descriptastorus           2.2.0                      py_0    rmg
43c44
< future                    0.17.1                   py37_0  
> future                    0.18.2                   py37_0  
45c46
< glib                      2.56.2               hd408876_0  
> glib                      2.63.1               h5a9c865_0  
47c48
< google-pasta              0.1.7                      py_0  
> google-pasta              0.1.8                      py_0  
58c59
< hyperopt                  0.1.2                      py_0    conda-forge
> hyperopt                  0.2.2                      py_0    conda-forge
60c61
[...]

PR #1851, opened 2019-12-13
Update required version of rdkit to avoid severe memory leaks

kspieks:
### Motivation or Problem
Significant memory leaks have been observed when using an rmg_env containing rdkit version 2019.03.4.0 as referenced in issue #1850. The issue is documented on the rdkit page [https://github.com/rdkit/rdkit/issues/2639](https://github.com/rdkit/rdkit/issues/2639) 

Memory leaks are not observed when using a more recent rdkit version of 2019.09.1.0 (and it looks like they've already released a version 2019.09.2.0)

### Description of Changes
Updated the minimum required version number for rdkit in the environment.yml file

### Testing
I used this environment.yml file to create an rmg_env. Doing `conda list rdkit` produced the following output, confirming that the change is implemented correctly

```
# Name                    Version                   Build  Channel
rdkit                     2019.09.2.0      py37h65625ec_1    rdkit
```

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -42,7 +42,7 @@ dependencies:
   - python >=3.7
   - pyyaml
   - quantities
-  - rdkit >=2018
+  - rdkit >=2019.09.1.0
   - scikit-learn
   - scipy
   - symmetry


alongd:
The updated Linux x64 version is v2019.09.2.0

mliu49:
I think this looks good. Alon, were you suggesting any changes, or confirming that this is ok?

alongd:
Suggesting to change 2019.09.1.0 into 2019.09.2.0

mliu49:
Ah, ok. I think this is fine though, since 2019.09.1.0 is the release which fixes the issue, so anything newer is acceptable.

mliu49:
Of course, unless there are other reasons that 2019.09.1.0 does not work for us, in which case we can increment it to 2019.09.2.0.

kspieks:
> Of course, unless there are other reasons that 2019.09.1.0 does not work for us, in which case we can increment it to 2019.09.2.0.

Both versions have not had memory leaks for me, so I think this PR is good to go

PR #1857, opened 2019-12-31
adding G4 model chemistry to Arkane

davidfarinajr:
### Motivation or Problem
Added `g4` (and `m062x-gd3/jun-cc-pvtz`) model chemistry to Arkane.  The G4 method was used to calculate the thermochemical data for the species in the Nitrogen_G4 library in the RMG-database PR `Adding Nitrogen ThermoLibrary, Benson Groups, and HBIs #376`

### Description of Changes
- added halogen atom spin orbit corrections taken from `https://cccbdb.nist.gov/elecspin.asp`
- added halogen atom energies to `m06-2x/cc-pvtz`
- added atom energies and bond additivity corrections for `g4`
- added `m062x-gd3_jun-cc-pvtz` atom energies
- added `g4` frequency scale factor from `J. Chem. Phys. 126, 084108 (2007); doi: 10.1063/1.2436888`
- added 'g4` to gaussian parser


arkane/encorr/data.py
```diff
@@ -100,6 +100,12 @@
         'H': -0.5010030, 'N': -54.564343, 'O': -75.030991, 'C': -37.827717, 'P': -341.116432, 'S': -397.961110
     },
 
+    # g4 atom energies taken from [J. Chem. Phys. 126, 084108 (2007); doi: 10.1063/1.2436888]
+    'g4': {
+        'Br': -2573.58537, 'C': -37.83417, 'Cl': -460.01505, 'F': -99.70498,
```
alongd: Could you add `S`, and `P`? (or, is there a reason not to add them?)
Could you also order the elements like the other model chemistries entries, so it's easier for users to compare what's missing?

arkane/encorr/data.py
```diff
@@ -384,6 +390,16 @@
         'N#N': -2.0, 'O=O': -0.2, 'H-H': 1.1,  # Unknown source
     },
 
+    'g4': {
```
alongd: Thanks! I appreciate this contribution.
Could you please add a brief comment of the data source used to derive these values?

arkane/encorr/data.py
```diff
@@ -106,6 +106,12 @@
         'H': -0.50142, 'N': -54.57367, 'O': -75.04550
     },
 
+    'm062x-gd3_jun-cc-pvtz': {
```
alongd: I'm familiar with `d3`, not `gd3`. Just making sure the `g` here is not a typo. Would you know what it stands for?
Also, please replace the underscore with a `/`

davidfarinajr: yes, g stands for `Grimme`.  The keyword in Gaussian 16 is `GD3`, which is why I included the `g` in there.  However, I think it is more commonly known just `D3`, so I'll remove the `g`.

arkane/encorr/data.py
davidfarinajr: will do

arkane/encorr/data.py
davidfarinajr: Thinking about this more...`d3` shouldn't have an effect on the atom energies.  So `m062x-d3/jun-cc-pvtz` atom energies should be the same as `m062x/jun-cc-pvtz`.  Should I remove `d3` from this model chemistry name?

alongd: Really? I thought it should. If it doesn't then we might as well remove it, I'd love to learn more about this issue.
@dranasinghe / @cgrambow, could you also comment? Does adding dispersion really not affect the energy?

cgrambow: Dispersion is an interaction between atoms, so individual atoms are not affected.

dranasinghe: I agree that dispersion will not effet atom energies but here we are considering atom correction such that energies will agree with the standard state. Since dispersion effect H2, O2, and molecules "correction" for the atom energy will be affected.

dranasinghe: I did a quick calculation for butanol m062x/jun-cc-pvtz and m062x-GD3/jun-cc-pvtz energy in Hartrees was -233.638900464 Eh and -233.639125055 Eh respectively. If you consider the same atom corrections,  the standard formation enthalpy for m062x-GD3/jun-cc-pvtz  will be wrong by ~0.58 kJ/mol.

alongd: Thanks all for your comments!
To conclude: in this case, calculated atom corrections w/ and w/o D3 are the same. However, if we take the approach of deriving atom corrections from the standard states as @dranasinghe mentioned, then it does matter. But this is probably not our case (I think here we let BAC's correct for that, though we do have instances in Arkane where atom corrections relate to standard states).
We do, however, need to keep the `D3` in the key label here, since these corrections must correspond to the BAC's, where dispersion does matter of course.

alongd:
@davidfarinajr, what's the status of this PR?
Could you wrap up the leftover comments and rebase?

davidfarinajr:
> @davidfarinajr, what's the status of this PR?
> Could you wrap up the leftover comments and rebase?

I need to remove the bond corrections.  It turns out that there is a bug in our version of Gaussian that causes it to print the final G4 energy in the wrong spot.  So I was parsing out the wrong energies. Sorry about that

PR #1861, opened 2020-01-08
T3 related modifications

alongd:
### Motivation or Problem
We'd like to be able to automate model refinement using electronic structure calculations, and couple it to model generation.

### Description of Changes
This PR adds the fist stage of automatically calculating species thermodynamic properties by calling RMG and [ARC ](https://github.com/ReactionMechanismGenerator/ARC) iteratively. Various species selection criteria were implemented.

- Crucial dependencies for ARC were added to the RMG environment (ase, paramico, py3dmol).
- Some minor modifications to RMG's API were made.
- The tandem tool was added as a root module in RMG-Py, since it imports from `rmgpy`.
- SA in Arkane will now also output the results in a YAML format (otherwise we would have needed to write a parser for the current txt output)

Run T3 by typing, e.g.:
```
python $rmgpy/rmg.py input.py -t3 arc.yml
```
where `arc.yml` is your favorite ARC input file (with no species defined), with the following additional (all optional!) directives (default values are given here):
```yaml
SA observables:
  - label: label1
    smiles: <smiles>
  - label: label2
    adj: <adjacency list>
SA species: 10
SA reactions: 10
SA pdep threshold: 0.01
collision violators: True
all core species: True
RMG tolerances: [0.05, 0.01, 0.001]
max tandem iterations: 10
max RMG exceptions allowed: 10
max RMG walltime: 01:00:00:00
```

This PR is still missing:
- Add uncertainty analysis
- Add support for RMS and/or Cantera for performing SA in different reactors and for global observables
- An appropriate addition to RMG's documentation (though for users who'd like to try it out right away, the docstrings are relatively clear)
- Add examples
- Modifications to the RMG-tests repo to accommodate ARC's dependencies

### Testing
Tests were added for all major functionalities.

### Reviewer Tips
Run the tests. See that the code makes sense. Try it out!

--------------------------------

T3 works according to the following general scheme:
image

### DIFF
--- a/arkane/sensitivity.py
+++ b/arkane/sensitivity.py
@@ -35,6 +35,7 @@ of kinetics and pressure-dependent jobs.
 import logging
 import os
 import string
+import yaml
 
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


mjohnson541:
So while the changes to RMG that make it possible to use the RMG-ARC tandem tool in this PR are perfectly welcome, I'm not convinced the tool itself should be part of RMG and not its own repository. RMG is getting too big and interconnected and this is something that simply uses RMG as a dependency rather than being a part of RMG, which makes it easy to keep separate. Max and I discussed this the last time it got brought up.

alongd:
Good point, @mjohnson541, I agree and will open a new repo, converting this PR into just the required modifications to RMG.

alongd:
@kspieks, @mjohnson541, could you take a look at this PR?
It only contains RMG-revevant changes, T3 itself was relocated to a different repo.
Thanks!!

kspieks:
Looks good! Just a quick question, why are we saving the data in multiple formats? (txt and yaml?)

alongd:
Good point. The SA data in text format could be deprecated, although it is nice and more readable that YAML. I didn't do that to be "backward compatible" if it makes sense. The YAML SA output is important so the data is more easily machine readable.

alongd:
Thanks both!

PR #1870, opened 2020-01-16
Move Arkane Data to RMG-database

amarkpayne:
### Motivation or Problem
Currently we store both atom energies and BACs in a separate data file, but not frequency scale factors even though these are data as well. This improves code organization, and allows for checking which frequency scale factors are currently in Arkane.

**Prior Motivation (see update below, this is out of date):**
I am running into problems with ARC where currently it does not check which model chemistries in Arkane have frequency scaling factors (instead this is a hard coded check against several well known ones) . Part of the problem is that this data is not accessible because it is contained as a local variable of a function. This PR changes that, and I hope to follow-up shortly with a PR to ARC that allows for this automatic checking.


amarkpayne:
Quick note: Oscar appears to have beaten me to this task, I just wasn't on the latest version of ARC. I still think this is a worthwhile move though

cgrambow:
Following @mjohnson541's suggestion, should we just go ahead and move all of this data to the RMG-database?

amarkpayne:
We could go ahead and move it all over there, but we should be smart in how we do it. For example, we don't want to have to load the full RMG-database to access just these parameters, as this would significantly slow down Arkane.

alongd:
I think that all of Arkane's data should be in one repo, either -Py or -db. It's OK to relocate to -db if it can be loaded separately than the kinetics and thermo estimators (and quickly). But I'd prefer to keep it in the same repo. I'm OK with transitioning the scaling factors out of statmech and into `encorr.data`

mjohnson541:
It's easy to load one database part, you just have to call the individual load function for that particular part of the database. I generally like the idea of putting everything in the RMG database, because it makes things very easy to find.

amarkpayne:
Sounds good, I'll rework this PR then with that change.

amarkpayne:
@cgrambow I have pushed all of the changes needed to make this work as we discussed. This PR should be merged in after #1935 , but I have already incorporated some of the changes into the database PR. The way this was implemented means that your BAC PR might not even be affected

cgrambow:
Before I review this, maybe we should merge this prior to #1935 and make #1935 a database PR instead? That way we could proceed with merging the BAC PR and deriving other corrections without being dependent on the status of #1935. What do you think?

arkane/encorr/data.py
```diff
@@ -28,391 +28,24 @@
...
+spec.loader.exec_module(data)
+
+# Assign the data here so that it can be imported
+atom_hf = data.atom_hf
+atom_thermal = data.atom_thermal
+SOC = data.SOC
+atom_energies = data.atom_energies
+pbac = data.pbac
+mbac = data.mbac
+freq_dict = data.freq_dict
```
cgrambow, line 51: To do this flexibly, you could consider something like
```python
for k, v in vars(data).items():
    if not k.startswith('__'):
        vars()[k] = v
```
Do you think that sacrifices readability too much?

arkane/encorr/data.py
```diff
@@ -28,391 +28,24 @@
...
-mbac = {}
+import importlib.util
+import os
+
+from rmgpy import settings
+
+quantum_corrections_path = os.path.join(settings['database.directory'], 'quantum_corrections', 'data.py')
+spec = importlib.util.spec_from_file_location("quantum_calculations", quantum_corrections_path)
+data = importlib.util.module_from_spec(spec)
+spec.loader.exec_module(data)
```
cgrambow, line 42: I like this approach because it's a neat way of loading data and having it immediately available. Do you think there are any issues with storing the data at the module level? I guess the other option would be writing some kind of database class that would store the data as attributes. Would that be better?

arkane/encorr/data.py
amarkpayne: Not at all, I like this a lot, thanks for the suggestion!

arkane/encorr/data.py
amarkpayne: I thought about this as well, but in theory no one should be modifying corrections data at runtime, so I think storing at the module level should be fine. I am fine implementing a quick database structure, but I am not sure that it is needed. Worthwhile discussion though

amarkpayne:
I am fine merging in this PR first and then converting the DLPNO BAC PR to a database PR, that way we don't rush the discussion that are going on for the DLPNO PR.

arkane/encorr/data.py
cgrambow: Generally agree, although I guess the `write_to_database` code in my BAC PR does modify at runtime, if desired, although it uses the same mechanism as you use here to reload the data.

PR #1882, opened 2020-02-05
Adding new surface attributes

mazeau:
This goes with [this rmg-database pull request](https://github.com/ReactionMechanismGenerator/RMG-database/pull/384) to add in a few new attributes to both thermo and kinetics, describing which metal, facet, and site the data was obtained/calculated from.  When the tests are run with my new-attributes database, all tests do pass.  I also wrote some more unit tests to make sure that all surface species at least have a corresponding metal.


rmgpy/data/base.py
```diff
@@ -101,6 +109,15 @@ def __init__(self,
         self.long_desc = long_desc
         self.rank = rank
         self.nodal_distance = nodal_distance
+        # if metal is not None:
```
rwest: Remove the commented out lines

rmgpy/data/kinetics/family.py
```diff
@@ -889,6 +889,9 @@ def save_training_reactions(self, reactions, reference=None, reference_type='',
                 short_desc=str(short_desc[i]),
                 long_desc=str(long_desc[i]),
                 rank=rank[i],
+                # metal=metal,
```
rwest: If not needed here, remove them

rmgpy/data/kinetics/family.py
```diff
@@ -1177,6 +1180,9 @@ def add_rules_from_training(self, thermo_database=None, train_indices=None):
                 reference=entry.reference,
                 short_desc="Rate rule generated from training reaction {0}. ".format(entry.index) + entry.short_desc,
                 long_desc="Rate rule generated from training reaction {0}. ".format(entry.index) + entry.long_desc,
+                # metal=metal,
```
rwest: and elsewhere (I won't comment on the others)

rmgpy/data/kinetics/family.py
```diff
@@ -2462,6 +2471,7 @@ def get_kinetics_from_depository(self, depository, reaction, template, degenerac
                     depository.label,
                     '[{0}]'.format(';'.join([g.label for g in template])))
                 kinetics.comment += "\nfamily: {}".format(self.label)
+                # todo: add in comments about which metal/site/facet it came from?
```
rwest: Can you do the todo?

rmgpy/data/kinetics/kineticsTest.py
```diff
@@ -147,6 +147,43 @@ def test_load_families_correct(self):
         except DatabaseError:
             self.fail("Unable to load families using list ['H_Abstraction', 'pah']")
 
+    def test_kinetics_libraries_have_surface_attributes(self):
```
rwest: I wonder if we should check every library in the 'Surface' folder, and enery family starting with 'Surface' ?

rmgpy/data/kinetics/library.py
```diff
@@ -561,6 +573,9 @@ def save(self, path):
         f.write('longDesc = """\n')
         f.write(self.long_desc.strip() + '\n')
         f.write('"""\n')
+        f.write('metal = "{0}"\n'.format(self.metal))
```
rwest: Preferable only to write these lines if defined? Otherwise we'll be printing "metal = None" etc. on the gas phase stuff?  I guess it's only once per file, so perhaps this is fine. 

But that makes me wonder: could you (and might you want to) have entry-level attributes not file-level attributes? Eg. a single library that contains different data for different facets or metals?

rmgpy/data/kinetics/kineticsTest.py
mazeau: fixed this

rmgpy/data/kinetics/family.py
mazeau: included this

rmgpy/data/kinetics/library.py
mazeau: I was under the impression I had both entry level attributes and file level attributes...  I do have a Nitrogen library that has data from different metals.

rmgpy/data/kinetics/library.py
```diff
@@ -573,9 +573,12 @@ def save(self, path):
         f.write('longDesc = """\n')
         f.write(self.long_desc.strip() + '\n')
         f.write('"""\n')
-        f.write('metal = "{0}"\n'.format(self.metal))
-        f.write('site = "{0}"\n'.format(self.site))
-        f.write('facet = "{0}"\n'.format(self.facet))
+        if not isinstance(self.metal, type(None)):
```
rwest: How about `if self.metal:`?

rmgpy/data/kinetics/family.py
```diff
@@ -2477,7 +2477,12 @@ def get_kinetics_from_depository(self, depository, reaction, template, degenerac
                     depository.label,
                     '[{0}]'.format(';'.join([g.label for g in template])))
                 kinetics.comment += "\nfamily: {}".format(self.label)
-                # todo: add in comments about which metal/site/facet it came from?
+                if not isinstance(entry.metal, type(None)):
```
rwest: How about `if entry.metal:` ?

rmgpy/data/kinetics/kineticsTest.py
rwest: Thanks for addressing that. I now realize that this sort of testing probably belongs more in the RMG-database tests (the script in `RMG-Py/testing/databaseTest.py` rather than the RMG-Py unit tests.  i.e. in this testing file (`rmgpy/data/kinetics/kinetictsTest.py`) we should be checking that RMG-Py can correctly read, write, access, understand, manipulate a well-formed database.  And in the other place (`RMG-Py/testing/databaseTest.py`) we should be checking that the database is well-formed.

rmgpy/data/thermoTest.py
```diff
@@ -75,6 +75,63 @@ def test_failing_loads_thermo_libraries(self):
         with self.assertRaises(Exception):
             database.load_libraries(os.path.join(path, 'libraries'), libraries)
 
+    def test_surface_thermo_libraries_have_surface_attributes(self):
+        """
+        Test that each entry in the surface thermo database has a 'metal' and 'facet' attribute, describing which metal
```
rwest: Again, if we're testing that the whole database is well-formed, it should probably be in `RMG-Py/testing/databaseTest.py`. This file should be for testing that we can read/write/manipulate a well-formed database properly (much of which, this test does)

rmgpy/data/kinetics/library.py
mazeau: fixed this

rmgpy/data/kinetics/family.py
mazeau: fixed this

rmgpy/data/kinetics/kineticsTest.py
mazeau: I see what you mean, because this goes and checks each entry, but I think a test does belong in rmg-py to make sure that database entries can have these surface attributes.  If someone breaks the way this is currently written, it makes more sense for it to show up in the rmg-py tests.  I can make the more robust (check each surface family entry) in the database tests, but that seems a bit inefficient because you're still loading at least one database twice and running super similar tests.

mazeau: maybe you're right.  I don't think any database tests check thermo libraries yet..

mazeau:
closing and renaming the branch

PR #1884, opened 2020-02-06
Adding capability to have reverse surfaceArrhenius training reactions

mazeau:
I was having issues with surfaceArrhenius training reactions that were in the reverse direction, as they would all be overwritten as ArrheniusEP instead of SurfaceArrheniusBEP.  This change fixes that and all tests pass locally and I am able to successfully generate models!

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1232,12 +1232,25 @@ class KineticsFamily(Database):
             item.template = self.get_reaction_template_labels(item)
             new_degeneracy = self.calculate_degeneracy(item)
 
+            if isinstance(entry.data, SurfaceArrhenius):
+                data = SurfaceArrheniusBEP(
+                    #  analogous to Arrhenius.to_arrhenius_ep
+                    A=deepcopy(data.A),
+                    n=deepcopy(data.n),
+                    alpha=0,
+                    E0=deepcopy(data.Ea),
+                    Tmin=deepcopy(data.Tmin),
+                    Tmax=deepcopy(data.Tmax)
+                )
+            else:
+                data = data.to_arrhenius_ep()
+
             new_entry = Entry(
                 index=index,
                 label=';'.join([g.label for g in template]),
                 item=Reaction(reactants=[g.item for g in template],
                               products=[]),
-                data=data.to_arrhenius_ep(),
+                data=data,
                 rank=entry.rank,
                 reference=entry.reference,
                 short_desc="Rate rule generated from training reaction {0}. ".format(entry.index) + entry.short_desc,
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1959,7 +1959,9 @@ class KineticsFamily(Database):
 
             # ToDo: try to remove this hard-coding of reaction family name..
             if 'adsorption' in self.label.lower() and forward:
-                if molecules_a[0].contains_surface_site() and molecules_b[0].contains_surface_site():
+                if 'Surface_Dual_Adsorption_vdW' is self.label and forward:
+                    pass
+                elif molecules_a[0].contains_surface_site() and molecules_b[0].contains_surface_site():
                     # Can't adsorb something that's already adsorbed.
                     # Both reactants either contain or are a surface site.
                     return []
--- a/rmgpy/reaction.py
+++ b/rmgpy/reaction.py
@@ -53,8 +53,9 @@ import rmgpy.constants as constants
 from rmgpy.exceptions import ReactionError, KineticsError
 from rmgpy.kinetics import KineticsData, ArrheniusBM, ArrheniusEP, ThirdBody, Lindemann, Troe, Chebyshev, \
     PDepArrhenius, MultiArrhenius, MultiPDepArrhenius, get_rate_coefficient_units_from_reaction_order, \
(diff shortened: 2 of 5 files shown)


rmgpy/reaction.py
```diff
@@ -787,6 +788,28 @@ def reverse_arrhenius_rate(self, k_forward, reverse_units, Tmin=None, Tmax=None)
         kr.fit_to_data(Tlist, klist, reverse_units, kf.T0.value_si)
         return kr
 
+    def reverse_surface_arrhenius_rate(self, k_forward, reverse_units, Tmin=None, Tmax=None):
+        """
+        Reverses the given k_forward, which must be an Arrhenius type.
```
rwest: Docstring says it must be an Arrhenius type, but it should in fact be SurfaceArrhenius ?

rmgpy/reaction.py
```diff
@@ -787,6 +788,28 @@ def reverse_arrhenius_rate(self, k_forward, reverse_units, Tmin=None, Tmax=None)
...
+        cython.declare(Tlist=np.ndarray, klist=np.ndarray, i=cython.int)
+        kf = k_forward
+        assert type(kf) is SurfaceArrhenius  # Only reverses SurfaceArrhenius rates
+        if Tmin is not None and Tmax is not None:
+            Tlist = 1.0 / np.linspace(1.0 / Tmax.value, 1.0 / Tmin.value, 50)
+        else:
+            Tlist = 1.0 / np.arange(0.0005, 0.0034, 0.0001)
+        # Determine the values of the reverse rate coefficient k_r(T) at each temperature
+        klist = np.zeros_like(Tlist)
+        for i in range(len(Tlist)):
```
rwest: is there a reason this isn't 
  for i, T in enumerate(Tlist):
then use the T directly instead of looking up Tlist[i] twice?

rmgpy/reaction.py
mazeau: I just copied and pasted from the reverse_arrhenius_rate one.  I'll fix the comment

rmgpy/reaction.py
mazeau: I just copied this directly from the reverse_arrhenius_rate and that's how it is there

mazeau:
what unit tests would you like to see?

mazeau:
The last commit specifically treats the surface migration family differently, because it is a special case where a surface species changes where it is bonded to the surface.  It does not subtract switch from m3/mol to m2/mol, because it was mistakenly switching it even though it was the only surface species.

rwest:
Found a better way to fix the two problems that were earlier avoided by skipping unit tests.

I rebased it, so the first two commits on this branch are actually in https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1894
which fixes an isomorphism bug that was causing problems.

And I edited the last commit, so it fixes the test rather than skipping it for a hard-coded family name.

rwest:
This now looks ok to me. Would be nice to see #1894 merged, then this rebased on top, then I'm fine with it. But I was the last to edit it so probably someone else should review it too :)

rmgpy/reaction.py
```diff
@@ -787,6 +788,28 @@ def reverse_arrhenius_rate(self, k_forward, reverse_units, Tmin=None, Tmax=None)
...
+    def reverse_surface_arrhenius_rate(self, k_forward, reverse_units, Tmin=None, Tmax=None):
+        """
+        Reverses the given k_forward, which must be a SurfaceArrhenius type.
+        You must supply the correct units for the reverse rate.
+        The equilibrium constant is evaluated from the current reaction instance (self).
+        """
+        cython.declare(kf=SurfaceArrhenius, kr=SurfaceArrhenius)
+        cython.declare(Tlist=np.ndarray, klist=np.ndarray, i=cython.int)
+        kf = k_forward
+        assert type(kf) is SurfaceArrhenius  # Only reverses SurfaceArrhenius rates
```
amarkpayne: Instead of assert, do you mind do a `isinstance` check and raising an exception? For example:

```
if not isinstance(kf, SurfaceArrhenius):
    raise TypeError(f'Expected a SurfaceArrhenius object for k_forward but received {kf}')
```

rmgpy/reaction.py
```diff
@@ -825,7 +849,10 @@ def generate_reverse_rate_coefficient(self, network_kinetics=False, Tmin=None, T
             return kr
 
         elif isinstance(kf, Arrhenius):
-            return self.reverse_arrhenius_rate(kf, kunits, Tmin, Tmax)
+            if type(kf) is SurfaceArrhenius:
```
amarkpayne: Can you change this to `if isinstance(kf, SurfaceArrhenius):`?

rmgpy/data/kinetics/family.py
```diff
@@ -1232,12 +1232,25 @@ def add_rules_from_training(self, thermo_database=None, train_indices=None):
             item.template = self.get_reaction_template_labels(item)
             new_degeneracy = self.calculate_degeneracy(item)
 
+            if type(data) is SurfaceArrhenius:
```
amarkpayne: Can you also change this to `if isinstance(SurfaceArrhenius)`?

amarkpayne:
Thanks for the changes @mazeau ! Matt is looking over the isomorphism change now

rwest:
With the latest commit this is no longer dependent on changing the isomorphism code (#1894) and so has been rebased to master.  I think @mazeau is in the process of writing a unit test for the new reversing method.

mazeau:
I've added in the new unit test

PR #1887, opened 2020-02-07
Warn the user if the RMG environment is not activated

amarkpayne:
### Motivation or Problem
It is a very common mistake for even experienced RMG developers to make to forget to activate their RMG environment before working with RMG. While experienced users can quickly figure out what went wrong, we should print out dialog that helps the user fix this issue whenever we detect that the conda environment is not correct so that new users can overcome this issue quickly.

### Description of Changes
We already had a utility for checking that vital dependencies could be found, so I just made sure that this code is always called whenever we run any of the unit tests like `make test`. 

Further, I added a method that specifically checks the name of the environment to make sure that this all checks out.

### Testing
I have tested running `make test` with and without my environment activated and it appears to work as expected

### Note to Reviewer
When checking the environment, I have two messages: one for if the environment is "base" and one for if the environment is simply not named "rmg_env". My intention with this was that "base" is almost certainly wrong, and we shouldn't even bother proceeding further, while for any other name it is possible that the user simply named their environment differently.

However, I currently did not make it such that "base" prevents the tests from running further. Therefore, I think I either need to consolidate the two messages, or implement this blocking for "base". Let me know which one you prefer.

Also, let me know if my formatting was too excessive for the warning statement (i.e. too many "\n").

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,28 +57,28 @@ decython:
 	find . -name *.so ! \( -name _statmech.so -o -name quantity.so -o -regex '.*rmgpy/solver/.*' \) -exec rm -f '{}' \;
 	find . -name *.pyc -exec rm -f '{}' \;
 
-test-all:
+test-all: check
 ifneq ($(OS),Windows_NT)
 	mkdir -p testing/coverage
 	rm -rf testing/coverage/*
 endif
 	nosetests --nocapture --nologcapture --all-modules --verbose --with-coverage --cover-inclusive --cover-erase --cover-html --cover-html-dir=testing/coverage --exe rmgpy arkane
 
-test test-unittests:
+test test-unittests: check
 ifneq ($(OS),Windows_NT)
 	mkdir -p testing/coverage
 	rm -rf testing/coverage/*
 endif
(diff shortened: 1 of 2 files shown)


mjohnson541:
So I can't easily test this because my base env is python 2 and I get other errors before the warnings, but now that I've thought about this more, it's perfectly valid to install RMG in your base environment. A better alternative might be to try-catch the first import error someone is likely to get running in a typical base environment and log a message telling the user this may have happened because they weren't in the right environment, this should be much more helpful (as it will only print the error message if they were having a problem anyway and will do so directly connected with the error).

rwest:
Agree with Matt - users should be allowed to install in base, or an enivronment called anything they choose. I'm in favor of helpful error messages that tell people likely causes, but only raising them when there's an actual error, and not being overly confident in the diagnosis (eg. don't want to imply it's definitely the wrong environment if it might be something else).

mjohnson541:
Should this be closed?

amarkpayne:
I think that this can be closed, I'll do so now

PR #1888, opened 2020-02-07
Detect and handle NaNs in solve_ss_network

mjohnson541:
If the solution of the linear system gives NaNs originally the NaN got passed as the concentrations in the network breaking flux based network reduction. This small fix detects this case and returns None instead causing flux based network reduction to fall back to rate coefficient based methods in these cases. 

Input file with the issue this fixes is attached:  
[input.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/4172916/input.txt)


mjohnson541:
Rebased, should be good to merge.

PR #1357, opened 2018-05-05
Allow PDep reactions and non-template library reactions in PDep networks

alongd:
This PR allows PDep networks to use kinetics from a LibraryReaction in two cases:
1. When a unimolecular library reaction does not correspond to any kinetics family (but is flagged as `elementary`)
2. When a unimolecular library reaction has PDep kinetics, in which case a high pressure limit rate is generated and saved as a new attribute of the Reaction. This rate is then used in the network.

A test was added for the high pressure limit kinetic generation. This PR involved relatively modest changed to pdep.py, so direct pdep tests were currently not added.

PR #1894, opened 2020-02-20
Find and fix a bug in Molecule isomorphism

rwest:
### Motivation or Problem
If you ran `Molecule.is_isomorphic` with an incorrect initial map, it would assume nothing was wrong with the map, and if it found no non-isomorphic things in the _rest_ of the molecule, would conclude the molecules were isomorphic.

This was causing some incorrect failing database tests.

### Description of Changes
If given an initial mapping, check it's valid.

### Testing
Made a unit test that fails, then fixed it and it passes.

### DIFF
--- a/rmgpy/molecule/isomorphismTest.py
+++ b/rmgpy/molecule/isomorphismTest.py
@@ -485,3 +485,49 @@ def test_isotope_subgraph_isomorphism_molecule_and_group():
     assert_true(methanei.is_subgraph_isomorphic(group_methane))
     assert_true(methane.is_subgraph_isomorphic(group_methane))
     assert_false(methanei.is_isomorphic(methane))
+
+def test_isomorphism_wrong_mapping():
+    """
+    Checks isomorphism finds things not isomorphic if given a wrong mapping.
+    """
+    # These molecules are not isomorphic
+    n1butane = Molecule().from_adjacency_list("""
+        1 *1 X u0 p0 c0 {2,S}
+        2 *2 C u0 p0 c0 {1,S} {3,S} {5,S} {6,S}
+        3 *3 C u0 p0 c0 {2,S} {4,S} {7,S} {8,S}
+        4 *4 H u0 p0 c0 {3,S}
+        5    H u0 p0 c0 {2,S}
+        6    H u0 p0 c0 {2,S}
+        7    H u0 p0 c0 {3,S}
+        8    C u0 p0 c0 {3,S} {9,S} {10,S} {11,S}
+        9    H u0 p0 c0 {8,S}
+        10   H u0 p0 c0 {8,S}
+        11   C u0 p0 c0 {8,S} {12,S} {13,S} {14,S}
+        12   H u0 p0 c0 {11,S}
+        13   H u0 p0 c0 {11,S}
+        14   H u0 p0 c0 {11,S}
+        """)
+
+    n2butane = Molecule().from_adjacency_list("""
+        1 *1 X u0 p0 c0 {3,S}
+        2 *2 C u0 p0 c0 {3,S} {5,S} {6,S} {4,S}
+        3 *3 C u0 p0 c0 {2,S} {1,S} {7,S} {8,S}
+        4 *4 H u0 p0 c0 {2,S}
+        5    H u0 p0 c0 {2,S}
+        6    H u0 p0 c0 {2,S}
+        7    H u0 p0 c0 {3,S}
+        8    C u0 p0 c0 {3,S} {9,S} {10,S} {11,S}
+        9    H u0 p0 c0 {8,S}
+        10   H u0 p0 c0 {8,S}
+        11   C u0 p0 c0 {8,S} {12,S} {13,S} {14,S}
+        12   H u0 p0 c0 {11,S}
+        13   H u0 p0 c0 {11,S}
+        14   H u0 p0 c0 {11,S}
+        """)
+    assert_false(n1butane.is_isomorphic(n2butane))
+    assert_false(n1butane.is_isomorphic(n2butane, generate_initial_map=True))
+    
+    mapping = {}
+    for label in ['*1', '*2', '*3', '*4']:
+        mapping[n1butane.get_labeled_atoms(label)[0]] = n2butane.get_labeled_atoms(label)[0]
+    assert_false(n1butane.is_isomorphic(n2butane, initial_map=mapping))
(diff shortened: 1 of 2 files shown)


mjohnson541:
I think @mliu49 and I talked about doing this before and decided not to. I think at the time the argument was that if one cares about the correctness of the initial map (assuming the map is unique) one can simply use the generate_initial_map option which will give the right result. However inside RMG we already know the map and know it is correct the vast majority of times so in all of those situations it doesn't make sense to run this check. I'm unsure how much speed difference this would make, but given how often we use is_isomorphic it could be significant.

rwest:
Calling with `generate_initial_map` doesn't enforce all labeled atoms are in both molecules.
For example in [databaseTest.py line 569](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/0198164c55a3dad187872ec72d64503b8c8c2a8f/testing/databaseTest.py#L569) `kinetics_check_adjlists_nonidentical` creates a mapping and then passes it to `is_isomorphic`.
If you replace this code with just calling `is_isomorphic` with `generate_initial_mapping=True` instead of passing in the initial mapping, then a bunch of families fail the tests, because things with and without labeled atoms are marked as isomorphic.

There are other ways to rewrite this particular unit test to work around the problem instead of the solution currently in this PR, but then we still have the fact that a function called `is_isomorphic` sometimes returns `True` for two molecules that are clearly not isomorphic.  I would argue that to accept this it should be 1) definite speed benefit, and 2) very well documented.

mliu49:
I do remember having that discussion, but I don't remember much of the details.

In one of my many attempts to refactor these comparisons, I wrote an entirely new method with multiple arguments to handle all comparison types. The code is still [here](https://github.com/mliu49/RMG-Py/blob/species_comparison/rmgpy/molecule/molecule.py#L1228), and it has an option to verify the initial mapping, whether it was provided or generated. I ended up abandoning that branch because it seemed like a big function with a ton of options was less efficient than having multiple individual ones.

I think it would be worthwhile to add this option to `is_isomorphic`, though it might be a good idea to check whether it does impact performance.

rwest:
So https://github.com/mazeau/RMG-Py/commit/8c78e91e558a03078ef6a071851f223cb2e36a2b now works around this problem so  #1884 is no longer dependent on this PR.

alongd:
looking good! please rebase

mjohnson541:
Are we certain we want to make this check non-optional?

rwest:
We could add a `assume_initial_map_is_correct=False` argument? 
If the method is called with `assume_initial_map_is_correct=True` then it skips the check?

When would we then use the feature? I suppose if we have just algorithmically generated the initial map, but I'm not sure where that's the case.

mjohnson541:
I'm not sure I understand...aren't all initial maps in an RMG job algorithmically generated?

rwest:
If there are labelled atoms then the labels are used to generate the initial mapping. If those atoms are NOT actually isomorphic, we never check them. See the example in the unit test that is added in this PR. 
It shows butane adsorbed to the surface through the first or the second C atom. Clearly different graphs, but were identified as isomorphic

```
CH2-CH2-CH2-CH3
|
X 

CH3-CH-CH2-CH3
    |
    X 
```

This was causing tests in the database to incorrectly fail.

mjohnson541:
Oh sorry, this is just Molecule.is_isomorphic, not all of them.

rwest:
Am I right that the only remaining concern is that checking isomorphism properly might slow things down a bit?

The continuous-integration/rmg-tests show these timings:

Observables Test Case: Aromatics Comparison
Execution time, Benchmark:
Execution time (DD:HH:MM:SS): 00:00:00:56
Execution time, Tested:
Execution time (DD:HH:MM:SS): 00:00:00:56

Observables Test Case: NC Comparison
Execution time, Benchmark:
 Execution time (DD:HH:MM:SS): 00:00:01:31
Execution time, Tested:
Execution time (DD:HH:MM:SS): 00:00:01:32

Observables Test Case: SO2 Comparison
Execution time, Benchmark:
Execution time (DD:HH:MM:SS): 00:00:01:00
Execution time, Tested:
Execution time (DD:HH:MM:SS): 00:00:01:00

Observables Test Case: EG1 Comparison
Execution time, Benchmark:
Execution time (DD:HH:MM:SS): 00:00:00:57
Execution time, Tested:
Execution time (DD:HH:MM:SS): 00:00:00:56


All these jobs are only ~1 minute, so not a convincing argument, but I don't see evidence that we should worry. One is 1 second faster, and one is 1 second slower, and the other two are the same.

Do we have (or should we have?)  a more rigorous or accurate away to benchmark speed changes in pull requests?



Weirdly, looking carefully at these logs, there appears to be some non-identical thermo in the rmg/nitrogen test
```
Non-identical thermo!
1119original:	O1[C]=N1
1120tested:	O1[C]=N1
1121Hf(300K)  |S(300K)   |Cp(300K)  |Cp(400K)  |Cp(500K)  |Cp(600K)  |Cp(800K)  |Cp(1000K) |Cp(1500K) 
1122    129.73|     51.79|      9.14|      9.59|      9.99|     10.34|     10.90|     11.38|     12.40
1123    154.90|     56.55|      9.79|      9.15|      8.58|      8.33|      9.02|     10.26|     10.96
1124thermo: Thermo group additivity estimation: missing(O2s-CdN3d) + group(N3s-CsHH) + group(Cds-CdsCsCs) + ring(Cyclopropene) + radical(Cds_P)
1125thermo: Thermo group additivity estimation: missing(O2s-CdN3d) + group(N3s-CsHH) + group(Cds-CdsCsCs) + ring(oxirene) + radical(Cds_P)
```
But the same oxirene and Cyclopropene interchange was noticed on many different PRs:
- https://github.com/ReactionMechanismGenerator/RMG-database/pull/255#issuecomment-374435434 
- https://github.com/ReactionMechanismGenerator/RMG-database/pull/257#issuecomment-374014008
- https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1357#issuecomment-387726781
- https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1276#issuecomment-375012342
so I think is an underlying non-deterministic behavior that should be addressed separately, in #2010

mjohnson541:
I think that's sufficient for this case, this one doesn't get as much use as its is_subgraph_isomorphic counterpart. 

It would be nice for us to have some speed benchmarks...as you've noted though RMG-tests examples aren't necessarily in the regime we're worried about. I think we'd need to add an example using the restart feature to start at a higher species load. To make it test properly on faster/slower processors I think we'd need to run an RMG independent speed test at the beginning that we could scale the RMG runtime to.

rwest:
Thanks

PR #1899, opened 2020-02-25
Added a "symmetric top" rotor

alongd:
### Motivation or Problem
So far Arkane treated species as either linear or non-linear ("asymmetric top"). Although RMG's statmech had functionality to treat "spherical top" rotors, it wasn't used when parsing ESS output files (it is used only if a user explicitly specifies it in the input file).

### Description of Changes
First a "spherical top" rotor was added. Here we also make Arkane identify the correct rotor using the parsed or computed rotational constants.

Nomenclature:
- Spherical Top rotor: A species with all three rotational constants equal (e.g., CH4).
- Symmetric Top rotor: A species with only two of the three rotational constants equal (e.g., NH3).
- Asymmetric Top rotor: A (non linear) species with three unique rotational constants. In RMG this is called a `NonlinearRotor`, we might want to consider renaming it to `AsymmetricRotor` (not renamed in this PR).

### DIFF
--- a/arkane/ess/gaussian.py
+++ b/arkane/ess/gaussian.py
@@ -39,7 +39,13 @@ import os.path
 import numpy as np
 
 import rmgpy.constants as constants
-from rmgpy.statmech import IdealGasTranslation, NonlinearRotor, LinearRotor, HarmonicOscillator, Conformer
+from rmgpy.statmech import (Conformer,
+                            HarmonicOscillator,
+                            IdealGasTranslation,
+                            LinearRotor,
+                            NonlinearRotor,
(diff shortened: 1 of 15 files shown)


JacksonBurns:
This PR is abandoned but up-to-date and simple enough that it could be merged. I am going to change the base branch to main, update it, and see what happens.

JacksonBurns:
@alongd the tests failed above are from (1) attempts to use parts of the API which have since been removed and (2) errors in values related to the hindered rotors, which are presumably actually not failures since the new values should be more accurate.

If you have time, I think this would be great to get merged!

Issue #1881, opened 2020-02-04
Database unit tests should test that the training reactions work and can be averaged


mazeau:
I'm finding that when I'm trying to add training reactions into families, all database tests pass but when I go to run RMG, sometimes it struggles with averaging them.  Maybe this should be added to the database tests?

PR #1900, opened 2020-02-25
Adding in a new database test to check all surface training reactions are useable

mazeau:
### Motivation or Problem
This is related to #1881, when I found that adding training reactions to the database would pass all database tests but then fail when used to generate rate rules when trying to run rmg.  This adds a check to catch these problem training reactions.
This also removes the hard-coded family work around, introduced in e8e292875bb412c8690b0a0584d2dc31c700d5b9

### Description of Changes
I have added a new kinetics check called `kinetics_check_surface_training_reactions_can_be_used` that checks all surface family training reactions.
I have changed the hard coded family exception to work for any family that has both reactants vdW bonded to the surface

### Testing
All tests pass!

### DIFF
--- a/testing/databaseTest.py
+++ b/testing/databaseTest.py
@@ -116,6 +116,14 @@ class TestDatabase(object):  # cannot inherit from unittest.TestCase if we want 
             self.compat_func_name = test_name
             yield test, family_name
 
+            # tests for surface families
+            if 'surface' in family_name.lower():
+                test = lambda x: self.kinetics_check_surface_training_reactions_can_be_used(family_name)
+                test_name = "Kinetics surface family {0}: entries can be used to generate rate rules?".format(family_name)
+                test.description = test_name
+                self.compat_func_name = test_name
+                yield test, family_name
+
             # these families have some sort of difficulty which prevents us from testing accessibility right now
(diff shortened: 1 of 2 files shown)


testing/databaseTest.py
```diff
@@ -116,6 +116,14 @@ def test_kinetics(self):
             self.compat_func_name = test_name
             yield test, family_name
 
+            # tests for surface families
+            if 'Surface' in family_name:
```
nateharms:
```suggestion
            if 'surface' in family_name.lower():
```

Possibly make this a `lower` to avoid missing case sensitivity.

amarkpayne:
I believe this looks fine, but I want to see the RMG-tests pass just to confirm. @mazeau can you squash the third commit into the first (since this is just a patch on the first commit anyways) and force-push to update? If the tests pass I'll merge in this PR

mazeau:
sounds good!

amarkpayne:
I just remembered (after talking/confirming with @mliu49) that RMG-tests fails if their is a hyphen in the branch name. Since this PR mostly affects unit tests, I am fine with merging this in (I'll rebase and merge right after I merge 1855 in.)

PR #1906, opened 2020-03-10
Liquid test improvements

amarkpayne:
Gets rid of the liquidTest restart file that is generated by the unit tests. Also uses the setUpClass method, which is what is actually desired here.

### DIFF
--- a/arkane/explorer.py
+++ b/arkane/explorer.py
@@ -321,9 +321,10 @@ class ExplorerJob(object):
             self.pdepjob.network = network
 
             if len(self.networks) > 1:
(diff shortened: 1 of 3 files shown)


amarkpayne:
I pushed a few more commits that were slightly related, but I am done adding commits now. @mjohnson541 let me know if this is all good to go

amarkpayne:
@mjohnson541 do you have time today to look at this? It should be a quick PR to get in

mjohnson541:
Sure

amarkpayne:
Thanks @mjohnson541 !

PR #1907, opened 2020-03-12
Adding new surface attributes

mazeau:
This goes with [this rmg-database pull request](https://github.com/ReactionMechanismGenerator/RMG-database/pull/391)  to add in a few new attributes to both thermo and kinetics, describing which metal, facet, and site the data was obtained/calculated from. When the tests are run with my new-attributes database, all tests do pass. I also wrote some more unit tests to make sure that all surface species at least have a corresponding metal.

This the same as #1882 , but changed the branch name so the rmg-tests will pass


rmgpy/data/base.py
```diff
@@ -1363,6 +1386,9 @@ def load_entry(self, label, group=None, molecule=None, species=None, shortDesc='
             item=item,
             short_desc=shortDesc,
             long_desc=longDesc.strip(),
+            metal=metal,
+            facet=facet,
+            site=site,
         )
 
     def save_entry(self, f, entry, name='entry'):
```
mjohnson541, line 1394: Why don't you need to edit this to save the metal, facet and site attributes? Won't it delete the attribute value if someone saves the database?

rwest: Thanks for catching this. @mazeau please can you take a look and fix or explain?

mazeau: Alright, I think I've fixed this! Thanks for catching it

amarkpayne:
I think the RMG-Tests are just failing because the benchmark branch can't use the current database (an issue solved by the temp commits we just removed). I'll merge in this PR now, but if it is an issue we can fix this

PR #1916, opened 2020-04-08
Added CSTR reactor type

hwpang:
### Motivation or Problem
I added a cstr reactor type where the user can specify a desired residence time. This allowed to continuous introduce new reactants and may be helpful for certain type of reactions and models, e.g. polymerization.

### Description of Changes
I added a ContinuousStirredTankReactor as cstr.pyx. The corresponding unit-test is cstrTest.py. Multiple other files were modified to include this new reactor type.

### Testing
The unit-test passed. I ran a input file with a mixture of dienes with a residence time of 10 min. The output model, sensitivity, and simulation all seemed to be reasonable.


mjohnson541:
Can you make your commit messages more specific? This is very important for enabling people to cherry-pick individual changes and read history down the line.  "Debugged for cstr.pyx" doesn't tell us anything about the changes contained inside other than that you changed the cstr.pyx file. Ideally your commits should tell the reader what changes you've made in them and sometimes why you made them.  I think this PR has some good examples of commit messages: https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1262

mliu49:
There's also a [wiki page about commit style](https://github.com/ReactionMechanismGenerator/RMG-Py/wiki/Git-Commit-Style-Guide) :)

kspieks:
> There's also a [wiki page about commit style](https://github.com/ReactionMechanismGenerator/RMG-Py/wiki/Git-Commit-Style-Guide) :)

Thanks for sharing Max! I also get lazy on my commit messages (whoops) so I'll look over this wiki too :)

hwpang:
> Can you make your commit messages more specific? This is very important for enabling people to cherry-pick individual changes and read history down the line. "Debugged for cstr.pyx" doesn't tell us anything about the changes contained inside other than that you changed the cstr.pyx file. Ideally your commits should tell the reader what changes you've made in them and sometimes why you made them. I think this PR has some good examples of commit messages: #1262

Thanks for all the above comments! I've added more detailed description in my commits. @mjohnson541

rmgpy/solver/cstr.pyx
```diff
@@ -767,5 +767,5 @@ cdef class ContinuousStirredTankReactor(ReactionSystem):
                         if ir[j, 2] != -1:
                             pd[ir[j, 2], ip[j, 2]] += deriv
 
-        self.jacobian_matrix = pd + cj * np.identity(num_core_species, np.float64)
+        self.jacobian_matrix = pd + cj * np.identity(num_core_species, np.float64) - F * np.identity(num_core_species, np.float64)
```
mjohnson541: This change isn't mentioned in the commit title. 

Also I'm pretty confident this doesn't fix the jacobian calculation as most jacobian entries should have units (1/s) while I'm reasonably sure you have units of (mol/s) in the new term.

hwpang: > So I did a quick initial overview of this. The first thing I noticed is that there are a lot of hidden changes in your commits that aren't mentioned or explained in the commit messages.
> 
> I unfortunately also think this design as a whole is unnecessary code duplication. The thermodynamics and rates of this reactor are identical to the liquid reactor it seems that it would make much more sense from a design perspective to add inflow and outflows to the liquid reactor instead of adding an entirely new reactor.
> 
> I also think it would be nice for the inlet and outlet to be able to be specified independently this should involve very few changes in the code, but allow users to run semi-batch and draining liquid reactors in addition to CSTR type reactors.

Thanks for the comment! I think modifying the liquid reactor is a good idea. I will make the change later.

> This change isn't mentioned in the commit title.
> 
> Also I'm pretty confident this doesn't fix the jacobian calculation as most jacobian entries should have units (1/s) while I'm reasonably sure you have units of (mol/s) in the new term.

Yeah I realized this mistake and corrected it in a later commit.

hwpang:
@mjohnson541 I've made the CSTR changes in liquid reactor and added very detailed commits. Would you please review these? Thanks.

hwpang:
> Hao-Wei, Thank you for your contribution. Before I do further reviews, I find some of the checks are not passed. The error in Travis seems to related to your modification
> `Test the liquid batch reactor/cstr with a simple kinetic model. ERROR`. Can you take a look and address this? Besides, in Codacy, there are also 4 minor issues, address these as well. Afterwards, rebase your branch, and let me know. I will do another review on that.

@xiaoruiDong Thank you for your review and comment! I've addressed the Travis and Codacy issues, and I also updated my commits to make them reader friendly. Would you please review them? Thanks!

rmgpy/solver/liquid.pyx
```diff
@@ -754,5 +811,9 @@ cdef class LiquidReactor(ReactionSystem):
                         if ir[j, 2] != -1:
                             pd[ir[j, 2], ip[j, 2]] += deriv
 
+        if residence_time:
+            pd -= 1/residence_time * np.identity(num_core_species, np.float64)
+
```
xiaoruiDong, line 816: I think for semi-batch there should be an extra term to account for the `dilution effect`, since the Volume change. I haven't done a complete proving but the term is something like `(-v_in/V) * np.identity(num_core_species, np.float64)`. I could be wrong. let me know your thoughts

rmgpy/rmg/input.py
```diff
@@ -341,6 +345,28 @@ def liquid_reactor(temperature,
                                  "entries of the same format")
             initialConcentrations[spec] = [Quantity(conc[0]), Quantity(conc[1])]
 
+    if inletVolumetricFlowRate or initialVolume or inletConcentrations or residenceTime:
+        if (inletVolumetricFlowRate or initialVolume or inletConcentrations) and residenceTime:
```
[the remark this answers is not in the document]
hwpang: Thanks for the comment! My original thought was to make the assumption of initial concentrations == inlet concentrations since it's usually the case for many industrial unit ops. I agree that having the optional inlet concentrations for CSTR can be beneficial to users too. I will modify the code.

rmgpy/rmg/input.py
```diff
@@ -1070,7 +1099,15 @@ def save_input_file(path, rmg):
...
-                f.write('        "{0!s}": ({1:g},"{2!s}"),\n'.format(spcs.label, conc.value, conc.units))
+                f.write('        "{0!s}": ({1:g},"mol/m^3"),\n'.format(spcs.label, conc))
+            if system.residence_time:
+                f.write('    residenceTime = ({0:g}, "s"),\n').format(system.residence_time)
+            if not system.constant_volume:
+                f.write('    inletVolumetricFlowRate = ({0:g}, "m^3/s"),\n'.format(system.v_in))
+                f.write('    inletConcentrations={\n')
+                for spcs, conc in system.inlet_concentrations.items():
+                    f.write('        "{0!s}": ({1:g},"mol/m^3"),\n'.format(spcs.label, conc))
+                f.write('    initialVolume = ({0:g}, "m^3"),\n'.format(system.V_0))
```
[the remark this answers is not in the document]
hwpang: Thanks for the suggestion! It is a good idea. I will change all the string formatting in this function to f-string.

rmgpy/solver/liquid.pyx
```diff
@@ -87,15 +101,21 @@ cdef class LiquidReactor(ReactionSystem):
...
-        for label, moleFrac in self.initial_concentrations.items():
+        for label, conc in self.initial_concentrations.items():
             if label == 'T':
                 continue
-            initial_concentrations[species_dict[label]] = moleFrac
+            initial_concentrations[species_dict[label]] = conc
         self.initial_concentrations = initial_concentrations
 
+        inlet_concentrations = {}
+        if self.inlet_concentrations:
```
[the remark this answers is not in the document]
hwpang: Thanks for the comment! I assumed that initial concentrations == inlet concentrations for CSTR previously, but I agree that it will be useful to have the optional inlet concentrations for CSTR too. I will modify the code.

rmgpy/solver/liquid.pyx
hwpang: Thanks for your review and comment! And yes, you're right. I will modify the code.

hwpang:
Thanks for the review and the comments! @xiaoruiDong They are very helpful. I've modified the code according to your comments. I made the inlet concentrations to be optional for the CSTR. If inletConcentrations is provided in the input file, it will use the provided ones. If not, it will assume inletConcentrations equals to initialConcentrations. I also corrected the equation of semi-batch. 

Please let me know if you think the new version is reasonable. If you agree, I will start writing the documentation for this new feature. Thanks!

hwpang:
> Great job, Hao-Wei. You did a good job! I just have 2 minor comments want to discuss. Otherwise, I think the changes look good.

Thanks for your review and your comments! @xiaoruiDong I agreed that inlet_concentrations can have better initialization. I initialized the self.inlet_concentrations to be an empty dictionary, the self.residence_time, self.v_in, self.V_0 to be -1 to avoid this concern. I also corrected the res in residual. Thanks!

PR #1918, opened 2020-04-09
Debugged liquid phase range based reactor

hwpang:
### Motivation or Problem
Range based reactor didn't work for liquid phase before. It's because RMG used `reaction_system.T` to calculate viscosity for liquid phase. However, `reaction_system.T` is `None` when the input T is a list (range). 

### Description of Changes
I modified the code such that RMG will check whether `reaction_system.T` exists. If so, viscosity will be calculated when the model is initialized (i.e. not range based). If not, then viscosity will be calculated in each run of `n_sims` with the corresponding T stored in the condition dictionary from `rmg_memories.get_cond()`. `diffusion_limiter` is still enabled in the initiation stage, no matter the liquid phase job is range based or not.

### Testing
I tested with an old liquid reactor input file. It works now. Output seems reasonable.

### DIFF
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -523,7 +523,8 @@ class RMG(util.Subject):
 
             # Set solvent viscosity for reaction filtering
             for reaction_system in self.reaction_systems:
-                reaction_system.viscosity = solvent_data.get_solvent_viscosity(reaction_system.T.value_si)
+                if reaction_system.T:
+                    reaction_system.viscosity = solvent_data.get_solvent_viscosity(reaction_system.T.value_si)
 
         try:
             self.walltime = kwargs['walltime']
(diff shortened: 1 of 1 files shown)


rmgpy/rmg/main.py
```diff
@@ -770,6 +771,14 @@ def execute(self, initialize=True, **kwargs):
                     for p in range(reaction_system.n_sims):
                         reactor_done = True
                         objects_to_enlarge = []
+
+                        if self.solvent:
```
mjohnson541, line 775: Is it possible to have this code in base.pyx or liquid.pyx instead of in main?

mjohnson541:
It also looks like based on RMG-tests this fix breaks non-ranged liquid reactors?

rmgpy/rmg/main.py
hwpang: I think it is possible, but I'm not sure if that would be desirable...In that case, RMG will need to load database again in `liquid.pyx` because it needs `solvent_data = self.database.solvation.get_solvent_data(self.solvent)` to calculate viscosity, which will take a longer time in total. I guess another way would be pass in the `self.database` attribute. However, it looks like the current RMG design tries to let `main.py` handle things that require the database information, so I think it is more consistent to keep this code in `main.py`. Let me know your thoughts! @mjohnson541

xiaoruiDong:
@mjohnson541 Matt, I am not authorized to merge this PR, can you help me?

mjohnson541:
Shouldn't these commits be squashed together?  Isn't the second one a fix for the first one? If not the second one needs a more specific commit message that explains why it's needed.

hwpang:
> Shouldn't these commits be squashed together? Isn't the second one a fix for the first one? If not the second one needs a more specific commit message that explains why it's needed.

I squashed the second commit and added a bit detailed description in the first one!

hwpang:
> Shouldn't these commits be squashed together? Isn't the second one a fix for the first one? If not the second one needs a more specific commit message that explains why it's needed.

@mjohnson541 I modified the commit message a bit more to explain why the conditions is not None check was added.

PR #1922, opened 2020-04-14
Add A units for trimolecular reactions in ArrheniusBM model fits

oscarwumit:
Now output correct units for trimolecular reactions

### Motivation or Problem
`A` unit for a trimolecular reaction was not correctly assigned. 

### Description of Changes
Updated unit dictionary, now supports a trimolecular reaction.

### DIFF
--- a/rmgpy/kinetics/arrhenius.pyx
+++ b/rmgpy/kinetics/arrhenius.pyx
@@ -650,10 +650,9 @@ cdef class ArrheniusBM(KineticsModel):
         A = np.exp(lnA)
 
         # fill in parameters
-        if len(rxns[0].reactants) == 1:
-            self.A = (A, 's^-1')
-        elif len(rxns[0].reactants) == 2:
-            self.A = (A, 'm^3/(mol*s)')
+        A_units = ['', 's^-1', 'm^3/(mol*s)', 'm^6/(mol^2*s)']
+        order = len(rxns[0].reactants)
+        self.A = (A, A_units[order])
 
         self.n = n
         self.w0 = (w0, 'J/mol')


alongd:
The PR looks good.

In other cases (e.g.: https://github.com/ReactionMechanismGenerator/RMG-Py/blob/d4f77cdde2ff9072b82abe1cfcbd4e25344c5a6b/rmgpy/chemkinTest.py#L311) RMG has a wider range of A factor units. Perhaps RMG's Architect (@amarkpayne ?) could comment about a desired future-proof implementation?

rmgpy/kinetics/arrhenius.pyx
```diff
@@ -650,10 +650,9 @@ cdef class ArrheniusBM(KineticsModel):
         A = np.exp(lnA)
 
         # fill in parameters
-        if len(rxns[0].reactants) == 1:
-            self.A = (A, 's^-1')
-        elif len(rxns[0].reactants) == 2:
-            self.A = (A, 'm^3/(mol*s)')
+        A_units = {1: 's^-1', 2: 'm^3/(mol*s)', 3: 'm^6/(mol^2*s)'}
```
amarkpayne, line 653: You could make this a list because the keys are integers in increasing order (make `''` the zeroth element), but I don't have a strong preference if you want to keep it this way.

amarkpayne:
> The PR looks good.
> 
> In other cases (e.g.:
> 
> https://github.com/ReactionMechanismGenerator/RMG-Py/blob/d4f77cdde2ff9072b82abe1cfcbd4e25344c5a6b/rmgpy/chemkinTest.py#L311
> ) RMG has a wider range of A factor units. Perhaps RMG's Architect (@amarkpayne ?) could comment about a desired future-proof implementation?

Unit conversions between stuff like cm and m should be handled by the ScalarQuantity objects that we are creating, so this specific instance should be fine. As far a future-proofing goes, what did you have in mind? The only thing I can think of would be making it so that a hypothetical order 4 and higher reaction are automatically handled, but I don't think this is necessary as I am not sure that we will ever need this in pure RMG.

alongd:
@amarkpayne, perhaps supported reaction orders along with the corresponding units could be defined once somewhere in RMG instead of having it defined in several locations, each considering different reaction orders.

oscarwumit:
@amarkpayne I agree with @alongd that we should import the units from some common file. This will improve code maintainability. I can make all the changes if that's the way we want to go.

amarkpayne:
> @amarkpayne, perhaps supported reaction orders along with the corresponding units could be defined once somewhere in RMG instead of having it defined in several locations, each considering different reaction orders.

I guess the issue here is that we might prefer giving different units in different circumstances. For example, Chemkin files use 'cm' based units, but I think it is safer for RMG to use 'm' based units (even though in theory ScalarQuantity objects should handle this properly).

We might have some of the defaults defined in quantities.py. If you are going to add a default unit for A factors, here might be a good place to define it. Thoughts?

rmgpy/kinetics/arrhenius.pyx
oscarwumit: I changed it to a list

oscarwumit:
@amarkpayne quantities.py might work, we can define two lists there, one for A with cm units, another one for m units. Since there are many places need to be changed, we should do that in a separate PR. Let's merge this one so we fix the particular issue for ArrheniusBM and then work on consolidating the units. What do you think?

amarkpayne:
@oscarwumit this sounds like a good plan. The changes are fine, but I want to do a rebase-and-merge after the ESS PR gets merged in, so that neither PR needs to be updated. I'll merge this in tonight. Thanks again for the PR!

PR #1924, opened 2020-04-16
Gav comments

### LINKED ISSUE #1877: Should add metadata if a GAV group isempty or missing
### Motivation or Problem
It's hard to asses the uncertainty of GAV contribution to thermo. I open this issue to specifically highlight cases where we might be missing a group altogether for an atom in a molecule (I encountered it several times in the past, but don't have a good example at hand), as well as cases where the group used drastically differs than the functional group it should describe (this is very common with nitrogen, for example a `C=NOO` motif will get its thermo contribution for the N atom from the `N3s-CHH` group.)

### Desired Solution
Cases as highligted above should be clearly marked in the thermo comment, so that model refinement tools could pick that up and try to calculate better thermo for such species. Of course, for the long run we'd like to have a quantifiable uncertainty attached to each GAV estimate, but this is a start.
Another desired solution, and more specific, is to expand/improve GAV, specifically for nitrogen.

### Additional Context
The species
image
was found to have the following GAV estimation: `group(O2s-OsH) + group(N3s-CsHH) + group(Cs-(Cds-Cds)HHH) + group(Cs-(Cds-Cds)HHH) + group(Cds-CdsCsCs) + group(Cdd-N3dCd) + radical(ROOJ)`
which resulted in an enthalpy of ~200 kJ/mol, where in fact its calculated enthalpy is closer to 600 kj/mol

alongd:
fixes #1877

### Motivation or Problem
We'd like to add metadata to GAV when a group is empty or missing.

### Description of Changes
- Added an `is_all_zeros` method to `ThermoData`
- Add a `missing(<group>)` comment to GAV ThermoData when relevant

### Testing
tests were added

### Reviewer Tips
generate GAV thermo for the molecule in the test or for the molecules in #1877, look at the thermo data comment

### DIFF
--- a/rmgpy/thermo/thermodata.pyx
+++ b/rmgpy/thermo/thermodata.pyx
@@ -402,3 +402,21 @@ cdef class ThermoData(HeatCapacityModel):
         :class:`NASAPolynomial` objects.
         """
         return self.to_wilhoit().to_nasa(Tmin, Tmax, Tint, fixedTint, weighting, continuity)
+
+    cpdef bint is_all_zeros(self):
+        """
+        Check whether a ThermoData object has all zero values, e.g.: 
+        
+            ThermoData(
+                Tdata=([300, 400, 500, 600, 800, 1000, 1500], "K"),
+                Cpdata=([0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0], "J/(mol*K)"),
+                H298=(0.0, "kJ/mol"),
+                S298=(0.0, "J/(mol*K)"),
+            )
+    
+        Returns:
+            bool: Whether all values are zeroes or not.
+        """
+        if self.H298.value_si == 0.0 and self.S298.value_si == 0.0 and all([cp == 0.0 for cp in self.Cpdata.value_si]):
+            return True
+        return False
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -1966,13 +1966,21 @@ class ThermoDatabase(object):
             # Iterate over heavy (non-hydrogen) atoms
             if atom.is_non_hydrogen():
                 # Get initial thermo estimate from main group database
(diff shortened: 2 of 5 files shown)


rmgpy/thermo/thermodata.pyx
```diff
@@ -402,3 +402,21 @@ cdef class ThermoData(HeatCapacityModel):
...
+                H298=(0.0, "kJ/mol"),
+                S298=(0.0, "J/(mol*K)"),
+            )
+    
+        Returns:
+            bool: Whether all values are zeroes or not.
+        """
+        if self.H298.value_si == 0.0 and self.S298.value_si == 0.0 and all([cp == 0.0 for cp in self.Cpdata.value_si]):
+            return True
+        return False
```
kspieks, line 422: Great addition and thanks for asking me to review :) Did you ever have problems with some, but not all, of the values being 0? Should that feature be added as an additional check?

alongd: I thought @mjohnson541 commented in yesterday's meeting that these groups were probably intended to be populated (original authors though they are important)

alongd: Sorry, I see now that you meant "some but not all". Then, no, I didn't have problems like that, if that's the case then it indeed makes things more complicated.

kspieks: Okay yea just wanted to check if that was a case we should also consider, but I didn't know if it was a common enough problem. My understanding from yesterday was that the "all zeros case" was affecting others as well so thanks for the fix!

alongd:
thanks all for your comments!
I added a fixup, let me know if you have additional comments, otherwise I'll squash.

mjohnson541:
Can you squash?

alongd:
squashed, thanks!

PR #1926, opened 2020-04-17
Fix labels of missing thermo groups

alongd:
Related to PR #1924

Put 'H' last, and don't repeat a 'Cb' label for neighboring atoms of a Cb atom.

These small label corrections will allow easier group interpretation in T3.

### DIFF
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -1975,12 +1975,17 @@ class ThermoDatabase(object):
                     logging.error(molecule.to_adjacency_list())
                     raise
                 if not data_added:
-                    neighbors = ''.join(sorted([atom2.atomtype.label for atom2 in atom.edges.keys()]))
-                    if thermo_data.comment:
-                        thermo_data.comment += f' + missing({atom.atomtype.label}-{neighbors})'
-                    else:
-                        thermo_data.comment = f'Thermo group additivity estimation: ' \
-                                              f'missing({atom.atomtype.label}-{neighbors})'
+                    neighbors = ''.join(sorted([atom2.atomtype.label for atom2 in atom.edges.keys()
+                                                if atom2.atomtype.label != 'H']))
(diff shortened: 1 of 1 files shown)


alongd:
@mjohnson541, could you take a look at this PR?

rmgpy/data/thermo.py
```diff
@@ -1975,12 +1975,17 @@ def compute_group_additivity_thermo(self, molecule):
...
-                    neighbors = ''.join(sorted([atom2.atomtype.label for atom2 in atom.edges.keys()]))
-                    if thermo_data.comment:
-                        thermo_data.comment += f' + missing({atom.atomtype.label}-{neighbors})'
-                    else:
-                        thermo_data.comment = f'Thermo group additivity estimation: ' \
-                                              f'missing({atom.atomtype.label}-{neighbors})'
+                    neighbors = ''.join(sorted([atom2.atomtype.label for atom2 in atom.edges.keys()
+                                                if atom2.atomtype.label != 'H']))
+                    neighbors += 'H' * len(['H' for atom2 in atom.edges.keys() if atom2.atomtype.label == 'H'])
+                    if atom.atomtype.label == 'Cb':
```
mjohnson541, line 1981: Could this handling of Cb be confusing to a user since they might expect it to behave like the other labels?

alongd: I thought I was following an RMG-db convention: https://github.com/ReactionMechanismGenerator/RMG-database/blob/master/input/thermo/groups/group.py#L49457

alongd:
@mjohnson541, I think the handling of `Cb` is following an existing convention in RMG-database. If that's not an issue, and unless there are addiitonal comments, this PR should be ready to go.

alongd:
done and rebased

rmgpy/data/thermo.py
mjohnson541: Okay makes sense!

PR #1929, opened 2020-04-21
Update copyright headers for 2020

amarkpayne:
### Description of Changes
I ran the `update_headers` function in `utilities.py` after updating our license for the year 2020. I went through the files to make sure there were no unnecessary changes.

### DIFF
--- a/Arkane.py
+++ b/Arkane.py
@@ -4,7 +4,7 @@
 #                                                                             #
 # RMG - Reaction Mechanism Generator                                          #
 #                                                                             #
-# Copyright (c) 2002-2019 Prof. William H. Green (whgreen@mit.edu),           #
+# Copyright (c) 2002-2020 Prof. William H. Green (whgreen@mit.edu),           #
 # Prof. Richard H. West (r.west@neu.edu) and the RMG Team (rmg_dev@mit.edu)   #
 #                                                                             #
 # Permission is hereby granted, free of charge, to any person obtaining a     #
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,4 +1,4 @@
-Copyright (c) 2002-2019 Prof. William H. Green (whgreen@mit.edu),
+Copyright (c) 2002-2020 Prof. William H. Green (whgreen@mit.edu),
 Prof. Richard H. West (r.west@neu.edu) and the RMG Team (rmg_dev@mit.edu)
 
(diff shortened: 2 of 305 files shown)


alongd:
Why do we need to have these headers in each file? Isn't having a license enough?
I do understand having these on standalone scripts, though.

amarkpayne:
I know we have discussed this in the past, but I cannot find a written record of it in the PRs and Issues of the past, so perhaps we only ever discussed this in person. We certainly want them on the scripts and main files (rmg.py and arkane.py), and then be consistent across the remaining files either way. I think a lot of mainstream packages out there don't put headers like this on their files.

Since we already have the headers and a script to automate adding them, it is whatever we prefer stylistically. Let's poll the other developers and see what they think. Do people like having the headers there or does it just waste space?

mliu49:
I think it's important to have a copyright header in every file, since it's possible that people might take a random file and modify it for their purposes. Including the full license isn't really necessary, and is mainly a choice which was carried over from [RMG-Java](https://github.com/ReactionMechanismGenerator/RMG-Java/blob/master/source/RMG/jing/chem/Species.java).

For comparison, both [RDKit](https://github.com/rdkit/rdkit/blob/master/Code/GraphMol/Atom.cpp) and [Cantera](https://github.com/Cantera/cantera/blob/master/src/thermo/IdealGasPhase.cpp) include copyright headers which reference their license file.

xiaoruiDong:
@amarkpayne Just in case you are waiting for me to merge this PR, I am not authorized to merge this pull request.

amarkpayne:
> @amarkpayne Just in case you are waiting for me to merge this PR, I am not authorized to merge this pull request.

Thanks @xiaoruiDong ! Sorry, I should have followed up. I want to let some the of the other developers weigh in before we merge this in. @mjohnson541 what are your thoughts? Should we keep the headers that we have or follow a style closer to that of RDKit/Cantera?

mjohnson541:
I mean the fundamental idea is that if someone copies a file over by default they'll retain the copyright/license. I'm not strongly opinionated on copyright + license vs copyright + reference to license, but the real difference is just how in your face we want to make it. I think sticking with what we have currently is fine.

amarkpayne:
Sounds good, thanks for weighing in everyone! I'll go ahead and merge this in then

PR #1931, opened 2020-04-21
BugFix: load_negative_frequency() undesired error behavior

xiaoruiDong:
Very minor. If no negative frequency, an error will be raised due to slice index 0 from an empty list, and `if frequency is None` will never happens.


kspieks:
Sorry to not think of this and thanks for catching it!

xiaoruiDong:
> Sorry to not think of this and thanks for catching it!

@kspieks no worry, I just saw this quite randomly. lol.

alongd:
@xiaoruiDong, please rebase

Issue #1932, opened 2020-04-25
mergeModels.py fails to guarantee unique indices and prevents plotting trajectories in Cantera


kspieks:
### Bug Description
I recently merged 4 models using the mergeModels.py script and was trying to use the new chem.inp, species_dictionary.txt, and tran.dat to obtain plots with Cantera. However, I haven’t been able to get a plot with Cantera due to numerous errors about reactions being unbalanced, such as the example shown below:

```
CanteraError: 
***********************************************************************
CanteraError thrown by checkReactionBalance:
The following reaction is unbalanced: C18H31(83) + S(118) <=> C11ene(46) + PDD(1)
  Element    Reactants    Products
  C           54           29
  H           90           52
***********************************************************************
```

This effectively renders the merged model useless since it cannot be simulated/plotted as is or restarted with RMG to run further. It seems that multiple models contained `S(<index>)` that were unique within the model, but not between the models, and these differences were overwritten while merging such that they now all point to one species, which is incorrect. I'm unsure if there are other indexing issues..

### How To Reproduce
Run the mergeModels.py script: `python RMG-Py/scripts/mergeModels.py <insert arguments for model 1, 2, etc>`

### Expected Behavior
Ensure unique indexing so the merged model can either be simulated with Cantera or restarted to run further.

PR #1933, opened 2020-04-25
Fix indexing while merging models

kspieks:
### Motivation or Problem
The script mergeModels.py fails to guarantee unique indices, which prevents the merged models from being simulated or restarted as described in issue #1932. 

### Description of Changes
mergemodels.py was reverted back to how it was before PR #1649. The motivation of PR #1649  was to avoid changing chemkin or cantera names at the expense of not ensuring unique indices. 

### Testing
An unnecessary check was removed from mergemodelsTest.py. All unittests passed locally.


kspieks:
Thanks for reviewing! I just rebased so the PR can be merged.

amarkpayne:
I'll merge once the tests pass

PR #1934, opened 2020-04-26
Profiler tools doc

kspieks:
### Description of Changes
Updates the RMG documentation to include updates from the merged `profiler_tools` branch, PR #1789. I look forward to your feedback.

### DIFF
--- a/documentation/source/users/rmg/running.rst
+++ b/documentation/source/users/rmg/running.rst
@@ -12,19 +12,74 @@ Basic run::
 
 	python rmg.py input.py
 
-Run by restarting from a seed mechanism ::
+.. _inputflags:
 
-    python rmg.py input.py -r path/to/seed/
+Input flags
+-----------
+
+The options for input flags can be found in ``/RMG-Py/rmgpy/util.py``. Running ::
+
+ 	python rmg.py -h
+
+at the command line will print the documentation from ``util.py``, which is reproduced below for convenience::
+
+	usage: rmg.py [-h] [-q | -v | -d] [-o DIR] [-r path/to/seed/] [-p] [-P]
+              [-t DD:HH:MM:SS] [-i MAXITER] [-n MAXPROC] [-k]
+              FILE
+
+	Reaction Mechanism Generator (RMG) is an automatic chemical reaction mechanism
+	generator that constructs kinetic models composed of elementary chemical
+	reaction steps using a general understanding of how molecules react.
+
+	positional arguments:
+	  FILE                  a file describing the job to execute
+
+	optional arguments:
+	  -h, --help            show this help message and exit
+	  -q, --quiet           only print warnings and errors
+	  -v, --verbose         print more verbose output
+	  -d, --debug           print debug information
+	  -o DIR, --output-directory DIR
+	                        use DIR as output directory
+	  -r path/to/seed/, --restart path/to/seed/
+	                        restart RMG from a seed
+	  -p, --profile         run under cProfile to gather profiling statistics, and
+	                        postprocess them if job completes
+	  -P, --postprocess     postprocess profiling statistics from previous
+	                        [failed] run; does not run the simulation
+	  -t DD:HH:MM:SS, --walltime DD:HH:MM:SS
+	                        set the maximum execution time
+	  -i MAXITER, --maxiter MAXITER
+	                        set the maximum number of RMG iterations
+	  -n MAXPROC, --maxproc MAXPROC
+	                        max number of processes used during reaction
+	                        generation
+	  -k, --kineticsdatastore
+	                        output a folder, kinetics_database, that contains a
+	                        .txt file for each reaction family listing the
+	                        source(s) for each entry
+
+Some representative example usages are shown below.
+
+Run by restarting from a seed mechanism::
+
+    python rmg.py -r path/to/seed/ input.py
 
 Run with CPU time profiling::
 
-    python rmg.py input.py -p
+    python rmg.py -p input.py
 
 Run with multiprocessing for reaction generation and QMTP::
 
     python rmg.py -n <Max number of processes allowed> input.py 
 
-We recommend you make a job-specific directory for each RMG simulation. Some jobs can take quite a while to complete, so we also recommend using a job scheduler if working in a linux environment. 
+Run with setting a limit on the maximum execution time::
+
+	python rmg.py -t <DD:HH:MM:SS> input.py
+
+Run with setting a limit on the maximum number of iterations::
+
+	python rmg.py -i <Max number of desired iterations> input.py
 
 
 Details on the multiprocessing implementation
--- a/documentation/source/users/rmg/running.rst
+++ b/documentation/source/users/rmg/running.rst
@@ -89,3 +89,20 @@ Currently, multiprocessing is implemented for reaction generation and the genera
 
 In python 3.4 new forking contexts 'spawn' and 'forkserver' are available. These methods will create new processes which share nothing or limited state with the parent and all memory passing is explicit. Once RMG is transferred to python 3 it is recommended to use the spawn or forkserver forking context to potentially allow for an increased number of processes.
 
+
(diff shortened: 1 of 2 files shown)


documentation/source/users/rmg/running.rst
```diff
@@ -4,33 +4,107 @@
...
 
+
+Details on profiling RMG jobs
+-----------------------------
+
+Here, we explain how to profile an RMG job. For starters, use the ``saveSeedModulus`` option in the input file, as described in the Section :ref:`Miscellaneous Options <miscellaneousoptions>`, to save the seed mechanism at regular intervals, perhaps every 50 or 100 iterations depending on the size of the mechanism. This flag is particularly important for saving intermediate steps when working with large mechanisms; it may be prudent to save and examine how the chemistry changes over mechanism development rather than just obtaining the final seed mechanism.
+
+
+Additionally, these seeds can then be restarted with use of the ``-i`` flag, as described in the Section :ref:`Input Flags <inputflags>` above. Restarting these seeds allows examination of how computational effort, time spent in each module, individual processor memory consumption if using the the ``-n`` flag, and overall memory consumption change over the course of mechanism development. To time profile, one could use::
```
rwest, line 101: Talking about restarting seeds, I thought you would refer to the `-r` option. But then the example is about something else. Do you mean to discuss seeds being restarted, or are you talking about something else? Sorry if I seem obtuse, I'm just a bit confused.

documentation/source/users/rmg/running.rst
```diff
@@ -4,33 +4,107 @@
...
 
+Run with setting a limit on the maximum execution time::
+
+	python rmg.py input.py -t <DD:HH:MM:SS>
+
+Run with setting a limit on the maximum number of iterations::
+
+	python rmg.py input.py -i <Max number of desired iterations>
+
 We recommend you make a job-specific directory for each RMG simulation. Some jobs can take quite a while to complete, so we also recommend using a job scheduler if working in a linux environment.
```
rwest, line 84: This final sentence probably belongs better in the _Running a Job_ section than the _Input Flags_ section, as it is not to do with input flags.

kspieks: Good point. Thanks!

documentation/source/users/rmg/running.rst
kspieks: Thanks for the review! I see that I could’ve explained that better. The full syntax that I should’ve included for these two examples are:

`rmg.py restart_from_seed.py -r <path_to_seed>/seed -p -i 15`

and 

`mprof run --multiprocess rmg.py -n 3 restart_from_seed.py -r <path_to_seed>/seed -p -i 15`

My previous use of `input.py` was incorrect since this paragraph was discussing how to pair the profiling tools with a restart job. I additionally wasn’t clear regarding my omission of the `-r` flag. Since the default argument for `-r` is an empty string, by default RMG looks for the required `seed` directory for a restart job in the same directory as the user is running `restart_from_seed.py`. To help me be organized, I often place the `seed` directory in the same directory as `restart_from_seed.py`, which allows me to omit the `-r` flag by taking advantage of its default argument. I’ve updated this PR with the proper syntax. 


Please let me know if I can update this further to be more clear. We’d love to have people understand the restart feature and new profiling tools 🙂

documentation/source/users/rmg/running.rst
```diff
@@ -4,33 +4,105 @@
...
 
 Currently, multiprocessing is implemented for reaction generation and the generation of QMfiles when using the QMTP option to compute thermodynamic properties of species. The processes are spawned and closed within each function. The number of processes is determined based on the ratio of currently available RAM and currently used RAM. The user can input the maximum number of allowed processes from the command line. For each reaction generation or QMTP call the number of processes will be the minimum value of either the number of allowed processes due to user input or the value obtained by the RAM ratio. The RAM limitation is employed, because multiprocessing is forking the base process and the memory limit (SWAP + RAM) might be exceeded when using too many processors for a base process large in memory.
 
 In python 3.4 new forking contexts 'spawn' and 'forkserver' are available. These methods will create new processes which share nothing or limited state with the parent and all memory passing is explicit. Once RMG is transferred to python 3 it is recommended to use the spawn or forkserver forking context to potentially allow for an increased number of processes.
```
rwest, line 90: Not relevant to this PR, but I just noticed this comment about something that should be done "once RMG is transferred to Python 3". Is this on anyone's radar? Should it be in the issue tracker or on a kanban?

documentation/source/users/rmg/running.rst
```diff
@@ -4,33 +4,105 @@
...
+Some representative example usages are shown below.
+
+Run by restarting from a seed mechanism::
 
     python rmg.py input.py -r path/to/seed/
 
-Run with CPU profiling::
+Run with CPU time profiling::
 
     python rmg.py input.py -p
```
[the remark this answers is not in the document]
rwest: Oh, perhaps this isn't better. I was just trying to be consistent with the usage instructions above. Does it matter at all what order they come in? The examples are rather inconsistent. Maybe it's not a big deal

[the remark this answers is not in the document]
kspieks: Indeed, either placement of the flags yields the intended result. I’ve now pushed changes with the consistent format.

documentation/source/users/rmg/running.rst
kspieks: Interesting point to bring up. It was my understanding that a previous member thoroughly explored multiprocessing options since we transitioned to python 3 awhile ago; the group has been using it for around a year I think or at least since I’ve arrived. I’m actually unsure if that paragraph is still relevant.. a quick search seems to show that both spawn and forkserver are inherently much slower than fork ([link1](https://docs.python.org/3/library/multiprocessing.html) , [link2](https://riptutorial.com/python/example/14842/the-multiprocessing-module
)), but I’ll ask around.

rwest: Then maybe the paragraph just needs fixing/removing. Thanks for looking into it

kspieks: After digging through our [Wiki](https://github.com/ReactionMechanismGenerator/RMG-Py/wiki/Multiprocessing), I found an updated version of this paragraph:

In python 3.4 new forking contexts 'spawn' and 'forkserver' are available. These methods will create new processes which share nothing or limited state with the parent and all memory passing is explicit. Using 'spawn' or 'forkserver' in set_start_method('forkserver') raises a DatabaseError: rmgpy.exceptions.DatabaseError: Could not get database with name: kinetics. We suspect that necessary database resources are identified as unnecessary resources and therefore not inherited even though they are needed by the child process.

This result is now included in the recently pushed version of this branch. Let me know if I can help update anything else before merging 🙂

mjohnson541: Is the documentation really the right place to leave comments like these? I feel like this belongs somewhere else.

rwest:
Interesting find about the multiprocessing, but my vote would be delete from the documentation, because it tells users nothing about how to use RMG.

kspieks:
Sounds good. We'll leave the information in the Wiki. The PR has been updated

documentation/source/users/rmg/running.rst
```diff
@@ -28,7 +28,7 @@ We recommend you make a job-specific directory for each RMG simulation. Some job
 
 
 Details on the multiprocessing implementation
-----------------------------------------------
```
amarkpayne, line 31: Any reason for the whitespace change?

rwest: Not just whitespace. They're trying to make the underline the same length as the line it's underlining.

rwest:
Looks good to me

kspieks:
@amarkpayne Ah I didn't realize the source code was automatically compiled into the documentation pdf. The last commit has now been dropped

amarkpayne:
Thanks for the documentation changes @kspieks , and thanks @rwest for catching this and reviewing!

PR #1936, opened 2020-04-27
Use Kekulized structures for Transport Estimation

mjohnson541:
Because there are no Joback groups for aromatics currently we can't estimate transport properly for benzene, this keuklizes aromatic molecules so they can be estimated using transport Joback groups.


amarkpayne:
@mjohnson541 as we discussed offline, I have modified the unit test so that future developers will know that they need to revert the changes in this PR if they add groups for benzene. I am fine with your changes so if you are fine with the modifications to the unit tests feel free to merge this in.

PR #1938, opened 2020-04-30
Update reference database

amarkpayne:
### Motivation or Problem
This PR adds in a function that greatly simplifies updating the reference database. The user can put all of the ArkaneSpecies YAML files in one directory and call the function by passing this path.

Furthermore, this PR fixes a few bugs that had previously gone unnoticed..

### Testing
I used the function in this PR to load in the DLPNO CCSD(T) calculations done by @dranasinghe


arkane/encorr/reference.py
```diff
@@ -665,5 +669,57 @@ def _is_valid_calculated_data(data_dictionary):
...
+            default to the ``level_of_theory`` parameter of the ArkaneSpecies objects
+        ref_paths (list): A list to the paths of reference sets to load. By default only the main reference set is
+            loaded
+
+    Returns:
+        None
+    """
+    spcs_list = []
+    for f in os.listdir(path):
+        if any(file_extension in f.lower() for file_extension in ['.yml', '.yaml']):
```
cgrambow, line 692: This is super nitpicky and very much a personal preference, but it would save one line. I typically just loop through the output of calling `glob.glob` on `'*.yml'` and `'*.yaml'`. I guess that means you'd also have to import `glob`, so you wouldn't actually save a line... Feel free to ignore this comment completely.

arkane/encorr/reference.py
```diff
@@ -665,5 +669,57 @@ def _is_valid_calculated_data(data_dictionary):
...
+            spc = ArkaneSpecies.__new__(ArkaneSpecies)
```
cgrambow, line 693: This isn't so much an issue with this PR but more of an issue with the `ArkaneSpecies` class. Initializing an empty instance in this way seems a little strange. I feel like there should be a mechanism to just initialize an instance by providing a yaml file directly.

arkane/encorr/reference.py
```diff
@@ -665,5 +669,57 @@ def _is_valid_calculated_data(data_dictionary):
...
+        species = Species().from_adjacency_list(arkane_spc.adjacency_list, raise_atomtype_exception=False,
+                                                raise_charge_exception=False)
+        # Try to generate resonance structures
+        try:
+            species.generate_resonance_structures()
+        except (AtomTypeError, ValueError):  # This can fail for charged or unusual species. Move on without it
+            pass
+
+        for ref_spcs in ref_spcs_list:
+            if ref_spcs.is_isomorphic(species, generate_resonance_structures=False):
```
cgrambow, line 711: I'm not confident that `is_isomorphic` will work. From what I can tell, `is_isomorphic` does not consider charge, so would, for example, methylium ([CH3+]) and methanide ([CH3-]) be isomorphic?

arkane/encorr/reference.py
```diff
@@ -165,6 +166,40 @@ def as_dict(self):
 
         return dictionary
 
+    def is_isomorphic(self, species: Union[ArkaneSpecies, Species], generate_resonance_structures: bool = True) -> bool:
```
cgrambow, line 169: Regarding my previous comment about the issues with `is_isomorphic`, maybe the charge comparison can be built into this function?

arkane/encorr/reference.py
amarkpayne: I'll add it, just because it is worth learning about glob so that I can use it again in the future when needed. Thanks for the suggestion!

arkane/encorr/reference.py
amarkpayne: I agree. Part of the issue is that you can't initialize an ArkaneSpecies object without passing arguments to __init__. Having a separate function for this feels clunky as well. I wonder if we can get around this by having `load_yaml` return the ArkaneSpecies object, so that this method can be called on the class rather than an instance. I'll investigate and if it works I'll push a commit

arkane/encorr/reference.py
amarkpayne: Interesting, I forgot to check to see if charge was considered in the isomorphism checks. When I was loading in Duminda's calc last night though, I did check that all of the species got mapped to one and only one reference species, so it seems like it was able to differentiate somehow. I'll investigate further, but at the very least add a check for charge (and multiplicity if needed while I am at it) to the isomorphism check method

cgrambow:
Another consideration: In addition to having the function, you could also add a script to update the reference database. Alternatively, the script could be `reference.py` itself by adding some code after the `if __name__ == '__main__':` part.

amarkpayne:
@JacksonBurns I might want this one to still make it in. If I get to it on a random weekend in the next 90 days great but if not then we can just let this one die out

JacksonBurns:
Thanks for the update @amarkpayne! The [conflicts](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1938/conflicts) look simple to resolve - I think this could make it in easily.

Could you elaborate on the bug you mention in the opening comment?

JacksonBurns:
Temporarily re-opening this to change base to main, I just realized it's pointed at master right now.

JacksonBurns:
Resolved some simple merge conflicts just so it doesn't get forgotten in the future, but leaving this on the long-term board and closing for now.

Issue #1937, opened 2020-04-29
equilibrium constant for surface reaction


davidfarinajr:
### Bug Description
The `reaction.get_equilibrium_constant` method currently assumes an ideal gas mixture, however it is being used for surface reactions to get `Kc` when we `generate_rate_coefficients` for the `SurfaceReactor`.  I think this method works for surface reactions if the number of surface sites are balanced on either side of the reactions and "cancel out".  However, for some bidentate reaction families, the surface sites don't cancel.  Since we assume ideal gas to determine the reference concentration, this may cause problems.

PR #1945, opened 2020-05-07
Reference concentration for surface species when calculating Kc

davidfarinajr:
### Motivation or Problem
This is related to issue https://github.com/ReactionMechanismGenerator/RMG-Py/issues/1937
We currently assume ideal gas mixture to determine Kc.  We should be using different reference concentrations for gas and surface species.  

### Description of Changes
Use reference concentration of (surface site density) for surface site species and ideal gas (P/RT) for gas species.  That way, gas concentration scales with temperature and surface species concentration is independent of temperature.  Therefore, we use the change in mols of gas in the reaction to determine Kc (instead of len(reactants) - len(products)).  However, the species object needs a molecule to determine its phase (if it contains surface site), so I added a try and except for this.

### Testing
I added unit test for determining Kc for a surface reaction. I also modified the gas Kc test to test reactions with species that contain and don't contain a molecule.

### Reviewer Tips
I don't really like my try and except block.  I'm interested in suggestions to improve this.

### DIFF
--- a/rmgpy/reaction.py
+++ b/rmgpy/reaction.py
@@ -555,27 +555,55 @@ class Reaction:
                 raise
         return dGrxn
 
-    def get_equilibrium_constant(self, T, type='Kc'):
+    def get_equilibrium_constant(self, T, type='Kc', surface_site_density=2.5e-05):
         """
         Return the equilibrium constant for the reaction at the specified
-        temperature `T` in K. The `type` parameter lets	you specify the
-        quantities used in the equilibrium constant: ``Ka`` for	activities,
-        ``Kc`` for concentrations (default), or ``Kp`` for pressures. Note that
-        this function currently assumes an ideal gas mixture.
+        temperature `T` in K and reference `surface_site_density`
+        in mol/m^2 (2.5e-05 default) The `type` parameter lets you specify
+        the quantities used in the equilibrium constant: ``Ka`` for activities,
+        ``Kc`` for concentrations (default), or ``Kp`` for pressures.  This
+        function assumes a reference pressure of 1e5 Pa for gas phases species
+        and uses the ideal gas law to determine reference concentrations. For
+        surface species, the `surface_site_density` is the assumed reference.
         """
         cython.declare(dGrxn=cython.double, K=cython.double, C0=cython.double, P0=cython.double)
         # Use free energy of reaction to calculate Ka
         dGrxn = self.get_free_energy_of_reaction(T)
         K = np.exp(-dGrxn / constants.R / T)
         # Convert Ka to Kc or Kp if specified
+        # Assume a pressure of 1e5 Pa for gas phase species
         P0 = 1e5
+        # Determine the number of gas phase reactants and products. For gas species,
+        # we will use 1e5 Pa and ideal gas law to determine reference concentration.
+        try:
+            number_of_gas_reactants = len([spcs for spcs in self.reactants if not spcs.contains_surface_site()])
+            number_of_gas_products = len([spcs for spcs in self.products if not spcs.contains_surface_site()])
+        except IndexError:
+            logging.warning("Species do not have an rmgpy.molecule.Molecule "
+                            "Cannot determine phases of species. We will assume "
+                            "ideal gas mixture when calculating Kc and Kp.")
+            number_of_gas_reactants = len(self.reactants)
+            number_of_gas_products = len(self.products)
+
+        # Determine the number of surface reactants and products.  For surface species,
+        # we will use the provided `surface_site_density` as the reference
+        number_of_surface_reactants = len(self.reactants) - number_of_gas_reactants
+        number_of_surface_products = len(self.products) - number_of_gas_products
+
+        # Determine the change in the number of mols of gas and surface species in the reaction
+        dN_surf = number_of_surface_products - number_of_surface_reactants # change in mols of surface spcs
+        dN_gas = number_of_gas_products - number_of_gas_reactants # change in mols of gas spcs
+
         if type == 'Kc':
             # Convert from Ka to Kc; C0 is the reference concentration
-            C0 = P0 / constants.R / T
-            K *= C0 ** (len(self.products) - len(self.reactants))
+            if dN_gas:
+                C0 = P0 / constants.R / T
+                K *= C0 ** dN_gas
+            if dN_surf:
+                K *= surface_site_density ** dN_surf
         elif type == 'Kp':
             # Convert from Ka to Kp; P0 is the reference pressure
-            K *= P0 ** (len(self.products) - len(self.reactants))
+            K *= P0 ** dN_gas
         elif type != 'Ka' and type != '':
             raise ReactionError('Invalid type "{0}" passed to Reaction.get_equilibrium_constant(); '
                                 'should be "Ka", "Kc", or "Kp".'.format(type))
--- a/rmgpy/reactionTest.py
+++ b/rmgpy/reactionTest.py
@@ -258,6 +258,17 @@ class TestSurfaceReaction(unittest.TestCase):
 
         self.assertEqual(self.rxn1s.generate_reverse_rate_coefficient().A.units, 'm^2/(mol*s)')
 
+    def test_equilibrium_constant_surface_kc(self):
+        """
+        Test the Reaction.get_equilibrium_constant() method for Kc of a surface reaction.
+        """
+        Tlist = numpy.arange(400.0, 1600.0, 200.0, numpy.float64)
+        Kclist0 = [15375.20186, 1.566753, 0.017772, 0.0013485,
+                   0.000263180, 8.73504e-05]
+        Kclist = self.rxn1s.get_equilibrium_constants(Tlist, type='Kc')
+        for i in range(len(Tlist)):
+            self.assertAlmostEqual(Kclist[i] / Kclist0[i], 1.0, 4)
+
 class TestReaction(unittest.TestCase):
     """
     Contains unit tests of the Reaction class.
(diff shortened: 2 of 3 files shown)


rmgpy/reaction.py
```diff
@@ -560,22 +560,37 @@ def get_equilibrium_constant(self, T, type='Kc'):
         Return the equilibrium constant for the reaction at the specified
         temperature `T` in K. The `type` parameter lets	you specify the
         quantities used in the equilibrium constant: ``Ka`` for	activities,
-        ``Kc`` for concentrations (default), or ``Kp`` for pressures. Note that
-        this function currently assumes an ideal gas mixture.
+        ``Kc`` for concentrations (default), or ``Kp`` for pressures.  This
+        function assumes a reference pressure of 1e5 Pa for gas phases species
+        and uses the ideal gas law to determine reference concentrations. For 
+        solid surface site species, a reference concentration of 1 is assumed.
```
skrsna, line 566: added the reference from Cantera's source code https://github.com/Cantera/cantera/blob/master/include/cantera/kinetics/InterfaceKinetics.h#L467

rmgpy/reaction.py
```diff
@@ -560,22 +560,37 @@ def get_equilibrium_constant(self, T, type='Kc'):
...
         dGrxn = self.get_free_energy_of_reaction(T)
         K = np.exp(-dGrxn / constants.R / T)
         # Convert Ka to Kc or Kp if specified
+        # Assume a pressure of 1e5 Pa for gas phase species
         P0 = 1e5
+        # Determine the number of gas phase reactants and products. The change in mols
+        # of gas will be used to determine Kc and Kp since solid species have reference
+        # concentration of 1
+        try:
+            gas_reacts = [spcs for spcs in self.reactants if not spcs.is_surface_site()]
```
davidfarinajr, line 579: Just realized I need to change `is_surface_site()` to `contains_surface_site()`

rmgpy/reaction.py
```diff
@@ -560,22 +560,37 @@ def get_equilibrium_constant(self, T, type='Kc'):
...
         # Convert Ka to Kc or Kp if specified
+        # Assume a pressure of 1e5 Pa for gas phase species
         P0 = 1e5
+        # Determine the number of gas phase reactants and products. The change in mols
+        # of gas will be used to determine Kc and Kp since solid species have reference
+        # concentration of 1
+        try:
+            gas_reacts = [spcs for spcs in self.reactants if not spcs.contains_surface_site()]
+            gas_prods = [spcs for spcs in self.products if not spcs.contains_surface_site()]
+        except IndexError:
```
nateharms, line 581: Okay, after looking through `species.py`, it looks like the `IndexError` is appropriate here. I understand why you'd feel it was wrong at first, I had to do some digging myself.

rwest:
This looks like it was approved three times three months ago. @davidfarinajr please could you rebase it once more and post in slack to try to get it merged? thanks

davidfarinajr:
I'm still not confident about using 1 for reference concentration of surface species.  1 is typically used for solids and liquids since their concentration does't change in the reaction, but the concentration of our adsorbates changes a lot (ranges from 0 to the surface site density).   I don't think we should be using ideal gas to determine reference concentration (as our surface concentrations don't change with temp), but perhaps we should be using the surface site density as the reference instead of 1?

rwest:
Ok. Maybe we consult @cfgoldsmith ?

rwest:
Could also compare with Cantera.

skrsna:
AFAIK, David's code is based on Cantera's code so I don't know how I feel about comparing it to Cantera's values.

davidfarinajr:
According to this review paper, it looks like this topic, of what to use for standard concentrations of adsorbates for multi dentate adsorptions reactions (where the surface species don't cancel out), is somewhat disputed:
https://pubs.acs.org/doi/pdf/10.1021/es305180e

davidfarinajr:
After discussion with @cfgoldsmith , we decided to use `surface site density` for reference concentration instead of 1.  I have rebased and made this change.  I added `surface_site_density` as a parameter to the `get_equilibrium_constant` method with a default value of `2.5e-05` mol/m^2 (approximately the value for Pt111).

[rmgcat_reference_surface_concentration.pdf](https://github.com/ReactionMechanismGenerator/RMG-Py/files/5138752/rmgcat_reference_surface_concentration.pdf)

rmgpy/reaction.py
```diff
@@ -555,27 +555,50 @@ def get_free_energy_of_reaction(self, T):
...
-        quantities used in the equilibrium constant: ``Ka`` for	activities,
-        ``Kc`` for concentrations (default), or ``Kp`` for pressures. Note that
-        this function currently assumes an ideal gas mixture.
+        temperature `T` in K and reference `surface_site_density` 
+        in mol/m^2 (2.5e-05 default) The `type` parameter lets you specify 
+        the quantities used in the equilibrium constant: ``Ka`` for activities,
+        ``Kc`` for concentrations (default), or ``Kp`` for pressures.  This
+        function assumes a reference pressure of 1e5 Pa for gas phases species
+        and uses the ideal gas law to determine reference concentrations. For 
+        surface site species, a reference concentration of 1 is assumed.
```
rwest, line 567: Docstring needs updating to reflect latest changes.

rmgpy/reaction.py
```diff
@@ -555,27 +555,50 @@ def get_free_energy_of_reaction(self, T):
...
         cython.declare(dGrxn=cython.double, K=cython.double, C0=cython.double, P0=cython.double)
         # Use free energy of reaction to calculate Ka
         dGrxn = self.get_free_energy_of_reaction(T)
         K = np.exp(-dGrxn / constants.R / T)
         # Convert Ka to Kc or Kp if specified
+        # Assume a pressure of 1e5 Pa for gas phase species
         P0 = 1e5
+        # Determine the number of gas phase reactants and products. The change in mols
+        # of gas will be used to determine Kc and Kp since solid species have reference
+        # concentration of 1
```
rwest, line 578: update comment

rmgpy/reaction.py
```diff
@@ -555,27 +555,50 @@ def get_free_energy_of_reaction(self, T):
...
+        try:
+            gas_reacts = len([spcs for spcs in self.reactants if not spcs.contains_surface_site()])
+            gas_prods = len([spcs for spcs in self.products if not spcs.contains_surface_site()])
+        except IndexError:
+            logging.warning(f"Species do not have an rmgpy.molecule.Molecule "
```
rwest, line 583:
```suggestion
            logging.warning("Species do not have an rmgpy.molecule.Molecule "  
```

rmgpy/reaction.py
```diff
@@ -555,27 +555,50 @@ def get_free_energy_of_reaction(self, T):
...
+        except IndexError:
+            logging.warning(f"Species do not have an rmgpy.molecule.Molecule "  
+                            "Cannot determine phases of species. We will assume "
+                            "ideal gas mixture when calculating Kc and Kp.")
+            gas_reacts = len(self.reactants)
+            gas_prods = len(self.products)
+
+        surf_reacts = len(self.reactants) - gas_reacts
+        surf_prods = len(self.products) - gas_prods
+        v_surf = surf_prods - surf_reacts # change in mols of surface spcs
```
rwest, line 591: not a big fan of these variable names. I guess you were going for "nu" but the v makes me start asking "why vee? is it volume?". Maybe `dN_gas` or `delta_stoichiometry_gas` or `change_in_moles_of_gas`?

rmgpy/reaction.py
```diff
@@ -555,27 +555,50 @@ def get_free_energy_of_reaction(self, T):
...
+        # of gas will be used to determine Kc and Kp since solid species have reference
+        # concentration of 1
+        try:
+            gas_reacts = len([spcs for spcs in self.reactants if not spcs.contains_surface_site()])
+            gas_prods = len([spcs for spcs in self.products if not spcs.contains_surface_site()])
+        except IndexError:
+            logging.warning(f"Species do not have an rmgpy.molecule.Molecule "  
+                            "Cannot determine phases of species. We will assume "
+                            "ideal gas mixture when calculating Kc and Kp.")
+            gas_reacts = len(self.reactants)
```
rwest, line 586: I'd rather it be called `gas_reactants` or even `number_of_gas_reactants`  (same for `gas_prods` below)

rmgpy/reactionTest.py
```diff
@@ -257,6 +257,18 @@ def test_methyl_adsorption_sticking_coefficient(self):
     def test_get_rate_coefficient_units_from_reaction_order(self):
 
         self.assertEqual(self.rxn1s.generate_reverse_rate_coefficient().A.units, 'm^2/(mol*s)')
+        
+    def test_equilibrium_constant_surface_kc(self):
+        """
+        Test the Reaction.get_equilibrium_constant() method.
+        """
+        Tlist = numpy.arange(400.0, 1600.0, 200.0, numpy.float64)
+        Kclist0 = [float(v) for v in
```
rwest, line 266: why do this float(v) for v in list of strings, instead of just doing a list of floats?

davidfarinajr: Not sure, I notice other tests were written this way, so I followed that

davidfarinajr:
Thanks for your review @rwest ! I believe I addressed all of your comments, could you take another look?

mazeau (dismissed):
This looks fine to me!

rwest (dismissed):
I just reviewed and made a few tiny changes while rebasing to the latest head.
If the tests pass I think this should be merged.

davidfarinajr:
Thanks Richard!

rwest (approved):
Regarding the discussion earlier in the comment thread, @davidfarinajr,  @cfgoldsmith, and I agreed we should be using surface site density as the appropriate reference. I have checked the implementation, and reviewed the code. All tests pass.

PR #1947, opened 2020-05-11
BugFix: correct AtomTypeError

xiaoruiDong:
### Motivation or Problem
When developing RMG-website, I found that the error message of AtomTypeError is incorrect. The problem is due to item mismatching when generating the error message.

### Description of Changes
Correct the error message.

### Testing
Passed all tests

### DIFF
--- a/rmgpy/molecule/atomtype.py
+++ b/rmgpy/molecule/atomtype.py
@@ -755,7 +755,7 @@ def get_atomtype(atom, bonds):
         single, _, r_double, o_double, s_double, triple, quadruple, benzene, lone_pairs, charge = mol_feature_list
 
         raise AtomTypeError(
-            'Unable to determine atom type for atom {0}, which has {1:d} single bonds, {2:d} double bonds to C, '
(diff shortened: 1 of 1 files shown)


rmgpy/molecule/atomtype.py
```diff
@@ -755,7 +755,7 @@ def get_atomtype(atom, bonds):
         single, _, r_double, o_double, s_double, triple, quadruple, benzene, lone_pairs, charge = mol_feature_list
 
         raise AtomTypeError(
-            'Unable to determine atom type for atom {0}, which has {1:d} single bonds, {2:d} double bonds to C, '
-            '{3:d} double bonds to O, {4:d} double bonds to S, {5:d} triple bonds, {6:d} benzene bonds, '
-            '{7:d} lone pairs, and {8:d} charge.'.format(
-                atom, single, r_double, o_double, s_double, triple, quadruple, benzene, lone_pairs, charge))
```
xiaoruiDong, line 761: There are ten items within `format()`, but only 9 slots in the string.

PR #1952, opened 2020-05-15
Debugged global uncertainty analysis

hwpang:
### Motivation or Problem
1. Currently global uncertainty analysis throws an KeyError message for liquid reactor because it tries to access `reaction_system.sens_conditions['P']` and `del mol_frac_list[0]['P']` in `rmgpy/rmg/main.py/RMG.run_uncertainty_analysis()`. 
2.  Currently, if terminationTime is not assigned in the reactor options block,  `rmgpy/rmg/main.py/RMG.run_uncertainty_analysis()` assigns the `time_criteria` to be `self.uncertainty['time']`. However, if terminationTime is also not provided in the uncertainty options block, `self.uncertainty['time']` would be None and causes an AttributeError. It is desirable for RMG to throw meaningful InputError here. 

### Description of Changes
I assigned `Plist = ([1e8],'Pa')` for liquid reactor, which uses `self.P = 1e5 #kPa` in liquid.pyx. I also added an informative InputError for terminationTime.

### Testing
I performed global uncertainty analysis by adding the uncertainty options block to the toy model provided in `examples/rmg/liquid_phase_constSPC/input.py`. The results seem normal.

### DIFF
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -1130,10 +1130,17 @@ class RMG(util.Subject):
 
                                     
                             Tlist = ([reaction_system.sens_conditions['T']], 'K')
-                            Plist = ([reaction_system.sens_conditions['P']], 'Pa')
+                            try:
+                                Plist = ([reaction_system.sens_conditions['P']], 'Pa')
+                            except KeyError:
+                                #LiquidReactor
+                                Plist = ([1e8], 'Pa')
+                                logging.warning('Using 1e8 Pa as the reaction system pressure to approximate liquid phase density.')
+
                             mol_frac_list = [reaction_system.sens_conditions.copy()]
                             del mol_frac_list[0]['T']
-                            del mol_frac_list[0]['P']
+                            if 'P' in mol_frac_list[0]:
+                                del mol_frac_list[0]['P']
 
                             # Set up Cantera reactor
                             job = Cantera(species_list=uncertainty.species_list, reaction_list=uncertainty.reaction_list,
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -61,7 +61,7 @@ from rmgpy.data.base import Entry
 from rmgpy.data.kinetics.family import TemplateReaction
 from rmgpy.data.kinetics.library import KineticsLibrary, LibraryReaction
 from rmgpy.data.rmg import RMGDatabase
-from rmgpy.exceptions import ForbiddenStructureException, DatabaseError, CoreError
+from rmgpy.exceptions import ForbiddenStructureException, DatabaseError, CoreError, InputError
 from rmgpy.kinetics.diffusionLimited import diffusion_limiter
 from rmgpy.kinetics import ThirdBody
 from rmgpy.molecule import Molecule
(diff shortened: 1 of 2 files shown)


rmgpy/rmg/main.py
```diff
@@ -1121,12 +1121,27 @@ def run_uncertainty_analysis(self):
...
+
+                                    
                             Tlist = ([reaction_system.sens_conditions['T']], 'K')
-                            Plist = ([reaction_system.sens_conditions['P']], 'Pa')
+                            try:
+                                Plist = ([reaction_system.sens_conditions['P']], 'Pa')
+                            except KeyError:
+                                #LiquidReactor
+                                Plist = ([1e8], 'Pa')
+
```
xiaoruiDong, line 1138: A minor comment: Can you add a warning mentioning P = 1e8 Pa is used?

rmgpy/rmg/main.py
```diff
@@ -1121,12 +1121,27 @@ def run_uncertainty_analysis(self):
...
                             mol_frac_list = [reaction_system.sens_conditions.copy()]
                             del mol_frac_list[0]['T']
-                            del mol_frac_list[0]['P']
+                            try:
+                                del mol_frac_list[0]['P']
+                            except KeyError:
+                                pass
```
xiaoruiDong, line 1144: A minor Question for Mark @amarkpayne: What is the common way to deal with things like this in RMG? Use `if` to check if a key in the `dict`, or use `try...except` to catch the key error when a key is not in a `dict`. Thanks.

hwpang: I think this is a very interesting discussion, and I would love to know @amarkpayne's thought too. 

I chose to use try/except for two reasons (correct me if I'm wrong)
(1) It is easier to ask for forgiveness than permission. (https://docs.python.org/3.4/glossary.html)
(2) For a performance point of view, `try dict[key]` will be faster than `if key in dict.keys()` **if we have lots of keys** in the dictionary because `dict[key]` is a hash table so it's O(1) and `if key in dict.keys()` is a list so it's O(n). (Of course we won't have lots of keys in this dict so this will barely make a difference)

If we have some convention in RMG, I will happily change it!

rmgpy/rmg/main.py
hwpang: @xiaoruiDong Thanks for your review and comment. I've added the warning in logging.info for this.

xiaoruiDong: Hao-Wei, is there a reason you choose to use `logging.info` instead of `logging.warning`?

rmgpy/rmg/main.py
mjohnson541: There's no defined way to do this in RMG. There some discussion of the pluses and minuses here:  https://stackoverflow.com/questions/4512557/if-key-in-dict-vs-try-except-which-is-more-readable-idiom

My personal opinion is unless we think it might be important for performance we should just use the conditional statement. But it doesn't matter than much.

hwpang:
> This looks good, you can remove the note of the file you modified in the title of the commit messages though, in most formats you'll look at commits it'll be easy to know what files were involved.

Thanks for the review! I updated the commit messages to be more comprehensive!

PR #1953, opened 2020-05-18
Make reordering and reindexing of entries optional when saving database

davidfarinajr:
### Motivation or Problem
When loading and saving the database, the entries are automatically re-ordered and re-indexed using the `get_entries_to_save` function.  While it is nice to have this option, sometimes we do not want to reorder and reindex when adding new entries because the entire file may be rewritten in the git commit history.

### Description of Changes
This PR makes the reordering and reindexing optional when saving.

### Testing
I used this branch when saving the thermo groups in PR https://github.com/ReactionMechanismGenerator/RMG-database/pull/414.

### DIFF
--- a/rmgpy/data/base.py
+++ b/rmgpy/data/base.py
@@ -322,7 +322,7 @@ class Database(object):
                 f.write(species_dict[label].molecule[0].to_adjacency_list(label=label, remove_h=False))
                 f.write('\n')
 
-    def save(self, path):
+    def save(self, path, reindex=True):
         """
         Save the current database to the file at location `path` on disk. 
         """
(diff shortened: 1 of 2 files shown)


rmgpy/data/base.py
```diff
@@ -322,22 +322,26 @@ def save_dictionary(self, path):
...
+            entries = self.entries.values()
+
 
         f = codecs.open(path, 'w', 'utf-8')
         f.write('#!/usr/bin/env python\n')
         f.write('# encoding: utf-8\n\n')
         f.write('name = "{0}"\n'.format(self.name))
-        f.write('shortDesc = "{0}"\n'.format(self.short_desc))
-        f.write('longDesc = """\n')
+        f.write('shortDesc = u"{0}"\n'.format(self.short_desc))
```
nateharms, line 343: Not sure if we need this string to be unicode 🤔. Is this the reason why the tests are failing?

rmgpy/data/thermo.py
```diff
@@ -123,14 +123,17 @@ def save_entry(f, entry):
             f.write('        CpInf = {0!r},\n'.format(entry.data.CpInf))
         f.write('    ),\n')
     else:
-        f.write('    thermo = {0!r},\n'.format(entry.data))
+        if entry.data:
+            f.write('    thermo = u{0!r},\n'.format(entry.data))
```
nateharms, line 127: Either that or this might be causing things to break? The `u` might be causing things to break.

PR #1954, opened 2020-05-19
Give warning for large negative Ea and make it zero

hwpang:
### Motivation or Problem
Rate rule generated large negative Ea for certain reactions. In my case, it was this reaction

`Arrhenius(A=(5.14225e+11,'s^-1'), n=0.346, Ea=(-36.417,'kcal/mol'), T0=(1,'K'), comment="""Estimated using template [R7_cyclic;doublebond_intra;radadd_intra_csHCs] for rate rule [Rn3c6_gamma;doublebond_intra_secDe_HNd;radadd_intra_csHCs]
Euclidian distance = 2.8284271247461903
family: Intra_R_Add_Endocyclic""")`

But current RMG didn't show warning.

### Description of Changes
If a rate generated by rate rule has a E0 more negative than -5 kcal/mol, a warning will be added in the species comment, and the E0 will be changed to 0 cal/mol. 0 cal/mol is chosen because the reaction is at most barrier-less (except for some special liquid phase reactions that have an energy well with hydrogen bond involved to help lower the energy of the transition state geometry, but this is not included in current RMG).  This -5 kcal/mol threshold was chosen because the proportionation family can generate moderate negative Ea around -2 kcal/mol.

The new comment looks like this:

`Arrhenius(A=(5.14225e+11,'s^-1'), n=0.346, Ea=(0,'kcal/mol'), T0=(1,'K'), comment="""Estimated using template [R7_cyclic;doublebond_intra;radadd_intra_csHCs] for rate rule [Rn3c6_gamma;doublebond_intra_secDe_HNd;radadd_intra_csHCs]
Euclidian distance = 2.8284271247461903
family: Intra_R_Add_Endocyclic
Warning: The estimated E0 = -152.36971378627635 kJ/mol is probably wrong. It is thus made 0 kJ/mol.""")`

### Testing
I tested with the generatereaction.ipynb and ran a RMG job that originally generate the reaction mentioned above.

### DIFF
--- a/ipython/generate_reactions.ipynb
+++ b/ipython/generate_reactions.ipynb
@@ -125,7 +125,7 @@
    "outputs": [],
    "source": [
     "# Execute generate reactions\n",
-    "from rmgpy.tools.generate_reactions import RMG, execute\n",
+    "from rmgpy.tools.generatereactions import RMG, execute\n",
     "kwargs = {\n",
     "    'walltime': '00:00:00:00',\n",
     "    'kineticsdatastore': True\n",
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -1220,8 +1220,10 @@ class RMG(util.Subject):
                         '\n{2}\n{3}'.format(spc.label, spc2.label, spc.to_adjacency_list(), spc2.to_adjacency_list())
                     )
 
-        # Check all core reactions (in both directions) for collision limit violation
+        # Check all core reactions (in both directions) for collision limit violation and large negative Ea reactions
         violators = []
+        large_negative_activation_energy_rxns = []
+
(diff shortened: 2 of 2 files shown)


mjohnson541:
I don't understand why this is being approached in a post processing way. Why aren't you just adding data to fix this or changing the tree? 

It's easy to say "5 kcal/mol" is unphysical and it isn't untrue, but the reality is that all of our rates are fitted and not necessarily representative of the physical quantities, occasionally a very low Ea is compensated by having a very large n especially over short temperature ranges. In general these post processing things complicate the rate estimation process and make it harder for users to figure out what's going on, why stuff is being estimated the way it is and easier to make mistakes. Especially considering we have a lot of capabilities to control rate estimation modularly within RMG, my opinion is that we should avoid these post processing things as much as we can on master.

PR #1960, opened 2020-05-22
Slightly modified automatic tree generater

hwpang:
### Motivation or Problem
The py2 version dict.keys() gives a copy of the list of dictionary's keys, so it's ok to loop through them and changes the key (in this case `del template_rxn_map[L]`. However, the py3 version dict.keys() gives an iterator, so it will have RuntimeError if a key is removed during the for loop. Instead, one can use list(dict) or list(dict.keys()) to force it to return a copy of the list of dictionary's keys to avoid the RuntimeError.

### Description of Changes
`template_rxn_map.keys()` was changed to list(template_rxn_map.keys()).
The max_batch_size was changed to 1000 for Intra_R_Add_Endocyclic tree generation.

### Testing
After this change, the ATG successfully ran for Intra_R_Add_Endocyclic.


rmgpy/data/kinetics/family.py
```diff
@@ -3123,7 +3123,7 @@ def extend_node(self, parent, template_rxn_map, obj=None, T=1000.0, ):
         return True
 
     def generate_tree(self, rxns=None, obj=None, thermo_database=None, T=1000.0, nprocs=1, min_splitable_entry_num=2,
-                      min_rxns_to_spawn=20, max_batch_size=800, outlier_fraction=0.02, stratum_num=8,
+                      min_rxns_to_spawn=20, max_batch_size=1000, outlier_fraction=0.02, stratum_num=8,
```
mjohnson541, line 3126: So I chose 800 because it seemed to be the best number working with the cascade algorithm it's just nicer in some cases if the number of training reactions is just a little over 800 to just run all of them in one batch. There's probably a nice way (looking at # training reactions / max_batch_size being >1 but maybe less than 1.3) to auto detect these cases in the script.

hwpang:
> Thanks Hao Wei! just some comments on the proposed max_batch_size change.

Thanks for the review and comment @mjohnson541 ! Your comment about the max_batch_size makes lots of sense. I modified the generateTree.py script such that if the number of training reactions is only slightly larger than the max_batch_size, max_batch_size will be set to num_training_rxns.

hwpang:
> Thanks Hao Wei! just some comments on the proposed max_batch_size change.

Hi @mjohnson541 I also removed the npcore option for now since the parallelization isn't working in py3 for some reasons. Please review when you have time. Thanks!

PR #1961, opened 2020-05-24
Draw hydrogen bonds.

rwest:
### Motivation or Problem
Hydrogen atoms were not drawn explicitly, and (therefore) hydrogen bonds were not drawn. Structures held together (or apart) with hydrogen bonds would just collapse on top of themselves. See before/after example below.

### Description of Changes
1. Don't remove an H if it has an H bond.
2. Draw H-bonds as dotted lines.
3. Fix a problem that occurred with a complicated system of fused rings.

### Testing

```
from rmgpy.molecule import Molecule
example =  Molecule().from_adjacency_list("""
example
1  O u0 p3 c-1 {2,S} {10,H}
2  N u0 p0 c+1 {1,S} {3,D} {4,S}
3  O u0 p2 c0 {2,D}
4  O u0 p2 c0 {2,S} {7,S}
5  N u0 p1 c0 {6,S} {8,S} {9,S} {7,H}
6  O u0 p2 c0 {5,S} {10,S}
7  H u0 p0 c0 {4,S} {5,H}
8  H u0 p0 c0 {5,S}
9  H u0 p0 c0 {5,S}
10 H u0 p0 c0 {6,S} {1,H}
""")
example
display(example)
for m in example.generate_h_bonded_structures():
    display(m)
```

Before:
Screen Shot 2020-05-23 at 11 13 05 PM

After:
Screen Shot 2020-05-23 at 11 19 48 PM

### DIFF
--- a/rmgpy/molecule/draw.py
+++ b/rmgpy/molecule/draw.py
@@ -176,7 +176,8 @@ class MoleculeDrawer(object):
         surface_sites = []
         for atom in self.molecule.atoms:
             if atom.is_hydrogen() and atom.label == '':
-                atoms_to_remove.append(atom)
+                if not any(bond.is_hydrogen_bond() for bond in atom.bonds.values()):
+                    atoms_to_remove.append(atom)
             elif atom.is_surface_site():
                 surface_sites.append(atom)
         if len(atoms_to_remove) < len(self.molecule.atoms) - len(surface_sites):
@@ -385,6 +386,18 @@ class MoleculeDrawer(object):
                     # need to keep self.coordinates and coordinates referring to the same object
                     self.coordinates = coordinates = np.dot(coordinates, rot)
 
+            # If two atoms lie on top of each other, push them apart a bit
+            # This is ugly, but at least the mess you end up with isn't as misleading
+            # as leaving everything piled on top of each other at the origin
+            import itertools
+            for atom1, atom2 in itertools.combinations(backbone, 2):
+                i1, i2 = atoms.index(atom1), atoms.index(atom2)
+                if np.linalg.norm(coordinates[i1, :] - coordinates[i2, :]) < 0.5:
+                    coordinates[i1, 0] -= 0.3
+                    coordinates[i2, 0] += 0.3
+                    coordinates[i1, 1] -= 0.2
+                    coordinates[i2, 1] += 0.2
+
             # Center backbone at origin
             xmin = np.min(coordinates[:, 0])
             xmax = np.max(coordinates[:, 0])
@@ -557,6 +570,8 @@ class MoleculeDrawer(object):
                     if count == 1 or count == 2:
                         if cycle is None or len(cycle0) > len(cycle): cycle = cycle0
             cycle0 = cycle1
+            if cycle is None:
+                break
             atoms.remove(cycle)
 
             # Shuffle atoms in cycle such that the common atoms come first
@@ -1037,16 +1052,22 @@ class MoleculeDrawer(object):
         self.right += padding
         self.bottom += padding
 
-    def _draw_line(self, cr, x1, y1, x2, y2, dashed=False):
+    def _draw_line(self, cr, x1, y1, x2, y2, dashed=False, dash_sizes=None):
         """
         Draw a line on the given Cairo context `cr` from (`x1`, `y1`) to
         (`x2`,`y2`), and update the bounding rectangle if necessary.
+
+        For a dashed line set ``dashed=True``.
+        Then the optional `dash_sizes` can be a list of on/off segment lengths,
+        which defaults to [3.5, 3.5] if not specified.
         """
 
         cr.set_source_rgba(0.0, 0.0, 0.0, 1.0)
         cr.set_line_width(1.0)
         if dashed:
-            cr.set_dash([3.5])
+            if dash_sizes is None:
+                dash_sizes = [3.5, 3.5]
+            cr.set_dash(dash_sizes)
         cr.set_line_cap(cairo.LINE_CAP_ROUND)
         cr.move_to(x1, y1)
         cr.line_to(x2, y2)
@@ -1137,6 +1158,9 @@ class MoleculeDrawer(object):
                 dv *= 1.6
                 self._draw_line(cr, x1 - du, y1 - dv, x2 - du, y2 - dv)
                 self._draw_line(cr, x1 + du, y1 + dv, x2 + du, y2 + dv, dashed=True)
+            elif bond.is_hydrogen_bond():
+                # Draw a dashed line
+                self._draw_line(cr, x1, y1, x2, y2, dashed=True, dash_sizes=[0.5, 3.5])
             else:
                 self._draw_line(cr, x1, y1, x2, y2)
         else:
--- a/rmgpy/molecule/draw.py
+++ b/rmgpy/molecule/draw.py
@@ -385,6 +385,18 @@ class MoleculeDrawer(object):
                                     [-math.sin(angle), math.cos(angle)]], np.float64)
                     # need to keep self.coordinates and coordinates referring to the same object
                     self.coordinates = coordinates = np.dot(coordinates, rot)
+            
+            # If two atoms lie on top of each other, push them apart a bit
+            # This is ugly, but at least the mess you end up with isn't as misleading
+            # as leaving everything piled on top of each other at the origin
+            import itertools
+            for atom1, atom2 in itertools.combinations(backbone, 2):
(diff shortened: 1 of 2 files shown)


rmgpy/molecule/draw.py
```diff
@@ -1137,6 +1147,9 @@ def _render_bond(self, atom1, atom2, bond, cr):
                 dv *= 1.6
                 self._draw_line(cr, x1 - du, y1 - dv, x2 - du, y2 - dv)
                 self._draw_line(cr, x1 + du, y1 + dv, x2 + du, y2 + dv, dashed=True)
+            elif bond.is_hydrogen_bond():
+                # Draw a dashed line
+                self._draw_line(cr, x1, y1, x2, y2, dashed=[0.5,3.5])
```
rwest, line 1152:
```suggestion
                self._draw_line(cr, x1, y1, x2, y2, dashed=[0.5, 3.5])
```

rwest: to fix the nagging codacity "code quality" thing

rmgpy/molecule/draw.py
```diff
@@ -557,6 +560,8 @@ def _generate_ring_system_coordinates(self, atoms):
                     if count == 1 or count == 2:
                         if cycle is None or len(cycle0) > len(cycle): cycle = cycle0
             cycle0 = cycle1
+            if cycle is None:
+                break
```
alongd, line 564: could you explain this addition?

rmgpy/molecule/draw.py
```diff
@@ -1041,12 +1046,17 @@ def _draw_line(self, cr, x1, y1, x2, y2, dashed=False):
         """
         Draw a line on the given Cairo context `cr` from (`x1`, `y1`) to
         (`x2`,`y2`), and update the bounding rectangle if necessary.
+
+        For a dashed line, `dashed` be a list of on/off segment lengths,
+        or `True` in which case [3.5,3.5] is used.
```
alongd, line 1051: Minor: please use double "`" marks for `True`, a single mark on each side represents a link in rst (double, a code)

rmgpy/molecule/draw.py
```diff
@@ -1041,12 +1046,17 @@ def _draw_line(self, cr, x1, y1, x2, y2, dashed=False):
...
         """
 
         cr.set_source_rgba(0.0, 0.0, 0.0, 1.0)
         cr.set_line_width(1.0)
         if dashed:
-            cr.set_dash([3.5])
+            if dashed is True:
+                dashed = [3.5]
```
alongd, line 1058: The docsting says `[3.5, 3.5]`, here it's `[3.5]`, which is correct?
Could you also provide a general explanation for why `dash` is sometimes a `bool` and sometimes a `list`? I understand it wasn't introduced by this PR.

rmgpy/molecule/drawTest.py
```diff
@@ -126,6 +126,28 @@ def test_draw_non_standard_bonds(self):
...
+6  O u0 p2 c0 {5,S} {10,S}
+7  H u0 p0 c0 {4,S} {5,H}
+8  H u0 p0 c0 {5,S}
+9  H u0 p0 c0 {5,S}
+10 H u0 p0 c0 {6,S} {1,H}
+11 X u0 p0 c0
+        """
+        )
+        try:
+            from cairocffi import PDFSurface
```
alongd, line 148: Are `cairocffi` and `cairo` OS-dependent? If not, can we assume one of them should be present?

rmgpy/molecule/drawTest.py
```diff
@@ -98,7 +98,7 @@ def test_draw_polycycle(self):
         if os.path.exists(path):
             os.unlink(path)
         polycycle = Molecule(smiles="C123CC4CC1COCC2CCC34")
-        surface, cr, (xoff, yoff, width, height) = self.drawer.draw(self.molecule, file_format='pdf', target=path)
+        surface, cr, (xoff, yoff, width, height) = self.drawer.draw(polycycle, file_format='pdf', target=path)
```
alongd: 👍

rmgpy/molecule/draw.py
```diff
@@ -176,7 +176,10 @@ def draw(self, molecule, file_format, target=None):
         surface_sites = []
         for atom in self.molecule.atoms:
             if atom.is_hydrogen() and atom.label == '':
-                atoms_to_remove.append(atom)
+                for bond in atom.bonds.values():
```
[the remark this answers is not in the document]
rwest: I was going to say I prefer my loop because it can break on finding the first H bond instead of having to evaluate all the bonds and then check the list, which your list comprehension does. Then I realized you can pass a _generator_ to the `any()` method instead of a list, and then it can abort the loop on the first hit! Just remove the `[]` from your code!

Any time you see `any([...])` or `all([...])` you should be very suspicious that the square brackets are necessary. You can almost always remove them and get code that runs faster and uses less memory.
https://eklitzke.org/generator-comprehensions-and-using-any-and-all-in-python

rmgpy/molecule/draw.py
rwest: Without it trying to draw
```
1  O u0 p2 c0 {5,S} {8,S}
2  O u0 p2 c0 {6,S} {7,S}
3  O u0 p3 c-1 {6,S} {8,H}
4  O u0 p2 c0 {6,D} {9,H}
5  N u0 p1 c0 {1,S} {7,H} {9,S} {10,S}
6  N u0 p0 c+1 {2,S} {3,S} {4,D}
7  H u0 p0 c0 {2,S} {5,H}
8  H u0 p0 c0 {1,S} {3,H}
9  H u0 p0 c0 {4,H} {5,S}
10 H u0 p0 c0 {5,S}
```
leads to
```
  File "/Users/rwest/Code/RMG-Py/rmgpy/molecule/draw.py", line 209, in draw
    self._generate_coordinates()
  File "/Users/rwest/Code/RMG-Py/rmgpy/molecule/draw.py", line 366, in _generate_coordinates
    self._generate_ring_system_coordinates(backbone)
  File "/Users/rwest/Code/RMG-Py/rmgpy/molecule/draw.py", line 564, in _generate_ring_system_coordinates
    atoms.remove(cycle)
ValueError: list.remove(x): x not in list
```

If you replace the `break` with a `continue` then you enter and endless loop (due to the `while`)

With the `break` we get the image
image

...which is not great - the H is being drawn exactly on top of an O which hides it - but at least it doesn't crash.

rmgpy/molecule/draw.py
rwest: Other places in the code were already calling this method with `dashed=True` and get a default dash size of 3.5. I wanted a way to specify the dash size.  I guess it could be done with an extra optional argument `dash_size` instead of overloading this one if you think that's better.

(`[3.5]` is equivalent to `[3.5, 3.5]` when specifying dash sizes in cairo, but to avoid people needing to know that I've changed it to `[3.5, 3.5]` in a fixup commit)

rmgpy/molecule/drawTest.py
rwest: I don't know - I just copied boilerplate from all the other drawing tests. Changing them all is out of scope of this PR.

rwest:
I addressed your suggestions with fixup commits (so you can review them independently) but they should be squashed on rebasing to merge.

rwest:
I added another commit which pushes atoms apart if they would have otherwise been on top of each other due to the failings of the algorithm for fused multicycles.

Now the image in the PR description up top looks like this:

Screen Shot 2020-06-11 at 1 29 10 AM
at first glance it's not as neat and pretty as before, but it's actually less misleading. Now that the atoms aren't all piled on top of each other at the origin, you can at least count and see them. These pictures all have 4 O atoms showing!

rmgpy/molecule/draw.py
alongd: If this arg is currently being used only as a `bool`, let's make a new one, unless you think it's very redundant

alongd:
@rwest, please do what you think is right re ading `dash_size` (I added a minor comment), then feel free to squash

rmgpy/molecule/draw.py
rwest: This better? https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1961/commits/c695be3f596d2f81960b379bfb1baeb29d499984

alongd: yes, thanks!

rwest:
OK, all fixups rebased, tests passing again. Should I merge, or wait for someone else to?

PR #1962, opened 2020-05-24
Avoid SMILES translation errors for reference species

amarkpayne:
### Motivation or Problem
If you load and then immediately save the reference database, some items in the database change, including SMILES strings, InChI strings/keys, and floating point errors for ScalarQuantity objects, even though the user made no changes to the files.

This PR fixes the issues related to the identifiers by resetting these attributes in the __init__ method if they were supplied (as they would be when loading the database). This avoids the translation errors that occur when creating the species in the call to the super class ArkaneSpecies.

This PR does NOT fix the issues with the floating point errors. It is noteworthy that these errors only occur inside of ScalarQuantity objects. The root cause of this is that quantities with non-SI units are first converted to SI units (ScalarQuantity.value is just a property that reads/updates value_si), and this arithmetic leads to floating point errors when converting back to the non-SI unit when saving the database. I tired implementing a fix for this by refactoring our quantity code so that we don't convert to SI units until needed, but this would involve a major refactoring of our quantities module (the current limiting factor is that we need a way to update the value property if the user changes the units, but this is difficult because the units are there own class). 

### Testing
I tested loading and saving the database, and only floating point error changes occurred.

### DIFF
--- a/arkane/encorr/reference.py
+++ b/arkane/encorr/reference.py
@@ -60,7 +60,7 @@ class ReferenceSpecies(ArkaneSpecies):
     selections for use in isodesmic reaction calculations
     """
 
-    def __init__(self, species=None, smiles=None, adjacency_list=None, inchi=None, reference_data=None,
+    def __init__(self, species=None, smiles=None, adjacency_list=None, inchi=None, inchi_key=None, reference_data=None,
                  calculated_data=None, preferred_reference=None, index=None, label=None, cas_number=None,
                  symmetry_number=None, default_xyz_chemistry=None, **kwargs):
         """
@@ -71,6 +71,7 @@ class ReferenceSpecies(ArkaneSpecies):
             smiles (str): SMILES string representing the reference species
             adjacency_list (str): An RMG adjacency list representation of the reference species
             inchi (str): InChI string representing the reference species
+            inchi_key (str): InChI key hash of the InChI string
             reference_data (dict): Formatted as {'source_string': ReferenceDataEntry, ...}
             calculated_data (dict): Formatted as {'model_chemistry': CalculatedDataEntry, ...}
             preferred_reference (str): The source string key for the reference data to use for isodesmic reactions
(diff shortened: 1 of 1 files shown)


arkane/encorr/reference.py
```diff
@@ -111,6 +112,12 @@ def __init__(self, species=None, smiles=None, adjacency_list=None, inchi=None, r
         if symmetry_number:
             self.symmetry_number = symmetry_number
 
+        # Reset to the provided identifiers to avoid translation errors
+        self.smiles = smiles or self.smiles
+        self.adjacency_list = adjacency_list or self.adjacency_list
+        self.inchi = inchi or self.inchi
+        self.inchi_key = inchi_key or self.inchi_key
```
cgrambow, line 119: Assignment using `or` will assign the second value if the first one is falsy. Is that your desired behavior for all possible falsy values of the first variable in these assignments (e.g., if `smiles` is `''`)? If instead you wrote this as a shortcut for checking whether the first variable is `None` or not, then assignment using this `or` construct is discouraged, as comparisons against `None` should always be done using `is`. For example, `self.smiles = self.smiles if smiles is None else smiles`.

Alternatively, since `self.smiles` is already assigned prior to this, you can just do
```python
if smiles is not None:
    self.smiles = smiles
```
Potentially saves one assignment but adds an extra line. (Last alternative: since the expression is short, PEP 8 might allow `if smiles is not None: self.smiles = smiles` on a single line.)

amarkpayne: For documentation purposes, we tried the approach of `if smiles is not None: self.smiles = smiles`. However, we ran into the following issue during testing:

image

We think that somewhere `''` is getting set as a SMILES string, which is being treated like a bool by RDKit

We decided to go back to `self.smiles = smiles or self.smiles`, because all of these identifiers should only be strings, and `''` is not a valid identifier whether it be a SMILES string, adjlist, or InChI. Therefore, we shouldn't run into issues where a valid identifier gets evaluated as False.

At some point we should track down where the smiles string default is being set as `''`

Issue #1963, opened 2020-05-25
Possibly regenerating resonances structures much more than needed


rwest:
### Bug Description

I had put some logging in during resonance structure generation, and found it occurring far more than I thought reasonable.

It looks like every time any species is reacted with any other species in any reaction family, all the resonance structures are regenerated every time.

I think this commit
https://github.com/ReactionMechanismGenerator/RMG-Py/commit/24d68fc46ce65f8ec3e51015e030dc2b47f35509#diff-547a65786a17bb742b1ec0865478baa0R259
which, according to the commit message, was just intended to make it possible to NOT generate resonance isomers in some occasions, in fact made them occur far more often.
Before the commit, they'd only be regenerated during  ensure_independent_atom_ids if identical atom ids were found between species, and the IDs needed to be regenerated.
After the commit, everything in input_species has generateResonanceIsomers called on it every time (unless the method is called with `resonance=False`, which it is not typically).

So I think we're doing much more resonance generation than needed. Most of these species have already had their resonance structures made.

Would anyone familiar with resonance stuff and/or reaction generation and/or interested in speed issues like to take a look?

### Installation Information
RMG version information:   - RMG-Py: current master https://github.com/ReactionMechanismGenerator/RMG-Py/commit/5b285428fdc085a54703bc2bd5bcdc6acc9cde10

PR #1964, opened 2020-05-25
Do resonance structure generation less often.

rwest:
### Motivation or Problem
This addresses https://github.com/ReactionMechanismGenerator/RMG-Py/issues/1963 so read that issue for context.

### Description of Changes
The change in this PR stops a lot of calls to `generate_resonance_structures`.
I'm not entirely sure some of it wasn't being counted on by something, but I am hoping that by making this pull request, the automated tests will tell us if it changes the outcome. 

Now the method `ensure_independent_atom_ids` only generates
resonance structures if doing so is needed to ensure independent
atom ids, and not as a side effect on all species it is called with.

I believe that if some code was relying on resonance structures to be generated
that code should instead explicitly call `generate_resonance_structures`, and that can be added to this PR.
The method name `ensure_independent_atom_ids` doesn't suggest it generates all resonance structures, and methods that are called one thing but relied on to do another are a code smell.

Before this change, resonances structures were generated a lot,
because `ensure_independent_atom_ids` is called a lot. Hopefully this speeds things up. Do the automated tests also check runtimes?

If we *do* need to generate resonance structures more than this to ensure we get all reactions, maybe we can call it from somewhere else so it gets done only once for each species.

### Testing
I already ran the unit test suite, which passes. Hoping the automated regression tests will be informative.

### DIFF
--- a/rmgpy/data/kinetics/common.py
+++ b/rmgpy/data/kinetics/common.py
@@ -208,10 +208,6 @@ def ensure_independent_atom_ids(input_species, resonance=True):
                 species.generate_resonance_structures(keep_isomorphic=True)
             if len(unreactive_mol_list):
                 species.molecule.extend(unreactive_mol_list)
-    elif resonance:
-        # IDs are already independent, generate resonance structures if needed
-        for species in input_species:
-            species.generate_resonance_structures(keep_isomorphic=True)
 
 
 def find_degenerate_reactions(rxn_list, same_reactants=None, template=None, kinetics_database=None,


mliu49:
The fact that `ensure_independent_atom_ids` would previously generate resonance structures only if the atom ids were not unique meant inconsistent behavior, since you wouldn't know if resonance structures would be generated. 24d68fc46ce65f8ec3e51015e030dc2b47f35509 made the behavior more consistent and also allowed the resonance structure generation part to be skipped.

The resonance structure generation step is independent from ensuring unique ids, and is not a necessary step even if the ids are not independent. I agree that it's not great for such a specifically named function to be doing other stuff, and this could theoretically be completely removed from the function.

I think a more appropriate quick fix would be to specify `resonance=False` in calls to `ensure_independent_atom_ids` where resonance structure generation is not necessary. I don't think removing just this section from the function is a good solution since it brings back the inconsistent behavior.

PR #1965, opened 2020-05-25
Add family.auto_generated check in Arkane

hwpang:
### Motivation or Problem
arkane.inpuy doesn't have the same fix that rmgpy.main does, i.e. skipping add_rules_from_training if the tree is auto-generated.

### Description of Changes
Added a check that if family.auto_generated is True, Arkane will skip add_rules_from_training.

### Testing
Tested with test_arkane_examples in inputTest.py locally and it passed.


mjohnson541:
Can you rebase this?

hwpang:
> Can you rebase this?

Thanks for your review. Rebased!

PR #1968, opened 2020-05-28
BAC cross-validation

cgrambow:
### Motivation or Problem
Evaluate BAC fits with leave-one-out cross-validation.

### Description of Changes
Adds a new `CrossVal` class to perform cross-validation for BACs. Currently, only implements leave-one-out cross-validation, but can be simply extended to other types of cross-validation by adding other classes from `sklearn.model_selection`. Also adds new functionality for specifying/excluding indices of reference species when fitting BACs.

### Testing
Added unit tests for `CrossVal` class and ran some cross-validation jobs.

### DIFF
--- a/arkane/encorr/bac.py
+++ b/arkane/encorr/bac.py
@@ -140,6 +140,8 @@ class BACJob:
         with open(output_file2, 'w') as f:
             writer = csv.writer(f)
             writer.writerow([
+                'Index',
+                'Label',
                 'Smiles',
                 'InChI',
                 'Formula',
@@ -152,6 +154,8 @@ class BACJob:
             ])
             for d in self.bac.dataset:
                 writer.writerow([
+                    d.spc.index,
+                    d.spc.label,
                     d.spc.smiles,
                     d.spc.inchi,
                     d.spc.formula,
--- a/arkane/encorr/bac.py
+++ b/arkane/encorr/bac.py
@@ -584,6 +584,8 @@ class BAC:
     def fit(self,
             weighted: bool = False,
             db_names: Union[str, List[str]] = 'main',
+            idxs: Union[Sequence[int], Set[int], int] = None,
+            exclude_idxs: Union[Sequence[int], Set[int], int] = None,
             exclude_elements: Union[Sequence[str], Set[str], str] = None,
             charge: Union[Sequence[Union[str, int]], Set[Union[str, int]], str, int] = 'all',
             multiplicity: Union[Sequence[int], Set[int], int, str] = 'all',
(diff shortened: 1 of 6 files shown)


kspieks:
Reviving this PR. Thanks to Colin and Mark for previously working on it :) Things look good to me and some test jobs completed fine. 

As a minor change, I propose using the `KFold()` cross-validator as has the same capability of `LeaveOneOut()`; additionally, it lets us leave 2 out (by setting the number of folds to ~200) or leave 3 out (by using ~135 folds), which give similar error estimates as using leave-one-out but run a bit faster. The corresponding unit test for KFold was added.

The previous implementation used the BACs trained on the last fold but then evaluated on the entire reference dataset to give the cross-validation results. I updated it to give mean +- 1 std of the RMSE and MAE of testing set from each fold.

As another minor change, I added "Training" and "Testing" to print statements so it was more clear which statistic was being reported.

@amarkpayne Can you review this when you get a chance?

arkane/encorr/bac.py
```diff
@@ -107,14 +110,14 @@ def execute(self, output_directory: str = None, plot: bool = False, jobnum: int
         logging.info(f'Running BAC job {jobnum}')
         self.bac.fit(**self.kwargs)
 
-        if output_directory is not None:
+        if output_directory is not None and self.crossval_n_folds == 1:
             os.makedirs(output_directory, exist_ok=True)
```
amarkpayne, line 114: @kspieks can you explain this change really quick?

kspieks: Colin had added a [change](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1968/commits/9e3a19484c2458f7196d06a30336ff342aa4630d#diff-a4352d67c1b492a6db2aa46f53d796701efb51a956c2a119b3928f6b8088d730R117) to only write to the database if not doing cross-validation, since we only want to add values that have been trained on the entire dataset.

Extending this line of thinking, I thought it was potentially misleading to print `output.py` with BAC values as well as the summary csv for BACs that were only trained on the last fold from cross-validation i.e. they were not trained on the entire reference set. Do follow up if this doesn't make sense.

kspieks:
Upon closer inspection, I was incorrectly understanding how the original code behaves. The original implementation gives one aggregate RMSE value for all folds since it uses the following formula: 



I propose adding another RMSE metric to give the mean +- 1 std which is another common way to report the results from k-fold cross-validation:

I added this approach [here](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1968/commits/302a2caf66df0acc6137ec87a812ee6472b33077#diff-a4352d67c1b492a6db2aa46f53d796701efb51a956c2a119b3928f6b8088d730R1045-R1052). However, @amarkpayne I believe we discussed reporting both RMSE metrics. Are you still working on this branch? Can we finalize changes next week so this branch can be merged?

amarkpayne (dismissed):
The additional changes look good to me, go ahead and squash

amarkpayne (approved):

PR #1969, opened 2020-05-28
BAC parameter confidence intervals

cgrambow:
### Motivation or Problem
BAC jobs only save the correlation matrix, but not confidence interval estimates.

### Description of Changes
Add these estimates as a comment in `output.py` of an Arkane job.

### Testing
Tested by running an Arkane job.

### Reviewer Tips
Run a BAC job to test that the CIs are included in the output and check the formulas.


amarkpayne (approved):

amarkpayne:
Thanks @cgrambow from the PR from a long while ago! Sorry this is just now getting in.

PR #1975, opened 2020-06-08
Updated the Pt binding energy values based on recent DFT calculations

kblondal:
The updated values are consistent with how the atomic binding energies on the other 9 metals are calculated (to be included in the database). [PR#387](https://github.com/ReactionMechanismGenerator/RMG-database/pull/387)
The DFT calculations were conducted by Katrin Blondal and Bjarne Kreitz at Brown University in the fall 2019.


mjohnson541:
Can you rebase?

kblondal:
I did not mean to dismiss your reviews, it seems to have happened automatically when I used git push -f. This is the same as @rwest and @mjohnson541 approved, but rebased.

PR #1978, opened 2020-06-09
Used fsolve for BM if only 1 training reaction

hwpang:
### Motivation or Problem
1. If there's only one training reaction in the node, we should be able to solve E0 for ArrheniusBM with A and n from the training reaction.
2. The `"BM rule fitted to {0} training reactions at node {1}".format(len(rxnlists[i]), entry.label)` always show BM rule fitted to 2 training reactions right now

### Description of Changes
1. If `len(rxns) == 1` in `fit_to_reactions`, E0 will be solved numerically with Ea from the one training reaction. Otherwise it will perform the original fitting.
2. Changed to `"BM rule fitted to {0} training reactions at node {1}".format(len(rxnlists[i][0]), entry.label)` because `rxnlists[i]` is a tuple containing a list of training reactions and the entry label `([rxns], entry.label)`.

### Testing
One-training-reaction node is very sensitive for the initial guess of E0 fed to curve_fit. For a one-training-reaction node that I'm working on, using fit_curve with the current initial guess of E0 gives this plot: 
image

Using a better initial guess for E0 gives this:
image

Using this PR, it gives this plot with both initial guess:
image

### DIFF
--- a/rmgpy/kinetics/arrhenius.pyx
+++ b/rmgpy/kinetics/arrhenius.pyx
@@ -30,7 +30,7 @@
 import numpy as np
 cimport numpy as np
 from libc.math cimport exp, sqrt, log10
-from scipy.optimize import curve_fit
+from scipy.optimize import curve_fit, fsolve
 
 cimport rmgpy.constants as constants
 import rmgpy.quantity as quantity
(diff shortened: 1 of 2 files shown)


rmgpy/kinetics/arrhenius.pyx
```diff
@@ -597,57 +597,80 @@ cdef class ArrheniusBM(KineticsModel):
...
-        A = np.exp(lnA)
+        if len(rxns) == 1:
+            T = 1000.0
+            rxn = rxns[0]
+            dHrxn = rxn.get_enthalpy_of_reaction(T)
+            A = rxn.kinetics.A.value_si
+            n = rxn.kinetics.n.value_si
+            Ea = rxn.kinetics.Ea.value_si
+            
+            def kfcn(E0):
```
mjohnson541, line 608: What happens when abs(dHrxn) > 4*E0? Can we just guess E0 = w/10.0 if that's the case?

hwpang: > What happens when abs(dHrxn) > 4*E0? Can we just guess E0 = w/10.0 if that's the case?

Thanks for the comment @mjohnson541 ! I changed the code accordingly.

hwpang:
> If you squash the second and the last and then rebase I think this should be good.

I squashed the commits!

mjohnson541:
Can you rebase this?

hwpang:
rmg-tests fails because of issue ReactionMechanismGenerator/RMG-Py#2011

PR #1979, opened 2020-06-09
Minor: Update group attribute name to py3

hwpang:
### Motivation or Problem
1. hasWildCards should be has_wildcards
2. standardize_atomtype() function throws attribute error because Group() doesn't have has_wildcards attribute. It really should be a checking done on the atoms.

### Description of Changes
1. Changed from hasWildCards to has_wildcards
2. Add a for loop to loop through self.atoms and changed self.has_wildcards to atom.has_wildcards.

### DIFF
--- a/rmgpy/molecule/group.py
+++ b/rmgpy/molecule/group.py
@@ -2074,7 +2074,7 @@ class Group(Graph):
         for index, atom in enumerate(self.atoms):
             claimed_atom_type = atom.atomtype[0]
             # Do not perform is this atom has wildCards
-            if atom.hasWildCards:
+            if atom.has_wildcards():
                 continue
             elif claimed_atom_type is ATOMTYPES['CO'] or claimed_atom_type is ATOMTYPES['CS']:
                 for bond12 in atom.bonds.values():
--- a/rmgpy/molecule/group.py
+++ b/rmgpy/molecule/group.py
@@ -1952,6 +1952,19 @@ class Group(Graph):
                         return False
         return True
 
+    def has_wildcards(self):
+        """
+        This function is a Group level wildcards checker.
+
+        Returns a 'True' if any of the atoms in this group has wildcards.
+        """
+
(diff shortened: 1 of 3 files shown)


rmgpy/molecule/group.py
```diff
@@ -1901,7 +1901,8 @@ def standardize_atomtype(self):
         modified = False
 
         # If this atom or any of its ligands has wild cards, then don't try to standardize
-        if self.hasWildCards: return modified
+        for atom in self.atoms:
```
mjohnson541, line 1904: This might be better done in a separate cythonized function since it is an independent operation and the for loop would benefit from cythonization.

hwpang: @mjohnson541 Thanks for the comment! I added a Group level wildcards checker to loop through the atoms and check wildcards. This method is also added to group.pxd so it should be cythonized!!

hwpang:
> This is fine but the changes order doesn't make much sense. Can you split the creation of the has_wildcards function into one commit and the naming/calling changes in to a second commit?

Thanks for the review! Commits updated!

mjohnson541:
Can you rebase?

hwpang:
> Can you rebase?

Rebased!

mjohnson541:
Sorry, can you rebase again?

mjohnson541:
Sorry, can you rebase again?

hwpang:
@mjohnson541 Rebased!

hwpang:
Rebased again!

kspieks (approved):
Assuming the tests pass, this LGTM. If convenient, it'd be great to add a small unit test so things like this don't go unnoticed in the future

hwpang:
@kspieks Thanks for the review! The unit test is just added!

kspieks:
@hwpang can you please rebase? I'll merge immediately since all tests have already passed

kspieks (approved):

PR #1980, opened 2020-06-11
Minor change rotor

xiaoruiDong:
### Motivation or Problem
Minor changes to Arkane's rotor related functions.

For commit 1, ARC is currently using load_scan_enerigies(). However, load_scan_enerigies() is always assuming the scan is 0 - 360, which may not be true.
For commit 2, When fitting scanned PES to a Fourier series, we are now fitting:
`k_0 * cos(0 * phi) + ... + k_n * cos(n * phi) + k_n+1  * sin(0 * phi) + .... + k_2n * sin(n * phi) = E`
and 
`k_n+1 + k_n+2 ... + k_2n = 0`
However, I don't think we should include `k_n+1`, we are fitting a coefficient before a variable that always equal to zero. This will not helpful to the fitting, but also weaken the constraint we put `k_n+1 + k_n+2 ... + k_2n = 0`, since `k_n+1` literally can be any number. 

### Description of Changes
For commit 1, the script uses the actual scan step size to build up the `angle` array.
For commit 2, I remove the sin(0*phi)s from A matrix and vectorize some of the calculations.

### Testing
The unit tests are passed. I also did some tests on rotor related calculations. `load_scan_energies` works as designed, and the fitted curves has very tiny differences compared to previous ones, but dV/dangle(0) is closed to zero (makes the result better in some cases).

### DIFF
--- a/arkane/ess/gaussian.py
+++ b/arkane/ess/gaussian.py
@@ -398,9 +398,9 @@ class GaussianLog(ESSAdapter):
             vlist = vlist[:-1]
 
         # Determine the set of dihedral angles corresponding to the loaded energies
-        # This assumes that you start at 0.0, finish at 360.0, and take
-        # constant step sizes in between
-        angle = np.arange(0.0, 2 * math.pi + 0.00001, 2 * math.pi / (len(vlist) - 1), np.float64)
+        # This assumes that all of the angles are evenly spaced with a constant step size
+        scan_res = math.pi / 180 * self._load_scan_angle()
+        angle = np.arange(0.0, scan_res * (len(vlist) - 1) + 0.00001, scan_res, np.float64)
 
         return vlist, angle
 
--- a/rmgpy/statmech/torsion.pyx
+++ b/rmgpy/statmech/torsion.pyx
@@ -526,32 +526,30 @@ cdef class HinderedRotor(Torsion):
         numterms = 6
         cdef bint negative_barrier
         negative_barrier = True
-        # numterms is actually half the number of terms. It is called numterms 
+        # numterms is actually half the number of terms. It is called numterms
         # because it is the number of terms of either the cosine or sine fit
 
         maxterms = np.floor(len(angle) / 3.0)
         while negative_barrier and numterms < maxterms:
             # Fit Fourier series potential
             N = V.shape[0]
-            A = np.zeros((N + 1, 2 * numterms), np.float64)
-            b = np.zeros(N + 1, np.float64)
-            for i in range(N):
-                phi = angle[i]
-                for m in range(numterms):
-                    A[i, m] = cos(m * phi)
-                    A[i, numterms + m] = sin(m * phi)
-                    b[i] = V[i]
+            # A: [1, cos(phi), ..., cos(M * phi), sin(phi), ..., sin(M * phi)]
+            A = np.zeros((N + 1, 2 * numterms - 1), np.float64)
+            A[:-1, 0] = 1
+            for m in range(1, numterms):
+                A[:-1, m] = np.cos(m * angle)
+                A[:-1, numterms + m - 1] = np.sin(m * angle)
             # This row forces dV/dangle = 0 at angle = 0
-            for m in range(numterms):
-                A[N, m + numterms] = 1
+            A[N, numterms:] = 1
+            b = np.concatenate((V, np.array([0.])))
+
             x, residues, rank, s = np.linalg.lstsq(A, b, rcond=RCOND)
             fit = np.dot(A, x)
             x *= 0.001
-            # This checks if there are any negative values in the forier fit.
-            # This part of the algorithm is replicated from the fucntion HinderedRotor(Torsion)
+            # This checks if there are any negative values in the Fourier fit.
             negative_barrier = False
             V0 = 0.0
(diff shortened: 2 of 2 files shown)


rmgpy/statmech/torsion.pyx
```diff
@@ -526,32 +526,30 @@ cdef class HinderedRotor(Torsion):
         numterms = 6
         cdef bint negative_barrier
         negative_barrier = True
-        # numterms is actually half the number of terms. It is called numterms 
+        # numterms is actually half the number of terms. It is called numterms
         # because it is the number of terms of either the cosine or sine fit
 
         maxterms = np.floor(len(angle) / 3.0)
         while negative_barrier and numterms < maxterms:
```
davidfarinajr, line 533: Can you change the `<` to `<=`?  Since we don't increment the numterms until after trying the fit and seeing if there is a negative_barrier, we never try maxterms.  For example, if your maxterms is 8 and you get negative barrier for 6 terms, since 8 is not less than 8, we won't try the fit with 8 terms.

xiaoruiDong: Yes, you are right.

davidfarinajr:
Screen Shot 2020-06-17 at 4 12 33 PM

looks good!

arkane/ess/gaussian.py
```diff
@@ -398,9 +398,9 @@ def load_scan_energies(self):
             vlist = vlist[:-1]
 
         # Determine the set of dihedral angles corresponding to the loaded energies
-        # This assumes that you start at 0.0, finish at 360.0, and take
-        # constant step sizes in between
-        angle = np.arange(0.0, 2 * math.pi + 0.00001, 2 * math.pi / (len(vlist) - 1), np.float64)
+        # This assumes that all of the angles are evenly spaced with a constant step size
+        scan_res = math.pi / 180 * self._load_scan_angle()
+        angle = np.arange(0.0, scan_res * (len(vlist) - 1) + 0.00001, scan_res, np.float64)
```
alongd, line 403: The angles still start at 0. Did you not intend to change that?

rmgpy/statmech/torsion.pyx
```diff
@@ -530,7 +530,7 @@ cdef class HinderedRotor(Torsion):
         # because it is the number of terms of either the cosine or sine fit
 
         maxterms = np.floor(len(angle) / 3.0)
-        while negative_barrier and numterms < maxterms:
+        while negative_barrier and numterms <= maxterms:
```
alongd, line 533: would you like to squash this commit into the previous one?

arkane/ess/gaussian.py
```diff
@@ -398,9 +398,9 @@ def load_scan_energies(self):
             vlist = vlist[:-1]
 
         # Determine the set of dihedral angles corresponding to the loaded energies
```
alongd, line 400: Do any of these require similar modifications?
https://github.com/ReactionMechanismGenerator/RMG-Py/blob/627777034e82462f0570b93b458d65c547adc6e9/arkane/ess/qchem.py#L307
https://github.com/ReactionMechanismGenerator/RMG-Py/blob/fe6e69a94afcf40220e4980b0b772029f66fb571/arkane/ess/terachem.py#L275

arkane/ess/gaussian.py
```diff
@@ -378,6 +378,11 @@ def load_scan_energies(self):
                 # to the optimized geometry
                 if 'Optimization completed' in line:
                     vlist.append(energy)
+                # In some cases, the optimization cannot converge within the given steps.
+                # Then, the geometry is not optimized. we need to exclude these values.
+                if 'Optimization stopped' in line:
+                    non_optimized.append(len(vlist))
+                    vlist.append(energy)
```
amarkpayne, line 385: Does `non_optimized` ever get initialized anywhere? I think this is the reason why the Travis tests are failing, unless I am missing something obvious.

xiaoruiDong: Yeah, you are definitely correct! Thanks, Mark.

arkane/ess/gaussian.py
xiaoruiDong: Is there a reason, we want to use the actual dihedral angle or other values?

rmgpy/statmech/torsion.pyx
xiaoruiDong: Sure thing.

arkane/ess/gaussian.py
xiaoruiDong: I think there is no `_load_scan_angle` method for qchem and terachem logs. I have no experience on these two packages and prefer not to change these methods in the current PR.

xiaoruiDong:
@amarkpayne, Hi mark, I add two commits.

One is to read the actual angles from output files (previously, we assume angles are a list of equally distributed with a step predefined). The actual angles are slightly different.

The other one is to mainly fix a bug in the nDTorsion. There are two things, 1. the final point is no longer necessary to be 2pi exactly now, so I simply extend the phis by `phis[:-1]` (-1 used to refer to 2pi in the previous PR); 2. Since we are grabbing `[:-1]`, we should also extend Es by `Es[:-1]` instead of `Es[1:]`.

Please help me review these when you are available. Thanks!

xiaoruiDong:
@davidfarinajr Thanks for rejuvenating this PR. I made some changes according to your suggestions.

PR #1982, opened 2020-06-12
Minor: Remove unicode encoding & use f-string in reference

hwpang:
### Motivation or Problem
The string attribute .encode("utf-8") returns bytes object, and thus lines written by the old code can't be read & written with this same function again, which is inconvenient. 

### Description of Changes
1. Removed unicode encoding since the string in python 3 is unicode. 
2. Converted format to f-string.

### DIFF
--- a/rmgpy/data/reference.py
+++ b/rmgpy/data/reference.py
@@ -88,9 +88,9 @@ class Reference(object):
         """
         string = self.get_author_string()
         if self.title != '':
-            string += u' *{0}*'.format(self.title)
+            string += f' *{self.title}*'
         if self.year != '':
-            string += u' ({0})'.format(self.year)
+            string += f' ({self.year})'
         if string and string[-1] != '.':
             string += '.'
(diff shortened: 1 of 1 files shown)


hwpang:
> This looks good, how did you test the modifications?

Thanks for the review! This is the script I used for testing:

```
from rmgpy.data.kinetics import KineticsDatabase
from rmgpy import settings
import os

family_name = "Intra_R_Add_Endocyclic"

db = KineticsDatabase()
db.load(path=os.path.join(settings["database.directory"],"kinetics"),families=[family_name])

family = db.families[family_name]

family.get_training_depository().save('./test.py')
```

The current version would have `AttributeError: 'bytes' object has no attribute 'encode'`. This PR can run the script without error.

hwpang:
@mjohnson541 Rebased, thanks!

hwpang:
I rebased again per @amarkpayne request!

PR #1990, opened 2020-07-23
Return signal.SIGINT to the default after using lpsolve

mjohnson541:
Lpsolve steals the interrupt signal when it runs, to handle this the original configuration grabbed the current SIGINT signal and returned it to that value. That worked well when running python normally, but this does not work with python-jl since
python-jl gives SIGINT to julia which causes grabbing the signal to return None. 

Setting it to the default (raise KeyBoardInterrupt) after running lpsolve should work in both cases.

### DIFF
--- a/rmgpy/molecule/resonance.py
+++ b/rmgpy/molecule/resonance.py
@@ -935,10 +935,6 @@ def _clar_optimization(mol, constraints=None, max_num=None):
                    a=list, objective=list, status=cython.int, solution=list, innerSolutions=list)
 
     from lpsolve55 import lpsolve
-    import signal
-
-    # Save the current signal handler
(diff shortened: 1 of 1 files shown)


mliu49:
During the Python 3 transition, @amarkpayne and I were unable to reproduce the signal hijacking behavior of lpsolve. Since Mark was modifying the lpsolve code for Python 3 compatibility, we thought it would be a good opportunity to remove the signal hijacking, but were unable to reproduce the behavior.

I had meant to look into it further, but never got around to it. If the signal handler is now restored properly without the manual juggling on our end, it would be really nice to remove the code completely.

mjohnson541:
Lets try removing the handling entirely then!

Issue #1872, opened 2020-01-20
Solution phase models might include with PDep rates from libraries


alongd:
### Bug Description
If a kinetic library has pressure-dependent rates, and this library is used in a simulation involving a LiquidReactor, some of these rates end up in the core, making RMG crash when it checks the model:
```
  File "/home/alongd/Code/RMG-Py/rmgpy/rmg/main.py", line 982, in execute
    self.check_model()
  File "/home/alongd/Code/RMG-Py/rmgpy/rmg/main.py", line 1202, in check_model
    p_min=self.Pmin, p_max=self.Pmax)
  File "rmgpy/reaction.py", line 1217, in rmgpy.reaction.Reaction.check_collision_limit_violation
  File "rmgpy/reaction.py", line 1250, in rmgpy.reaction.Reaction.check_collision_limit_violation
  File "rmgpy/kinetics/arrhenius.pyx", line 775, in rmgpy.kinetics.arrhenius.PDepArrhenius.get_rate_coefficient
  File "rmgpy/kinetics/arrhenius.pyx", line 785, in rmgpy.kinetics.arrhenius.PDepArrhenius.get_rate_coefficient
ValueError: No pressure specified to pressure-dependent PDepArrhenius.get_rate_coefficient().
```

Although this is a conflict in the user's definitions (use LiquidReactor and use a library with PDep expressions), I still think this is a bug, since we don't warn not to do so, and RMG should have caught that on time, either crashing or dealing with it appropriately.

### How To Reproduce
Run a LiquidReactor simulation with the NOx2018 library including methanol in the initial species (so that OH + CH3 = CH3OH, for example, is taken from that library)

### Expected Behavior
I would expect RMG to at least try taking the high pressure limit rate from a pdep expression, if available (we have the functionality to do that: https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/data/kinetics/library.py#L150). If unavailable, we need to decide whether to crash, or warn the user while ignoring that library reaction.

hwpang:
So current LiquidReactor does have a `P` attribute that defaults to P=1e8 Pa to get reactions to the high pressure limit. This error is actually implicitly caused by line 677 in main.py because LiquidReactor doesn't have `Prange` attribute.

```
        try:
            self.Pmin = min([x.Prange[0].value_si if x.Prange else x.P.value_si for x in self.reaction_systems])
[...]

PR #1994, opened 2020-07-29
Modify Pmin/Pmax code in main.py & Feed in species_list for ThirdBody type kinetics to allow collision limit violation check

hwpang:
### Motivation or Problem
See #1872.

Essentially the original code intended to get reactions to high pressure limit by setting P=1e8 Pa in LiquidReactor, but it couldn't perform the collision limit violation check for PDep type reactions because LiquidReactor doesn't have Prange attribute. (main.py line 677)

```
        try:
            self.Pmin = min([x.Prange[0].value_si if x.Prange else x.P.value_si for x in self.reaction_systems])
            self.Pmax = max([x.Prange[1].value_si if x.Prange else x.P.value_si for x in self.reaction_systems])
        except AttributeError:
            # For LiquidReactor, Pmin and Pmax remain with the default value of `None`
            pass
```

### Description of Changes
Added dummy variable `Prange` in LiquidReactor. With this, Pmin = Pmax = P = 1e8 Pa to get reactions to high pressure limit for LiquidReactor.

### Testing

This is my testing script. The original code gave an AttributeError when trying to access x.Prange. This PR's code can perform the collision limit violation check.

```
from rmgpy.rmg.main import RMG, RMGDatabase
from rmgpy.solver.liquid import LiquidReactor
from rmgpy.kinetics import PDepArrhenius
from rmgpy import settings

rmg = RMG()
rmg.reaction_systems = [LiquidReactor(298,{})]

rmg.Tmin = min([x.Trange[0].value_si if x.Trange else x.T.value_si for x in rmg.reaction_systems])
rmg.Tmax = max([x.Trange[1].value_si if x.Trange else x.T.value_si for x in rmg.reaction_systems])
rmg.Pmin = min([x.Prange[0].value_si if x.Prange else x.P.value_si for x in rmg.reaction_systems])
rmg.Pmax = max([x.Prange[1].value_si if x.Prange else x.P.value_si for x in rmg.reaction_systems])

db = RMGDatabase()
db.load(
    path=settings['database.directory'],
    thermo_libraries=['primaryThermoLibrary'],
    transport_libraries=[],
    reaction_libraries=['NOx2018'],
    seed_mechanisms=[],
    kinetics_families='default',
    kinetics_depositories=['training'],
    depository=False,
)

for rxn in db.kinetics.libraries["NOx2018"].get_library_reactions():
    if isinstance(rxn.kinetics,PDepArrhenius):
        for i, spc in enumerate(rxn.reactants):
            rxn.reactants[i].thermo = db.thermo.get_thermo_data(spc)
        for i, spc in enumerate(rxn.products):
            rxn.products[i].thermo = db.thermo.get_thermo_data(spc)
        PDeprxn = rxn
        break

PDeprxn.check_collision_limit_violation(t_min=rmg.Tmin, t_max=rmg.Tmax,p_min=rmg.Pmin, p_max=rmg.Pmax)
```

### DIFF
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -63,6 +63,7 @@ from rmgpy.data.kinetics.library import KineticsLibrary, LibraryReaction
 from rmgpy.data.rmg import RMGDatabase
 from rmgpy.exceptions import ForbiddenStructureException, DatabaseError, CoreError
 from rmgpy.kinetics.diffusionLimited import diffusion_limiter
+from rmgpy.kinetics import ThirdBody
 from rmgpy.molecule import Molecule
 from rmgpy.qm.main import QMDatabaseWriter
 from rmgpy.reaction import Reaction
@@ -1244,7 +1245,10 @@ class RMG(util.Subject):
                                   '"Violation factor" is the ratio of the rate coefficient to the collision limit'
                                   ' rate at the relevant conditions\n\n')
                 for violator in violators:
-                    rxn_string = violator[0].to_chemkin()
+                    if isinstance(violator[0].kinetics, ThirdBody):
+                        rxn_string = violator[0].to_chemkin(self.reaction_model.core.species)
+                    else:
+                        rxn_string = violator[0].to_chemkin()
                     direction = violator[1]
                     ratio = violator[2]
                     condition = violator[3]
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -674,12 +674,8 @@ class RMG(util.Subject):
         # determine min and max values for T and P (don't determine P values for liquid reactors)
         self.Tmin = min([x.Trange[0].value_si if x.Trange else x.T.value_si for x in self.reaction_systems])
         self.Tmax = max([x.Trange[1].value_si if x.Trange else x.T.value_si for x in self.reaction_systems])
-        try:
-            self.Pmin = min([x.Prange[0].value_si if x.Prange else x.P.value_si for x in self.reaction_systems])
-            self.Pmax = max([x.Prange[1].value_si if x.Prange else x.P.value_si for x in self.reaction_systems])
-        except AttributeError:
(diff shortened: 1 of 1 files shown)


hwpang:
During investigating above mentioned issue, I realize that the current collision limit violation check doesn't work for ThirdBody type kinetics, because the species_list isn't fed in. The original mainTest before this PR didn't have a ThirdBody type kinetics violator.

alongd:
I wonder if we could make this fix "cleaner". It's good that you added the comment next to `self.Prange = None`, because accidentally deleting or changing it is indeed a weakness of this implementation. But it's implementing a solution far away (code-wise) from the problem.
If you agree, I'd recommend directly addressing the issue in [main.py](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/rmg/main.py#L677), so that it correctly deals with any scenario (and won't break if `LiquidReactor` ever changes).
What do you say about something like:

```python
self.Pmin = min([x.Prange[0].value_si if x.hasattr('Prange') and x.Prange
                 else x.P.value_si for x in self.reaction_systems])
```

hwpang:
> I wonder if we could make this fix "cleaner". It's good that you added the comment next to `self.Prange = None`, because accidentally deleting or changing it is indeed a weakness of this implementation. But it's implementing a solution far away (code-wise) from the problem.
> If you agree, I'd recommend directly addressing the issue in [main.py](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/rmg/main.py#L677), so that it correctly deals with any scenario (and won't break if `LiquidReactor` ever changes).
> What do you say about something like:
> 
> ```python
> self.Pmin = min([x.Prange[0].value_si if x.hasattr('Prange') and x.Prange
>                  else x.P.value_si for x in self.reaction_systems])
> ```

Thanks for the comment! @alongd I agree that this is a much more elegant way to do it. I modified the code for Pmin/Pmax and removed the now stale try/except block in main.py.

PR #1997, opened 2020-08-03
halogen additions (simultaneous update with DB)

davidfarinajr:
This PR goes with https://github.com/ReactionMechanismGenerator/RMG-database/pull/434.  Importantly, this PR adds the lone pairs to make a sample bromine atom, which was previously causing the DB Travis build to fail.  I also added a new atom type `R!H!Val7!`, which will be used in kinetics trees, and halogen identification functions (i.e `is_bromine`) to the molecule and group Atom classes.

### DIFF
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -2138,8 +2138,8 @@ class ThermoDatabase(object):
         cyclic = molecule.is_cyclic()
         # Generate estimates of the thermodynamics parameters
         for atom in molecule.atoms:
-            # Iterate over heavy (non-hydrogen) atoms
-            if atom.is_non_hydrogen():
+            # Iterate over atoms and skip hydogens and halogens (since there are no groups centered on these atomtypes)
+            if atom.is_non_hydrogen() and not atom.is_halogen():
                 # Get initial thermo estimate from main group database
                 data_added = False
                 try:
--- a/rmgpy/molecule/moleculeTest.py
+++ b/rmgpy/molecule/moleculeTest.py
@@ -208,6 +208,17 @@ class TestAtom(unittest.TestCase):
             else:
                 self.assertFalse(atom.is_chlorine())
 
+    def test_is_bromine(self):
+        """
+        Test the Atom.is_bromine() method.
+        """
+        for element in element_list:
+            atom = Atom(element=element, radical_electrons=1, charge=0, label='*1', lone_pairs=3)
+            if element.symbol == 'Br':
+                self.assertTrue(atom.is_bromine())
+            else:
+                self.assertFalse(atom.is_bromine())
+
     def test_is_iodine(self):
         """
         Test the Atom.is_iodine() method.
--- a/rmgpy/molecule/moleculeTest.py
+++ b/rmgpy/molecule/moleculeTest.py
@@ -120,6 +120,17 @@ class TestAtom(unittest.TestCase):
             else:
                 self.assertTrue(atom.is_non_hydrogen(), "Atom {0!r} isn't reporting is_non_hydrogen()".format(atom))
 
+    def test_is_halogen(self):
+        """
+        Test the Atom.is_halogen() method.
+        """
+        for element in element_list:
+            atom = Atom(element=element, radical_electrons=1, charge=0, label='*1', lone_pairs=3)
+            if element.symbol in ['F', 'Cl', 'Br', 'I']:
+                self.assertTrue(atom.is_halogen())
+            else:
+                self.assertFalse(atom.is_halogen(), "Atom {0!r} is reporting is_halogen(), but it shouldn't be".format(atom))
+
     def test_is_carbon(self):
         """
         Test the Atom.is_carbon() method.
--- a/documentation/source/reference/molecule/atomtype.rst
+++ b/documentation/source/reference/molecule/atomtype.rst
@@ -23,6 +23,7 @@ Atom type       Description
 ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 ``R``           any atom with any local bond structure
 ``R!H``         any non-hydrogen atom with any local bond structure
+``R!H!Val7``    any non-hydrogen and non-halogen atom (a non-terminal atom)
 *Hydrogen atom types*
 ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 ``H``           hydrogen atom with up to one single bond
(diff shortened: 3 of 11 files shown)


davidfarinajr:
looks like 3 Arkane tests failed due to `Exception: Running symmetry on the point group calculation has failed. Please check if symmetry program is installed on your system in RMG-Py/bin/symmetry or on your path.` and possibly `Error: [Errno 12] Cannot allocate memory`.  Not sure how to go about fixing that

amarkpayne:
> looks like 3 Arkane tests failed due to `Exception: Running symmetry on the point group calculation has failed. Please check if symmetry program is installed on your system in RMG-Py/bin/symmetry or on your path.` and possibly `Error: [Errno 12] Cannot allocate memory`. Not sure how to go about fixing that

I have ran into this issue before--I think it is  mix of us pushing the memory limit of Travis with the current size of our database and more importantly poorly written Arkane functional tests that seem to not tear-down properly (At least I have suspected this in the past, but was only able to make some progress towards fixing this).

This is almost certainly not a fault of your own PR--but it does mean that we need to look into fixing the Arkane functional tests, and I need to wrap up my GitHub Actions PR (assuming that GitHub's testing service offers more memory, which we haven't confirmed).

davidfarinajr:
https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/227596283

rwest:
If this doesn't belong in this PR please open a new issue for it, but we really ought to update the documentation of how to write adjacency lists. The list of permitted elements in molecules is out of date, and there's no mention of the special atom types that are allowed in groups (eg. `R!H` etc.). 

- http://reactionmechanismgenerator.github.io/RMG-Py/reference/molecule/adjlist.html#rmgpy-molecule-adjlist
- https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/documentation/source/reference/molecule/adjlist.rst

rmgpy/molecule/moleculeTest.py
```diff
@@ -120,6 +120,17 @@ def test_is_non_hydrogen(self):
...
+    def test_is_halogen(self):
+        """
+        Test the Atom.is_halogen() method.
+        """
+        for element in element_list:
+            atom = Atom(element=element, radical_electrons=1, charge=0, label='*1', lone_pairs=3)
+            if element.symbol in ['F', 'Cl', 'Br', 'I']:
+                self.assertTrue(atom.is_halogen())
+            else:
+                self.assertFalse(atom.is_halogen(), "Atom {0!r} isn't reporting is_halogen()".format(atom))
```
rwest, line 132:
```suggestion
                self.assertFalse(atom.is_halogen(), "Atom {0!r} is reporting is_halogen() but it shouldn't be".format(atom))
```

rmgpy/data/thermo.py
```diff
@@ -2020,7 +2020,7 @@ def compute_group_additivity_thermo(self, molecule):
         # Generate estimates of the thermodynamics parameters
         for atom in molecule.atoms:
             # Iterate over heavy (non-hydrogen) atoms
-            if atom.is_non_hydrogen():
+            if atom.is_non_hydrogen() and not atom.is_halogen():
```
rwest, line 2023: Is this temporary or permanently the right thing? H atoms seem to be treated specially below around line 2034; should halogens be treated similarly?   Is H special because it's terminal, one-short-of-valent (is there a term for that?) or because it's light, or something else?   
If this is all correct, and we're right to just be iterating over any non-terminal atoms, or non-valence-7 atoms (rather than "heavy" atoms) then please just update the comment on line 2022 which says "heavy (non-hydrogen)".  Or if this is pending future work (eg. more groups) open a new issue so we don't forget.

rwest (commented):
Besides my comments above/below, it looks good to me and I approve.

mliu49:
> If this doesn't belong in this PR please open a new issue for it, but we really ought to update the documentation of how to write adjacency lists. The list of permitted elements in molecules is out of date, and there's no mention of the special atom types that are allowed in groups (eg. `R!H` etc.).
> 
> * http://reactionmechanismgenerator.github.io/RMG-Py/reference/molecule/adjlist.html#rmgpy-molecule-adjlist
> * https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/documentation/source/reference/molecule/adjlist.rst

The full list of atom types is documented in both the user's guide and the API reference for `AtomType` (note that this means changes need to be made in two places). The adjlist page should still be updated though.
- http://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/database/introduction.html#group-definitions
- http://reactionmechanismgenerator.github.io/RMG-Py/reference/molecule/atomtype.html#rmgpy.molecule.AtomType

rwest:
> The full list of atom types is documented in both the user's guide and the API reference for `AtomType` (note that this means changes need to be made in two places). The adjlist page should still be updated though.
> 

Thanks Max. So some documentation updates should be made (in this PR) and maybe the Adjacency List page could just be updated to link to the other two places, so that we don't need to maintain the list in _three_ places moving forwards.

rmgpy/data/thermo.py
davidfarinajr: This is temporary, until we added hypervalent chlorine, bromine and iodine groups (which we don't currently have atom types for yet).  Currently, halogens are terminal, and therefore the thermo is accounted for by central atom (like CF3 group in CF3-R).  We have dummy halogen groups in the thermo database so that, when we descend the tree with a halogen, it hits the dummy node with zeros.  I added `not atom.is_halogen()` here to avoid descending the tree at all, because it will be zeros.  However, since we will want to descend the tree if/when we add hypervalent groups, perhaps I should remove this since the dummy halogen groups are already there to account for terminal halogens.

mazeau (approved):
I'm not a huge fan of the R!H!Val7 new atomtype and I can't help but think there should be a nicer way to build trees without that atomtype but I can't think of a better way so I'll approve this

davidfarinajr:
This DB test is failing because nitrogen `adsorptionPt111` entry is not accessible.  We need to merge this DB PR https://github.com/ReactionMechanismGenerator/RMG-database/pull/450 first, then the tests should pass

mazeau (approved):
I'm super glad that we're testing species with nitrogen in them now! Thank you very much

rmgpy/data/thermo.py
rwest: The comment was updated and makes more sense. resolving conversation.

rwest (approved):
Looks good to me!

PR #1998, opened 2020-08-06
Readded try-except for Pmin and Pmax determination

kblondal:
This commit prevents error message when running with surfaceReactor
For now, Pmin and Pmax of surfaceReactor remain with the default value of `None`

See comments on https://github.com/ReactionMechanismGenerator/RMG-Py/commit/219590135056ceb98571a6bd04a47b6955456eeb


hwpang:
Thanks for the PR! This changes looks good to me. I don't have the authority to merge PRs. @alongd would you please merge it?

rwest:
Thanks for fixing this, Katrin. 
I wonder if we could have more regression/unit tests for catalyssis. It's strange that a bug that entirely breaks all catalytic simulations got introduced to the master branch.

PR #2001, opened 2020-08-10
Allow Specification of an Initial Set of Reacts

mjohnson541:
Currently there's no way to force RMG to react species that we know are important. This PR adds a react() block that takes in a list of tuples of species labels. Each tuple corresponds to a group of species that are reacted at the beginning of the run.

### DIFF
--- a/rmgpy/rmg/input.py
+++ b/rmgpy/rmg/input.py
@@ -174,7 +174,17 @@ def inchi(string):
 def adjacency_list(string):
     return Molecule().from_adjacency_list(string)
 
-
+def react(tups):
+    if not isinstance(tups, list):
+        raise InputError("React takes a list of tuples of species strings.")
+    for item in tups:
+        if not isinstance(item, tuple):
+            raise InputError("React takes a list of tuples of species strings.")
+        for it in item:
+            if not isinstance(it, str):
+                raise InputError("React takes a list of tuples of species strings.")
+    rmg.init_react_tuples = tups
+            
 # Reaction systems
 def simple_reactor(temperature,
                    pressure,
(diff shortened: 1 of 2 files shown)


rmgpy/rmg/main.py
```diff
@@ -1701,7 +1701,32 @@ def initialize_reaction_threshold_and_react_flags(self):
                         self.bimolecular_react[:num_restart_spcs, :num_restart_spcs] = False
                         if self.trimolecular:
                             self.trimolecular_react[:num_restart_spcs, :num_restart_spcs, :num_restart_spcs] = False
-
+            
+            if self.filter_reactions and self.init_react_tuples:
```
xiaoruiDong, line 1705: Not sure about the reason why `self.filter_reactions` is needed. Do you think we can move this block to the beginning of the function so that values assigned for `self.xxmolecular_react` will not be overwritten by this block. Since we are not really `initialize_reaction_threshold_and_react_flags`(name of the function) in this block. Do you think, it may be better to have a separate function for this block and call the new function in `initialize()`

mjohnson541: If we aren't filtering reactions there's no need to run this block because everything is going to be reacted anyway and it much better to not tamper with variables that shouldn't be touched in a non-filter run. I've adjusted the block to process identically to the react/threshold update function that should allow it to operate nicely with the restart feature, the values should be changed by this block. I've also isolated it in a separate function as suggested.

rmgpy/rmg/input.py
```diff
@@ -174,7 +174,17 @@ def inchi(string):
...
+def react(tups):
+    if not isinstance(tups,list):
+        raise ValueError("react takes a list of tuples of species strings")
+    for item in tups:
+        if not isinstance(item,tuple):
+            raise ValueError("react takes a list of tuples of species strings")
+        for it in item:
+            if not isinstance(it,str):
+                raise ValueError("react takes a list of tuples of species strings")
+    rmg.init_react_tuples = tups
```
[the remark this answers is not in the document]
mjohnson541: fixed

mjohnson541:
It isn't possible to write a reasonable unit test for this function as it doesn't operate independently of the react functionality.

Issue #2003, opened 2020-08-14
Bidentate adsorbates incorrectly estimated using Rj-Rj biradical not R=R molecule in gas phase


kblondal:
For monodentate adsorbates, the gas-phase precursor is obtained by breaking the bond to the surface, leaving a radical at the binding atom. Similarly, the gas-phase precursor of a bidentate adsorbate is currently obtained by breaking both bonds to the surface, leaving a radical at each binding atom. In some cases this does not match the adsorption corrections, like the one shown here:
![Bidentate_adsorbate](https://user-images.githubusercontent.com/38664427/90238652-49509d00-ddf4-11ea-9e9b-a55170d61d23.png)
The adsorption correction assumes that the precursor has a triple bond between the two binding atoms. Therefore, when breaking both bonds of a bidentate adsorbate to the surface to form the gas-phase precursor, the suggestion is to increase the bond order between the binding atoms rather than leaving the two radicals (if the binding atoms are connected).

PR #2006, opened 2020-08-19
Fixes to adsorption thermochemistry estimation, including bidentate.

rwest:
### Motivation or Problem
Issue #2003 identifies that  the gas-phase precursor of a bidentate adsorbate is currently obtained by breaking both bonds to the surface, leaving a radical at each binding atom. This leads to a gas phase estimate of the biradical, instead of the closed shell species with a higher bond order. 
When fixing that issue, I created some unit tests for adsorbate thermo, which revealed another problem: When CH3X was removed from the surface to make CH3j, it was not matching the CH3j species in the gas phase thermo library. The issue was the atom's connectivity values had not been reset by the call to Molecule.update().
### Description of Changes
To fix this I
- explicitly call update_connectivity_values() after removing the adsorbate from the surface
- fix the docstrings on .update() so it is no longer misleading about what it does and doesn't do
- ~made the underlying add/remove vertex/edge methods on the Graph class reset (i.e. mark as invalid) the cached connectivity values. This should trigger a recalculation of them if/when they're needed, rather than code just assuming they're correct when they're not.~ [Moved this to #2009 because was causing a unit test failure]

Also
- added the unit tests, as mentioned
- fixed the bidentate adsorbate issue (so CH2X-CH2X is estimated from CH2=CH2 not CH2j-CH2j )
- increased detail on the Linear Scaling Relationship comment on thermo estimates.

### Testing
Added some unit tests, which now pass.

### DIFF
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -1487,6 +1487,11 @@ class ThermoDatabase(object):
                 bond.increment_order()
                 adsorbed_atoms[0].decrement_radical()
                 adsorbed_atoms[1].decrement_radical()
+                if adsorbed_atoms[0].radical_electrons and adsorbed_atoms[1].radical_electrons:
+                    # There are still spare adjacenct radicals, so do it again
+                    bond.increment_order()
+                    adsorbed_atoms[0].decrement_radical()
+                    adsorbed_atoms[1].decrement_radical()
                 
         dummy_molecule.update_connectivity_values()
         dummy_molecule.update()
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -1469,8 +1469,8 @@ class ThermoDatabase(object):
                     bonded_atom.increment_lone_pairs()
                 else:
                     raise NotImplementedError("Can't remove surface bond of type {}".format(bond.order))
-
             dummy_molecule.remove_atom(site)
+        dummy_molecule.update_connectivity_values()
         dummy_molecule.update()
 
         logging.debug("Before removing from surface:\n" + molecule.to_adjacency_list())
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -1415,10 +1415,13 @@ class ThermoDatabase(object):
             find_cp0_and_cpinf(species, thermo)
 
         # now edit the adsorptionThermo using LSR
+        comments = []
         for element in 'CHON':
-            change_in_binding_energy = self.delta_atomic_adsorption_energy[element].value_si * normalized_bonds[element]
-            thermo.H298.value_si += change_in_binding_energy
-        thermo.comment += " Binding energy corrected by LSR."
+            if normalized_bonds[element]:
+                change_in_binding_energy = self.delta_atomic_adsorption_energy[element].value_si * normalized_bonds[element]
+                thermo.H298.value_si += change_in_binding_energy
+                comments.append(f'{normalized_bonds[element]:.2f}{element}')
+        thermo.comment += " Binding energy corrected by LSR ({})".format('+'.join(comments))
         return thermo
 
     def get_thermo_data_for_surface_species(self, species):
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -1435,13 +1435,14 @@ class ThermoDatabase(object):
         if species.is_surface_site():
             raise DatabaseError("Can't estimate thermo of vacant site. Should be in library (and should be 0).")
 
-        logging.debug(("Trying to generate thermo for surface species"
-                       " with these {} resonance isomer(s):").format(len(species.molecule)))
+        logging.debug("Trying to generate thermo for surface species with these %d resonance isomer(s):",
+                      len(species.molecule))
         molecule = species.molecule[0]
         # only want/need to do one resonance structure,
         # because will need to regenerate others in gas phase
         dummy_molecule = molecule.copy(deep=True)
         sites_to_remove = []
+        adsorbed_atoms = []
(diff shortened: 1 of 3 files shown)


rmgpy/data/thermo.py
```diff
@@ -1470,6 +1472,19 @@ def get_thermo_data_for_surface_species(self, species):
                 else:
                     raise NotImplementedError("Can't remove surface bond of type {}".format(bond.order))
             dummy_molecule.remove_atom(site)
+
+        if len(adsorbed_atoms) == 2:
+            # Bidentate adsorption.
+            # Try to turn adjacent biradical into a bond.
+            try:
+                bond = adsorbed_atoms[0].bonds[adsorbed_atoms[1]]
+            except IndexError:
```
davidfarinajr, line 1484: Since `adsorbed_atoms[0].bonds` is a dict, I think this should be `KeyError` instead of `IndexError` (@kblondal discovered this when running RMG with this branch)

rwest: Thanks. making a fixup now

rmgpy/data/thermo.py
```diff
@@ -1470,6 +1472,19 @@ def get_thermo_data_for_surface_species(self, species):
...
+                pass # the two adsorbed atoms are not bonded to each other
+            else:
+                bond.increment_order()
+                adsorbed_atoms[0].decrement_radical()
+                adsorbed_atoms[1].decrement_radical()
+
```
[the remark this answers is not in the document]
rwest: Good point. Done in 982065d39130c9375a5e1fc294a596ba54c75512

kblondal:
I ran this with a database branch including some bidentate families and the run completed without any errors after the fixups. Bidentate adsorbate thermo is reasonable and what is great is that now bidentate families are being added from the edge to the core which didn't happen before since the thermo was off.
What seems to be an additional consequence of the changes made, which I am not sure are intentional (currently) is that RMG is considering adsorbates with even more than two binding sites (3 and some 4 Xs in the edge). These don't get in the core so it probably isn't a big issue but the thermo is questionable. 

200820_3_4_X

Otherwise everything looks good to me!

rwest:
Thanks. Can you see how those poly-dentate species are being made? Maybe we block at the reaction family level. Or else just try to estimate their thermo accurately. Or put a max element species constraint of “no more than 2 X atoms”.

kblondal:
They all seem to be a product of the Surface_DoubleBond_to_Bidentate family that is included on the bidentate database branch I was running with. So blocking at the reaction family level should work.

amarkpayne:
Just want to make sure I understand the situation--this PR needs to wait until we resolve #2009 , right? Or can it be merged in independently?

rwest:
I split #2009 off so that it would be independent. This PR no longer needs it.

amarkpayne:
That makes sense, I just wanted to confirm. I wasn't seeing the error in the Travis tests. All of the remaining errors are just from the matplotlib stuff that got merged in a while back. If you rebase this PR I'll take a final look and then we can merge this in

rwest:
Added one more commit, rebased, ran tests locally, (Ran 1789 tests in 197.992s OK (SKIP=44)) and force pushed.

rmgpy/data/thermo.py
```diff
@@ -1487,6 +1487,11 @@ def get_thermo_data_for_surface_species(self, species):
                 bond.increment_order()
                 adsorbed_atoms[0].decrement_radical()
                 adsorbed_atoms[1].decrement_radical()
+                if adsorbed_atoms[0].radical_electrons and adsorbed_atoms[1].radical_electrons:
+                    # There are still spare adjacenct radicals, so do it again
+                    bond.increment_order()
+                    adsorbed_atoms[0].decrement_radical()
+                    adsorbed_atoms[1].decrement_radical()
```
amarkpayne: I am assuming that something like X#R-R#X would never happen, right? If so the atoms would have lone pairs instead of radicals. Though even then I guess they would keep their lone pairs instead of forming another bond? I am sure this is fine, it is just something that I was curious about.

rmgpy/data/thermoTest.py
```diff
@@ -707,6 +707,38 @@ def test_adsorbate_thermo_generation_bidentate_double(self):
...
+9 H u0 p0 c0 {3,S}
+10 X u0 p0 c0 {2,S}
+11 X u0 p0 c0 {3,S}""")])
+        spec.generate_resonance_structures()
+        initial = list(spec.molecule)  # Make a copy of the list
+        thermo = self.database.get_thermo_data(spec)
+
+        self.assertEqual(len(initial), len(spec.molecule))
+        self.assertEqual(set(initial), set(spec.molecule))
+        #self.assertTrue('library' in thermo.comment, 'Thermo not found from library, test purpose not fulfilled.')
```
amarkpayne, line 815: Minor: remove comment

rmgpy/data/thermo.py
```diff
@@ -1484,15 +1484,28 @@ def get_thermo_data_for_surface_species(self, species):
...
+                    if (adsorbed_atoms[0].lone_pairs and
+                            adsorbed_atoms[1].lone_pairs and 
+                            bond.order < 3):
+                        # X#C-C#X will end up with .:C-C:. in gas phase
+                        # and we want to get to .C#C. but not :C=C:
+                        bond.increment_order()
+                        adsorbed_atoms[0].decrement_lone_pairs()
+                        adsorbed_atoms[0].increment_radical()
+                        adsorbed_atoms[1].decrement_lone_pairs()
+                        adsorbed_atoms[1].increment_radical()
```
amarkpayne, line 1507: Never mind, this answers my previous question

PR #2015, opened 2020-08-27
Cantera thermo sa

kspieks:
### Description of Changes
Added the option to run thermo SA using the Cantera class.

### Testing
`canteramodelTest.py` passes and I ran `cantera_sensitivity_comparison.ipynb` to verify the feature. Without the new `thermo_SA` flag, it preserves the same behavior as before. With the new `thermo_SA` flag, it correctly produces the data and plot for thermo SA. Feel free to checkout the notebook :)


rmgpy/tools/canteramodel.py
```diff
@@ -363,7 +368,12 @@ def plot(self, data, top_species=10, top_sensitive_reactions=10):
                 ReactionSensitivityPlot(x_var=time,
                                         y_var=reaction_sensitivity_data[j * num_ct_reactions:(j + 1) * num_ct_reactions],
                                         num_reactions=top_sensitive_reactions).barplot(
-                    os.path.join(self.output_directory, '{0}_{1}_sensitivity.png'.format(i + 1, species.to_chemkin())))
+                    os.path.join(self.output_directory, '{0}_{1}_reaction_sensitivity.png'.format(i + 1, species.to_chemkin())))
+                if self.thermo_SA:
+                    ThermoSensitivityPlot(x_var=time,
+                                          y_var=thermodynamic_sensitivity_data[j * num_ct_species:(j + 1) * num_ct_species],
+                                          num_species=top_sensitive_species).barplot(
+                        os.path.join(self.output_directory, '{0}_{1}_thermo_sensitivity.png'.format(i + 1, species.to_chemkin())))
```
xiaoruiDong, line 376: Is Cantera calculate sensitivity wrt enthalpy? Is that is the case, then should we assign the argument `xlabel` to `'dln(c)/d(H_i) [(kcal/mol)^-1]'` (maybe unit based on mass fraction instead?)?

rmgpy/tools/canteramodel.py
```diff
@@ -497,19 +532,33 @@ def simulate(self):
...
                     reaction_sensitivity_data.append(reaction_sensitivity_generic_data)
 
-            all_data.append((time, condition_data, reaction_sensitivity_data))
+            # save thermo data as generic data object
+            thermodynamic_sensitivity_data = []
+            if self.thermo_SA:
+                for index, species in enumerate(self.sensitive_species):
+                    for j in range(num_ct_species):
+                        thermo_sensitivity_generic_data = GenericData(
+                            label='dln[{0}]/dG[{1}]'.format(species, self.model.species()[j].name),
```
xiaoruiDong, line 554: Should we also change here?

rmgpy/tools/canteramodel.py
amarkpayne: So @mjohnson541 and I talked about this once, and I think we decided that thermo sensitivity with respect to Gibbs free energy is exactly the same as with respect to enthalpy. It is completely fair to change the label, but I don't think we need to worry if not.

amarkpayne: Also, I don't think it is possible to convert this based on mass fraction, as H_i does not depend on the mass fraction. You could use mass based units like `kcal/g` but I don't think this is as useful. We tend to think about the thermochemistry of the species involved in terms of molar units. Correct me if I am wrong though, I don't know if perhaps there is a better way that RMS or someone else uses.

rmgpy/tools/canteramodel.py
amarkpayne: I agree, we should be consistent either way

kspieks:
Thank you both for catching the units and for our offline discussion. I just rebased and pushed the branch with the updated units :)

rmgpy/tools/canteramodel.py
```diff
@@ -363,7 +367,13 @@ def plot(self, data, top_species=10, top_sensitive_reactions=10):
                 ReactionSensitivityPlot(x_var=time,
                                         y_var=reaction_sensitivity_data[j * num_ct_reactions:(j + 1) * num_ct_reactions],
                                         num_reactions=top_sensitive_reactions).barplot(
-                    os.path.join(self.output_directory, '{0}_{1}_sensitivity.png'.format(i + 1, species.to_chemkin())))
+                    os.path.join(self.output_directory, '{0}_{1}_reaction_sensitivity.png'.format(i + 1, species.to_chemkin())))
+                if self.thermo_SA:
+                    ThermoSensitivityPlot(x_var=time,
+                                          y_var=thermodynamic_sensitivity_data[j * num_ct_species:(j + 1) * num_ct_species],
+                                          xlabel='dln(c)/d(H_i) [(J/mol)^-1]',
```
mjohnson541, line 374: I think kcal/mol is much preferable to J/mol for most kineticists, it has better known benchmarks, definitely should at least use kJ/mol.

kspieks: Thanks for sharing. I updated the units to plot in [(kcal/mol)^-1]

mjohnson541:
Looks good! Can you squash the minor commit in? Then I think this is ready to merge.

amarkpayne:
@kspieks I just want to make sure before this gets merged. Where do you convert the units from kmol/J to mol/kcal ? I don't see this unit conversion anywhere, though perhaps it is handled somewhere by ArrayQuantity objects or similar

amarkpayne:
Nevermind, I see it

rmgpy/tools/canteramodel.py
```diff
@@ -370,8 +370,8 @@ def plot(self, data, top_species=10, top_sensitive_reactions=10, top_sensitive_s
                     os.path.join(self.output_directory, '{0}_{1}_reaction_sensitivity.png'.format(i + 1, species.to_chemkin())))
                 if self.thermo_SA:
                     ThermoSensitivityPlot(x_var=time,
-                                          y_var=thermodynamic_sensitivity_data[j * num_ct_species:(j + 1) * num_ct_species],
-                                          xlabel='dln(c)/d(H_i) [(J/mol)^-1]',
+                                          y_var=thermodynamic_sensitivity_data[j * num_ct_species:(j + 1) * num_ct_species]*4184,
```
amarkpayne, line 373: I think the conversion factor is 4184000

kspieks: Ah sorry I'll update this now. Thanks for catching that

PR #2019, opened 2020-08-28
Save metal attributes in adsorption thermo groups file.

rwest:
Small change. The new attributes were not saved when saving thermo groups.
The fact that we didn't know it was missing shows that database i/o is not 100% tested.


mazeau (approved):

rwest:
This was added in 2da61c8000cd58c91fe1e356bca73c97ee5a7307 which was merged via a different pull request. Closing this one.

PR #1983, opened 2020-06-14
Update anacondaDeveloper.rst

mefuller:
Combine Ubuntu installation commands for gcc, g++, git and make to one line;
add equivalent for installation instruction for Fedora since this is a major distro on the Red Hat side (vs. Debian side of the Linux world with Ubuntu).
This should add clarity for newer users.

### Motivation or Problem
The documentation for the developer install on Linux would benefit from providing instructions specific to Fedora and not just Ubuntu since this is another widely-used distro and serves as a template for most Red Hat derivatives.

### Description of Changes
The necessary command to install gcc, g++, make, and git was added.
Also the four Ubuntu commands were compressed to one line.
Both changes should make installation slightly easier either for those copying and pasting or novice users.

### Testing
The suggested command was executed successfully on a Fedora 32 system.
Installation of multiple packages in one line with apt was confirmed on a current Ubuntu 20.04 system.
These steps have been followed to set up RMG developer environments on Ubuntu 20.04 and Fedora 32.

### Reviewer Tips
Suggestions for verifying that this PR works or other notes for the reviewer.

PR #2020, opened 2020-08-28
Update anacondaDeveloper.rst

mefuller:
Replaces pull request #1983

Combine Ubuntu installation commands for gcc, g++, git and make to one line;
add equivalent for installation instruction for Fedora since this is a major distro on the Red Hat side (vs. Debian side of the Linux world with Ubuntu).
This should add clarity for newer users.
Motivation or Problem

The documentation for the developer install on Linux would benefit from providing instructions specific to Fedora and not just Ubuntu since this is another widely-used distro and serves as a template for most Red Hat derivatives.
Description of Changes

The necessary command to install gcc, g++, make, and git was added.
Also the four Ubuntu commands were compressed to one line.
Both changes should make installation slightly easier either for those copying and pasting or novice users.
Testing

The suggested command was executed successfully on a Fedora 32 system.
Installation of multiple packages in one line with apt was confirmed on a current Ubuntu 20.04 system.
These steps have been followed to set up RMG developer environments on Ubuntu 20.04 and Fedora 32.
Reviewer Tips


amarkpayne:
Thanks @mefuller for the PR!

PR #2034, opened 2020-10-22
Bidentate CO treatment improved

kblondal:
### Problem
Before the changes, when running with bidentate surface families the mechanism generation would in some cases crash due to the following resonance structure of gas-phase CO when obtaining the gas-phase counterpart of bidentate CO:
```
Error: Could not update atomtypes for this molecule:
multiplicity 1
1 O u1 p1 c0 {2,T}
2 C u1 p0 c0 {1,T}
```

The surface species in question is:
```
Before removing from surface:
1 O u0 p2 c0 {2,S} {4,S}
2 C u0 p0 c0 {1,S} {3,T}
3 X u0 p0 c0 {2,T}
4 X u0 p0 c0 {1,S}
```

### Description of Changes
1. This adds a unit test for the treatment of the bidentate CO adsorbate that previously had the strange gas-phase counterpart.
2. More descriptive messages that help to debug this kind of error when obtaining surface thermo.
3. Adds an if statement to correct the resonance structure of gas-phase CO (should be specific to CO) after removing bidentate CO from the surface.

### Testing
I tried running the same input file again with these changes and the mechanism generation completed successfully without the crash. I saw in the RMG.log file that the correct gas-phase counterpart for CO was formed:
```
1 C u0 p1 c-1 {2,T}
2 O u0 p1 c+1 {1,T}
```

### DIFF
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -1219,9 +1219,14 @@ class ThermoDatabase(object):
             return thermo0
 
         if species.contains_surface_site():
-            thermo0 = self.get_thermo_data_for_surface_species(species)
-            thermo0 = self.correct_binding_energy(thermo0, species)
-            return thermo0
+            try:
+                thermo0 = self.get_thermo_data_for_surface_species(species)
+                thermo0 = self.correct_binding_energy(thermo0, species)
+                return thermo0
+            except:
+                logging.error("Error attempting to get thermo for species %s with structure \n%s", 
+                    species, species.molecule[0].to_adjacency_list())
+                raise
 
         try:
             quantum_mechanics = get_input('quantum_mechanics')
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -1458,9 +1458,10 @@ class ThermoDatabase(object):
         if species.is_surface_site():
             raise DatabaseError("Can't estimate thermo of vacant site. Should be in library (and should be 0).")
 
-        logging.debug("Trying to generate thermo for surface species with these %d resonance isomer(s):",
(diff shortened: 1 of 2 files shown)


mazeau (dismissed):
Yay! I'm glad you guys were able to fix this so quickly.  Looks great to me

rmgpy/data/thermo.py
```diff
@@ -1219,9 +1219,14 @@ def get_thermo_data(self, species, training_set=None):
...
-            thermo0 = self.get_thermo_data_for_surface_species(species)
-            thermo0 = self.correct_binding_energy(thermo0, species)
-            return thermo0
+            try:
+                thermo0 = self.get_thermo_data_for_surface_species(species)
+                thermo0 = self.correct_binding_energy(thermo0, species)
+                return thermo0
+            except:
+                logging.error("Error attempting to get thermo for species %s with structure \n%s", 
+                    species, species.molecule[0].to_adjacency_list() )
```
mazeau, line 1228: There's an extra space at the end here causing it to fail in the code standards

kblondal: Thank you! Fixed it.

rwest (approved):
This whole section of code for making gas-phase counterparts is not very satisfying at the moment (a growing list of exceptions), but until we properly deal with resonance structures of adsorbates, I think this is the best solution. Your pull request is definitely a net improvement, even if we improve it again soon, so in my opinion this should be merged as is. Thanks @kblondal.

amarkpayne (approved):
Looks good to me, I'll merge this in right now

rwest:
Thanks Mark!

Issue #2030, opened 2020-10-14
Could improve diagnostic error messages to fix user/database errors


rwest:
### Motivation or Problem
Error messages like this one
![image](https://user-images.githubusercontent.com/93807/95997645-fdc16c00-0e01-11eb-9545-b33e3be1869a.png)
are not very helpful to fix the problem(s) in the database or input file (or RMG, if indeed that was the cause).

### Desired Solution
Error handling in various strategic places could, when handling an exception and about to crash, report things like

1. full adjacency list of the molecule
2. where it came from (if it was a library, for example)
3. which reaction created it (if it was made by a reaction)
4. where that reaction came from

rwest:
Also error messages like "Reaction ___ in kinetics library ___ was not balanced!" could be much worse (yay!), but could be even better, eg. could state what the imbalance is, and/or print the adjacency lists.

Tingchenlee:
There are also different molecules errors:

**1. If the reactant contained H2O:**
For reaction generation 1 process is used.
Generating thermo for new species...
Traceback (most recent call last):
  File "/Users/lee.ting/Code/RMG-Py/rmg.py", line 118, in <module>
    main()
  File "/Users/lee.ting/Code/RMG-Py/rmg.py", line 112, in main
    rmg.execute(**kwargs)
  File "/Users/lee.ting/Code/RMG-Py/rmgpy/rmg/main.py", line 922, in execute
    trimolecular_react=self.trimolecular_react)
  File "/Users/lee.ting/Code/RMG-Py/rmgpy/rmg/model.py", line 610, in enlarge
    self.apply_thermo_to_species(procnum)
  File "/Users/lee.ting/Code/RMG-Py/rmgpy/rmg/model.py", line 820, in apply_thermo_to_species
    self.generate_thermo(spc, rename=True)
  File "/Users/lee.ting/Code/RMG-Py/rmgpy/rmg/model.py", line 827, in generate_thermo
    submit(spc, self.solvent_name)
  File "/Users/lee.ting/Code/RMG-Py/rmgpy/thermo/thermoengine.py", line 175, in submit
    spc.thermo = evaluator(spc, solvent_name=solvent_name)
  File "/Users/lee.ting/Code/RMG-Py/rmgpy/thermo/thermoengine.py", line 160, in evaluator
    thermo = generate_thermo_data(spc, solvent_name=solvent_name)
[...]

PR #2035, opened 2020-10-22
Improve debugging error messages

rwest:
### Motivation or Problem
As described in #2030, the error messages and strack traces are not always super helpful in finding and fixing the cause of the problem. This PR attempts to add a few more diagnostics.

### Description of Changes
The main approach is to catch an exception as it passes through a section of code that knows something helpful, log a diagnostic error message, and then re-raise the exception so that it carries on as before. 
For example, if you get a bad atom type when trying to estimate adsorption thermo for a species that was included in a reaction library, then it's helpful to know which species, which reaction library, as well as which atom.  But the atom typing code doesn't know this info when it raises the exception, so we have to catch and re-raise it a few times. 

### Testing
I ran the unit tests. Then added a unit test that causes several of the errors that trigger the new debugging messages. All tests now pass.

### Reviewer Tips
As this code is only supposed to work when something goes wrong, testing it is interesting.
I hope that @tignchenlee can try this branch on her broken reaction libraries, input files, etc. (before she figured out how to fix them) and see if the new error messages work and would have helped. Perhaps @katrinblondal and @mazeau have some input files with recent versions of databases etc. (or before the bidentate problems were fixed) that cause exceptions too.
_Update: I added unit tests._

### Copy this PR! (or add commits)
If you have recently come across a bug (or found one in the issue tracker) feel free to take this approach (catch an exception, log something helpful, and raise it again) and apply it elsewhere. This could be just the start. The focus was on thermo problems.

### DIFF
--- a/rmgpy/rmg/model.py
+++ b/rmgpy/rmg/model.py
@@ -418,8 +418,12 @@ class CoreEdgeReactionModel:
         """
 
         # Determine the proper species objects for all reactants and products
-        reactants = [self.make_new_species(reactant, generate_thermo=generate_thermo)[0] for reactant in forward.reactants]
-        products = [self.make_new_species(product, generate_thermo=generate_thermo)[0] for product in forward.products]
+        try:
+            reactants = [self.make_new_species(reactant, generate_thermo=generate_thermo)[0] for reactant in forward.reactants]
+            products = [self.make_new_species(product, generate_thermo=generate_thermo)[0] for product in forward.products]
+        except:
+            logging.error(f"Error when making species in reaction {forward:s} from {forward.family:s}")
+            raise
         if forward.specific_collider is not None:
             forward.specific_collider = self.make_new_species(forward.specific_collider)[0]
 
--- a/rmgpy/data/kinetics/database.py
+++ b/rmgpy/data/kinetics/database.py
@@ -562,6 +562,8 @@ and immediately used in input files without any additional changes.
                 except:
                     logging.error("Problem family: {}".format(label))
                     logging.error("Problem reactants: {}".format(molecules))
+                    for m in molecules:
+                        logging.error(f"{m}\n{m.to_adjacency_list()}")
                     raise
 
         for reactant in molecules:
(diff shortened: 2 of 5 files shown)


mazeau (approved):
This looks great. Is there a reason for switching the `.format` to `f` strings?

rwest:
> Is there a reason for switching the `.format` to `f` strings?

f-strings are easier to read, easier to debug, and faster to execute. Generally better.
https://realpython.com/python-f-strings/

rmgpy/rmg/model.py
```diff
@@ -418,8 +418,12 @@ def make_new_reaction(self, forward, check_existing=True, generate_thermo=True):
         """
 
         # Determine the proper species objects for all reactants and products
-        reactants = [self.make_new_species(reactant, generate_thermo=generate_thermo)[0] for reactant in forward.reactants]
-        products = [self.make_new_species(product, generate_thermo=generate_thermo)[0] for product in forward.products]
+        try:
+            reactants = [self.make_new_species(reactant, generate_thermo=generate_thermo)[0] for reactant in forward.reactants]
+            products = [self.make_new_species(product, generate_thermo=generate_thermo)[0] for product in forward.products]
+        except:
+            logging.error("Error when making species in reaction {forward:s} from {forward.family:s}")
```
rwest, line 425:
```suggestion
            logging.error(f"Error when making species in reaction {forward:s} from {forward.family:s}")
```

Shall I pretend I left this mistake in to check our code review process?

mazeau: Ah, I should have caught this! Sorry

mazeau (approved):
Ok, I see no issues with merging this in

rwest:
Yes! I got the code coverage test to pass (even though this is updating error-handling code that therefore shouldn't normally be run). OK, now all tests pass, this could be reviewed and merged?  If folks want to add more debugging messages of this type, then can be in a new PR.

mjohnson541 (dismissed):
This looks good to me, once it passes RMG-tests should be good to merge.

rwest:
The previous commit passed all tests. This one has stalled for some reason, and I can't see how to trigger it again.

rwest:
@mjohnson541 said good to merge once tests pass. The only change since then was a pretend change (tweak a commit message) to trigger the build scripts again, which had stalled. Now they have completed. All passed, so I merged.

Issue #2039, opened 2020-11-09
Problem in Surface_Bidentate_Dissociation causing charged N adsorbate 


Tingchenlee:
### Bug Description
This minimal input file below leads to this crash caused by the reaction family Surface_Bidentate_Dissociation:
```
For reaction generation 1 process is used.
Error: Could not update atomtypes for this molecule:
multiplicity -187
1 *4 X u0 p0 c0 {2,S}
2 *1 N u0 p1 c-1 {1,S} {3,T}
3 *2 N u0 p1 c-1 {2,T} {4,S}
4 *3 H u0 p0 c0 {3,S}

Error: Problem family: Surface_Bidentate_Dissociation
Error: Problem reactants: (Molecule(smiles="[Pt]"), Molecule(smiles="[Pt]N=N[Pt]"))
Error: <Molecule "[Pt]">
1 *3 H u0 p0 c0 {2,S}
2 *6 X u0 p0 c0 {1,S}

Error: <Molecule "[Pt]N=N[Pt]">
1 *1 N u0 p1 c0 {2,D} {3,S}
2 *2 N u0 p1 c0 {1,D} {4,S}
3 *4 X u0 p0 c0 {1,S}
4 *5 X u0 p0 c0 {2,S}

Traceback (most recent call last):
  File "/Users/lee.ting/Code/RMG-Py/rmg.py", line 118, in <module>
    main()
  File "/Users/lee.ting/Code/RMG-Py/rmg.py", line 112, in main
    rmg.execute(**kwargs)
  File "/Users/lee.ting/Code/RMG-Py/rmgpy/rmg/main.py", line 922, in execute
    trimolecular_react=self.trimolecular_react)
  File "/Users/lee.ting/Code/RMG-Py/rmgpy/rmg/model.py", line 599, in enlarge
    procnum=procnum)
  File "/Users/lee.ting/Code/RMG-Py/rmgpy/rmg/react.py", line 172, in react_all
    return react(spc_fam_tuples, procnum), [fam_tuple[0] for fam_tuple in spc_fam_tuples]
  File "/Users/lee.ting/Code/RMG-Py/rmgpy/rmg/react.py", line 66, in react
    reactions = list(map(_react_species_star, spc_fam_tuples))
  File "/Users/lee.ting/Code/RMG-Py/rmgpy/rmg/react.py", line 79, in _react_species_star
    return react_species(*args)
[...]

Issue #2038, opened 2020-11-09
Problem in Surface_Adsorption_Abstraction_vdW causing charged N adsorbate


rwest:
### Bug Description

This minimal input file below leads to this crash caused by the reaction family Surface_Adsorption_Abstraction_vdW:

```
Generating initial reactions...
For reaction generation 1 process is used.
Error: Could not update atomtypes for this molecule:
multiplicity -187
1 *4 N u0 p1 c+1 {2,D}
2 *6 X u0 p0 c0 {1,D}

Error: Problem family: Surface_Adsorption_Abstraction_vdW
Error: Problem reactants: (Molecule(smiles="N#N.[Pt]"), Molecule(smiles="[Pt]N=N"))
Error: <Molecule "N#N.[Pt]">
1 *2 N u0 p1 c0 {2,T}
2 *3 N u0 p1 c0 {1,T}
3 *1 X u0 p0 c0

Error: <Molecule "[Pt]N=N">
1 *4 N u0 p1 c0 {2,D} {4,S}
2 *5 N u0 p1 c0 {1,D} {3,S}
3    H u0 p0 c0 {2,S}
4 *6 X u0 p0 c0 {1,S}

Traceback (most recent call last):
  File "/Users/rwest/Code/Cat/RMG-Py/rmg.py", line 118, in <module>
    main()
  File "/Users/rwest/Code/Cat/RMG-Py/rmg.py", line 112, in main
    rmg.execute(**kwargs)
  File "/Users/rwest/Code/Cat/RMG-Py/rmgpy/rmg/main.py", line 732, in execute
    trimolecular_react=self.trimolecular_react)
  File "/Users/rwest/Code/Cat/RMG-Py/rmgpy/rmg/model.py", line 599, in enlarge
    procnum=procnum)
  File "/Users/rwest/Code/Cat/RMG-Py/rmgpy/rmg/react.py", line 172, in react_all
    return react(spc_fam_tuples, procnum), [fam_tuple[0] for fam_tuple in spc_fam_tuples]
  File "/Users/rwest/Code/Cat/RMG-Py/rmgpy/rmg/react.py", line 66, in react
    reactions = list(map(_react_species_star, spc_fam_tuples))
  File "/Users/rwest/Code/Cat/RMG-Py/rmgpy/rmg/react.py", line 79, in _react_species_star
    return react_species(*args)
[...]

PR #2041, opened 2020-11-12
Test reaction family recipes when testing database

rwest:
### Motivation or Problem
There were recently some problems detected in reaction family recipes and templates (eg. #2038, #2039), and I wanted to see if these could have been found earlier with better testing of the database.

### Description of Changes
The new test, which runs on every reaction family, uses the machinery of the existing test for descending the tree, to create a sample molecule from the group definition of every node. It then takes these sample molecules, and tries reacting them according to the reaction recipe. It doesn't do every combination (I thought that might be too many) but reacts each molecule at least once. It reports any errors, from things like charged species, atoms that fail atom-type generation, etc. 

### Testing
The new test fails (correctly) on the master database at commit 2aae843, because it detects the problem causing #2039 in Surface_Bidentate_Dissociation, which was fixed by RMG-database PR https://github.com/ReactionMechanismGenerator/RMG-database/pull/445.
With that PR (now merged) it passes.

Unfortunately it was not able to detect the problem causing #2038 in Surface_Adsorption_Abstraction_vdW, which will be fixed by RMG-database PR https://github.com/ReactionMechanismGenerator/RMG-database/pull/446 because the test always uses the simplest possible sample molecule, and that bug only shows up with a more complex reactant.  I can't think of a way to test every possible reactant in less than infinite time.

So this new test doesn't guarantee to catch all recipe bugs, but it catches some, and is not too horribly slow.

### Note
~Don't merge until after https://github.com/ReactionMechanismGenerator/RMG-database/pull/445~ (it's now merged)

### DIFF
--- a/testing/databaseTest.py
+++ b/testing/databaseTest.py
@@ -1300,6 +1300,7 @@ Origin Group AdjList:
         So unfortunately, problems can still slip through this test.
         """
         family = self.database.kinetics.families[family_name]
+        test1 = []
 
         # ignore any products
         ignore = []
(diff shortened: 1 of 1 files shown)


rwest:
This is now ready for review (now that the database has been fixed and the tests pass). The github algorithm is suggesting @mjohnson541 so I'm requesting a review, but feel free to re-assign!

rwest:
On latest rebase, on my 2015 iMac (i7) it adds  41 seconds, or 19%, on to the database tests.
Without this extra test: `Ran 1394 tests in 218.115s` 
With this extra test: `Ran 1477 tests in 259.294s`

mjohnson541 (approved):
This looks fine, I don't think the runtime difference is a big deal given the current tree sizes, but I wonder how the automatically generated trees that have many more nodes will affect this. It's conceivable this could become the dominant operation in the database tests when H_Abstraction switches over to automatic generation.

PR #2054, opened 2020-12-22
Species label is wrongly being assinged the multiplicity string

alongd:
### Motivation or Problem
The problem is that when a `Species` is defined `from_adjacency_list()` and the adjlist has the multiplicity in the first line, the species label is being assigned with the multiplicity values, e.g. , `spc.label = 'multiplicity 2'`.

### Description of Changes
Now checking that the keyword 'multiplicity' is not in the label being assigned from the adjlist.

### Testing
A simple test was added.

### DIFF
--- a/rmgpy/species.py
+++ b/rmgpy/species.py
@@ -356,7 +356,7 @@ class Species(object):
                 break
         else:
             label = ''
-        if len(label.split()) > 0 and not label.split()[0].isdigit():
+        if len(label.split()) > 0 and not label.split()[0].isdigit() and 'multiplicity' not in label:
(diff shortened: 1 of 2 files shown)


amarkpayne (changes requested):
Thanks for the catch @alongd ! Can you also add a quick test to make sure that it can get a label from the adjacency list when the multiplicity is also given? Just add another species for this in the unit test

alongd:
@amarkpayne, thanks for the quick feedback!
I added another test as you suggested.

amarkpayne (approved):

PR #2056, opened 2020-12-27
Use adjlist as the species structure instead of SMILES in an Arkane PDep network file

alongd:
### Motivation or Problem
Species structures saved in Arkane network files is given in SMILES, which sometimes can be ambiguous. The case I came across was CH2(S) which has the same SMILES as CH2(T), and reading it automatically by T3 failed. Of course in this particular case the multiplicity, which is also given, could be used to differentiate between these states, but we could think of other cases where adjacency list is superior to SMILES in terms of ambiguity (e.g., smiles for hyper-valence species is at times interpreted wrong, whereas adjlists are usually/always better.).

### Description of Changes
Added an attribute to RMG passed to `CoreEdgeReactionModel` and then to `PDepNetwork` for requesting an adjlist structure form in Arkane network files instead of SMILES.

### DIFF
--- a/arkane/pdep.py
+++ b/arkane/pdep.py
@@ -665,7 +665,7 @@ class PressureDependenceJob(object):
                 f.write('species(\n')
                 f.write('    label = {0!r},\n'.format(str(spec)))
                 if len(spec.molecule) > 0:
-                    f.write('    structure = SMILES({0!r}),\n'.format(spec.molecule[0].to_smiles()))
+                    f.write(f'    structure = adjacencyList("""{spec.molecule[0].to_adjacency_list()}"""),\n')
                 if spec.conformer is not None:
                     if spec.conformer.E0 is not None:
                         f.write('    E0 = {0!r},\n'.format(spec.conformer.E0))


kspieks (approved):
Thanks for the PR. I like that you preserved the original behavior. This looks good to me

arkane/pdep.py
```diff
@@ -641,9 +641,11 @@ def draw(self, output_directory, file_format='pdf'):
 
         NetworkDrawer().draw(self.network, file_format=file_format, path=path)
 
-    def save_input_file(self, path):
+    def save_input_file(self, path, use_adj_list=False):
```
mjohnson541, line 644: Is there a reason we need to have this be an option. What's the downside to just making it always be an adjlist?

mjohnson541 (commented):
This broadly looks fine, but Arkane input files are our own format, so I would think there wouldn't be any reason to not always use adjlists?

alongd:
I agree we should simply use adjlists (always) instead of SMILES. This extremely simplifies this PR.
I tested the updated functionality by running an RMG PDep job that generates a bunch of Arkane network files, and using one of those files to run an Arkane job (so reading in the adjlist structures). Works smoothly.
Thanks!

mjohnson541 (approved):

PR #2066, opened 2021-02-03
Updated the amu values of common elements in gas phase chemistry

alongd:
### Motivation or Problem
Our element mass value for Ar is outdated.
The element mass is used in Arkane to compute the molar mass of a species, affecting calculated pressure-dependent rates.

### Description of Changes
The element mass for Ar as well as other common elements was updated.
Data taken from IUPAC (https://applets.kcvs.ca/IPTEI/IPTEI.html)
Element isotope composition was averaged when needed, comments were added.
For most other elements (not Ar), the change was not significant. I used the significant digits from the IUPAC source.

### DIFF
--- a/rmgpy/molecule/element.py
+++ b/rmgpy/molecule/element.py
@@ -179,32 +179,32 @@ X = Element(0,    'X', 'surface_site'   , 0.0)
 
 # Period 1
 # Hydrogen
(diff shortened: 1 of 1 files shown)


alongd:
@oscarwumit or @kfir4444, could you take a look at this small PR?

kfir4444 (approved):

PR #2072, opened 2021-02-12
Create CI.yml

mjohnson541:
Github actions for running unittests.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -6,30 +6,31 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       max-parallel: 5
+    defaults:
+      run:
+        shell: bash -l {0}
 
     steps:
-    - uses: actions/checkout@v2
-    - name: Set up Python 3.7
-      uses: actions/setup-python@v2
-      with:
-        python-version: 3.7
-    - name: Add conda to system path
-      run: |
-        # $CONDA is an environment variable pointing to the root of the miniconda directory
-        echo $CONDA/bin >> $GITHUB_PATH
-    - name: Install dependencies run tests
-      run: |
-        cd ..
-        git clone https://github.com/ReactionMechanismGenerator/RMG-database.git
-        cd RMG-Py
-        conda env create -q -f environment.yml
-        source activate rmg_env
-        conda install -y -c conda-forge codecov
-        conda list
-        yes 'Yes' | /usr/share/miniconda/envs/rmg_env/bin/mopac $MOPACKEY > /dev/null
-        git clone -b arkanepy3 https://github.com/mjohnson541/Q2DTor.git external/Q2DTor
-        make
-        make test-unittests
-        make test-functional
-        make test-database
-        codecov
+      - uses: actions/checkout@v2
+      - uses: conda-incubator/setup-miniconda@v2
+        with:
+          environment-file: environment.yml
+          python-version: 3.7
+          activate-environment: rmg_env
+      - name: Conda info
+        run: |
+          conda info
+          conda list
+      - name: Compile and test
+        run: |
+          cd ..
+          git clone https://github.com/ReactionMechanismGenerator/RMG-database.git
+          cd RMG-Py
+          conda install -y -c conda-forge codecov
+          yes 'Yes' | ${CONDA_PREFIX}/bin/mopac $MOPACKEY > /dev/null
+          git clone -b arkanepy3 https://github.com/mjohnson541/Q2DTor.git external/Q2DTor
+          make
+          make test-unittests
+          make test-functional
+          make test-database
+          codecov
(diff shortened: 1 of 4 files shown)


.github/workflows/CI.yml
```diff
@@ -0,0 +1,36 @@
...
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
+    - name: Add conda to system path
+      run: |
+        # $CONDA is an environment variable pointing to the root of the miniconda directory
+        echo $CONDA/bin >> $GITHUB_PATH
+    - name: Install dependencies run tests
```
skrsna, line 21: May I suggest conda incubator action! https://github.com/marketplace/actions/setup-miniconda

rwest: can you elaborate? or demonstrate?

skrsna: https://github.com/conda-incubator/setup-miniconda/blob/master/action.yml 
This elaborates more on the options and things you can do with miniconda action. In general, this action helps us setup a conda env more efficiently.

rwest: can you push a commit(s) to this PR that makes the changes you suggest?.

rwest:
Would be nice to split the "install dependencies" part from the "run tests" part(s) if possible (for the sake of looking at the progress, results, and log)

.github/workflows/CI.yml
skrsna: I have a question about MOPAC, is it needed for tests and how is the env variable setup? is it a secret? I updated the CI with miniconda setup. Takes about [~5 min](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2072/checks?check_run_id=2143247106)s to install conda env from the file

mliu49: Yeah, the mopac key has to be set up for the tests to run, otherwise they're skipped (like the gaussian tests). It was previously defined in .travis.yml.

Here are explanations for all of the secrets that the Travis build has, as listed in the build logs.

Secret environment variables which were configured via the Travis website:
```
# Old keys for triggering RMG-tests, no longer used
$ export encrypted_244ac3091cff_key=[secure]
$ export encrypted_244ac3091cff_iv=[secure]
# GitHub access token for triggering RMG-tests, replaced above keys, set up from RMG Dev user account
$ export GH_TOKEN=[secure]
# Conda push token for automatically building RMG conda binaries and pushing to anaconda.org, set up from RMG organization
$ export CONDA_TOKEN=[secure]
```

Secret environment variables from [.travis.yml](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/.travis.yml#L3), should match the order they're listed in the file
```
# MOPAC key, I don't know what the original value is, but any valid key would work
$ export MOPACKEY=[secure]
# Thermo Central Database connection info and credentials - the server is no longer running, and the unit tests are skipped automatically, so these probably aren't needed anymore
$ export TCD_USER=[secure]  
$ export TCD_PW=[secure]
$ export TCD_HOST=[secure]
$ export TCD_PORT=[secure]
```

rwest:
Looking at the admin settings, I'm not seeing any "[secrets](https://docs.github.com/en/actions/reference/encrypted-secrets)" for github actions. Did you get as far as trying that, Matt? 
If not, that'd explain why the MOPAC install step is failing.

rwest:
I can't select @mjohnson541 as a reviewer because you created the pull request, but I think this is different enough now that you'd be an appropriate reviewer.

skrsna (approved):
Thanks, now that this is in we can think about [caching](https://github.com/marketplace/actions/setup-miniconda#caching) to speed up conda

.travis.yml
```diff
@@ -34,30 +33,6 @@ before_install:
 
 jobs:
   include:
-    - stage: test
```
mjohnson541, line 37: If you do this I believe RMG-tests won't trigger anymore which is why it isn't triggering on this PR

rwest: I've re-enabled the trigger.

rwest:
I've re-enabled the RMG-tests trigger, using Travis. If someone with access to @RMGdev account can put an authentication token in the secrets available to GithubActions then they/we could move the triggering over to Github Actions and remove Travis from the RMG-Py repository.  

Actually moving the RMG-tests over to Github Actions is another whole project.

rwest:
OK, I think this is ready to merge (after one more review by Matt?), and then switch the required-for-merge flag from the travis to the github actions check.

mjohnson541 (commented):
Looks good! Can you squash and rebase?

rwest:
I have now switched the branch protection rules for the master branch to require the `build-and-test-linux` tests to pass, instead of the Travis ones.

PR #2075, opened 2021-02-14
Allow RMS to output surface mechanisms and bugfixes

mjohnson541:
This PR fixes a bug where RMG can output mechanisms where two species have the same name and properly handles surface reactor phases.

### DIFF
--- a/rmgpy/yml.py
+++ b/rmgpy/yml.py
@@ -86,6 +86,8 @@ def obj_to_dict(obj, spcs, label="solvent"):
     if isinstance(obj, Species):
         result_dict["name"] = obj.label
         result_dict["type"] = "Species"
+        if obj.contains_surface_site():
+            result_dict["adjlist"] = obj.molecule[0].to_adjacency_list()
         result_dict["smiles"] = obj.molecule[0].to_smiles()
         result_dict["thermo"] = obj_to_dict(obj.thermo, spcs)
         result_dict["radicalelectrons"] = get_radicals(obj)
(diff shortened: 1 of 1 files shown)


rmgpy/yml.py
```diff
@@ -63,15 +63,37 @@ def write_yml(spcs, rxns, solvent=None, solvent_data=None, path="chem.yml"):
...
+            break
+    if not is_surface:
+        result_dict = dict()
+        result_dict["Units"] = dict()
+        result_dict["Phases"] = [dict()]
+        result_dict["Phases"][0]["name"] = "phase"
+        result_dict["Phases"][0]["Species"] = [obj_to_dict(x, spcs, names=names) for x in spcs]
+        result_dict["Reactions"] = [obj_to_dict(x, spcs, names=names) for x in rxns]
+        if solvent_data:
+            result_dict["Solvents"] = [obj_to_dict(solvent_data, spcs, names=names, label=solvent)]
```
davidfarinajr, line 84: Currently we can't do solvent corrections for rmg-cat, but this may be something we implement in the future.  In this case, perhaps we add the "Solvents" to `results_dict` regardless of whether `is_surface` is true or not.  If `is_surface` is True, then `solvent_data` should be None, so it's not added anyways, but if we implement solvent corrections for surface mechs in the future, then we should add it to yaml.

davidfarinajr (approved):
Looks good Matt! Thanks for adding surface to yaml format!  My only comment was regarding the solvent_data for surfaces. This is not a bug since we can't do solvent corrections for surface mechs yet, but could potentially be if we do implement this.  Since this change is not currently needed, I approve.  I think it would also be helpful to include the surface attributes for `X` from the MetalDB in the yaml, this is something I can work on adding, hopefully in the near future.

davidfarinajr:
And almost forgot, should we add the `surface_site_density` to the yaml?  This is needed to evaluate StickingCoefficient rate constants. It is specified in RMG input file or taken from the metalDB.

PR #2077, opened 2021-02-14
Change Chebyshev Functions to cpdef

mjohnson541:
Having them cdef is really inconvenient when writing tools and the operation is definitely not a performance bottleneck.


lily90502:
All the tests are passed. I think it's fine for this PR.

rwest (approved):

Issue #2079, opened 2021-02-18
Kinetics search on the RMG website can't find the reactions rates


Tingchenlee:
### Bug Description
Kinetics search on the RMG website can't calculate the reactions rates for some mechanisms, which look like this one:
![image](https://user-images.githubusercontent.com/56702445/108298101-5c10a980-716a-11eb-8757-0b3b6ae3fbc4.png)
While their reverse reactions rates can be found:
![image](https://user-images.githubusercontent.com/56702445/108298166-7ea2c280-716a-11eb-9ee6-4f67be189ff4.png)
### How to Reproduce
```
1 N u0 p1 c0 {2,S} {3,S} {4,S}
2 H u0 p0 c0 {1,S}
3 H u0 p0 c0 {1,S}
4 X u0 p0 c0 {1,S}
``` 
react with
```
1 N u0 p1 c0 {2,S} {3,S} {4,S}
2 H u0 p0 c0 {1,S}
3 H u0 p0 c0 {1,S}
4 X u0 p0 c0 {1,S}
```

### Expected Behavior
RMG can calculate the rates of these reactions

### Installation Information
   - RMG-website
  
### Additional Context
other examples:
 - N2H4_X ⇔ N2H4 + X , in Surface_Adsorption_vdW  
 - H2_X ⇔ H2 + X , in Surface_Adsorption_vdW  
 - H_X + NH2_X ⇔ NH3 + X + X , in Surface_Adsorption_Dissociative  
 - N2H_X ⇔ H_X + N2  , in Surface_EleyRideal_Addition_Multiple_Bond
 - X_O2_X ⇔ O2 + X + X , in Surface_Adsorption_Bidentate 
 - N_X + N_X ⇔ N2 + X + X , in Nitrogen

rwest:
Probably a problem reversing the rate from the sticking coefficient. In order to calculate the reverse rate, from a sticking coefficient, it will need to assume a surface site density. Which might not be specified in the context of a generic web search.

rwest:
Perhaps this is OK to not calculate the reverse, and just tell users to use it in the forward (sticking) direction. 

But in that case the message displayed to users on the website could be clearer.

xiaoruiDong:
Sorry, I just saw your reply, @rwest.

Some of my investigations are included in 
https://github.com/ReactionMechanismGenerator/RMG-website/issues/212#issuecomment-781636694

[...]

PR #2080, opened 2021-02-19
Reverse kinetics for sticking coefficient

davidfarinajr:
### Motivation or Problem
RMG had no method to reverse Sticking Coefficient kinetics.  I believe this is related to this website issue https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2079.  Adding a method to reverse Sticking Coeffs could be useful to compare Sticking Coeff kinetics with literature rates that are expressed as Arrhenius in the reverse direction.  This method could also be useful if we need to reverse sticking coefficient kinetics in training reactions.

### Description of Changes
Added StickingCoefficient kinetics handling to reaction `get_rate_coefficient` method so that we can use this method for all types of kinetics.
Added a method to reverse StickingCoefficient kinetics with default surface site density.

### Testing
Added a unit test for reversing StickingCoefficient kinetics

### DIFF
--- a/rmgpy/reaction.py
+++ b/rmgpy/reaction.py
@@ -53,9 +53,9 @@ import rmgpy.constants as constants
 from rmgpy.exceptions import ReactionError, KineticsError
 from rmgpy.kinetics import KineticsData, ArrheniusBM, ArrheniusEP, ThirdBody, Lindemann, Troe, Chebyshev, \
     PDepArrhenius, MultiArrhenius, MultiPDepArrhenius, get_rate_coefficient_units_from_reaction_order, \
-    StickingCoefficient, SurfaceArrheniusBEP, StickingCoefficientBEP
+    SurfaceArrheniusBEP, StickingCoefficientBEP
 from rmgpy.kinetics.arrhenius import Arrhenius  # Separate because we cimport from rmgpy.kinetics.arrhenius
-from rmgpy.kinetics.surface import SurfaceArrhenius  # Separate because we cimport from rmgpy.kinetics.surface
+from rmgpy.kinetics.surface import SurfaceArrhenius, StickingCoefficient  # Separate because we cimport from rmgpy.kinetics.surface
 from rmgpy.kinetics.diffusionLimited import diffusion_limiter
 from rmgpy.molecule.element import Element, element_list
 from rmgpy.molecule.molecule import Molecule, Atom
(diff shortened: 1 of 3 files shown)


rmgpy/reaction.py
```diff
@@ -677,7 +679,7 @@ def get_rate_coefficient(self, T, P=0):
         else:
             return self.kinetics.get_rate_coefficient(T, P)
 
-    def get_surface_rate_coefficient(self, T, surface_site_density):
+    def get_surface_rate_coefficient(self, T, surface_site_density=2.5e-05):
```
rwest, line 682: I don't think this one is required, because it's set by the calling functions?  I'm not very happy with having a default, in general, as it increases the chance of someone forgetting to pass it when they should.  (There's no current bug, I'm just anticipating future bugs..)

rwest (changes requested):
Try to assume default values as little as possible.
In the context of the website, some surface site density must be assumed (which should be reported on the website), but otherwise we should have it specified in the reaction simulation, and should use that value. By putting default values in the method definitions, we risk forgetting to use the actual value.
But the fact that we didn't need these methods suggests we won't actually use them during a simulation, so it's perhaps a moot point. Although if we do end up using it, as you suggest in the PR comment, for reversing training reactions, then getting the surface site density from the metal attributes database would presumably be the right thing to do :shrug: Although perhaps doing it per site would be better than per square centimeter anyway. 
 
Anyway, in cases where you do need to use an assumed default, please make sure to specify it in the docstrings that this is happening.

davidfarinajr:
> Try to assume default values as little as possible.
> In the context of the website, some surface site density must be assumed (which should be reported on the website), but otherwise we should have it specified in the reaction simulation, and should use that value. By putting default values in the method definitions, we risk forgetting to use the actual value.
> But the fact that we didn't need these methods suggests we won't actually use them during a simulation, so it's perhaps a moot point. Although if we do end up using it, as you suggest in the PR comment, for reversing training reactions, then getting the surface site density from the metal attributes database would presumably be the right thing to do 🤷 Although perhaps doing it per site would be better than per square centimeter anyway.
> 
> Anyway, in cases where you do need to use an assumed default, please make sure to specify it in the docstrings that this is happening.

Thanks for your review Richard! Sounds good, I can take the default values out.  One of the reasons for the default is that for all of the kinetics types except Sticking Coeff in ‘generate_reverse_rate_constant’, we don’t need the surface site density (so I don’t want to force the user to specify one). In that case, maybe I should remove Sticking Coeff from that function and we should just call ‘ reverse_sticking_coeff_rate’.  It makes sense to have a separate function for this since it’s the only one dependent on surface site density. As for training reactions, I think most if not all of our families are in the surface adsorption direction (rather than the desorption direction) so we haven’t needed to reverse a sticking coeff yet, but if we do, I think it makes the most sense to use the surface densities in the Metal DB since we are also using atomic binding energies from there.

rwest:
Agree that requiring a parameter that is not used would be bad. Could do something like have it default to zero and raise an exception (ValueError?) if it's needed and is zero. But not complain if it's not needed.

mazeau (approved):
This looks good to me! I like richard's suggestion of not having a default value.

davidfarinajr:
With the last push, I added docstings and changed the default surface_site_density to 0 for methods that don't require a surface site density.  That way, the user doesn't need to provide it, but if it is needed in the function to calc Surface Coefficient rate constant, we will raise a value error if a nonzero (and non negative) surface site density is not provided.

rmgpy/reaction.py
```diff
@@ -845,11 +854,40 @@ def reverse_surface_arrhenius_rate(self, k_forward, reverse_units, Tmin=None, Tm
...
+        Reverses the given k_forward, which must be a StickingCoefficient type.
+        You must supply the correct units for the reverse rate.
+        The equilibrium constant is evaluated from the current reaction instance (self).
+        The surface_site_density in `mol/m^2` is used to evalaute the forward rate constant.
+        """
+        cython.declare(kf=StickingCoefficient, kr=SurfaceArrhenius)
+        cython.declare(Tlist=np.ndarray, klist=np.ndarray, i=cython.int)
+        kf = k_forward
+        if not isinstance(kf, StickingCoefficient): # Only reverse StickingCoefficient rates
+            raise TypeError(f'Expected a StickingCoefficient object for k_forward but received {kf}')
```
rwest, line 868: Probably this type check with a nice polite TypeError should check k_forward before we assign it to kf which has already been declared a StickingCoefficient in cython. Currently, if k_forward is not StickingCoefficient I guess it'll crash on line 866 and never reach 867 ?

rwest (changes requested):
Looking good!  thanks david. I just noticed one last thing.

rmgpy/reaction.py
davidfarinajr: ah yes, good catch.  I'll fix this

mazeau (dismissed):
It still looks good to me

Tingchenlee (approved):
I think I can understand this. Thanks, David! It looks good to me.

rwest (approved):

PR #2081, opened 2021-02-26
Add label swaps for own-reverse family Surface_Abstraction_single_vdW

rwest:
### Motivation or Problem
The new reaction family Surface_Abstraction_single_vdW in https://github.com/ReactionMechanismGenerator/RMG-database/pull/456
is its own reverse, so needs some hard-coding in RMG-Py to swap the atom labels.

Other families are also missing these label swaps.

### Description of Changes
I added label swaps for the new family.

I also added a check to see if other families are their own reverse and don't have this done, and quite a lot show up.
I turned the exception into a warning, but there are quite a few.
**_Should we fix these? Or should this be done as a database test?  Or do they really not matter and we should silence the warning?_**

### Testing
Database tests pass locally.

### Reviewer Tips
Some discussion is needed about the other families missing label swaps.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1528,6 +1528,9 @@ class KineticsFamily(Database):
                 atom_labels['*5'].label = '*1'
                 atom_labels['*2'].label = '*4'
                 atom_labels['*4'].label = '*2'
+            else:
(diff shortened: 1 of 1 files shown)


mazeau:
This has already been addressed so I'm going to close this PR

Issue #1864, opened 2020-01-14
cclib.parser error when using Gaussian for QMTP


rlawson:
General area which your question is related to.
 - [ ] Installation of RMG
 - [ ] Running an RMG job
 - [ ] Using RMG API
 - [ ] Arkane (formerly CanTherm)
 - [ ] Dependencies
 - [ ] An error message 

I'm trying to simulate a propane-air oxidation  using a simple reactor at 800K and 40 bar.

However, I encountered a problem when I used quantum mechanics for thermodynamic properties estimation which was "AttributeError: module 'cclib.parser' has no attribute 'Mopac'"

I have RMG installed on a macOS 10.15.2 
I installed using binary with anaconda
RMG version 3.0.0 with build py37he38e130_1 with the corresponding latest database.

Find attached my input file
[input.zip](https://github.com/ReactionMechanismGenerator/RMG-Py/files/4061913/input.zip)

amarkpayne:
This appears to be an issue with the on-the-fly quantum mechanics tool, QMTP. We had to update some of the code for the cclib parser when we switched to python 3, but it looks like there is an error that our unit tests didn't catch.

Looking at your input file though, it is very likely that you do not need to use the QMTP interface. In the past we relied on using this for polycyclics and fused ring species, which I don't think will arise in your system. Furthermore, I think that a lot of the improvements we have made over the past couple of years have really helped with this problem, including new algorithms for handling ring species and new groups, not to mention some of the machine learning strategies our group has implemented. I am tagging @cgrambow and @mjohnson541 who know more about these subjects than me in case they want to weigh in, but I think you should be fine if you remove the quantum mechanics block for your system.

That being said, we can continue to look at this bug in the meantime so the the functionality is there if you want/need it.

rlawson:
Thanks Max for the prompt response. I do agree with you the species of interest does not really require the use of QMTP.  After installing v 3.0.0, this was one of the simplest cases to just check if all the modules we'll be using in future runs were ok. I plan on simulating some cyclic structures which in the past have given good results only when QMTP interface is used, that's why I was interested in using the quantum mechanics block.

lynnrinkui:
Any update on this issue? I'm simulating a cyclic chemical combustion and need to use MOPAC to get more accurate estimates of thermo properties. Now I found out that this bug has been around for more than a year......

xrulesin:
[...]

PR #2088, opened 2021-03-15
Switch to OpenBabel 3

rwest:
### Motivation or Problem
The release notes say OpenBabel 3 a major improvement
https://github.com/openbabel/openbabel/releases/tag/openbabel-3-0-0
but it does break some things in the API
https://open-babel.readthedocs.io/en/latest/UseTheLibrary/migration.html#migrating-to-3-0
so subsequent commits on this pull request should fix RMG to be compatible.

This pull request is currently a stub. Please feel free to contribute commits!

### Description of Changes
- update the environment.yaml to use the official updated cclib and openbabel conda packages.
- change import statements
- when reading radicals from OpenBabel molecules, calculate spin multiplicity based on undervalence of atoms, because OpenBabel no longer does this for you
- Mark a couple of broken unit tests as work_in_progress, because they seem to have malformed InChIs that we can no longer read, and I'm not sure what the correct molecule is supposed to be.

### Testing
- Tests run locally on my Mac.
- Continuous Integration tests will be run.

### Reviewer Tips
- if someone can figure out the broken InChIs in the unit tests, that'd be great.

Further detailed commentary is in the individual commit messages.

Issue #1875, opened 2020-01-27
Cclib/Gaussian tests not passing


mazeau:
In the most recent version of rmg master, I get these gaussian/cclib errors when running tests on Discovery.  I have everything up to date with the most recent master commits.

```
======================================================================
ERROR: Test that generate_thermo_data() works correctly on gaussian PM3.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/scratch/westgroup/mazeau/Cat/RMG-Py/rmgpy/qm/gaussianTest.py", line 78, in test_generate_thermo_data
    self.qmmol1.generate_thermo_data()
  File "/scratch/westgroup/mazeau/Cat/RMG-Py/rmgpy/qm/molecule.py", line 359, in generate_thermo_data
    self.qm_data = self.generate_qm_data()
  File "/scratch/westgroup/mazeau/Cat/RMG-Py/rmgpy/qm/gaussian.py", line 265, in generate_qm_data
    success = self.run()
  File "/scratch/westgroup/mazeau/Cat/RMG-Py/rmgpy/qm/gaussian.py", line 100, in run
    return self.verify_output_file()
  File "/scratch/westgroup/mazeau/Cat/RMG-Py/rmgpy/qm/gaussian.py", line 170, in verify_output_file
    qm_data = self.parse()
  File "/scratch/westgroup/mazeau/Cat/RMG-Py/rmgpy/qm/molecule.py", line 334, in parse
    cclib_data = parser.parse()
  File "/home/mazeau.e/.conda/envs/rmg_env/lib/python3.7/site-packages/cclib/parser/logfileparser.py", line 363, in parse
    data = self.datatype(attributes=self.__dict__)
  File "/home/mazeau.e/.conda/envs/rmg_env/lib/python3.7/site-packages/cclib/parser/data.py", line 412, in __init__
    super(ccData_optdone_bool, self).__init__(*args, **kwargs)
  File "/home/mazeau.e/.conda/envs/rmg_env/lib/python3.7/site-packages/cclib/parser/data.py", line 203, in __init__
    self.setattributes(attributes)
  File "/home/mazeau.e/.conda/envs/rmg_env/lib/python3.7/site-packages/cclib/parser/data.py", line 416, in setattributes
    invalid = super(ccData_optdone_bool, self).setattributes(*args, **kwargs)
[...]

Issue #756, opened 2016-08-23
We should switch to the new CCLib


rwest:
CCLib have [just released](https://github.com/cclib/cclib/releases) version 1.5 which includes an updated MOPAC parser (based somewhat on our code, it seems: https://github.com/cclib/cclib/pull/257 )

If we can switch to using the official CCLib instead of our own fork, that would be great. Would avoid issues like https://github.com/ReactionMechanismGenerator/RMG-Py/issues/177. 
And if it's not good enough for us, then there is now a clear upstream place to send our improvements :)

mliu49:
We are a bit closer. During the Python 3 transition, @amarkpayne reconciled most of our changes with the official cclib v1.6. Those changes are located in his fork [amarkpayne/cclib](https://github.com/amarkpayne/cclib). However, I think he did not merge the MOPAC parsers, and both our version and the official version exist side-by-side (they were named differently). This is the version which was used to create the 1.6.1.rmg conda build of cclib currently in the RMG channel.

Ideally, we should identify the remaining differences and pull request to the official version.

rwest:
Would anyone  like to pick up the torch and help us switch for good?  (Or else update the "RMG" version and push a suite of new conda binaries).  I think the need to use this old conda build of cclib is holding up https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2088

PR #2091, opened 2021-03-18
Switch to newer cclib?

rwest:
### Motivation or Problem
As mentioned in https://github.com/ReactionMechanismGenerator/RMG-Py/issues/756 it would be nice to use the official [cclib](https://github.com/cclib/cclib) rather than a fork that we need (but fail) to maintain. For example the current fork has some bugs like #1864 #1875, and is incompatible with OpenBabel 3+, delaying #2088, and we don't even have [conda binaries](https://anaconda.org/rmg/cclib) for Mac OS X, meaning Mopac etc. has I think been broken for Mac OS users for ages.

### Description of Changes
Currently a work in progress, but the aim is to use the official cclib.

- switch the environment.yml to specify the conda-forge channel
- change the name of the MOPAC parser
- calculate molar mass if it's not parsed
- parse the molar mass from MOPAC files

### Testing
Running unit tests.
At one point an entropy calculated from MOPAC result changed 376.91 != 334.50 J/mol/K but Mark fixed the units in the rotational constants.

There may be other things that don't have good test coverage though. 
Is Gaussian parsing tested?

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -7,7 +7,7 @@ dependencies:
   - cairo
   - cairocffi
   - rmg::cantera >=2.3.0
-  - cclib
+  - conda-forge::cclib >=1.6.3
   - rmg::chemprop
   - coolprop
   - coverage
--- a/rmgpy/qm/molecule.py
+++ b/rmgpy/qm/molecule.py
@@ -331,8 +331,12 @@ class QMMolecule(object):
         """
         parser = self.get_parser(self.output_file_path)
         parser.logger.setLevel(logging.ERROR)  # cf. http://cclib.sourceforge.net/wiki/index.php/Using_cclib#Additional_information
+        parser.rotcons = [] # give it an attribute and it won't delete it, leaving it on the parser object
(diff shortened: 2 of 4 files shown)


amarkpayne:
@rwest and @mjohnson541 all of the tests passed on my local machine after adding one small fix, so this PR should be good to go. The change was that the new version of cclib returns rotcons in cm^-1, not GHz.

amarkpayne:
fixes #1864

rwest:
Yay, well done and thanks for figuring that out.

amarkpayne:
> Yay, well done and thanks for figuring that out.

Thanks for taking it 95% of the way there!

rwest:
So all the tests pass, but do you think we need any of the untested stuff? Or shall we just merge and fix stuff later if anyone complains?

amarkpayne:
My vote is that we go ahead and merge. I think it is better to have this change than to wait for us to add tests. If we are going to find remaining bugs it will happen when people use the feature

PR #2098, opened 2021-04-08
halogenated hydrocarbon Tb corrections

davidfarinajr:
added halogen corrections to boiling point when estimating critical properties using group additivity 

source:https://doi.org/10.1021/ie00008a029

### DIFF
--- a/rmgpy/data/transport.py
+++ b/rmgpy/data/transport.py
@@ -445,6 +445,12 @@ class TransportDatabase(object):
         
         Radicals are saturated with H atoms and the parent molecule properties
         are returned.
+
+        For halogenated hydrocarbons, a boiling point `Tb` correction is applied from the following source:
+        Sukumar Devotta and V. Rao Pendyala, 
+        "Modified Joback group contribution method for normal boiling point of aliphatic halogenated compounds"
+        Industrial & Engineering Chemistry Research 1992 31 (8), 2042-2046
+        DOI: 10.1021/ie00008a029
         """
         # For transport estimation we need the atoms to already be sorted because we
         # iterate over them; if the order changes during the iteration then we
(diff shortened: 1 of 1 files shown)


davidfarinajr:
This PR goes with https://github.com/ReactionMechanismGenerator/RMG-database/pull/466

ChrisBNEU (dismissed):
I checked that this worked with differently halogenated/fluorinated versions of 2,2-dimethylpropane, and it added the correct Tb values.

rwest (commented):
I feel like this should be documented somewhere, and the authors of that paper acknowledged. I couldn't find a good spot in the documentation though. Anyone have any ideas?   Bare minimum I guess is the docstring of the method you changed, but that's not very visible.   Should we acknowledge them in the longDesc of the estimated parameters?  Also, did you need to update the F group or anything else based on that paper?

davidfarinajr:
> I feel like this should be documented somewhere, and the authors of that paper acknowledged. I couldn't find a good spot in the documentation though. Anyone have any ideas? Bare minimum I guess is the docstring of the method you changed, but that's not very visible. Should we acknowledge them in the longDesc of the estimated parameters? Also, did you need to update the F group or anything else based on that paper?

Yes, I updated the F group on the database PR  ReactionMechanismGenerator/RMG-database#466, which has a slightly different `Tb` from this paper (6.31) than the Joback F group has (−0.03).  I can add citation to docstring.  Unfortunately there is not much transport documentation on the website.  I searched for `Joback` and got 1 result in `3. Creating Input Files`:

`sigma and epsilon are Lennard-Jones parameters, which can be estimated using the Joback method on the RMG website.`

Maybe I could add it there? And the paper will ultimately be cited when we write an RMG-halogens paper down the road

rwest:
Agreed - there's barely a stub to expand. For now I'd say just add a few lines to the docstring of `estimate_critical_properties_via_group_additivity` method. Mention the original Joback and the new paper.  I agree proper reference will be made in your upcoming paper 😀

PR #2103, opened 2021-04-15
Add species filter to output html

### LINKED ISSUE #1904: Maybe adding species filters to the output htmls
### Motivation or Problem
In extremely large models (especially when we're looking at the edge output), it would be nice to have a filter checkbox for species type (say, display only surface species or gas species) like we have for reaction families.  It would be even nicer to be able to filter species by whether they have radicals or not.  
@kblondal and I were talking about how this might make our lives a bit easier

### Desired Solution
Probably some tweaking in `rmgpy/rmg/output.py`

### Potential Alternatives
This feature could be deemed unnecessary in the majority of cases, and we could just leave it as is.

bjkreitz:
### Motivation or Problem
fixes #1904 

### Description of Changes
WIP

### Testing
WIP

### Reviewer Tips
WIP

### DIFF
--- a/rmgpy/rmg/output.py
+++ b/rmgpy/rmg/output.py
@@ -4,7 +4,7 @@
 #                                                                             #
 # RMG - Reaction Mechanism Generator                                          #
 #                                                                             #
-# Copyright (c) 2002-2020 Prof. William H. Green (whgreen@mit.edu),           #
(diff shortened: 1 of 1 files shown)


alongd:
was updated since the PR was opened, closing

PR #2107, opened 2021-04-16
v3.1.0

mjohnson541:


rwest:
Do you think we can get the openbabel3 upgrade in?

rwest:
Saw your note about the trouble building the Mac OS binaries on Github Actions. I haven't dug into this at all (and probably can't for at least a few days) but I had previously noticed that https://github.com/ReactionMechanismGenerator/RMG-Py/blob/v3.1.0/.conda/conda_build_config.yaml mentions "Travis" and things may be different on Github Actions.

rwest:
I'm not sure what should be done with this PR now, if it should be rebased (probably not, as then it'll include stuff added after the release), merged, merged to stable, closed, or what. But in any case, I cherry-picked to it the commits that fix up the release note formatting.

mjohnson541:
This was just for running tests easily, I merged stable back into master as Max's directions said to, so there should be no need for this PR.

PR #2110, opened 2021-04-20
Updating documentation and input file using metal database

mazeau:
This PR updates the surface documentation with instructions on how to use the metal database to specify certain metals instead of having to manually enter in binding energies and surface site density.

I also changed the ch4_o2 example input file to call `Pt111` in place of listing out Pt111 binding energies and surface site density

### DIFF
--- a/documentation/source/users/rmg/surfaces.rst
+++ b/documentation/source/users/rmg/surfaces.rst
@@ -24,6 +24,7 @@ This block should also contain the reference adatom binding energies
 
 One can also specify a specific metal of interest, and RMG will load corresponding binding energies from the
 metal database by using ``metal``.
+The full database of available metals can be found in ``input/surface/libraries/metal.py``.
 
 
 Here is an example catalyst properties block for Pt(111)::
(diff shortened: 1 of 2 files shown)


ChrisBNEU (approved):
This looks fine to me.

documentation/source/users/rmg/surfaces.rst
```diff
@@ -22,7 +22,16 @@ It varies depending on the catalyst in question, but is held constant across sim
 This block should also contain the reference adatom binding energies 
 (see linear scaling section below).
 
-Here is an example catalyst properties block for Pt(111)::
+One can also specify a specific metal of interest, and RMG will load corresponding binding energies from the
+metal database by using ``metal``.
+
+
+Here is an example catalyst properties block for Rh(111)::
```
rwest, line 29:
```suggestion
Here is an example catalyst properties block for Pt(111)::
```

rwest (changes requested):
Thanks. Would be good to tell people where to find the list of available metals and facets. 
Also there's a typo (Rh instead of Pt).

davidfarinajr:
Yep, looks good and I agree with Richard; we should tell people where to find available metals/facets in docs.  Perhaps we could have a `Metal Search` or something on the website so people can look it up and see what atomic binding energies would be used for metal of interest.

mazeau:
> Yep, looks good and I agree with Richard; we should tell people where to find available metals/facets in docs. Perhaps we could have a `Metal Search` or something on the website so people can look it up and see what atomic binding energies would be used for metal of interest.

I like the website idea! But I think that should be a separate PR. I somehow missed the part about telling people where to find the database, so I will add that in

davidfarinajr (approved):
Thanks @mazeu.  Looks good to me

rwest (approved):
Looks good now, and the example runs fine. Thanks.
I fixed some whitespace on rebasing, and will now merge

PR #2120, opened 2021-05-04
Halogens thermo - missing group test

davidfarinajr:
simultaneous with https://github.com/ReactionMechanismGenerator/RMG-database/pull/473

This PR changes the molecule to use for missing thermo group test since the thermo group for the old molecule is no longer missing (so the test was failing)


rwest (dismissed):
Looks fine to me.

PR #2126, opened 2021-05-10
optionally `generate_resonance_structures` when loading species dicts

davidfarinajr:
Resonance structure generation can slow down loading of species dictionaries.  Usually this is not significant, but it can be if you are loading a very large species dictionary from a model's edge.  This PR makes it optional (with defaulting to True), similar to how it is for `get_species` method for database

### DIFF
--- a/rmgpy/chemkin.pyx
+++ b/rmgpy/chemkin.pyx
@@ -799,12 +799,13 @@ def read_reaction_comments(reaction, comments, read=True):
 ################################################################################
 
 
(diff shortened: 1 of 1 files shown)


rwest (dismissed):
Looks OK to me. I guess the use case is some external script.

rwest (approved):
I added the default behaviors to the docstrings, and added reasoning to the commit message (taken from the pull request description). Ready to merge.

PR #2129, opened 2021-05-13
an empty surface site template group should only match an empty surface site

davidfarinajr:
### Motivation or Problem
Currently, empty surface site template groups match with vdw adsorbates since they both have the `Xv` atomtype.  This creates problems when generating reactions when a vdw adsorbate matches with the `Xv` group that is meant to be an empty site.  This problem has occurred in the reverse direction for the `Surface_Addition_Single_vdW` family which generates this reaction from its reverse template:

Screen Shot 2021-05-13 at 6 43 51 PM

### Description of Changes
Added a check to make sure that if the template group is an empty surface site, it should only match a structure that is an empty surface site

### Testing
the `Surface_Addition_Single_vdW`  no longer generates the reaction above.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1743,6 +1743,7 @@ class KineticsFamily(Database):
             struct = template_reactant
 
         reactant_contains_surface_site = reactant.contains_surface_site()
+        reactant_is_surface_site = reactant.is_surface_site()
 
         if isinstance(struct, LogicNode):
             mappings = []
@@ -1753,6 +1754,9 @@ class KineticsFamily(Database):
                 mappings.extend(reactant.find_subgraph_isomorphisms(child_structure, save_order=self.save_order))
             return mappings
         elif isinstance(struct, Group):
+            if struct.is_surface_site() != reactant_is_surface_site:
+                # An empty surface site group should not match an adsorbate
+                return []
             if struct.contains_surface_site() != reactant_contains_surface_site:
                 # An adsorbed template can't match a gas-phase species and vice versa
                 return []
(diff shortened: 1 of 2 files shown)


davidfarinajr:
Maybe we could also enforce that the denticity matches here 🤔, if a group is suppose to be mono dentate, make sure that the reactant only had 1 surface site

davidfarinajr:
just added unit test.  @ChrisBNEU , can you test this by merging it into your HAN branch and building a model?  If the model builds smoothly, we will have more confidence merging this in.

ChrisBNEU:
Current head for RMG-Py is [here](https://github.com/ReactionMechanismGenerator/RMG-Py/commits/new_han_paper), database is [here](https://github.com/ReactionMechanismGenerator/RMG-database/tree/han_paper_rebase). model is still running after 24 hours, previously it failed after 1.5 hours.

ChrisBNEU (commented):
See my other comment, this seems to fix the issue for the HAN model. Before I approve, does this work if an empty site is a product as well as a reactant?

davidfarinajr:
> See my other comment, this seems to fix the issue for the HAN model. Before I approve, does this work if an empty site is a product as well as a reactant?

That's a good point.  This will not check the products in the forward direction, and could result in Undeterminable kinetics for products in the reverse direction.  However, I do not think this is currently an issue since surface sites can only bond to one thing.  

If the empty surface site is a product in forward direction like this:
```
*2=*3       *4              *2-*3-*4
  :     +    |     ---->    |         +
~*1~       ~*5~           ~*1~            ~*5~
```
it will not be checked.  This would be a problem if we allow surface sites to bond to more than one thing (*5 X was bonded to 2 things for example, then the *5 product would not be empty), but we don't currently allow this.


If the empty surface site is a product in reverse direction like this:
```
 *1--*2                 *1     *2
  |            ---->    ||      |
~*3~ + ~*4~~           ~*3~ + ~*4~~
```
and *4 molecule was not an empty surface site, RMG would not be able to get the kinetics since it would not match the template reactant.

ChrisBNEU (approved):

ChrisBNEU (approved):

rwest (approved):

Issue #2122, opened 2021-05-07
Undeterminable Kinetics Error of Surface_Abstraction_vdW family


Tingchenlee:
### Bug Description
The reactions crashed when the bidentate species tried to match the reverse template of Surface_Abstraction_vdW family. There were three problem reactions so far:
![image](https://user-images.githubusercontent.com/56702445/117471687-220fa200-af26-11eb-9af8-5e1df87ad0d7.png)

### How To Reproduce
Maybe an easier way to reproduce these is using the **kinetic search** of the RMG website, here are the error messages
 - Reaction 1.
```
rmgpy.exceptions.UndeterminableKineticsError: 
(TemplateReaction(reactants=[Molecule(smiles="NO[O].[Pt]"), Molecule(smiles="[N-]([NH+]=[Pt])[Pt]")], 
products=[Molecule(smiles="[N-]([NH+](O[O])[Pt])[Pt]"), Molecule(smiles="N[Pt]")], 
pairs=[], family='Surface_Abstraction_vdW'), 'Kinetics could not be determined. 
Unable to find matching template for reaction 
<Molecule "NO[O].[Pt]"> + <Molecule "[N-]([NH+]=[Pt])[Pt]"> <=> <Molecule "[N-]([NH+](O[O])[Pt])[Pt]"> + <Molecule "N[Pt]"> 
in reaction family <KineticsGroups "Surface_Abstraction_vdW/groups">.
Trying to match [<Entry index=1 label="AdsorbateVdW">, <Entry index=2 label="Adsorbate1">] but matched [<Entry index=9 label="*N">]')
```
 - Reaction 2.
 ```
rmgpy.exceptions.UndeterminableKineticsError: 
(TemplateReaction(reactants=[Molecule(smiles="NN([Pt])[O].[Pt]"),
Molecule(smiles="O=[Pt]")], products=[Molecule(smiles="[Pt]ON([Pt])[O]"), Molecule(smiles="N[Pt]")], 
pairs=[], family='Surface_Abstraction_vdW'), 'Kinetics could not be determined.
 Unable to find matching template for reaction
 <Molecule "NN([Pt])[O].[Pt]"> + <Molecule "O=[Pt]"> <=> <Molecule "[Pt]ON([Pt])[O]"> + <Molecule "N[Pt]"> 
in reaction family <KineticsGroups "Surface_Abstraction_vdW/groups">.
[...]

PR #2131, opened 2021-05-14
check multiplicity of reverse products if the family template reactants have multiplicity constraints (vdW groups)

davidfarinajr:
This PR attempt to fix this issue https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2122

### Motivation or Problem
If we are applying a family in reverse and the template reactants restrict multiplicity, we need to make sure that
a reverse product matches the multiplicity-constrained template reactant because the forward template products
do not enforce the multiplicity restriction.  Without this check, we get an `UndeterminableKineticsError` if the reverse products have a multiplicity that does not match the multiplicity constraint of the template reactants.

### Description of Changes
Added a multiplicity check when applying the template in reverse.

### Testing
Need to add unit test.  Tested a few reactions which had `UndeterminableKineticsError` but now are no longer generated since they don't satisfy the multiplicity constraints in the template.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1602,6 +1602,34 @@ class KineticsFamily(Database):
                 lowest_labels.append(min(labels))
             product_structures = [s for _, s in sorted(zip(lowest_labels, product_structures))]
 
+        # if applying the family in reverse and the template reactants restrict multiplicity, we need to make sure that
+        # a reverse product matches the multiplicity-constrained template reactant because the forward template products
+        # do not enforce the multiplicity restriction
+        if not forward and isinstance(reactant_structure, Molecule):
+            for template in self.forward_template.reactants:
+                # iterate through the template reactants and check to see if they have a multiplicity constraint
+                if isinstance(template.item, Group):
+                    if template.item.multiplicity != []:
+                        # this template restricts multiplicity and needs to be checked
+                        for struct in product_structures:
+                            # iterate through the product structures to make sure that
+                            # it matches a mulitplicity-constrained template reactant
+                            match = self._match_reactant_to_template(struct, template)
+                            if match:
+                                # A product structure matches the template! 
+                                break
+                        if not match:
+                            # No product matched the template reactant
+                            # Therefore, this reaction is invalid
+                            logging.debug(
+                                f'No product structures matched {template} which has a multiplicity of {template.item.multiplicity}\n\n'
+                                'Template:\n\n'
+                                f'{template.item.to_adjacency_list()}\n\n'
+                                'Product structures:\n')
+                            for struct in product_structures:
+                                logging.debug(f'{struct}\n{struct.to_adjacency_list()}\n')
+                            return None
+
         # Return the product structures
         return product_structures
 
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1602,9 +1602,10 @@ class KineticsFamily(Database):
                 lowest_labels.append(min(labels))
             product_structures = [s for _, s in sorted(zip(lowest_labels, product_structures))]
 
-        # if applying the family in reverse and the template reactants restrict multiplicity, we need to make sure that
-        # a reverse product matches the multiplicity-constrained template reactant because the forward template products
-        # do not enforce the multiplicity restriction
+        # If applying the family in reverse and the template reactants restrict multiplicity,
+        # we need to make sure that a reverse product matches the multiplicity-constrained
+        # template reactant because the forward template products do not enforce the
+        # multiplicity restriction.
         if not forward and isinstance(reactant_structure, Molecule):
             for template in self.forward_template.reactants:
                 # iterate through the template reactants and check to see if they have a multiplicity constraint
@@ -1616,18 +1617,19 @@ class KineticsFamily(Database):
                             # it matches a mulitplicity-constrained template reactant
                             match = self._match_reactant_to_template(struct, template)
                             if match:
-                                # A product structure matches the template! 
+                                # A product structure matches the template!
                                 break
                         if not match:
                             # No product matched the template reactant
                             # Therefore, this reaction is invalid
-                            logging.debug(
-                                f'No product structures matched {template} which has a multiplicity of {template.item.multiplicity}\n\n'
-                                'Template:\n\n'
-                                f'{template.item.to_adjacency_list()}\n\n'
-                                'Product structures:\n')
-                            for struct in product_structures:
-                                logging.debug(f'{struct}\n{struct.to_adjacency_list()}\n')
+                            if logging.getLogger().isEnabledFor(logging.DEBUG):
+                                logging.debug(
+                                    'No product structures matched %s which has a multiplicity of %r\n'
+                                    'Template:\n%s\n'
+                                    'Product structures:\n',
+                                    template, template.item.multiplicity, template.item.to_adjacency_list() )
+                                for struct in product_structures:
+                                    logging.debug(f'{struct}\n{struct.to_adjacency_list()}\n')
                             return None
 
         # Return the product structures
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1339,7 +1339,7 @@ class KineticsFamily(Database):
     def apply_recipe(self, reactant_structures, forward=True, unique=True):
         """
         Apply the recipe for this reaction family to the list of
-        :class:`Molecule` objects `reactant_structures`. The atoms
+        :class:`Molecule` or :class:`Group` objects `reactant_structures`. The atoms
         of the reactant structures must already be tagged with the appropriate
         labels. Returns a list of structures corresponding to the products
         after checking that the correct number of products was produced.
--- a/rmgpy/data/kinetics/familyTest.py
+++ b/rmgpy/data/kinetics/familyTest.py
@@ -830,7 +830,7 @@ class TestGenerateReactions(unittest.TestCase):
             thermo_libraries=[],
             reaction_libraries=[],
             kinetics_families=['H_Abstraction', 'R_Addition_MultipleBond', 'Singlet_Val6_to_triplet', 'R_Recombination',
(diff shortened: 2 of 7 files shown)


rmgpy/data/kinetics/family.py
```diff
@@ -1602,6 +1604,24 @@ def apply_recipe(self, reactant_structures, forward=True, unique=True):
                 lowest_labels.append(min(labels))
             product_structures = [s for _, s in sorted(zip(lowest_labels, product_structures))]
 
+        # if applying the family in reverse and the template reactants restrict multiplicity, we need to make sure that
+        # a reverse product matches the multiplicity-constrained template reactant because the forward template products
+        # do not enforce the multiplicity restriction
+        if not forward and reactant_type == 'mol':
```
rwest, line 1610: If you do
```suggestion
        if not forward and isinstance(reactant_structure, Molecule):
```
could you do away with the reactant_type variable and remove the changes above?

rmgpy/data/kinetics/family.py
```diff
@@ -1602,6 +1604,24 @@ def apply_recipe(self, reactant_structures, forward=True, unique=True):
...
+                if isinstance(node.item, Group):
+                    if node.item.multiplicity != []:
+                        multiplicity_restricted_templates.append(node)
+            if multiplicity_restricted_templates:
+                for template in multiplicity_restricted_templates:
+                    for struct in product_structures:
+                        match = self._match_reactant_to_template(struct, template)
+                        if match:
+                            break
+                    if not match:
```
rwest, line 1622: Complicated nested `for` loops with `if` and `break` clauses etc.  usually benefit greatly from some comments. I think this is correct, but I had to think very hard to check it. Please make life easier for reviewers future programmers by adding some comments here.

rmgpy/data/kinetics/family.py
```diff
@@ -1602,6 +1604,24 @@ def apply_recipe(self, reactant_structures, forward=True, unique=True):
...
+        # if applying the family in reverse and the template reactants restrict multiplicity, we need to make sure that
+        # a reverse product matches the multiplicity-constrained template reactant because the forward template products
+        # do not enforce the multiplicity restriction
+        if not forward and reactant_type == 'mol':
+            multiplicity_restricted_templates = []
+            for node in self.forward_template.reactants:
+                if isinstance(node.item, Group):
+                    if node.item.multiplicity != []:
+                        multiplicity_restricted_templates.append(node)
+            if multiplicity_restricted_templates:
```
rwest, line 1616: I think this `if` is redundant. If the list is empty the following `for` will just run zero times.

rwest (changes requested):
As well as the minor changes to the code that I suggested, please improve your commit message. It should describe what you're trying to achieve, and why, and how.
One of the first things I often do when debugging something is check commit messages of when the code was last updated, and try to get into the head of whoever implemented it at the time (even when it's my past self, I usually need the commit messages to help me remember). Understanding _why_ something is done the way it is is immensely helpful, and this is usually best conveyed by commit messages.  (Discussion on github pull requests and linked issues are better than nothing, but may get lost and aren't instantly visible in an IDE or git client. The commit message is there forever, and easily accessible)

davidfarinajr:
Thanks for the review @rwest, I'll work on those changes

rwest:
Thanks for making this fix, @davidfarinajr.  It doesn't go un-noticed when you put effort into solving other people's problems. 🥇

davidfarinajr:
Also added debug logging which looks like this for the unit test
```
DEBUG:root:No product structures matched AdsorbateVdW which has a multiplicity of [1]

Template:

multiplicity [1]
1 *1 Xv u0 p0 c0
2 *2 R  ux {3,S}
3 *3 R  ux {2,S}


Product structures:

DEBUG:root:<Molecule "[CH3].[Pt]">
multiplicity 2
1 *2 C u1 p0 c0 {2,S} {3,S} {4,S}
2    H u0 p0 c0 {1,S}
3    H u0 p0 c0 {1,S}
4 *3 H u0 p0 c0 {1,S}
5 *1 X u0 p0 c0


DEBUG:root:<Molecule "[CH]=[Pt]">
multiplicity 2
1 *4 C u1 p0 c0 {2,S} {3,D}
2    H u0 p0 c0 {1,S}
3 *5 X u0 p0 c0 {1,D}
```

davidfarinajr:
This multiplicity check will not check logic nodes though.  I don't think we have many (or any) of these with surface families, but perhaps we should figure out how to check those as well.  You could have a logic node at the top which points to groups which have multiplicity restrictions (don't think we currently have any of these)

rwest:
I guess we could check that the products of a reverse reaction _always_ match the forward reaction, but perhaps that's a lot of checking and will slow things down too much ?

rwest:
How often do we think this will get hit, as currently implemented?
There's an optimization possible for debug logging (avoids  doing the string formatting if you're not going to actually log it - because usually we're not running with the verbose option) but if we very rarely hit this then it's maybe not worth it.

davidfarinajr:
> I guess we could check that the products of a reverse reaction _always_ match the forward reaction, but perhaps that's a lot of checking and will slow things down too much ?

We could check, but this is "checked" when we descend the tree to get the kinetics.  If it doesn't at least match the top node, then we get `UndeterminableKineticsError`.  Perhaps, we could check this when we apply the reaction recipe (as we are doing with the multiplicity check) so that we don't generate the reaction in the first place.  This will slow things down, although I'm not sure by how much

davidfarinajr:
> How often do we think this will get hit, as currently implemented?
> There's an optimization possible for debug logging (avoids doing the string formatting if you're not going to actually log it - because usually we're not running with the verbose option) but if we very rarely hit this then it's maybe not worth it.

🤔, not sure how prevalent this issue is.  We would be hitting this check quite often for catalysis, but the majority of the time we should get a match because we don't usually have radicals on the surface and the reactant groups are restricted to singlet.

rwest:
This looks OK to me, but I made the last 3 commits, so now need someone else to review it.

Tingchenlee (approved):
It looks good to me. Thanks, Richard and David! I used this branch to run one of the problem input files and the previous issue is gone now.

rwest (approved):

davidfarinajr:
Looks good to me as well! Thanks Richard and Ting-Chen!

rwest:
I've merged this because it fixes a known issue and has been approved by review. But it occurred to me there may be another approach, that we should perhaps discuss.  I tried looking at it for a while, but it was taking too long to figure out.

We generate the reverse template by applying the recipe to the forward template. This is done mostly using atom type actions, etc. so for example if you have a `Cs` atom type (carbon with all single bonds) and apply a `increment_bond` then it becomes a `Cd` atom type (carbon with one double bond) in the reverse template for example. Could we do something similar with the overall group attributes like multiplicity? i.e. if you have `multiplicity = [0,2]` enforced for the forward template, then apply an `increment_radical` action you end up with `multiplicity = [1,3]` in the product template.    

Thoughts?

mjohnson541:
While I think the solution in this PR is fine for now, I think tracking multiplicity when applying recipe actions is probably cleaner and more consistent long term.

davidfarinajr:
> I've merged this because it fixes a known issue and has been approved by review. But it occurred to me there may be another approach, that we should perhaps discuss. I tried looking at it for a while, but it was taking too long to figure out.
> 
> We generate the reverse template by applying the recipe to the forward template. This is done mostly using atom type actions, etc. so for example if you have a `Cs` atom type (carbon with all single bonds) and apply a `increment_bond` then it becomes a `Cd` atom type (carbon with one double bond) in the reverse template for example. Could we do something similar with the overall group attributes like multiplicity? i.e. if you have `multiplicity = [0,2]` enforced for the forward template, then apply an `increment_radical` action you end up with `multiplicity = [1,3]` in the product template.
> 
> Thoughts?

Yes, this was my original thought:  If we enforce multiplicity with the template reactants, we could determine the multiplicity of the template products from the actions.  I like this approach, but I think it's more difficult to implement than it may seem.

Consider this family for example:

```
name = "Surface_Adsorption_Abstraction_vdW/groups"
shortDesc = u""
longDesc = u"""
Adsorbtion of a vdw species to the surface with a surface species.

*2=*3    *4-*5        *2-*3-*5    *4
  :   +   |    ---->   |       +  ||
~*1~    ~*6~         ~*1~~       ~*6~

The rate, which should be in mol/m2/s,
will be given by k * (mol/m2) * (mol/m2)
so k should be in (m2/mol/s). We will use sticking coefficients.
"""

template(reactants=["AdsorbateVdW", "Adsorbate1"], products=["Adsorbate2","Adsorbate3"], ownReverse=False)

reverse = "Surface_Desorption_Abstraction_vdW"

reactantNum=2
productNum=2

recipe(actions=[
    ['BREAK_BOND', '*4', 1, '*5'],
    ['FORM_BOND', '*3', 1, '*5'],
    ['CHANGE_BOND', '*4', 1, '*6'],
])
```

`AdsorbateVdW` is restricted to multiplicity 1, and there is no multiplicity restriction on `Adsorbate1`.  None of the actions changes the radical count, so we don't have to worry about that.  In order to ensure `AdsorbateVdW` is mult 1,
*2 and *3 should have no radicals, and any atom they are bonded to in `AdsorbateVdW` should have no radicals, but *5 can have radicals since `Adsorbate1` has no multiplicity restriction.  This means that the multiplicity of the `Adsorbate2` product should be the number of radicals on *5 + 1 (which would be hard to implement in the group definition).  

Since the multiplicity is also determined by the atoms in the molecule that are not labeled in the group (an not affected by actions), it it difficult to restrict multiplicity in the template products.  

It also gets tricky if only one of the template reactants has a multiplicity constraint.  If, for example `Adsorbate1` was limited to mult 1, and the actions don't change radical count, then we would know both products should be mult 1.

This could also be affected by how the groups are written in the database.  For this family, the `AdsorbateVdW` group looks like:

```
entry(
    index = 1,
    label = "AdsorbateVdW",
    group =
"""
multiplicity [1]
1 *1 Xv  u0 p0 c0
2 *2 R!H ux px cx {3,[D,T]}
3 *3 R!H ux px cx {2,[D,T]}
""",
    kinetics = None,
)
```

*2 and *3 have `ux` but they should be written as `u0` since its mult 1, so this could get messy if we are using the group atom radicals to determine multiplicity.

Issue #2010, opened 2020-08-19
Non-deterministic choice between ring(Cyclopropene) and ring(oxirene) ?


rwest:
### Bug Description
Comments on several pull requests have noticed some non-identical thermo in the rmg/nitrogen test in the regression tests. It shows up something like this:
```
Non-identical thermo!
original:	O1[C]=N1
tested:	O1[C]=N1
Hf(300K)  |S(300K)   |Cp(300K)  |Cp(400K)  |Cp(500K)  |Cp(600K)  |Cp(800K)  |Cp(1000K) |Cp(1500K) 
    129.73|     51.79|      9.14|      9.59|      9.99|     10.34|     10.90|     11.38|     12.40
    154.90|     56.55|      9.79|      9.15|      8.58|      8.33|      9.02|     10.26|     10.96
thermo: Thermo group additivity estimation: missing(O2s-CdN3d) + group(N3s-CsHH) + group(Cds-CdsCsCs) + ring(Cyclopropene) + radical(Cds_P)
thermo: Thermo group additivity estimation: missing(O2s-CdN3d) + group(N3s-CsHH) + group(Cds-CdsCsCs) + ring(oxirene) + radical(Cds_P)
```

It appears in comments on these separate pull requests:
- https://github.com/ReactionMechanismGenerator/RMG-database/pull/255#issuecomment-374435434 
- https://github.com/ReactionMechanismGenerator/RMG-database/pull/257#issuecomment-374014008
- https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1357#issuecomment-387726781
- https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1276#issuecomment-375012342
- https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1894#issuecomment-676613335

The diff may appear in other pull requests but without anyone delving into the logs and commenting.

It is mostly just mentioned and then ignored, although @goldmanm discussed it briefly [here](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1276#issuecomment-375012342)
> One species had different thermo [C]1=NO1, which was due to whether RMG used the group additivity value of ring(Cyclopropene) or ring(oxirene), a 30 kcal difference. Neither of these groups is likely giving accurate values for the thermo, but the change either shows some indeterminate nature of RMG or that we are losing resonance structures in the simulation with these changes.

I suspect it's an indeterminate behaviour.

### How To Reproduce
[...]

PR #2136, opened 2021-05-21
Cat fixes - vdw radicals, bidentates, resonance

davidfarinajr:
### Motivation or Problem and Description of Changes

1) vdw radicals - We previously implemented a multiplicity check to ensure products in the reverse direction match the multiplicity of the template reactants . However, we are getting vdw products that do not obey multiplicity restriction in the forward direction (https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2122) , so we also need to check this direction as well. 

2) radicals on X - if resonance generation puts radicals/lone pairs/or a formal charge on `X`, RMG crashes.  I added a commit on this PR to filter out these structures

3) bidentate vdw - We were getting bidentate vdw structures. This PR makes them forbidden for surface families

### Testing
added unit test for forbidden vdw bidentates and successfully built an ammonia cat oxidation model that was previously crashing.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1714,6 +1714,13 @@ class KineticsFamily(Database):
         if self.forbidden is not None and self.forbidden.is_molecule_forbidden(molecule):
             return True
 
+        # forbid vdw multi-dentate molecules for surface families
+        if "surface" in self.label.lower():
+            if molecule.get_num_atoms('X') > 1:
+                for atom in molecule.atoms:
+                    if atom.atomtype.label == 'Xv':
+                        return True
+
         return False
 
     def _create_reaction(self, reactants, products, is_forward):
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1527,6 +1527,14 @@ class KineticsFamily(Database):
                 atom_labels['*3'].label = '*1'
                 atom_labels['*5'].label = '*2'
 
+            elif label == 'surface_abstraction_single_vdw':
+                # *3 migrates from *2-*1 to *4-*5
+                # so swap *1 with *5, swap *2 with *4
+                atom_labels['*1'].label = '*5'
+                atom_labels['*5'].label = '*1'
+                atom_labels['*2'].label = '*4'
+                atom_labels['*4'].label = '*2'
+
         if not forward:
             template = self.reverse_template
             product_num = self.reactant_num or len(template.products)
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1616,12 +1616,16 @@ class KineticsFamily(Database):
                 lowest_labels.append(min(labels))
             product_structures = [s for _, s in sorted(zip(lowest_labels, product_structures))]
 
-        # If applying the family in reverse and the template reactants restrict multiplicity,
-        # we need to make sure that a reverse product matches the multiplicity-constrained
-        # template reactant because the forward template products do not enforce the
-        # multiplicity restriction.
-        if not forward and isinstance(reactant_structure, Molecule):
-            for template in self.forward_template.reactants:
+        # If the template restricts multiplicity, we need to make sure that
+        # a product matches the multiplicity-constrained template
+        # because the template does not ensure that the
+        # multiplicity restriction is obeyed
+        if isinstance(reactant_structure, Molecule):
+            if forward:
+                template_groups = self.forward_template.products
+            else:
+                template_groups = self.forward_template.reactants
+            for template in template_groups:
                 # iterate through the template reactants and check to see if they have a multiplicity constraint
                 if isinstance(template.item, Group):
                     if template.item.multiplicity != []:
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1521,6 +1521,12 @@ class KineticsFamily(Database):
                 atom_labels['*2'].label = '*3'
                 atom_labels['*3'].label = '*2'
 
+            elif label == 'surface_abstraction':
+                atom_labels['*1'].label = '*3'
+                atom_labels['*2'].label = '*5'
+                atom_labels['*3'].label = '*1'
+                atom_labels['*5'].label = '*2'
+
         if not forward:
             template = self.reverse_template
             product_num = self.reactant_num or len(template.products)
--- a/rmgpy/molecule/resonance.py
+++ b/rmgpy/molecule/resonance.py
@@ -182,11 +182,12 @@ def generate_resonance_structures(mol, clar_structures=True, keep_isomorphic=Fal
                       " definition.".format(mol.to_adjacency_list()))
         raise
     if mol.get_net_charge() != 0:
-        raise ValueError("Got the following structure:\nSMILES: {0}\nAdjacencyList:\n{1}\nNet charge: {2}\n\n"
-                         "Currently RMG cannot process charged species correctly."
-                         "\nIf this structure was entered in SMILES, try using the adjacencyList format for an"
-                         " unambiguous definition.".format(mol.to_smiles(), mol.to_adjacency_list(), mol.get_net_charge()))
-
+        # logging.info("Got the following structure:\nSMILES: {0}\nAdjacencyList:\n{1}\nNet charge: {2}\n\n"
+        #                  "Currently RMG cannot process charged species correctly."
+        #                  "\nIf this structure was entered in SMILES, try using the adjacencyList format for an"
+        #                  " unambiguous definition. "
+        #                  "Returning the input mol".format(mol.to_smiles(), mol.to_adjacency_list(), mol.get_net_charge()))
+        return [mol]
     if not mol.reactive:
         raise ResonanceError('Can only generate resonance structures for reactive molecules! Got the following '
                              'unreactive structure:\n{0}Reactive = {1}'.format(mol.to_adjacency_list(), mol.reactive))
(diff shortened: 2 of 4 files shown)


mazeau (commented):
I think everything here looks fine. There definitely shouldn't be any bidentate things matching with vdw families. 

I don't think the best thing to do is to trash species after we generate charges/radicals/lone pairs on surface sites, but to prevent rmg from suggesting them in the first place.

davidfarinajr:
> I think everything here looks fine. There definitely shouldn't be any bidentate things matching with vdw families.
> 
> I don't think the best thing to do is to trash species after we generate charges/radicals/lone pairs on surface sites, but to prevent rmg from suggesting them in the first place.

Yes, I agree.  The fix on this PR is a band-aid to prevent RMG from crashing, and we should avoid generating these types of structures in the first place.  I think the filter on this PR is a sensible solution until we figure out how we want to generate resonance structs for adsorbates

mazeau (approved):
> > I think everything here looks fine. There definitely shouldn't be any bidentate things matching with vdw families.
> > I don't think the best thing to do is to trash species after we generate charges/radicals/lone pairs on surface sites, but to prevent rmg from suggesting them in the first place.
> 
> Yes, I agree. The fix on this PR is a band-aid to prevent RMG from crashing, and we should avoid generating these types of structures in the first place. I think the filter on this PR is a sensible solution until we figure out how we want to generate resonance structs for adsorbates

Yes, sorry, I forgot to change the bubble from comment to approve!

rmgpy/molecule/resonance.py
```diff
@@ -182,11 +182,12 @@ def generate_resonance_structures(mol, clar_structures=True, keep_isomorphic=Fal
                       " definition.".format(mol.to_adjacency_list()))
         raise
     if mol.get_net_charge() != 0:
-        raise ValueError("Got the following structure:\nSMILES: {0}\nAdjacencyList:\n{1}\nNet charge: {2}\n\n"
-                         "Currently RMG cannot process charged species correctly."
-                         "\nIf this structure was entered in SMILES, try using the adjacencyList format for an"
-                         " unambiguous definition.".format(mol.to_smiles(), mol.to_adjacency_list(), mol.get_net_charge()))
-
+        # logging.info("Got the following structure:\nSMILES: {0}\nAdjacencyList:\n{1}\nNet charge: {2}\n\n"
```
ChrisBNEU, line 185: is this staying commented out? or do we want it logged?

ChrisBNEU (commented):
This looks good, I agree with you and emily that this is a band-aid but hopefully when we have our surface resonance meetup with franklin we can revisit this. I tried it with an NO2/O2 on pt111 input file just to make sure it wouldn't give any errors. I had two concerns but other wise it looks ok
1. the commented out  logging in resonance.py (see my other comment)
2. should we update the resonanceTest so we capture what we are doing for charged gas/surface and radical surface species?

rmgpy/molecule/resonance.py
davidfarinajr: Logging it would probably be best.  I cherrypicked this commit from my electrocat branch where we do get charged species, but don't generate resonance structs for them (just return original molecule:  [mol])

davidfarinajr:
> 2. should we update the resonanceTest so we capture what we are doing for charged gas/surface and radical surface species?

Yes, we should add unit test that makes resonance structs with radicals on `X` and make sure we filter them out (with this PR).  We may have to change the test once we overhaul resonance generation for adsorbates, but that's ok.  It could actually help to make sure future changes don't break this.

rmgpy/molecule/resonance.py
```diff
@@ -299,11 +301,28 @@ def _generate_resonance_structures(mol_list, method_list, keep_isomorphic=False,
...
+            for x in [atom for atom in mol.atoms if atom.is_surface_site()]:
+                if x.radical_electrons != 0:
+                    mol_list.remove(mol)
+                    logging.warning('Resonance generation created a molecule {0} with {1} radicals on {2}\n'
+                    'Removing {0} from resonance structures'.format(mol.smiles,x.radical_electrons,x.symbol))
+                elif x.lone_pairs != 0:
+                    mol_list.remove(mol)
+                    logging.warning('Resonance generation created a molecule {0} with {1} lone pairs on {2}\n'
+                    'Removing {0} from resonance structures'.format(mol.smiles,x.lone_pairs,x.symbol))
+                elif x.charge != 0:
```
Tingchenlee, line 322: I have a question about this, would it prevent all the surface species with charges?
The representative structures of N2O in gas phase are
image

and it might have some surface reactions that eventually contain charged species 
image

 or maybe we can just assume N2O=X, rather than a bidentate:
```
1 O u0 p2 c0 {2,D}
2 N u0 p1 c0 {1,D} {3,S}
3 N u0 p1 c0 {2,S} {4,D}
4 X u0 p0 c0 {3,D}
```

davidfarinajr: It will not discard all surface species with formal charges. It only discards the ones that have a formal charge on the X atom like this one for example:
1 O u0 p2 c0 {2,S} {4,S}
2 N u0 p0 c+1 {1,S} {3,T}
3 N u0 p1 c0 {2,T}
4 X u0 p0 c-1 {1,S}

Tingchenlee (dismissed):
This looks good to me, thanks for the fixes!

rwest:
There was a comment up thread that a unit test should be added. Are we waiting for that before merging, or has it been done and I missed it?

davidfarinajr:
> There was a comment up thread that a unit test should be added. Are we waiting for that before merging, or has it been done and I missed it?

I still need to add that unit test before merging.  I also might get rid of the logging or change it to debug because it looks ugly in the RMG log files.  

I'm also not completely satisfied with the way the filter works (makes bad structs then removes them).  I can prevent them from being made, but I am not sure if that is a better long term solution since we may drastically change how resonance generation works for adsorbates 🤔

ChrisBNEU (dismissed):

davidfarinajr:
> > There was a comment up thread that a unit test should be added. Are we waiting for that before merging, or has it been done and I missed it?
> 
> I still need to add that unit test before merging. I also might get rid of the logging or change it to debug because it looks ugly in the RMG log files.
> 
> I'm also not completely satisfied with the way the filter works (makes bad structs then removes them). I can prevent them from being made, but I am not sure if that is a better long term solution since we may drastically change how resonance generation works for adsorbates 🤔

I added the unit test. I think the resonance fix on this PR is ok since it fixes a bug, but we should revisit resonance for adsorbates in the future.

davidfarinajr:
```
rmgpy.exceptions.ChemkinError: 
Unexpected species identifier Estimated encountered in flux pairs for 
reaction H(6) + C2H4(10) <=> C2H5(5).
```
getting strange chemkin error in integration tests 🤔

rwest:
I saw that. I think [this](https://github.com/ReactionMechanismGenerator/RMG-Py/commit/e4145148b2565b3180be59aac6f40b207319c46c) commit just got merged via https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2139 a couple hours ago - maybe even during the middle of running the tests. Let's rebase yet again and see if the problem goes away?

rmgpy/data/kinetics/family.py
```diff
@@ -1521,6 +1521,19 @@ def apply_recipe(self, reactant_structures, forward=True, unique=True):
                 atom_labels['*2'].label = '*3'
                 atom_labels['*3'].label = '*2'
 
+            elif label == 'surface_abstraction':
+                atom_labels['*1'].label = '*3'
+                atom_labels['*2'].label = '*5'
+                atom_labels['*3'].label = '*1'
+                atom_labels['*5'].label = '*2'
+
+            elif label == 'surface_abstraction_single_vdw':
```
rwest, line 1530: in #2081 I came up with a different set of label swaps here
https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2081/files
Please can you double check which of us is wrong? (or are they in fact equivalent?)

rwest (commented):
I requested you check one thing (atom label swaps), and chris had a comment (about logging stuff)) that's not been resolved yet.  Otherwise looks good to me. So, _very_ nearly ready to merge.

rmgpy/data/kinetics/family.py
davidfarinajr: yours is the right way.  Do you want to merge that PR in first, or maybe cherrypick that commit onto this one?

davidfarinajr: just cherrypicked that commit and rebased

rmgpy/molecule/resonance.py
davidfarinajr: I think leaving it commented out is fine.  Eventually we will be able to processed charged species and don't want to log this every time

davidfarinajr:
> I requested you check one thing (atom label swaps), and chris had a comment (about logging stuff)) that's not been resolved yet. Otherwise looks good to me. So, _very_ nearly ready to merge.

Thanks for the review Richard.  Your label swaps for `surface_abstraction_single_vdw` are correct, so I removed it from my commit and cherrypicked yours onto this PR.  The logging is not needed for resonance in my opinion, and is better left commented out.

rwest:
This was meant to be catalysis fixes, and yet the regression tests are showing a change in a gas phase molecule's thermo:

```
Non-identical thermo!
original:	O1[C]=N1
tested:	O1[C]=N1
Hf(300K)  |S(300K)   |Cp(300K)  |Cp(400K)  |Cp(500K)  |Cp(600K)  |Cp(800K)  |Cp(1000K) |Cp(1500K) 
    116.46|     53.90|     11.62|     12.71|     13.49|     13.96|     14.14|     13.85|     13.58
    141.64|     58.66|     12.26|     12.27|     12.09|     11.96|     12.26|     12.72|     12.15
thermo: Thermo group additivity estimation: group(O2s-CdN3d) + group(N3d-OCd) + group(Cd-HN3dO) + ring(Cyclopropene) + radical(CdJ-NdO)
thermo: Thermo group additivity estimation: group(O2s-CdN3d) + group(N3d-OCd) + group(Cd-HN3dO) + ring(oxirene) + radical(CdJ-NdO)
```

is this a false alarm? Could we have somehow changed from  ring(Cyclopropene) to  ring(oxirene) by tweaking a resonance algorithm, or is this a prior non-deterministic bug in RMG and we got unlucky? 

Looking at the code diff again, I can't think of anything that would cause this change. So maybe it's random?

...and so I check the issue tracker and sure enough: https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2010 

So, I'm satisfied it's not our fault, and am approving this PR.

rwest (approved):

PR #2137, opened 2021-05-22
Fix QM Parsers

kspieks:
### Motivation or Problem
A minor bug in Arkane's QChem parser would return 0 when parsing the energy rather than raising an error to alert the user. In general, our parsers should check for common errors before parsing the log files.

### Description of Changes
This PR adds common convergence checks for Arkane's QM parsers and fixes a minor bug with our QChem parser. Some parsers had already implemented brief checks for convergence but these were only within one method, such as `load_geometry()` or `load_conformer()`; other methods were not benefiting from error checking, and if only parsing the energy, no error would be raised. The new default behavior is to check the QM log file for common errors upon initializing the class so that the user can trust any values parsed using the class methods.

### Testing
I added unit tests to check the new feature for the software that I had errored log files for. 

### Reviewer Tips
While adding unit tests, I realize that all of our QChem log files used for unit tests actually have this [line](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/arkane/data/qchem/co.out#L2405
): `gprntSymmMtrx error report:`
The text below that line doesn't look like an error, but [Google](https://www.google.com/search?q=qchem+%22gprntSymmMtrx+error+report%22&client=firefox-b-1-e&sxsrf=ALeKk01nuZDEdp0orTOAahrY5Q_CczMwgQ%3A1621630064809&ei=cByoYILYMMHi5NoPmYq_4A4&oq=qchem+%22gprntSymmMtrx+error+report%22&gs_lcp=Cgdnd3Mtd2l6EAMyBAgjECc6BwgAEEcQsANQqoQHWJKUB2DRlgdoAXACeACAAVOIAe4BkgEBNJgBAKABAaoBB2d3cy13aXrIAQjAAQE&sclient=gws-wiz&ved=0ahUKEwiC0ODf0tvwAhVBMVkFHRnFD-wQ4dUDCA0&uact=5) doesn't recognize this phrase so admittedly I'm unsure whether to flag this as an error or not. I parsed over 2500 QChem files I had used for BACs recently with QChem version 5 and this phrase isn't present, so maybe it's only used in version 4? 

Do we want to continue [ignoring this phrase](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/fix_qm_parsers/arkane/ess/qchem.py#L78) in our QChem parser or do we normally want to raise this as an error and just set `check_for_errors=False` during the unit tests?


xiaoruiDong:
@kspieks Thanks Kevin, I am aware of this PR. I will look at those changes during the week, and I will let you know once I finish reviewing them.

kspieks:
No rush. Looking forward to your feedback :)

kspieks:
Based on the discussion at RMG subgroup today, I believe we decided that we should normally raise this as an error and if it happens to be a false alarm, as is the case with our current QChem log files used for testing, the user can simply set `check_for_errors=False` to parse the file. I can squash these fixup commits with any other feedback you may have

kspieks:
After further discussion offline, we think it is okay to not raise an error for `gprntSymmMtrx` since the following text does not indicate any error

arkane/ess/gaussian.py
```diff
@@ -57,8 +57,47 @@ class GaussianLog(ESSAdapter):
...
+            self.check_for_errors()
+
+    def check_for_errors(self):
+        """
+        Checks for common errors in a Gaussian log file.
+        If any are found, this method will raise an error and crash.
+        """
+        with open(os.path.join(self.path), 'r') as f:
+            lines = f.readlines()
+            error, keywords = None, None
```
xiaoruiDong, line 72: `keywords` is never used.

arkane/ess/gaussian.py
```diff
@@ -57,8 +57,47 @@ class GaussianLog(ESSAdapter):
...
+            for line in reversed(lines):
+                # check for common error messages
+                if 'termination' in line:
```
xiaoruiDong, line 75: I don't find an `else` here. Generally, without a `termination` shown up in the file. It is highly possible that the job is not finished. Besides, according to my memory, for a single job, `termination` will only show up once at the very end; but for a complex job (e.g., composite methods or opt freq jobs) it may appear at several places. So better only check the term `termination` at the last few lines (e.g. lines[-1] to lines[-100]). Also you can add a `break` once find the identifier to avoid going further.

arkane/ess/gaussianTest.py
```diff
@@ -57,6 +57,18 @@ def setUpClass(cls):
...
+        Uses Gaussian log files that had various errors
+        to test if errors are properly parsed.
+        """
+        with self.assertRaises(LogError):
+            GaussianLog(os.path.join(self.data_path, 'l913.out'))
+        with self.assertRaises(LogError):
+            GaussianLog(os.path.join(self.data_path, 'l9999.out'))
+        with self.assertRaises(LogError):
+            log = GaussianLog(os.path.join(self.data_path, 'error_termination.out'))
+
```
xiaoruiDong, line 71: Better to also add a check for cases when `check_for_errors = False`. Same for other Log classes.

arkane/ess/molpro.py
```diff
@@ -34,8 +34,8 @@
 
 import logging
 import math
-
 import numpy as np
+import os.path
```
xiaoruiDong, line 38: Maybe too picky (**if so neglect this comment**). In most `.py` file in RMG, we put built-in modules before exterior modules.  For here, `os` is a built-in module while `numpy` is not; so better put `import os.path` before `import numpy ..` and add an extra empty line.

arkane/ess/gaussian.py
```diff
@@ -57,8 +57,47 @@ class GaussianLog(ESSAdapter):
     arrays. GaussianLog is an adapter for the abstract class ESSAdapter.
     """
 
-    def __init__(self, path):
+    def __init__(self, path, check_for_errors=True):
         self.path = path
+        if check_for_errors:
+            self.check_for_errors()
```
xiaoruiDong, line 63: Since all `XXX Log` Classes share this same `__init__` function, can we simply modify the `__init__` for the parent class `ESSAdapter` ?

arkane/ess/molpro.py
```diff
@@ -56,8 +56,26 @@ class MolproLog(ESSAdapter):
...
+        """
+        Checks for common errors in a Molpro log file.
+        If any are found, this method will raise an error and crash.
+        """
+        with open(os.path.join(self.path), 'r') as f:
+            lines = f.readlines()
+            error = None
+            for line in reversed(lines):
+                # check for common error messages
+                if 'No convergence' in line:
```
xiaoruiDong, line 74: I am not familiar with `Molpro` better to have someone else. I found a few more cases in ARC, maybe we can add some of them as well?

xiaoruiDong (commented):
Thanks, @kspieks! This change is really helpful! Generally, I am not very experienced in QM packages other than Gaussian, so better to find more people reviewing some of the changes. I also have few more suggestions regarding the implementation.

arkane/ess/gaussian.py
kspieks: Thanks for catching this. I'll remove `keywords`

arkane/ess/gaussian.py
kspieks: That's a good suggestion. I pushed a change

arkane/ess/molpro.py
kspieks: Good point. I'll add more examples from ARC's `trsh.py`

arkane/ess/molpro.py
kspieks: Thanks for letting me know. I'll update this

arkane/ess/gaussian.py
kspieks: Thanks for the suggestions. I'll update this to only iterate through the last 100 lines in the log file and to break if the file terminated without errors. I pushed a fixup commit. Let me know if I should modify it further or if it's okay to squash

arkane/ess/gaussianTest.py
kspieks: Sorry help me understand this comment. What would I check for if `check_for_errors = False`?

arkane/ess/qchem.py
```diff
@@ -57,8 +57,64 @@ class QChemLog(ESSAdapter):
...
+
+                    # "Relative error" is captured as a warning later
+
+                    # If running on NERSC, also want to avoid lines containing 'zonesort'
+                    # Per the NERSC documenation:
+                    # The error message means that running the zonesort kernel failed for some reason.
+                    # The end result is that your code may have run less optimally.
+                    # Other than that, the message is usually harmless.
+                    # https://docs.nersc.gov/jobs/errors/
+                    error = 'SCF failed'
```
amarkpayne, line 88: Are you sure that this error will always be an SCF error?

arkane/ess/gaussianTest.py
amarkpayne: I think this is fine. I was going to suggest something similar as well, but then I realized that if `check_for_errors = False` then the program will just run, so nothing to check

arkane/ess/orca.py
```diff
@@ -32,8 +32,8 @@
 Used to parse Orca output files
 """
 
-import logging
 import numpy
+import os.path
```
amarkpayne, line 36: Fix this formatting as well

arkane/ess/terachem.py
```diff
@@ -57,8 +57,31 @@ class TeraChemLog(ESSAdapter):
...
+    def __init__(self, path, check_for_errors=True):
         self.path = path
+        if check_for_errors:
+            self.check_for_errors()
+
+    def check_for_errors(self):
+        """
+        Checks for common errors in a TeraChem log file.
+        If any are found, this method will raise an error and crash.
+        """
```
amarkpayne, line 64: If possible, can you add a unit test for TeraChem?

arkane/ess/molpro.py
```diff
@@ -56,8 +57,56 @@ class MolproLog(ESSAdapter):
...
+    def __init__(self, path, check_for_errors=True):
         self.path = path
+        if check_for_errors:
+            self.check_for_errors()
+
+    def check_for_errors(self):
+        """
+        Checks for common errors in a Molpro log file.
+        If any are found, this method will raise an error and crash.
+        """
```
amarkpayne, line 64: If possible, can you add a unit test for Molpro as well?

amarkpayne (changes requested):
Overall I think everything looks good. Thanks for going the extra mile and implementing the check error methods for all of the different QM parsers! I think you covered everything, at least that I am aware of.

One thing I am thinking of (that is probably outside the scope of this PR) is treating multi-job log files better. I think it would be worthwhile to split (in memory, not as physical files) the log files so that each parser instance is only dealing with 1 QM job at a time. This would allow us to report the status of each job independently, which might be useful (for example the formyl_azide QChem example. It would be nice to report to the user that the opt job failed, but the freq job succeeded, and that the energy comes from the freq job. The user could then decide if this needs to be addressed or not). 

Another thing I am not sure of is how this all faces to the user. You have an option to set `check_for_errors = False` in the API, but would a user be able to set this from the Arkane input file? How would the user actually use this feature?

amarkpayne:
@kspieks once you address the requested changes go ahead and squash in the fixups and rebase onto current master

arkane/ess/qchem.py
```diff
@@ -278,12 +278,12 @@ def load_energy(self, zpe_scale_factor=1.):
         """
         e_elect = None
         with open(self.path, 'r') as f:
-            a = b = 0
+            a = b = None
             for line in f:
                 if 'Final energy is' in line:
-                    a = float(line.split()[3]) * constants.E_h * constants.Na
+                    a = float(line.split()[-1]) * constants.E_h * constants.Na
```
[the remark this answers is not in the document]
kspieks: I updated the docstring for this method to describe our prioritization. Let me know if I should modify the wording :)

arkane/ess/qchem.py
kspieks: To my knowledge, I think so? I only have a few examples of QChem failing for me so perhaps there are exceptions. But this builds upon the logic in ARC's [troubleshooting](https://github.com/ReactionMechanismGenerator/ARC/blob/master/arc/job/trsh.py#L172-L207)

arkane/ess/molpro.py
kspieks: I added a unit test by using the failed log files from ARC's unit tests

arkane/ess/terachem.py
kspieks: I haven't used TeraChem before so I don't have any failed jobs, and it seems like ARC doesn't have any either. Feel free to send me some to add if you have them though

kspieks:
Thanks for the review! Indeed, I had only considered the API so thanks for reminding me to make the new argument accessible from the input file. I pushed a simple commit that modifies `statmech.py` so that users can write an input file with
```
geometry = Log(<filepath>, check_for_errors=False)
```
if desired.


Detecting multiple jobs and parsing them separately would likely be helpful for my scenario. Though ARC separates each job type as a separate input file, and ARC seems to be commonly used in our group so maybe this feature wouldn't be as used? I can try working on it in the future though

arkane/ess/qchem.py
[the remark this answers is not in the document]
amarkpayne: I still think it is a good idea to rename the variables in the code, as they are not descriptive, but I think the docstring looks good.

arkane/ess/qchem.py
```diff
@@ -272,9 +272,10 @@ def load_conformer(self, symmetry=None, spin_multiplicity=0, optical_isomers=Non
 
     def load_energy(self, zpe_scale_factor=1.):
         """
-        Load the energy in J/mol from a QChem log file. Only the last energy
-        in the file is returned. The zero-point energy is *not* included in
-        the returned value.
+        Load the energy in J/mol from a QChem log file. Prioritize the energy from a converged
+        geometry optimization (assigned to variable "a"). If the file does not contain an optimization job
+        or if the optimization hit the maximum cycles, return the next best energy, such as from a frequency job
```
amarkpayne, line 315: Actually, maybe change "return the next best source" to "return the next equivalent source"

arkane/ess/qchem.py
amarkpayne: Let's match ARC for now then

arkane/ess/adapter.py
```diff
@@ -45,6 +45,11 @@ class ESSAdapter(ABC):
     An abstract ESS Adapter class
     """
 
+    def __init__(self, path, check_for_errors=True):
+        self.path = path
+        if check_for_errors:
+            self.check_for_errors()
+
```
amarkpayne, line 52: If possible, squash this commit in to the ones before it, as this fixes a change within the same PR. If that proves difficult though I am fine with leaving this commit the way it is, this is a big ask for little gain.

I think this would mean that this addition to the ess adapter goes into the same commit as adding the check_for_errors method, and the deletions in this commit would be applied to the various other commits in this PR

arkane/statmech.py
```diff
@@ -234,7 +234,7 @@ def load(self, pdep=False, plot=False):
         path = self.path
         directory = os.path.abspath(os.path.dirname(path))
 
-        def create_log(log_path):
+        def create_log(log_path, check_for_errors=True):
             if not os.path.isfile(log_path):
```
amarkpayne, line 238: Thanks for adding this! Can you also modify one of the Arkane example files so that this gets tested out when we run the  functional tests? Add a comment in the input file letting the user know:

1. That this is the syntax for instructing Arkane to ignore any error messages in the file
2. This is not actually needed in this specific file, but added to show the syntax

arkane/statmech.py
```diff
@@ -234,7 +234,7 @@ def load(self, pdep=False, plot=False):
         path = self.path
         directory = os.path.abspath(os.path.dirname(path))
 
-        def create_log(log_path):
+        def create_log(log_path, check_for_errors=True):
```
amarkpayne, line 237: Can you document this functionality to the Arkane documentation? I think the best place is at the end of this [section](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/documentation/source/users/arkane/input.rst#option-1-automatically-parse-quantum-chemistry-calculation-output) and just above this [header](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/documentation/source/users/arkane/input.rst#without-hinderedfree-rotors)

When it mentions the Log class, mention that the first argument is the path to the log file itself. Optionally, you may supply the keyword argument `check_for_errors=False` to prevent Arkane from raising an exception if it finds an error in the QM job based on the statements in the Log.

amarkpayne (commented):
Thanks for the fixes! Few more quick changes. If we don't hear from anyone about Terachem files in the next few hours then go ahead and push these final fixes and I'll merge this PR in.

arkane/ess/qchem.py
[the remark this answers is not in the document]
kspieks: That makes sense. I updated those variable names

arkane/ess/qchem.py
kspieks: Updated the docstring :)

arkane/ess/adapter.py
kspieks: Thanks for catching this. I squashed my many incremental commits so they are now organized by "additions to our ESS log classes" and "additions to our ESS log tests".

arkane/statmech.py
kspieks: Done! I modified `testing/arkane/two_parameter_arrhenius_fit/CH3OO.py`

arkane/statmech.py
kspieks: Absolutely. Hopefully this is the last time I forget to update the corresponding documentation 😅 Let me know if I should modify the wording :)

amarkpayne (approved):
Thanks for the PR, @kspieks ! And thanks @xiaoruiDong for reviewing this as well. I'll go ahead and merge this in

PR #2143, opened 2021-05-28
Remove file_format kwarg from plt.savefig

amarkpayne:
### Motivation or Problem
In `ipython/kinetics_library_to_training.py`, we include the kwarg `file_format` when calling the Matplotlib API. This raises a deprecation warning stating that this kwarg is no longer supported in Matplotlib 3.3 or later. 

### Description of Changes
I have removed this keyword argument. I also searched the whole code base for other instances and I did not find any other occurrences. 

### Testing
I confirmed that the warning messages go away with these changes.


xiaoruiDong (dismissed):
Thanks @amarkpayne. It looks good. 

For record, `plt.savefig` is now using the argument `format` other than the `file_format` in newer versions. When `format` is unset and no extension is included in the `fig`, the default format will be `png`, so the behavior doesn't change.

amarkpayne:
Thanks for the review, @xiaoruiDong ! I went ahead and rebased to update the branch, so feel free to merge this in once the tests pass

xiaoruiDong (approved):

xiaoruiDong:
@amarkpayne, Since I don't have permission to merge the PR into master. Can you do that for me? Thanks.

PR #2144, opened 2021-05-28
Have WSL users install graphviz system wide

### LINKED ISSUE #2141: "graphviz/libgvplugin_pango.so.6" - file not found" missing dependency on WSL installations
### Topic
Kinetic Library Addition to RMG Training Data

### Context
I couldn't work with "kinetics_library_to_training.ipynb" because of "error while loading shared libraries: libltdl.so.7" message.
I tried to work with "kinetics_library_to_training_tools.py" but I didn't find how to use it.

### Question
How to use "kinetics_library_to_training_tools.py" ?

### Installation Information
 - OS : Windows 10, WSL, Ubuntu 18.04 
 - Installation method: Miniconda using & download from github files
 - RMG version information:
   - RMG-Py: 3.1.0-59-g74b176a77
   - RMG-database: 3.0.0-206-g833577ec

[kinetics_library_to_training_tools.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/6545207/kinetics_library_to_training_tools.txt)
[kinetics_library_to_training.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/6545220/kinetics_library_to_training.txt)

amarkpayne:
### Motivation or Problem
fixes #2141 . The graphviz dependency requires many render libraries to work properly. Unfortunately, these dependencies are not listed by this package, so conda does not
install them in the conda environment. Instead, the system wide libraries are used if
available. On most native Linux distro like Ubuntu, these libraries are installed by default.
However, these libraries are not installed by default in WSL versions of Ubuntu. This
means that WSL users are missing dependencies that they need to run many RMG
related scripts and features.

### Description of Changes
Updated WSL installation documentation to prompt user to install the graphviz package system wide. This is preferred over figuring out which dependencies graphviz needs that
are not currently installed by default, as these could change over time. In an ideal world
conda would deal with this for us, but the graphviz conda package does not do this, though
there appears to be some future work towards this end (see the discussion in #2141).

### Testing
I have confirmed that this solution works in a fresh WSL install, and this also fixed the
user's problem in #2141. I have also built the documentation locally to ensure the changes
look as desired.

### Reviewer Note
libxrender1 is a dependency of graphviz, and thus installed
together.

### DIFF
--- a/documentation/source/users/rmg/installation/linuxSubsystem.rst
+++ b/documentation/source/users/rmg/installation/linuxSubsystem.rst
@@ -45,7 +45,12 @@ Installing the Linux Subsystem
     sudo apt install gcc
     sudo apt install g++
     sudo apt install make
-    sudo apt-get install libxrender1
+    sudo apt-get install graphviz 
+
+   Note: installing graphviz as a system package makes sure that rendering libraries that are not included
+   in the WSL version of Ubuntu are installed. Unfortunately, the graphviz conda package does not install
+   these dependencies in the RMG conda environment (later in the installation), instead relying on system
+   libraries even in native Linux installations.
 
 5. Follow the instructions for either the binary (:ref:`anacondaUser`) or source installation (:ref:`anacondaDeveloper`)
    for the Linux Operating system. Follow these instructions from the point directly after installing Anaconda.


jonwzheng (dismissed):
Confirmed that this fixes the dependency issue pointed out in #2141 in WSL 1, in addition to the 2 confirmations in the same thread that this works for WSL 2.

amarkpayne:
Thanks for the review @jonwzheng ! I'll rebase to update this branch. If you don't mind approving this PR after the test pass again, I'll merge this in

jonwzheng (approved):
(duplicate of above review) - Confirmed that this fixes the dependency issue pointed out in #2141 in WSL 1, in addition to the 2 confirmations in the same thread that this works for WSL 2.

PR #2145, opened 2021-05-28
Training reactions constraints

davidfarinajr:
This PR added a `constraints` param to the `add_rules_from_training` method so that the user can use certain criteria to pick and choose which training reactions to include (and exclude) when creating rate rules for non-auto gen families when the rules are compiled at the start of an RMG job.  The constraints are:
-   `metal` : list of allowed metals (e.g ['Pt'])
-   `facet` : list of allowed facets (e.g ['111'])
-  `elements` : list of allowed elements (e.g ['C','H','O'])
-  `forward_only` : True/False, if True, only reactions in the forward direction are allowed 

Using these constraints will allow us to customize the kinetics trees for individual RMG jobs.  For example, we know that the facet can have a profound affect on activation barriers (https://www.nature.com/articles/s41467-019-12858-3 shows example with CO2), and using a `facet` constraint will enable use to constrain which facets we want to include when creating rate rules.

Although we don't have many surface training reactions in the database right now, adding constraints will be more useful as we continue to add more training data (as we are doing in the PR https://github.com/ReactionMechanismGenerator/RMG-database/pull/479)

### Testing
I have not added any unit tests yet, but I built a few surface ammonium oxidation models with and without training reactions constraints.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -1144,11 +1144,24 @@ class KineticsFamily(Database):
             raise ValueError('No entry for template {0}.'.format(template))
         return entry
 
-    def add_rules_from_training(self, thermo_database=None, train_indices=None):
+    def add_rules_from_training(self, thermo_database=None, train_indices=None, constraints={}):
         """
         For each reaction involving real reactants and products in the training
-        set, add a rate rule for that reaction.
+        set, add a rate rule for that reaction if it does not violate any provided `constraints`
+
+        `constraints` (dict):
+            "metal" : list of allowed metals (e.g ['Pt']) or None (all allowed)
+            "facet" : list of allowed facets (e.g ['111']) or None (all allowed)
+            "elements" : list of allowed elements (e.g ['C','H','O']) or None (all allowed)
+            "forward_only" : True/False, if True, only reactions in the forward direction are allowed 
         """
+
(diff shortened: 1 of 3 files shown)


davidfarinajr:
Note - this PR does not fix the "issue" that if a surface training reaction is isomorphic to a reaction in the model, the kinetics for that training reaction will be assigned to that reaction, regardless of metal or facet.  A further discussion is needed to determine what approach to take here:
1) strict - assume isomorphic iff metal and facet match
2) medium - use BEP to scale the Ea based on the thermo of reaction (could enforce that facets match here or not, not sure)
3) loose - ignore metal and facet and allow all isomorphic matches

mazeau (approved):
This looks good to me! Thanks for adding this in. I think you should add in more unit tests in general, but also to make sure you can use 2 constraints or more at once, like only wanting to use Pt111 things or something.

rwest:
We could make it _remove_ the training reactions that don't satisfy the constraints, so that they don't get turned into rules _and_ they don't get used later if fully isomorphic.  

And/or we could say anything that's not fully isomorphic (with metal and facet) then don't use the training reaction (from the wrong metal or facet) and instead use the tree-based estimate (which would at least be a specific node)

rwest:
What to do with Matt's auto-generated trees when we have different metals and facets is an unsolved problem to be discussed another day...

PR #2148, opened 2021-05-31
SurfaceArrhenius <=> StickingCoefficient

davidfarinajr:
### Motivation or Problem
Some surface reaction families have training data with mixed kinetics types (SurfaceArrhenius and StickingCoeff).  We need methods to convert SurfaceArrhenius to StickingCoeff and vice versa so that we can convert the training data to the same kinetics type so that we can average them.

### Description of Changes
- created methods to convert StickingCoeff to SurfaceArrhneius and SurfaceArrhenius to StickingCoeff
- added a check to make sure that the kinetics types are all the same when averaging kinetics 

### Testing
- Built a model with the DB branch for this PR https://github.com/ReactionMechanismGenerator/RMG-database/pull/479, where we have mixed kinetics types in training
- tested the Surf Arrh <--> StickingCoeff method in Jupyter notebook, but have not added unit tests yet

### DIFF
--- a/rmgpy/reaction.py
+++ b/rmgpy/reaction.py
@@ -729,6 +729,63 @@ class Reaction:
 
         raise NotImplementedError("Can't get_surface_rate_coefficient for kinetics type {!r}".format(type(self.kinetics)))
 
+    def surface_arrhenius_to_sticking_coeff(self, surface_site_density, Tmin=None, Tmax=None):
+        """
+        Converts `SurfaceArrhenius` kinetics to `StickingCoeff` kinetics using the provided
+        `surface_site_density` in SI units (mol/m^2).  The reaction's kinetics type must but be
+        `SurfaceArrhenius`.
+
+        Returns:
+            `StickingCoefficient` kinetics
+        """
+        cython.declare(kf=StickingCoefficient)
+        cython.declare(sticking_coefficient=cython.double, molecular_weight_kg=cython.double)
+        cython.declare(Tlist=np.ndarray, klist=np.ndarray, i=cython.int, number_of_sites=cython.int)
+        if not isinstance(self.kinetics, SurfaceArrhenius):
+            raise TypeError(f'Expected a SurfaceArrhenius object but received {self.kinetics}')
+        # determine the adsorbate and number of surface sites
+        adsorbate = None
+        number_of_sites = 0
+        for r in self.reactants:
+            if r.contains_surface_site():
+                number_of_sites += 1
+            else:
+                if adsorbate is None:
+                    adsorbate = r
+                else:
+                    logging.error("Error in kinetics for reaction {0!s}: "
+                                    "more than one adsorbate detected".format(self))
+                    raise ReactionError("More than one adsorbate detected")
+
+        if adsorbate is None or adsorbate.contains_surface_site():
+            logging.error("Problem reaction: {0!s}".format(self))
+            raise ReactionError("Couldn't find the adsorbate!")
+        molecular_weight_kg = adsorbate.molecular_weight.value_si
+
+        # generate temperature array to evaluate the rate coeff
+        if Tmin is not None and Tmax is not None:
+            Tlist = 1.0 / np.linspace(1.0 / Tmax.value, 1.0 / Tmin.value, 50)
+        else:
+            Tlist = 1.0 / np.arange(0.0005, 0.0034, 0.0001)
+
+        # Determine the values of the sticking coefficient at each temperature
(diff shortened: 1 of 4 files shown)


Tingchenlee:
I also ran a test(which included `CH4`, `NH3`, and `O2`species) with all new libraries and surface families, and the model was completed.

rwest:
These are useful methods to have, to convert back and forth, and we should keep them, but I feel like training reactions should all be in the same format in the database, and the person adding them to the database should convert at that point, rather than when loading the database at run time.  This will greatly simplify the averaging, and the interpretation of averaging, and Matt's auto tree generation, etc.

So I suggest a unit test added to the database checks to ensure all training reactions in adsorption reaction families are in StickingCoefficient form (and perhaps prints the helpful conversion in the error report if it's in SurfaceArrhenius, to facilitate changing it).

rwest:
Worth checking we do this: things entered in the reverse direction (i.e. desorption) will be a `SurfaceArrhenius` with units of `1/s`. These will need converting into a `StickingCoefficient`, not just a reversed `SurfaceArrhenius`.

rmgpy/data/kinetics/family.py
```diff
@@ -1189,8 +1200,14 @@ def add_rules_from_training(self, thermo_database=None, train_indices=None):
             if type(data) is Arrhenius:
                 # more specific than isinstance(data,Arrhenius) because we want to exclude inherited subclasses!
                 data = data.to_arrhenius_ep()
-            elif isinstance(data, StickingCoefficient):
-                data = StickingCoefficientBEP(
+            elif isinstance(data, SurfaceArrhenius):
+                if has_sticking_coeff:
+                    if surface_site_density is None:
+                        surface_site_density = 2.483e-09
```
mazeau, line 1206: I think instead of having this hard coded in, you should load it from the metal database

rmgpy/data/kinetics/family.py
```diff
@@ -1283,15 +1311,32 @@ def add_rules_from_training(self, thermo_database=None, train_indices=None):
             new_degeneracy = self.calculate_degeneracy(item)
 
             if isinstance(entry.data, SurfaceArrhenius):
-                data = SurfaceArrheniusBEP(
+                if has_sticking_coeff:
+                    if surface_site_density is None:
+                        surface_site_density = 2.483e-09
```
mazeau, line 1316: here as well

rmgpy/data/kinetics/family.py
mazeau: don't remember what the units are in the db, so this might affect your reaction.py additions

mazeau (changes requested):
This looks good! I just am not a fan of the hard coded stuff

rmgpy/reaction.py
```diff
@@ -729,6 +729,103 @@ def get_surface_rate_coefficient(self, T, surface_site_density):
...
+        adsorbate = None
+        number_of_sites = 0
+        for r in self.reactants:
+            if r.contains_surface_site():
+                number_of_sites += 1
+            else:
+                if adsorbate is None:
+                    adsorbate = r
+                else:
+                    logging.error("Error in kinetics for reaction {0!s}: "
```
ChrisBNEU, line 756: nitpicky, but do we have adsorption reactions that have multiple adsorbates? I think this is really testing if the reaction is the correct type (i.e. adsorption). maybe the text could say "multiple adsorbates detected, reaction probably isn't adsorption"?

rmgpy/reaction.py
```diff
@@ -729,6 +729,103 @@ def get_surface_rate_coefficient(self, T, surface_site_density):
...
+        else:
+            Tlist = 1.0 / np.arange(0.0005, 0.0034, 0.0001)
+
+        # Determine the values of the sticking coefficient at each temperature
+        klist = np.zeros_like(Tlist)
+        for i in range(len(Tlist)):
+            sticking_coefficient = self.kinetics.get_rate_coefficient(Tlist[i])
+            sticking_coefficient /= math.sqrt(constants.kB * Tlist[i] / (2 * math.pi * molecular_weight_kg))
+            klist[i] = sticking_coefficient
+        for i in range(number_of_sites):
```
ChrisBNEU, line 777: nitpicky, but this could just be klist *= surface_site_density**number_of_sites.

ChrisBNEU (approved):
I left some nitpicky comments, but these two new functions are tools that are not used in the main RMG model building code, so I think it is good to go.

PR #2153, opened 2021-06-07
fixups in `get_training_set` family method

davidfarinajr:
Some small fixups in the `get_training_set` method which is used for autogen tree method

1) make a deepcopy of the species to get the thermo.  This is needed to clear the atom labels on the molecule which can mess with getting thermo

2) parse out the metal from the entry

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -4091,17 +4091,33 @@ class KineticsFamily(Database):
                     metal = entry.metal + entry.facet
                 for j, react in enumerate(entry.item.reactants):
                     if rxns[i].reactants[j].thermo is None:
-                        react_copy = react.copy(deep=True)
-                        react_copy.molecule[0].clear_labeled_atoms()
-                        react_copy.generate_resonance_structures()
-                        rxns[i].reactants[j].thermo = tdb.get_thermo_data(react_copy, metal_to_scale_to=metal)
+                        label_dict = react.molecule[0].get_all_labeled_atoms()
+                        mol = react.molecule[0]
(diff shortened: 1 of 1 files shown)


Tingchenlee:
I think this PR fixed the errors of `KeyError: 'Node not found for atom {} in molecule <Molecule "O=CO.[Pt]"> in thermo database adsorptionPt111.'` in `Surface_Dissociation_vdW` family!
But I have another error in rmgpy.molecule.graph.Graph.get_edge `ValueError: The specified vertices are not connected by an edge in this graph` now...

rwest (approved):
This looks OK to me

mjohnson541:
Might it be faster to grab the labels with get_all_labeled_atoms, then clear the atoms and then reapply them afterwards? 

On the larger families this already takes a while to run and this adds an extra deepcopy not just for every species in the dictionary, but for the reactants and products of every reaction in the training set.

mjohnson541:
Just pushed a commit doing that.

davidfarinajr:
> Might it be faster to grab the labels with get_all_labeled_atoms, then clear the atoms and then reapply them afterwards?
> 
> On the larger families this already takes a while to run and this adds an extra deepcopy not just for every species in the dictionary, but for the reactants and products of every reaction in the training set.

so something like this?
```
mol = species.molecule[0]
labeled_atoms = mol.get_all_labeled_atoms()
mol.clear_labeled_atoms()
species.generate_resononace_structures()

> Just pushed a commit doing that.

Thanks, looks good to me

rwest (commented):
Can't see anything wrong

PR #2154, opened 2021-06-07
bond dissociation energies for surface bonds

davidfarinajr:
### Motivation or Problem
Bond dissociation energies (bdes) are needed for surface bonds in order to fit BM rules during automated tree gen.  However, we cannot currently get bdes for X-A (surface_site - adsorbate) bonds

### Description of Changes
Added 2 new `Bond` methods (`add_van_der_waals_bond` and `is_surface_bond`)
 and modified the `get_bde` method to use provided `metal` (default Pt111) to estimate the bdes based on the metal's
atomic binding energies from the surface metal library.

### Testing
Added a couple unit tests and tested getting bdes for adsorbates in jupyter notebook. Have not tried generating trees yet

### DIFF
--- a/rmgpy/molecule/molecule.py
+++ b/rmgpy/molecule/molecule.py
@@ -672,7 +672,29 @@ class Bond(Edge):
                 binding_energy = metal_binding_energies[adatom]
                 max_bond_order = {'C': 4., 'O': 2., 'N': 3., 'H': 1.}
                 normalized_bond_order = self.order / max_bond_order[adatom]
-                binding_energy = quantity.Energy(binding_energy).value_si * normalized_bond_order
+                # intercepts estimated from DOI:10.1103/PhysRevLett.99.016105
+                intercepts = {
+                    'C' : {
+                        1 : 0,
+                        2 : -1.3,
+                        3 : -1.1,
+                        4 : 0,
+                    },
+                    'N' : {
+                        1 : -0.5,
+                        2 : -0.75,
+                        3 : 0
+                    },
+                    'O' : {
+                        1 : -0.5,
+                        2 : 0
+                    },
+                    'H' : {
+                        1 : 0
+                    }
+                }
+                intercept = intercepts[adatom][self.order]
+                binding_energy = quantity.Energy(binding_energy).value_si * normalized_bond_order + intercept
                 return -1 * binding_energy
 
     def equivalent(self, other):
--- a/rmgpy/molecule/molecule.py
+++ b/rmgpy/molecule/molecule.py
@@ -45,13 +45,14 @@ from urllib.parse import quote
 import cython
 import numpy as np
 
+import rmgpy.quantity as quantity
 import rmgpy.constants as constants
 import rmgpy.molecule.converter as converter
 import rmgpy.molecule.element as elements
 import rmgpy.molecule.group as gr
 import rmgpy.molecule.resonance as resonance
 import rmgpy.molecule.translator as translator
-from rmgpy.exceptions import DependencyError
+from rmgpy.exceptions import DependencyError, DatabaseError
 from rmgpy.molecule.adjlist import Saturator
 from rmgpy.molecule.atomtype import AtomType, ATOMTYPES, get_atomtype, AtomTypeError
 from rmgpy.molecule.element import bdes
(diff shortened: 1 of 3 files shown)


rmgpy/molecule/molecule.py
```diff
@@ -1118,6 +1118,19 @@ def remove_van_der_waals_bonds(self):
...
+    def add_van_der_waals_bond(self):
+        """
+        Adds a single van der Waals bond to the graph to connect a surface site to the physisorbed absorbate.
+        """
+        cython.declare(bond=Bond,fragment=Molecule,fragments=list)
+
+        if self.contains_surface_site():
+            fragments = self.split()
+            if len(fragments) == 2:
+                if any([fragment.is_surface_site() for fragment in fragments]):
```
rwest, line 1175: If you remove the `[` and `]` from this line then the `any` will operate over the generator `fragment.is_surface_site() for fragment in fragments` and it'll be able to abort the loop as soon as one of them returns `True`. By creating a list first by wrapping the generator with `[ ]`, you force it to go to the end of the loop, and you add the overhead of creating the list object.

Here's an exaggerated example:
```python
In [3]: f = list(range(-3,100))

In [4]: any([i>0 for i in f])
Out[4]: True

In [5]: any(i>0 for i in f)
Out[5]: True

In [6]: %timeit any([i>0 for i in f])
3.9 µs ± 121 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [7]: %timeit any(i>0 for i in f)
503 ns ± 1.69 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

In this context it's not going to make a gigantic deal, but thought you might like to learn the trick :-)

rmgpy/molecule/molecule.py
```diff
@@ -627,16 +628,50 @@ def sorting_key(self):
...
+            if not self.is_surface_bond():
+                raise KeyError('Bond Dissociation energy not known for combination: '
+                               '({0},{1},{2})'.format(self.atom1.element.symbol, self.atom2.element.symbol, self.order))
+            else:
+                if self.atom1.is_surface_site():
+                    add_atom = self.atom2.element.symbol
+                else:
+                    add_atom = self.atom1.element.symbol
+                if self.is_van_der_waals():
+                    if add_atom == 'O':
```
rwest, line 651: I think you mean `adatom` not `add_atom`? 
> An adatom is an atom that lies on a crystal surface. [Wikipedia]

rmgpy/molecule/molecule.py
```diff
@@ -627,16 +628,50 @@ def sorting_key(self):
...
+                    metal_db = tdb.surface['metal']
+                except (DatabaseError, KeyError):
+                    from rmgpy.data.surface import MetalDatabase
+                    from rmgpy import settings
+                    metal_db = MetalDatabase()
+                    metal_db.load(os.path.join(settings['database.directory'], 'surface'))
+                metal_binding_energies = metal_db.find_binding_energies(metal=metal)
+                binding_energy = metal_binding_energies[add_atom]
+                max_bond_order = {'C': 4., 'O': 2., 'N': 3., 'H': 1.}
+                normalized_bond_order = self.order / max_bond_order[add_atom]
```
rwest, line 672: That raises an interesting question. Linear Scaling Relations tell us that moving from one metal to another the _change_ in adsorption energy scales with the normalized bond order (eg. 1/2 for H2C=X and 1/4 for H3C-X) but is it true that the actual adsorption itself is 1/2 and 1/4 of the C binding energy respectively?  

I think the answer is "no", or else these lines would all have a _y_ intercept of 0 ?
LSR slide

Is the BDE we're after the adsorption energy, in this case?

(If it's just for use in the Blowers Masel fitting, I have found it's very insensitive PROVIDED that w_0 > 2 * E^0, and have used an arbitrary high value of 10000 kJ/mol. i.e. it's more important to be high enough than to be accurate; OTOH if we have a method called "get_bde" future users/developers may assume that's what it does, and we probably shouldn't return arbitrary numbers.).

rwest (commented):
Looking good so far. Made a few comments.

rmgpy/molecule/molecule.py
davidfarinajr: Ah yes, you're right. They would go through the origin. Looks like the 1/4 bond has small y intercept, but the 2/4 and 3/4 bon d are shifted down by about 1 to 1.5 eV. We could just use the equations of these lines to get a better estimate of BDE based on atomic binding energies. Maybe something like this for C-X bonds:
1/4 bond bde = 1/4 x
2/4 bond bde = 1/2 x - 1.35 eV
3/4 bond bde = 3/4 x - 1.08 eV
Where x is C binding energy in DB.

davidfarinajr:
> Looking good so far. Made a few comments. 

Thanks Richard. I'll make some changes to address those comments

PR #2156, opened 2021-06-14
Fix spc doesn't have metal attribute error

hwpang:
### Motivation or Problem
Fix spc doesn't have metal attribute error

### Description of Changes
Changed `r.metal` to `metal`. Also changed dummy variable `r` to `rev_react` for a meaningful name & consistency.

### Testing
Species doesn't have metal attribute error is gone after the changes

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -4241,14 +4241,14 @@ class KineticsFamily(Database):
                 rrev.is_forward = False
 
                 if estimate_thermo:
-                    for r in rrev.reactants:
-                        if r.thermo is None:
-                            therm_spc = deepcopy(r)
+                    for rev_react in rrev.reactants:
+                        if rev_react.thermo is None:
+                            therm_spc = deepcopy(rev_react)
                             therm_spc.generate_resonance_structures()
-                            if r.metal:
-                                r.thermo = tdb.get_thermo_data(therm_spc, metal_to_scale_to=r.metal)
+                            if metal:
+                                rev_react.thermo = tdb.get_thermo_data(therm_spc, metal_to_scale_to=metal)
                             else:
-                                r.thermo = tdb.get_thermo_data(therm_spc)
+                                rev_react.thermo = tdb.get_thermo_data(therm_spc)
                 rxns[i] = rrev
 
         if self.own_reverse and get_reverse:
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -4051,6 +4051,28 @@ class KineticsFamily(Database):
                 if atm.label not in root_labels:
                     atm.label = ''
 
+        def get_reactant_thermo(reactant,metal):
+            """
+            Save the label of reactant and reapply them after thermo estimation to avoid deepcopying
+            """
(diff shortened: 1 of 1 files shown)


rmgpy/data/kinetics/family.py
```diff
@@ -4239,14 +4239,14 @@ def fix_labels_mol(mol, root_labels):
                 rrev.is_forward = False
 
                 if estimate_thermo:
-                    for r in rrev.reactants:
-                        if r.thermo is None:
-                            therm_spc = deepcopy(r)
+                    for rev_react in rrev.reactants:
+                        if rev_react.thermo is None:
+                            therm_spc = deepcopy(rev_react)
```
mjohnson541, line 4244: Can we avoid this deepcopy the same way we did in #2153 ?

mjohnson541 (changes requested):
Looks good! One thing.

rmgpy/data/kinetics/family.py
hwpang: Yeah sure!

hwpang:
@mjohnson541 Thanks for the review! I have added a function `get_react_thermo` to avoid deepcopy.

rmgpy/data/kinetics/family.py
```diff
@@ -4049,6 +4049,23 @@ def fix_labels_mol(mol, root_labels):
                 if atm.label not in root_labels:
                     atm.label = ''
 
+        def get_react_thermo(react,rxn_react,metal):
```
rwest, line 4052: From the function name and parameter names I can't guess what this is supposed to do. The meanings of `react`, `rxn_react`, and `metal` are hard to guess out of context. Maybe consider renaming them, and/or add a docstring to the function saying what it does?

rwest (commented):
I know you were just reducing code duplication by making that function, but there's an opportunity to make it better. As it stands, I found the function quite confusing based just on its name and parameter names.

davidfarinajr:
Thanks for fixing this @hwpang !  I like adding the function here, but perhaps it could have just 2 arguments (`react` and `metal`) and then, since it's a get function, it could return the thermo object instead of assigning it to rxn_react.thermo attr?  So something like 
```
 if rxns[i].reactants[j].thermo is None:
      rxns[i].reactants[j].thermo = get_react_thermo(react,metal)
```

davidfarinajr:
Also a little bit of documentation could help, but since it's a local function inside a function, I don't think we will use it too much

rmgpy/data/kinetics/family.py
hwpang: @rwest Good point! Thanks

hwpang:
@rwest @davidfarinajr Thanks for the comments! I changed the function name to `get_reactant_thermo` and added a docstring to give more explanation. I also changed the function to return the thermo object!

rwest (approved):

davidfarinajr (approved):
🙏 Thank you for fixing this @hwpang and addressing the comments !  Looks good to me!

mjohnson541:
@hwpang can you rebase?

hwpang:
> @hwpang can you rebase?

Rebased!

mjohnson541 (approved):

PR #2162, opened 2021-06-20
Added a save_order attribute to some methods, and make sure it is properly being passed around

alongd:
### Motivation or Problem
We try to perform atom labeling for species participating in a reaction while keeping the atom order constant (in the context of ARC). However, not all current methods support passing the (already existing) `save_order` argument around.

### Description of Changes
The `save_order` argument was added to methods that were identified as "problematic", and is now being passed on (eventually to the vf2 algorithm) appropriately.

### Testing
These additions were tested separately via ARC, yet unit tests were not added. This is a relatively "minor" modification, but I'm happy to hear thoughts on implementing tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,6 @@ external/Q2DTor
 
 # egg files (pip install -e)
 RMG_Py.egg-info/*
+
+# deleted files
+**/.fuse_hidden*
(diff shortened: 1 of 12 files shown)


mjohnson541 (approved):
Looks good to me!

PR #2167, opened 2021-06-25
correct barrier for library reactions

davidfarinajr:
### Motivation or Problem
We do not correct the barriers for library reactions.  For some library reactions, the Ea < H0 of reaction with RMG thermo which means the forward rate is too fast and the reverse rate is way way too fast.  This is particularly a problem for the `X + CHX -> HX + CX` in the Deutchmann kinetics library.

### Description of Changes
added an input file option `correct_library_kinetics` (default False) which when turned on, corrects the barriers for library reactions based on the thermo.

Screen Shot 2021-06-25 at 4 01 42 PM

### Testing
Tested this with @mazeau CPOX model, and the barrier for the Deutchmann reaction was corrected.

### DIFF
--- a/ipython/correct_barriers_in_kinetics_libraries.ipynb
+++ b/ipython/correct_barriers_in_kinetics_libraries.ipynb
@@ -1,0 +1,271 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "505f2fc9",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from rmgpy.data.rmg import RMGDatabase\n",
+    "from rmgpy.thermo.thermoengine import submit\n",
+    "from rmgpy.thermo import ThermoData\n",
+    "from copy import deepcopy\n",
+    "from rmgpy import settings\n",
+    "\n",
+    "from IPython.display import display\n",
+    "import os"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "fcf7c9e8",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "path = settings['database.directory']"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "ed58ea38",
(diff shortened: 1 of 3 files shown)


mazeau (dismissed):
Yay, this looks great! Thank you very much

rwest:
I still think we should check for which reaction rates this is true and just remove them from our libraries, or stop using those libraries, or fix them. Rather than just quietly altering the barrier (but not the A factor) and hence replacing one dodgy rate estimate with another even dodgier rate estimate (albeit one that is less fast and less prone to cause ODE solver crashes). 

But this looks like it should work as intended, and seems harmless, and will solve some of our problems. But if we're going to keep it then it probably needs documentation (maybe @mazeau could add). 


I suppose one approach would be to build a model with this on but then _check the log files and annotated chemkin files to identify all the reactions it applied to, and go look at them carefully and fix the libraries_.  Maybe that advice could be part of the documentation of this feature.

davidfarinajr:
Does the importer check if Ea >= H0? We might want to check it then. Although I think this is less of an issue for gas phase kinetics, and more of a problem for catalysis, especially if you are using a kinetics library on a different metal/facet with different thermo or a kinetics library from a model that is not thermodynamically consistent (doesn't use thermo to calculate reverse rates). @mazeau compared Deutchmanns thermo to ours and it was very different. We could remove these reactions from libraries, however if our families can't make them, we may have model truncation error. I'd rather have them in there and correct the barriers so that the rates are more sensible so they don't crash the solver.

mazeau:
Does anyone like the idea of keeping the unmolested deutschmann reactions in one direction in one library and then the "fixed" Ea in a separate "adjusted" deutschmann library? That way kids can choose whether they want the OG/as published rates vs the rmg fixed rates?

rwest:
image

mazeau:
will-you-press-the-button-all-of-your-friends-feel-28818188

mazeau (approved):
this all still looks good, but I think it might not be a good idea to let users easily overwrite their database in case they don't load the correct thermo libraries.

PR #2170, opened 2021-06-30
Ctc atomtype

davidfarinajr:
### Motivation or Problem
RMG crashes when trying to make `[C+]#[C-]` molecule because there is no atomtype for carbon with `p0 c+1 and triple bond`.   

### Description of Changes
added a `Ctc` atomtype and added  `[C+]#[C-]` to forbidden structures

### Testing
added atomtype test for `Ctc` and generated model

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       max-parallel: 5
     env: # update this if needed to match a pull request on the RMG-database
-      RMG_DATABASE_BRANCH: master
+      RMG_DATABASE_BRANCH: ctc
     defaults:
       run:
         shell: bash -l {0}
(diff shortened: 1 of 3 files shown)


rwest:
Are we adding it, or forbidding it, or both? I guess adding the atom type so that our forbidden structure can be read without crashing?  Does it not care about the C with a negative charge and a triple bond?  Also, why is `Cdc` only `charge=[+1]` and not also `-1`?  Probably I don't understand something here, which makes me nervous reviewing it quickly.

davidfarinajr:
RMG crashed with an atomtype error  because it made "[C+]#[C-]". We have "C2tc" atomtype for C with triple bond, 1 lone pair, and -1 charge, but no atomtype for C with triple bond, 0 lone pairs, and +1 charge. My solution on this PR was to make the new "Ctc" atomtype so we can make this molecule, and then forbid it so RMG doesn't crash if it makes "[C+]#[C-]". Unfortunately I don't remember how this molecule was made as it happened a while back, but adding this atomtype fixed it. Maybe there is a better way to forbid this without creating new atomtype?

PR #2171, opened 2021-07-01
Thermo engine metal support

davidfarinajr:
### Motivation or Problem
The thermo engine `submit` function is often used to get thermo from the DB, but currently it does not support different metals.  Therefore, we can only get adsorbate thermo on the default thermoDB metal 'Pt111' when using the thermo engine.

### Description of Changes
- added `metal` to the thermo engine's functions.  
- revised the `find_binding_energies` method to return the label of the metal entry so we know that we can report what metal we scaled to when correcting the binding energy using LSRs

### Testing
tested in notebook
Screen Shot 2021-07-01 at 1 25 22 PM

### DIFF
--- a/rmgpy/data/surface.py
+++ b/rmgpy/data/surface.py
@@ -233,8 +233,9 @@ class MetalDatabase(object):
 
     def find_binding_energies(self, metal):
         """
-        Tries to find a match in the database when given a metal and/or facet and returns the binding energies or a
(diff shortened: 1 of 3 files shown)


mazeau (dismissed):
I like this way of keeping track of there the binding energies came from, instead of just having the logging warning. thanks

PR #2173, opened 2021-07-02
Forming surface bonds

davidfarinajr:
### Motivation or Problem
We recently discovered molecules in our models that have multiple bonds to the surface through a single adatom
Screen Shot 2021-07-02 at 10 14 31 AM
https://rmg.mit.edu/database/molecule/1%20O%20u0%20p2%20c0%20%7B2%2CS%7D%20%7B3%2CS%7D%0A2%20X%20u0%20p0%20c0%20%7B1%2CS%7D%0A3%20X%20u0%20p0%20c0%20%7B1%2CS%7D%0A
I don't think we should be making these molecules.  

 

### Description of Changes
added `is_bonded_to_surface` Atom and GroupAtom methods and modified the `_apply` ReactionRecipe function so that it raises an `InvalidActionError` if trying to form a surface bond to an atom already bonded to the surface

### Testing
Not tested yet

### Reviewer Tips
see if these molecules are still in models

### DIFF
--- a/rmgpy/molecule/molecule.pxd
+++ b/rmgpy/molecule/molecule.pxd
@@ -79,6 +79,8 @@ cdef class Atom(Vertex):
     cpdef bint is_nos(self)
 
     cpdef bint is_surface_site(self)
+
+    cpdef bint is_bonded_to_surface(self)
     
     cpdef increment_radical(self)
 
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -346,15 +346,23 @@ class ReactionRecipe(object):
                                 else:
                                     atom2.atomtype = [ATOMTYPES['Xv']]
                 elif (action[0] == 'FORM_BOND' and forward) or (action[0] == 'BREAK_BOND' and not forward):
+                    # Form bond between atom1 and atom2
                     if struct.has_bond(atom1, atom2):
                         raise InvalidActionError('Attempted to create an existing bond.')
                     if info not in (1, 0):  # Can only form single or vdW bonds
                         raise InvalidActionError('Attempted to create bond of type {:!r}'.format(info))
+                    # we need to make sure we are not forming
+                    # a surface bond to an atom that is already bonded to the surface
+                    if atom1.is_surface_site() and atom2.is_bonded_to_surface():
+                        raise InvalidActionError('Attempted to form a surface bond to an atom already bonded to surface.')
+                    elif atom2.is_surface_site() and atom1.is_bonded_to_surface():
+                        raise InvalidActionError('Attempted to form a surface bond to an atom already bonded to surface.')
                     bond = GroupBond(atom1, atom2, order=[info]) if pattern else Bond(atom1, atom2, order=info)
                     struct.add_bond(bond)
                     atom1.apply_action(['FORM_BOND', label1, info, label2])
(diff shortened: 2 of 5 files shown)


rmgpy/molecule/molecule.pxd
```diff
@@ -79,6 +79,8 @@ cdef class Atom(Vertex):
     cpdef bint is_nos(self)
 
     cpdef bint is_surface_site(self)
+
+    cpdef bind is_bonded_to_surface(self)
```
rwest, line 83: This PR is consistent with the rest of the file so I'm not requesting you change it before we merge, but do you have any idea why `molecule.pxd` doesn't have the `except -2`  declarations but `group.pxd` does?

rwest (approved):
This looks OK to me (just looking at the code, not running it) but did you figure out why it's needed? What reaction recipe was messing up, and can it be fixed?

rmgpy/molecule/molecule.pxd
rwest:
```suggestion
    cpdef bint is_bonded_to_surface(self)
```

davidfarinajr: Thanks, I just fixed the type with force push, and nope, I am not sure why that is

mazeau (dismissed):
I know the tests haven't officially passed yet, but this looks very good to me! I wonder what changed for this to be needed though....

davidfarinajr:
I declared the GroupAtom `is_bonded_to_surface` function in the Group class and not the GroupAtom class. It's just one of those days 🤦 .

davidfarinajr:
🤞 attempt number 3

davidfarinajr:
> This looks OK to me (just looking at the code, not running it) but did you figure out why it's needed? What reaction recipe was messing up, and can it be fixed?

Looks like it was `Surface_Dissociation` which has this template
```
 *1--*2                 *1     *2
  |            ---->    ||      |
~*3~ + ~*4~~           ~*3~ + ~*4~~
```

It was applied to a bidentate molecule that had *2 bonded to `X`

rwest:
Hmm. Thanks. Are we stuck without a way to declare absence of a bond? i.e. how do we put "*2 cannot be bonded to an X" into the top level template?

rwest:
"Cannot convert rmgpy.molecule.group.GroupBond to rmgpy.molecule.group.GroupAtom"  Really not your day! 😢

davidfarinajr:
lol, I told @mazeau that I would drop out if the test fail again so I guess this is good bye 👋

davidfarinajr:
I'll fix it, get ready for another force push and attempt number 4

davidfarinajr:
> Hmm. Thanks. Are we stuck without a way to declare absence of a bond? i.e. how do we put "*2 cannot be bonded to an X" into the top level template?

Yes.  The most elegant solution would be to somehow put that in the top level group so that those bidentate reactants don't match the template and we don't try to create product structures.  With this PR, that bidentate reactant will match and the family will try to generate product structures, but it will fail to do so due to the `InvalidActionError` when applying the template.  However, I still think this is a better and more generalizable solution then adding forbidden structs to each family which are only checked after the products are made and the reaction recipe is applied.

mazeau (approved):
this still looks good to me, and all the test pass! I say it's merge time

PR #2177, opened 2021-07-13
Making Mopac calculations more robust

rwest:
### Motivation or Problem
@12Chao came across a few bugs when building models using Mopac. This tries to handle the errors better.

### Description of Changes
In one case, just trying the next geometry. In another, just reporting the molecule that caused the problem.

### Testing
So far, I have run unit tests.
I'm hoping @12Chao will try it out on the input files that caused the problems.


mjohnson541 (approved):
Looks good to me!

PR #2181, opened 2021-07-16
Load chemkin bug fix - extended elements and digits

davidfarinajr:
### Motivation or Problem
I tried to load RMG generated chemkin file, but errors were raised when
1) trying to read thermo block with extended elemental composition syntax
2) reading a reaction with a species the has a digit as its first character (i.e. 2-BTP)

### Description of Changes
1) check for extended syntax notation, and append the elemental comp. line to thermo block
2) check species dictionary before assuming that the leading digit corresponds to stoic coefficient

### Testing
I was able to load my chemkin files

### DIFF
--- a/rmgpy/chemkin.pyx
+++ b/rmgpy/chemkin.pyx
@@ -1190,6 +1190,16 @@ def read_thermo_block(f, species_dict):
             continue
 
         thermo_block += line
+
+        # check for extended elemental composition line
+        if line.rstrip().endswith('1&'):
+            # this thermo entry has extended elemental composition line
+            # read the elements line, append it to thermo_block, and continue 
+            line = f.readline()
+            thermo_block += line
+            line = f.readline()
+            continue
+
         if line[79] == '4':
             try:
                 label, thermo, formula = read_thermo_entry(thermo_block, Tmin=Tmin, Tint=Tint, Tmax=Tmax)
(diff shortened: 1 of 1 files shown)


ChrisBNEU (approved):
these look good, thanks for making them.

rmgpy/chemkin.pyx
```diff
@@ -1227,7 +1237,7 @@ def read_thermo_block(f, species_dict):
                     logging.warning('Skipping unexpected species "{0}" while reading '
                                     'thermodynamics entry.'.format(label))
             thermo_block = ''
-        if len(thermo_block.split('/n')) > 4:
+        if len(thermo_block.split('/n')) > 5:
             raise ChemkinError('Should only have 4 lines in a thermo block:\n{0}'.format(thermo_block))
```
ChrisBNEU, line 1241: nitpicky, but maybe change error message text: 
```suggestion
            raise ChemkinError('Should only have a maximum of 5  lines in a thermo block:\n{0}'.format(thermo_block))
```

ChrisBNEU (changes requested):
I missed this before, but we should change the message text. Otherwise this is good I think.

rmgpy/chemkin.pyx
davidfarinajr: Thanks Chris! Just fixed this

ChrisBNEU (approved):
looks good!

PR #2182, opened 2021-07-18
Added additional features to preserve atom order and template labels when generating reactions

alongd:
### Motivation or Problem
To assist in atom-mapping RMG reactions, some relatively minor modifications were made.

### Description of Changes
-  Use the save_order argument in generate_resonance_structures() 
- Added labeled_reactant/product_atom attribute to TemplateReaction
- Added relabel_atoms and delete_labels arguments to methods of KineticsFamily. These are used to optionally avoid relabeling atoms and to optionally avoid deleting atom labels, respectively. The original behavior is preserved.

In all cases, the original behavior was preserved.

### Testing
A test of retaining atom labels in a template reaction was added.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -65,6 +65,7 @@ from rmgpy.tools.uncertainty import KineticParameterUncertainty
 
 ################################################################################
 
+
 class TemplateReaction(Reaction):
     """
     A Reaction object generated from a reaction family template. In addition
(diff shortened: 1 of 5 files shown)


rmgpy/data/kinetics/family.py
```diff
@@ -1754,11 +1757,13 @@ def _create_reaction(self, reactants, products, is_forward):
 
         # Store the labeled atoms so we can recover them later
         # (e.g. for generating reaction pairs and templates)
-        labeled_atoms = []
+        reaction.labeled_reactant_atoms, reaction.labeled_product_atoms = list(), list()
         for reactant in reaction.reactants:
             for label, atom in reactant.get_all_labeled_atoms().items():
-                labeled_atoms.append((label, atom))
-        reaction.labeledAtoms = labeled_atoms
+                reaction.labeled_reactant_atoms.append((label, atom))
```
mjohnson541, line 1767: Would this not be easier to work with as a dictionary than a list of tuples?

mjohnson541 (commented):
This looks mostly good, my only question is about whether the labeled_reactant_atoms and labeled_product_atoms should be dictionaries instead of lists of tuples.

rmgpy/data/kinetics/family.py
alongd: Absolutely agree. I'll make the change.

alongd:
@mjohnson541, I modified the attribute to be a dict, now called `.labled_atoms` with `reactants` and `products` keys, respective values are `{label: atom}` dicts.
I removed a commit where the species constraints debug message was deleted (so this debug message is back now) because it gave some trouble in the tests.
I took the liberty to squash my fixup changes since they weren't major, I hope that's OK.
Thanks for the feedback!

mjohnson541 (approved):

PR #2202, opened 2021-09-28
Change default branch to `main`

amarkpayne:
### Motivation or Problem
The default branch name is hard coded in a few different places. To make the switch to `main`, we need to
change the default branch name in these locations.

### Description of Changes
I searched the whole repo for `master` and replaced it with `main` where appropriate.

### Testing
Hopefully the tests pass?


amarkpayne:
@mjohnson541 Let's merge this in at 12 EST on Friday. It looks like everything worked out okay here

Issue #2203, opened 2021-10-01
MultiArrhenius issue when loading reaction families


mazeau:
### Bug Description
Hi, I am getting this error, both locally and on discovery, when trying to use the BurkeH2O2 reaction library. The below pops up when trying to add the entire library into the model edge, and I believe it may be due to multiarrhenius.

```
Traceback (most recent call last):
  File "/home/mazeau.e/.conda/envs/rmg/lib/python3.7/site-packages/julia/pseudo_python_cli.py", line 308, in main
    python(**vars(ns))
  File "/home/mazeau.e/.conda/envs/rmg/lib/python3.7/site-packages/julia/pseudo_python_cli.py", line 59, in python
    scope = runpy.run_path(script, run_name="__main__")
  File "/home/mazeau.e/.conda/envs/rmg/lib/python3.7/runpy.py", line 263, in run_path
    pkg_name=pkg_name, script_name=fname)
  File "/home/mazeau.e/.conda/envs/rmg/lib/python3.7/runpy.py", line 96, in _run_module_code
    mod_name, mod_spec, pkg_name, script_name)
  File "/home/mazeau.e/.conda/envs/rmg/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/mazeau.e/Code/RMG-Py/rmg.py", line 118, in <module>
    main()
  File "/home/mazeau.e/Code/RMG-Py/rmg.py", line 112, in main
    rmg.execute(**kwargs)
  File "/scratch/westgroup/mazeau/RMG-Py/rmgpy/rmg/main.py", line 695, in execute
    self.initialize(**kwargs)
  File "/scratch/westgroup/mazeau/RMG-Py/rmgpy/rmg/main.py", line 577, in initialize
    self.reaction_model.add_reaction_library_to_edge(library)
  File "/scratch/westgroup/mazeau/RMG-Py/rmgpy/rmg/model.py", line 1725, in add_reaction_library_to_edge
    self.add_reaction_to_edge(rxn)
  File "/scratch/westgroup/mazeau/RMG-Py/rmgpy/rmg/model.py", line 1478, in add_reaction_to_edge
    self.edge.phase_system.phases["Default"].add_reaction(rxn)
  File "/scratch/westgroup/mazeau/RMG-Py/rmgpy/rmg/reactors.py", line 217, in add_reaction
[...]

PR #2205, opened 2021-10-04
adding EmptyRateUncertainty to rms

mazeau:
RMG was unable to load in multiArrhenius (and perhaps some others) to rms, so this should fix that

This goes with [this issue](https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2203)


mazeau:
this is addressed in #2208

PR #2212, opened 2021-10-07
Use python-jl for running RMG-Py

### LINKED ISSUE #2206: RMG Update
### Bug Description
Hi Guys! So, I had RMG installed, and worked perfectly! However, I've updated it and, since then, I've been getting the following error "name 'rms' is not defined". To solve it, I've tried to "make clean", update rms, and so on. Additionally, I've also removed the environment and installed all of it once more; but it didn't solve it. What should I do? 

### How To Reproduce
The error is coming specifically for "**rad = rms.getspeciesradius(atomnums, bondnum)"**

### Installation Information
 - Windows 10, with anaconda

amarkpayne:
### Motivation or Problem
fixes #2206 

Now that RMG relies on RMS/pyrms, we need to use python-jl instead of just python

### Description of Changes
Updated the shebang for the rmg.py executable. I'll also update the documentation but in the installation script PR

### Testing
This should be tested one more time on OSX.


amarkpayne:
Thanks for the review @sevyharris !!

PR #2229, opened 2021-10-31
Fix atom order issue in generate_resonance_structure

xiaoruiDong:
Previously, `generate_resonance_structure` will not keep the atom order when applying `keep_isomorphic=False` and `save_order=True`. The key issue is `Molecule.is_isomorphic` is called in the function and will sort atoms by default. The current change adds the correct argument to `is_isomorphic` and thus avoid such behavior. Moreover, `filter_structures` is removed from `_generate_resonance_structure`. since it is not used inside the function

### DIFF
--- a/rmgpy/molecule/resonance.pxd
+++ b/rmgpy/molecule/resonance.pxd
@@ -34,7 +34,7 @@ cpdef dict analyze_molecule(Molecule mol)
 
 cpdef list generate_resonance_structures(Molecule mol, bint clar_structures=?, bint keep_isomorphic=?, bint filter_structures=?, bint save_order=?)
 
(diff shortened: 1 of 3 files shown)


alongd (dismissed):
Thanks @xiaoruiDong! Looks good to me.

alongd:
@xiaoruiDong, could you add a test of a molecule that did not preserve it atom order when calling `_generate_resonance_structure` before, and now it does?

xiaoruiDong:
> @xiaoruiDong, could you add a test of a molecule that did not preserve it atom order when calling `_generate_resonance_structure` before, and now it does?

Thank you for the review. It is a good suggestion and a unit test is added.

xiaoruiDong:
@alongd all check passed

alongd (approved):
Thanks for this catch and fix!

PR #2237, opened 2021-12-01
set a maximum of 5 retries for model resurrection.

ChrisBNEU:
### Motivation or Problem
RMG can often become "stuck" if there is a particular reaction that is causing the solver to fail. to alleviate this, an edge species is added into the core. The problem is this continues indefinitely, and can cause RMG to add species indefinitely until the run is terminated, and can make a giant model that is mostly chemically irrelevant species. 

### Description of Changes
Added a hardcoded number of retries (5) so that RMG will eventually throw an error and crash. 

### Testing
I did not add a unit test per a conversation at developer hours, but I did test it by doing the following: 
add this reaction library to the eg0 input file:
```
reactionLibraries = ['BurkeH2O2inArHe'],
```
change the activation energy an A-factor in the first entry in that library to a very large value. Set the Ea to 0. 
```
entry(
    index = 1,
    label = "H + O2 <=> O + OH",
    degeneracy = 1,
    kinetics = Arrhenius(A=(1.04e+40, 'cm^3/(mol*s)'), n=0, Ea=(0, 'cal/mol'), T0=(1, 'K')),
    shortDesc = u"""Hong et al., Proc. Comb. Inst. 33:309-316 (2011)""",
)
```

### DIFF
--- a/rmgpy/solver/base.pyx
+++ b/rmgpy/solver/base.pyx
@@ -183,6 +183,10 @@ cdef class ReactionSystem(DASx):
         self.bimolecular_threshold = None
         self.trimolecular_threshold = None
 
+        # number of allowed model resurrections
+        self.retry = 0
+        self.max_retries = 5
+
     def __reduce__(self):
         """
         A helper function used when pickling an object.
@@ -617,6 +621,8 @@ cdef class ReactionSystem(DASx):
         cdef np.ndarray[np.float64_t, ndim=1] mole_sens, dVdk, norm_sens
         cdef list time_array, norm_sens_array, new_surface_reactions, new_surface_reaction_inds, new_objects, new_object_inds
 
+
+
         zero_production = False
         zero_consumption = False
         pdep_networks = pdep_networks or []
@@ -734,8 +740,14 @@ cdef class ReactionSystem(DASx):
                     if np.isnan(self.y).any():
                         raise DASxError("nans in moles")
                 except DASxError as e:
+                    self.retry += 1
                     logging.error("Trying to step from time {0} to {1} resulted in a solver (DASPK) error: "
                                   "{2!s}".format(prev_time, step_time, e))
+
+                    if self.retry > self.max_retries:
+                        raise RuntimeError(
+                            'Reached the maximum permitted number of retries due to solver errors ({0}), ending simulation'.format(self.max_retries))
+
 
                     logging.info('Resurrecting Model...')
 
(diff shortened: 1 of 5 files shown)


rmgpy/solver/base.pyx
```diff
@@ -746,20 +752,28 @@ cdef class ReactionSystem(DASx):
...
-                            obj = edge_reactions[ind]
-                            logging.info('At time {0:10.4e} s, Reaction {1} at dynamics number {2} was added to model core '
-                                         'in model resurrection process'.format(self.t, obj,total_div_accum_nums[ind]))
-                            invalid_objects.append(obj)
+                        if self.retry < self.max_retries:
+                            #species flux criterion
+                            if len(edge_species_rate_ratios) > 0:
+                                ind = np.argmax(edge_species_rate_ratios)
+                                obj = edge_species[ind]
+                                self.retry += 1
```
davidfarinajr, line 760: Maybe we should count each retry every time we encounter a `DASPK` error, regardless whether we enter the 2 if blocks which increment the self.retry?
```
 logging.error("Trying to step from time {0} to {1} resulted in a solver (DASPK) error: "
                                  "{2!s}".format(prev_time, step_time, e))
if self.retry >= self.max_retries:
    raise RuntimeError
self.retry += 1
```
Not sure 🤔.  Also since it's series of 3 ifs and not if/elifs, I'm not sure if we would ever enter more than one of those.

rmgpy/solver/base.pyx
```diff
@@ -746,20 +752,28 @@ cdef class ReactionSystem(DASx):
...
+                            if total_div_accum_nums and len(total_div_accum_nums) > 0:  #if dynamics data available
+                                ind = np.argmax(total_div_accum_nums)
+                                obj = edge_reactions[ind]
+                                self.retry += 1
+                                logging.info('At time {0:10.4e} s, Reaction {1} at dynamics number {2} was added to model core '
+                                            'in model resurrection process. Retry # {3} of {4}'.format(self.t, obj,total_div_accum_nums[ind],  self.retry, self.max_retries))
+                                invalid_objects.append(obj)
+                                
+                        else:
+                            raise RuntimeError ('too many retries, ending simulation')
```
davidfarinajr, line 775: I think RMG users might not know what `too many retries` means should they encounter it.  Perhaps we can make it a littler more descriptive:
`Reached the maximum permitted number of solver errors (self.max_retries), ending simulation`.

davidfarinajr (commented):
Thanks for this addition @ChrisBNEU !  I think this is a nice addition and should stop runaway bad chemistry.  My comments are about when we should increment `self.retry` and logging statements.  I'm also not sure whether we should make this an option in the input file and whether anyone would ever want to turn it off 🤔 or increase/decrease max_retries.

rmgpy/solver/base.pyx
ChrisBNEU: I changed it to just increment on DASPK errors because that makes more sense.

rmgpy/solver/base.pyx
ChrisBNEU: Agreed. I changed it to the say something similar, as shown in the attached traceback
[resurrect_retry_traceback.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/7687414/resurrect_retry_traceback.txt)
.

PR #2243, opened 2021-12-09
Allow ArrheniusBM Fitting up to 2000K for RMG-database Rate Trees

kspieks:
This small PR adds 2000K to the list of default temperatures used when doing an ArrheniusBM fit. This is relevant because the previous values only went up to 1500K, which forces our rate rules to extrapolate for any system that is operating above 1500K. Thanks to @davidfarinajr for identifying this issue as discussed in RMG-database PR [550](https://github.com/ReactionMechanismGenerator/RMG-database/pull/550).

It also adds a unit test for our ArrheniusBM class

### DIFF
--- a/rmgpy/kinetics/arrhenius.pyx
+++ b/rmgpy/kinetics/arrhenius.pyx
@@ -591,7 +591,7 @@ cdef class ArrheniusBM(KineticsModel):
         assert w0 is not None or recipe is not None, 'either w0 or recipe must be specified'
 
         if Ts is None:
-            Ts = [300.0, 500.0, 600.0, 700.0, 800.0, 900.0, 1000.0, 1100.0, 1200.0, 1500.0]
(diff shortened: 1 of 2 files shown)


davidfarinajr (commented):
Thanks @kspieks and thanks for adding the unit tests as well!  While we are adding tests for ArrBM here, could you also add a test for `get_activation_energy`?  Looks like you added test for most of the other ArrBM methods 🙏 .  I think `get_activation_energy` is important to test since we use it in RMG to covert ArrBM rules to Arr.  Thanks!

kspieks:
Good point thanks! Just added a test for `get_activation_energy`.

davidfarinajr (dismissed):

davidfarinajr:
Looks good!  I just rebased since it was out-of-date.  I'll merge when test pass this go around

davidfarinajr (approved):

kspieks:
Ah sorry forgot to rebase. Thank you!

PR #2260, opened 2022-01-10
BugFix: Draw CO as 'CO'

xiaoruiDong:
When read from a species dictionary, CO molecule has a different atom order than the one generated from SMILES. It further causes CO drawn as 'OC'. This commit ensures that CO is drawn correctly.

### DIFF
--- a/rmgpy/molecule/draw.py
+++ b/rmgpy/molecule/draw.py
@@ -153,7 +153,7 @@ class MoleculeDrawer(object):
         png. If `path` is given, the drawing is saved to that location on disk. The
         `options` dict is an optional set of key-value pairs that can be used to
         control the generated drawing.
-        
+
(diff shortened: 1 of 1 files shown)


hwpang (commented):
This looks good to me! Could you rebase?

hwpang:
Also I was wondering what is the difference between `atom.symbol` and `atom.element.symbol`, and they turn out to be the same.  From `class Atom`:

```
Additionally, the ``mass``, ``number``, and ``symbol`` attributes of the
    atom's element can be read (but not written) directly from the atom object,
    e.g. ``atom.symbol`` instead of ``atom.element.symbol``.
```

hwpang:
@xiaoruiDong Looks good! I will merge once the tests pass.

hwpang (approved):

PR #2263, opened 2022-01-13
give integer num of points to linspace

mjohnson541:
Small bugfix should fix #2256 in theory.

PR #2268, opened 2022-02-17
Several Arkane Torsion (Rotor) related changes

xiaoruiDong:
1. Correct the Fourier fitting Procedure. A function of the following form is to be fitted.
`f(x) = 1  + a cosx + b cos2x + ... + a' sinx + b' sin2x + ...`
In the fitting procedure, Arkane makes an assumption that f'(0) = 0 (last line in the regression matrix), which correspond to 
`f'(0) = a' + 2 b' + 3 c' ....=0.` 
But in the past 10 years, this was set to be 
`f'(0) = a' + b' + c' .. = 0.`
Now, it is corrected. In terms of influence, for most PES profiles, the previous mistake won't cause problems to a curve with good shape (like the PES from -CH3). For a more problematic PES profile (like the ones whose end conformed is significantly different from the initial one), the previous fitting approach may result in a tiny well with negative energies close to the origin. I haven't quantified the influence yet, but conceptually I don't think it is significant.

2. Enable parse scan log files with something like `L 1 2 3 B`. Currently, Arkane doesn't recognize this internal coordinate definition and will raise errors when parsing the file. So the 'L' defines a linear bend which can be helpful when there are ~180 angles in the molecule. Now, Arkane will simply pass the lines starting with an 'L'. 

3. Enable Assign PES (angles vs energies) directly in the Arkane input file. Previously, Arkane can only read the profile from a QM output. The new feature introduces more flexibility when conducting scan jobs, especially when you need to combine some results together. E.g., you may have a job stop in the midway, or you want to run jobs at each scan point separately, or if you want to combine PES profiles from both forward and backward scans, now after your jobs, you can simply prepare a list of angle, and a list of energies, and fit them to the Arkane. A Unit test corresponding to this feature is also added.

4. Fix a minor np.linspace argument issue. [fixed in another PR, removed the commit in this PR]


xiaoruiDong:
#2263 Just realize there is another fix for the linspace error

rmgpy/statmech/torsion.pyx
```diff
@@ -540,7 +540,7 @@ cdef class HinderedRotor(Torsion):
                 A[:-1, m] = np.cos(m * angle)
                 A[:-1, numterms + m - 1] = np.sin(m * angle)
             # This row forces dV/dangle = 0 at angle = 0
-            A[N, numterms:] = 1
+            A[N, numterms:] = np.arange(1., numterms)
```
alongd, line 543: I saw you related to this in the PR message, but could you please explain what this technically does?

arkane/statmech.py
```diff
@@ -158,6 +158,9 @@ def hinderedRotor(scanLog, pivots, top, symmetry=None, fit='best'):
     """Read a hindered rotor directive, and return the attributes in a list"""
     return [scanLog, pivots, top, symmetry, fit]
 
+def hinderedRotor1DArray(angles, energies, pivots, top, symmetry=None, fit='best'):
+    """Read a hindered rotor PES profile, and return the attributes in a list"""
+    return [angles, energies, top, pivots, symmetry, fit]
```
alongd, line 163: Is the order of `pivots` and `top` reversed on purpose between the input args and the output?

arkane/statmech.py
```diff
@@ -273,6 +276,7 @@ def create_log(log_path, check_for_errors=True):
             'True': True,
             'False': False,
             'HinderedRotor': hinderedRotor,
+            'HinderedRotor1DArray': hinderedRotor1DArray,
```
alongd, line 279: Should we add this option to Arkane's documentation?

alongd (commented):
Thanks! Please see some minor comments

arkane/statmech.py
xiaoruiDong: Thanks for finding it out. It should be a mistake.

arkane/statmech.py
xiaoruiDong: Thank you again for pointing me to the documentation. I just realize `HinderedRotor` may take a text file with angles and energies stored. It turns out I never find this before...So it actually does something very similar to my addition. Do you know it is still worth adding the class `HinderedRotor1DArray`?

alongd: The current option only allows to take in a list of angles and energies. Your option has more arguments. Perhaps keep the current option of passing a file to `ScanLog`. We can make this file more machine-friendly by allowing one to pass a yaml file, where you can add all your args and a list of data points.

arkane/statmech.py
xiaoruiDong: Fixed

arkane/statmech.py
xiaoruiDong: As suggested, I added the support of reading csv files and yaml files using `ScanLog`. For both method, I added unit tests and related descriptions in the documentation.

xiaoruiDong:
Thanks for the review @alongd. I made several changes as suggested:
- correct the argument sequence
- add support of reading csv files and yaml files using `ScanLog`
- add unit tests and documentations for all additions

arkane/statmech.py
```diff
@@ -92,38 +95,117 @@ def load(self):
...
+                angles = df[column]
+            elif 'energy' in column.lower():
+                try:
+                    energy_unit = column.split()[1][1:-1]
+                except IndexError:
+                    raise ValueError(f'Invalid headers ({column}) in the csv file.')
+                energies = df[column]
+        return angle_unit, energy_unit, angles, energies
+
+    def load_yaml(self):
```
alongd, line 153: Do we want to allow users to specify `pivots`, `top`, and `fit` directly in the YAML file?

alongd (commented):
Thanks! I added a few minor questions

arkane/statmech.py
xiaoruiDong: We could but, currently, this adds no benefit since we make these arguments required for `HinderedRotor`, as other file formats don't support reading the information from them. I decide to not add this into the scope of this PR.

xiaoruiDong:
@alongd Thank you for correcting me. I have removed those leftovers. I will squash the `fixup` PRs and rebase them after you are satisfied.

alongd:
Perfect, please go ahead and rebase

alongd:
We're merging PRs even though rmg-tests fail, right? Is it being looked into?

xiaoruiDong:
> We're merging PRs even though rmg-tests fail, right? Is it being looked into?

Not entirely sure about how much people fixed. #2271 solved some of the issues related to failures in rmg-test, if not any.

alongd:
Seems like the tests fail due to an issue with RMS:
`JULIA: MethodError: no method matching ReactionMechanismSimulator.Species(::String, ::Int64, ::String, ::String, ::String, ::ReactionMechanismSimulator.NASA{ReactionMechanismSimulator.EmptyThermoUncertainty}, ::Dict{Any, Any}, ::Int64, ::ReactionMechanismSimulator.StokesDiffusivity{Float64}, ::Float64, ::Int64, ::Float64)`

@mjohnson541 , @hwpang , could you take a look?

I think that principally we can merge this PR, let me know if that's acceptable with the current failing rmg-tests

mjohnson541:
Yeah, we accidentally merged the RMS part of a PR that should've been a twin RMG-Py-RMS PR. Although, I'm not sure if RMG-tests was working before that either. It should be fine to merge this one without RMG-tests.

alongd (approved):
Thanks!

xiaoruiDong:
@alongd @mjohnson541 Thanks everyone's help!

PR #2269, opened 2022-02-23
Add commit strings for RMG-Py and RMG-database

kspieks:
### Motivation or Problem
Currently, Arkane only logs the release version of RMG-Py. However, many calculations involve knowing exactly what values from RMG-database are being used. Since AECs, BACs, rate trees, etc can be changed/refit in between official release version updates, this PR allows Arkane to now log the git commit hash for both RMG-Py and RMG-database. This will make it easier for users to check their Arkane log files and know exactly what values were used for reproducibility.


ChrisBNEU:
I gave this a shot with examples/arkane/Benzyl and it works as expected. Ready for merge.

ChrisBNEU (approved):

PR #2273, opened 2022-02-28
Change rmgrc to a template file

amarkpayne:
This allows the user to make changes without these changes being reset. The previous solution of storing a permanent copy in $HOME/.rmg/rmgrc still works, but this is not a good solution when a user has multiple RMG installations that each point to different database locations

### Motivation or Problem
If a user wanted to point RMG to use a database at a non-default location, one way they can accomplish this is by changing this setting in the rmgrc file. Because we track this file though, any changes the user makes are likely to get overwritten constantly. This PR moves the file to a "template file" location. This allows the user to make their own rmgrc file, which is no longer tracked and in the .gitignore file. This approach has the advantage that multiple RMG installations on the same computer can be supported in this way. For example, this is useful for a git worktrees approach.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,3 +81,6 @@ RMG_Py.egg-info/*
 
 # deleted files
 **/.fuse_hidden*
+
+# RMG configuration file
+rmgpy/rmgrc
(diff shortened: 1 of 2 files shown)


sevyharris (approved):
This looks good to me!

sevyharris (approved):

PR #2278, opened 2022-02-28
New Pdep Methods

mjohnson541:
Adds the new simulation least squares (SLS) methods and the Georgievskii et al. 2013 CSE variant. Also enables negligence of energy transfer in high energy grains in the master equation matrix.


mjohnson541:
Replaced by #2315

PR #2280, opened 2022-03-05
Added Julia into the conda rmg_env

alongd:
### Motivation or Problem
New users installing RMG from source get stuck when executing
`python -c "import julia; julia.install(); import diffeqpy; diffeqpy.install()"`
since `Julia` isn't present in the `rmg_env`

### Description of Changes
Added Julia into the conda environment

### Reviewer Tips
From my experience this addition is required to run RMG when building a fresh env, I'm surprised that the current tests do pass w/o it, so perhaps I'm wrong. Could use another set of eyes.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -18,6 +18,7 @@ dependencies:
   - graphviz
   - h5py
   - jinja2
+  - conda-forge::julia
   - jupyter
   - rmg::lpsolve55
   - markupsafe


mjohnson541:
I've had a lot of issues with the conda-forge binaries, so I compiled my own (that are on the rmg channel) and pyjulia has them as a dependency so it shouldn't be strictly necessary to specify julia, although perhaps we should. The main issue I had with the conda-forge binaries as far as I'm aware is fixed now, but I could never get the environment to build with the binaries from conda-forge.

alongd:
I see. Also, neither of the tests liked my change, closing PR

JacksonBurns:
This ultimately came up again in #2415

PR #2285, opened 2022-03-14
remove check for adsorbate getting adsorbed again

ChrisBNEU:
### Motivation or Problem
Removes a check in reaction generation routine. If a species is already adsorbed, currently it cannot form a new bond. One of the upcoming bidentate families that @kblondal  is working on requires this:
```
 *1--*2--*3                   *1--*2    *3 
  |                   ---->    |   |     |
~*4~ + ~*5~ + ~*6            ~*4~~*5 + ~*6~~ 
```
### Description of Changes
Removed the checks specified above. they are no longer required. The original problem this was addressing was the mis-identification of an empty surface site as an adsorbed molecule. the routine ```_match_group_to_template``` does this check now, so it is no longer required.

### Testing
check with a surface mechanism and ensure it remains unchanged.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -2168,7 +2168,7 @@ class KineticsFamily(Database):
             This could be a surface reaction
                 A + X + X <=> BX + CX    (dissociative adsorption)
                 A + X + X <=> AXX        (bidentate adsorption)
-                ABX + X + X <=> AXX + BX (bidentate dissociation)
+                ABX + X + X <=> AXX + BX (dissociation to bidentate)
             or a termolecular gas phase reaction
                 A + B + C <=> stuff
             We check the two scenarios in that order.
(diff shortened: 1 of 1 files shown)


rmgpy/data/kinetics/family.py
```diff
@@ -2170,8 +2166,9 @@ def _generate_reactions(self, reactants, products=None, forward=True, prod_reson
         elif len(reactants) == 3 and len(template_reactants) == 3:
             """
             This could be a surface reaction
-                A + X + X <=> BX + CX  (dissociative adsorption)
-                A + X + X <=> AXX      (bidentate adsorption)
+                A + X + X <=> BX + CX    (dissociative adsorption)
+                A + X + X <=> AXX        (bidentate adsorption)
+                ABX + X + X <=> AXX + BX (bidentate dissociation)
```
kblondal, line 2171: Since there is another family called bidentate dissociation, I think it would be better to have "dissociation to bidentate" in the parentheses here for consistency with the name in the reaction family database.

kblondal (commented):
Thank you Chris! 
I tested this as well as the main branch without the new Surface_Dissociation_to_Bidentate family and there was no change in the surface mechanism. Also, with these changes there are no errors when running with the Surface_Dissociation_to_Bidentate family.

kblondal (approved):

Issue #2296, opened 2022-04-12
Calculate Reaction Degeneracy Fails in Intra_R_Add_Endocyclic Family


sevyharris:
### Bug Description
RMG fails to calculate a reaction's degeneracy  ( [CH2]C=CC=CC=C <=> [CH]1C=CC=CCC1 in reaction family Intra_R_Add_Endocyclic) while attempting to add_rules_from_training. This occurred while loading the database for an uncertainty calculation.

### How To Reproduce
Try running the following code:

```
from rmgpy.data.rmg import RMGDatabase
from rmgpy import settings

database = RMGDatabase()
database.load(
    settings['database.directory'],
    thermo_libraries=[
        'BurkeH2O2',
        'CurranPentane',
        'FFCM1(-)',
        'primaryThermoLibrary',
        'thermo_DFT_CCSDTF12_BAC',
        'DFT_QCI_thermo',
        'CBS_QB3_1dHR',
    ],
    kinetics_families=[
        'Intra_R_Add_Endocyclic',
    ],
)

for familyLabel, family in database.kinetics.families.items():
    family.add_rules_from_training(thermo_database=database.thermo)


```

### Expected Behavior
I expect the database to load without issue because it has in the recent past

### Installation Information
Describe your installation method and system information.
 - OS: WSL: Ubuntu 20.04.1
 - Installation method: source
 - RMG version information:
   - RMG-Py: 3.0.0-938-g4377398cf
   - RMG-database: 3.0.0-539-g89564b633

### Additional Context
Here's the error traceback:
```
Unable to calculate degeneracy for reaction [CH2]C=CC=CC=C <=> <Molecule "[CH]1C=CC=CCC1"> in reaction family Intra_R_Add_Endocyclic. Expected 1 reaction but generated 0
  File "/home/moon/rmg/RMG-Py/rmgpy/data/kinetics/family.py", line 1948, in calculate_degeneracy
    'but generated {2}').format(reaction, self.label, len(reactions)))
  File "/home/moon/rmg/RMG-Py/rmgpy/data/kinetics/family.py", line 1303, in add_rules_from_training
    new_degeneracy = self.calculate_degeneracy(item)
  File "/home/moon/rmg/my_examples/debug_degeneracy/min_err.py", line 22, in <module>
    family.add_rules_from_training(thermo_database=database.thermo)
```

xiaoruiDong:
[...]

PR #2297, opened 2022-04-13
Add tests to check num of reactant and product defined in ATG rate rules

xiaoruiDong:
### Motivation or Problem

Related to https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2296. The new ATG for Intra_R_Add families miss `reactantNum` and `productNum` defined. 

Add a unit test to check if `reactantNum` and `productNum` is defined inside `database/input/families/[ATG family]/groups.py`

If they are defined, family.reactant_num and family.product_num will not be `None` or `0`; Otherwise, if they are not defined, it will cause `family._generate_reactions` unexpectedly return empty list (see code around the ref link).

https://github.com/ReactionMechanismGenerator/RMG-Py/blob/4377398cf30eff9f28cf2ae65306cdd9dcb3d204/rmgpy/data/kinetics/family.py#L2000


testing/databaseTest.py
```diff
@@ -970,6 +976,20 @@ def kinetics_check_reactant_and_product_template(self, family_name):
             if boo:
                 raise ValueError("Error occured in databaseTest. Please check log warnings for all error messages.")
 
+    def kinetics_check_num_reactant_and_product(self, family_name):
+        """
+        This test checks that if the number of reactants and products are specified in the groups.py for
+        rate rules that are generated from the ATG.
+        """
+        family = self.database.kinetics.families[family_name]
+        if family.auto_generated:
```
kspieks, line 985: Can you help me understand why we only do this check if the tree was autogenerated? There are many surface reaction families that are missing the `autoGenerated=True` but they still define `reactantNum` and `productNum`. [This](https://github.com/ReactionMechanismGenerator/RMG-database/blob/2118733e3d2b8fcbfc8d23c15f665799b3920e6a/input/kinetics/families/Surface_Dissociation_vdW/groups.py#L21-L24
) is one example. Also halocarbene recombination ([link](https://github.com/ReactionMechanismGenerator/RMG-database/blob/f1a84af193e62347263de66b24802c3075b1b54a/input/kinetics/families/halocarbene_recombination/groups.py#L18-L22
))

I guess once all of our rate rules have been passed through ATG (hopefully in the near future), lines like [this](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/73dcb294fae0201da83407d822dd86fb89fd9411/rmgpy/data/kinetics/family.py#L739) will no longer be necessary since everything will rely on `reactantNum` and `productNum`?

testing/databaseTest.py
```diff
@@ -970,6 +976,20 @@ def kinetics_check_reactant_and_product_template(self, family_name):
...
+            if not getattr(family, 'reactant_num'):
+                logging.error(f'The number of reactants is not defined in the family {family_name}')
+            if not getattr(family, 'reactant_num'):
+                logging.error(f'The number of products is not defined in the family {family_name}')
+            if not getattr(family, 'reactant_num') or not getattr(family, 'reactant_num'):
```
kspieks, line 990: Should the second conditional say `getattr(family, 'product_num')` ?

kspieks (commented):
Thanks for adding this test and for fixing these two families on the `fix_intra_R` branch on RMG-database! I only have a few minor questions below. We can probably merge this PR today :)

testing/databaseTest.py
xiaoruiDong: My mistake. Thanks for catching that

xiaoruiDong:
The test won't pass since the PR for fixing database hasn't been created and merged.

xiaoruiDong:
@kspieks The RMG test failure is irrelevant to this addition. Feel free to approve this PR.  Looks like there is something hasn't been fixed in the Intra_R_Add_Exocyclic family. It looks like an issue of forbidden structure.

testing/databaseTest.py
kspieks: @xiaoruiDong Can you push a change that updates this line before we merge?

xiaoruiDong: Thanks! I am too careless about this.

kspieks (approved):

PR #2299, opened 2022-04-13
Correct ATG rate rules initialization

xiaoruiDong:
`add_rules_from_training` and `fill_rules_by_averaging_up` should be only called for non-ATG rate rules when initiating the RMG database rate rule estimator.
1. A full keyword search is done using these two function names. Wherever they are called, `if not autogenerated:` is made sure to be added before the block.
2. Bypass the calls of these two methods on ATG families inside these functions. Currently, there is no error raised to avoid breaking up user's private scripts and workflow. But this will be deprecated in the future version, and an error will be raised.

### DIFF
--- a/testing/databaseTest.py
+++ b/testing/databaseTest.py
@@ -395,8 +395,9 @@ class TestDatabase(object):  # cannot inherit from unittest.TestCase if we want 
     # These are the actual tests, that don't start with a "test_" name:
     def kinetics_check_surface_training_reactions_can_be_used(self, family_name):
         """Test that surface training reactions can be averaged and used for generating rate rules"""
-        family = self.database.kinetics.families[family_name]
-        family.add_rules_from_training(thermo_database=self.database.thermo)
+            family = self.database.kinetics.families[family_name]
+            if not family.auto_generated:
+                family.add_rules_from_training(thermo_database=self.database.thermo)
 
     def general_check_metal_database_has_catalyst_properties(self, library):
         """Test that each entry has catalyst properties"""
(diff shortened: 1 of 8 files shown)


testing/databaseTest.py
```diff
@@ -395,8 +395,9 @@ def test_metal_libraries(self):
     # These are the actual tests, that don't start with a "test_" name:
     def kinetics_check_surface_training_reactions_can_be_used(self, family_name):
         """Test that surface training reactions can be averaged and used for generating rate rules"""
-        family = self.database.kinetics.families[family_name]
-        family.add_rules_from_training(thermo_database=self.database.thermo)
+            family = self.database.kinetics.families[family_name]
+            if not family.auto_generated:
+                family.add_rules_from_training(thermo_database=self.database.thermo)
```
kspieks, line 400: Is there a reason we don't use `family.fill_rules_by_averaging_up(verbose=True)` for this surface kinetics unit test?

kspieks (commented):
Thanks for this PR! I only had one minor question but then we can merge :)

testing/databaseTest.py
xiaoruiDong: To be honest, I don't know why this isn't added. This test was written by a graduated student from the Northeastern Group. I personally feel there is no harm in testing the function you mentioned as well, so I just added it in a new commit. (I also quickly checked KineticsTest.py, and it seems to me there isn't a duplicate test for surface families there).

kspieks: Sounds good thanks

kspieks (approved):

PR #2303, opened 2022-05-09
Check isomorphism of lists with 4+ species

zhaoyangwx:
### Motivation or Problem
Sometimes need to compare reaction with 4+ reactants/products.

### Description of Changes
Added checking for list with 4+ reactants/products.
Added limitation of list length to prevent too long input causing hanging.
Improved performance by adjusting sequence to avoid unnecessary calculation of list length.

### Testing
Tested with simplified program:

```
list1 = [1,2,3,4,6,5]
list2 = [6,5,2,4,1,3]
list3 = [6,2,3,1,1,4]
list4 = [7,2,3,1,5,4]
def same(x,y):
	return (x==y)
def issamelist(l1,l2):
	l1=l1[:]
	l2=l2[:]
	x=0
	while x<len(l1):
		found_y = False
		for y in range(len(l2)):
			if same(l1[x],l2[y]):
				found_y = True
				del l1[x]
				x-=1
				del l2[y]
				break
		if not(found_y):
			return False
		x+=1
	return (len(l1)==0) and (len(l2)==0)
print(issamelist(list1,list2))
print(issamelist(list1,list3))
print(issamelist(list1,list4))
print(list1)

######output######
True
False
False
[1, 2, 3, 4, 6, 5]
```
### Reviewer Tips
maximum_length may be adjusted

### DIFF
--- a/rmgpy/reaction.py
+++ b/rmgpy/reaction.py
@@ -1501,7 +1501,7 @@ class Reaction:
 
 
 def same_species_lists(list1, list2, check_identical=False, only_check_label=False, generate_initial_map=False,
-                       strict=True, save_order=False, maximum_length=10):
+                       strict=True, save_order=False):
     """
     This method compares whether two lists of species or molecules are the same,
     given the comparison options provided. It is used for the `is_same` method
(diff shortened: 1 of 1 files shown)


zhaoyangwx:
Still some problem comparing lists with same elements but different number (with dupe elements)

Issue #2290, opened 2022-04-02
RMS error after commit 437739


alirmg:
I have just pulled the most recent RMG-Py and RMG-Database from git. when I run the superminimal example, I get the following error (full text below). (for both superminimal example and any other input files I generated) `NameError: name 'rms' is not defined `

mjohnson541:
The environment has been updated. You need to reinstall see: http://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/installation/anacondaDeveloper.html

IannLiu:
Hi, I got the same error after running the superminimal example. However, another error,
AttributeError: module 'julia.ReactionMechanismSimulator' has no attribute 'EmptyHenryLawConstant' occurred.

IannLiu:
Hi, I got the same error after running the superminimal example. However, another error,
AttributeError: module 'julia.ReactionMechanismSimulator' has no attribute 'EmptyHenryLawConstant' occurred after installing and linking Julia dependencies.

.

IannLiu:
> The environment has been updated. You need to reinstall see: http://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/installation/anacondaDeveloper.html
Hi, I got the same error after running the superminimal example. However, another error,
AttributeError: module 'julia.ReactionMechanismSimulator' has no attribute 'EmptyHenryLawConstant' occurred after installing and linking Julia dependencies.

mjohnson541:
Okay, in the environment run `julia` to open julia and then run 

`] add ReactionMechanismSimulator#main`

The `]` will disappear after you type it.

xiaoruiDong:
@ChrisBNEU Hi Chris, I remembered you brought up this (or something similar) during the developer hour this week (Apr 4, 2022). Can you update here with what you found and what troubleshooting you have done? If there is a PR related please link it as well. Many thanks!

ChrisBNEU:
So far, the problem appears to be that the ReactionMechanismSimulator package in julia does not track the github repo, despite it being specified in the pyrms wrapper. Have not started a PR, the farthest I have gotten with it is it might be due to the install step for julia: 
```
julia -e 'using Pkg; Pkg.add(PackageSpec(name="ReactionMechanismSimulator",version="0.4")); ...'
```
I think we can specify this instead: 
```
julia -e 'using Pkg; Pkg.add(PackageSpec(name="ReactionMechanismSimulator",rev="main")); ...'
```
[...]

PR #2305, opened 2022-05-09
update developer installation instructions to use the main rms branch

ChrisBNEU:
### Motivation or Problem
people are having trouble installing rmg because it is using an old rms version, see issue #2290

### Description of Changes
changed the documentation to use the rms main branch instead of v 0.4. 

### Testing
did a clean rmg install and verified that I could regenerate the issue with the old install, and that the issue was resolved with the updated instructions. 

### Reviewer Tips
Just verify that this is a reasonable way to do this. I think the root cause is the specification of "v 0.4" in the install supercedes any configuration we have in the pyrms wrapper config.

### DIFF
--- a/documentation/source/users/rmg/installation/anacondaDeveloper.rst
+++ b/documentation/source/users/rmg/installation/anacondaDeveloper.rst
@@ -119,7 +119,7 @@ Installation by Source Using Anaconda Environment for Unix-based Systems: Linux 
 
      python -c "import julia; julia.install(); import diffeqpy; diffeqpy.install()"
 
-     julia -e 'using Pkg; Pkg.add(PackageSpec(name="ReactionMechanismSimulator",version="0.4")); using ReactionMechanismSimulator;'
+     julia -e 'using Pkg; Pkg.add(PackageSpec(name="ReactionMechanismSimulator",rev="main")); using ReactionMechanismSimulator;'
 
    Note that this links your python to python-jl enabling calls to Julia through pyjulia. Occasionally programs will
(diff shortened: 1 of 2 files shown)


mjohnson541:
The instruction change looks good to me!

mjohnson541 (approved):

rwest:
Thanks Chris and Matt.
Should we update the instructions on [updating code](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/main/documentation/source/users/rmg/installation/updatingSourceCode.rst) so people can keep up to date with RMS? and is anything needed to "compile" it?

mjohnson541:
Yes. RMS doesn't require any special compilation step beyond updating the GitHub. However, the GitHub update step will look a bit different using the current instructions (since the GitHub repo it is setup to use there is managed by Julia. 

When I originally set it up I had assumed RMS would operate more as a dependency rather than a codeveloped repo. However, dual PRs with RMS are becoming somewhat common. So it might actually not be a bad idea to handle RMS at least in the developer install like we do RMG-database. When doing it that way I believe you would just add the repo to Julia as a path instead of a GitHub url.

ChrisBNEU:
I agree with Matt, It will take me a little bit but I can update the developer instructions to include setting up RMS independently, I have already done that successfully on my computer. I am fortunately working on an older OS for mac, I heard at the last hack-a-thon that people had trouble getting the local RMS build to work with the most recent mac os. that might take a little investigating.  

For now should I just update the "updating code" section and we can merge this? Just so external people don't keep getting the error from #2290

mjohnson541 (approved):
Looks good!

PR #2310, opened 2022-06-06
change shebang in update_headers to use python-jl instead of python3

sevyharris:
### Motivation or Problem
The script to automatically update the LICENSE header at the top of all files uses python3 instead of python-jl. This script is run when generating new binaries.

### Description of Changes
Changed the shebang in the utilities.py:update_headers function to use python-jl instead of python3


12Chao (commented):
looks good to me!

JacksonBurns:
Superseded by #2444

PR #2317, opened 2022-06-19
Transitory Edge Analysis

mjohnson541:
Adds transitory edge analysis species selection algorithm to RMG. RMS dual PR with https://github.com/ReactionMechanismGenerator/ReactionMechanismSimulator.jl/pull/196.

### DIFF
--- a/examples/rmg/nox_transitory_edge/input.py
+++ b/examples/rmg/nox_transitory_edge/input.py
@@ -1,0 +1,85 @@
+database(
+    thermoLibraries = ['BurkeH2O2','FFCM1(-)','primaryThermoLibrary','thermo_DFT_CCSDTF12_BAC','CBS_QB3_1dHR','DFT_QCI_thermo','primaryNS', 'NitrogenCurran'],
+    reactionLibraries = ['primaryNitrogenLibrary'],
+    seedMechanisms = ['BurkeH2O2inN2','FFCM1(-)'],
+    kineticsDepositories = 'default',
+    kineticsFamilies = 'default',
+    kineticsEstimator = 'rate rules',
+)
+
+species(
+    label='ethane',
+    reactive=True,	
+    structure=SMILES("CC"),
+)
+species(
+    label='O2',
+    structure=SMILES("[O][O]"),
+)
+species(
+    label='N2',
+    reactive=False,
+    structure=adjacencyList("""
+    1 N u0 p1 c0 {2,T}
+	2 N u0 p1 c0 {1,T}
+	"""),
+)
+
+species(
+   label='NO',
+   reactive=True,
+   structure=SMILES("[N]=O"),
+)
+
+constantTPIdealGasReactor(
+    temperature=(900,'K'),
+    pressure=(10.0,'bar'),
+    initialMoleFractions={
+        "N2": 4,
+        "O2": 1,
+        "ethane": 2.0/7.0,
+    },
+
+    terminationConversion={
+        'ethane': .90,
+    },
+   terminationTime=(40,'s'),
+)
+
+simulator(
+    atol=1e-16,
+    rtol=1e-8,
+)
+
+model(
+    toleranceMoveToCore=0.3,
+    toleranceInterruptSimulation=0.3,
+    maxNumObjsPerIter=1,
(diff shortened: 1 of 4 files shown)


rmgpy/rmg/reactors.py
```diff
@@ -463,10 +463,23 @@ def generate_reactor(self, phase_system):
         if self.evap_cond_conditions:
             kLA_kH_evap_cond = rms.kLAkHCondensationEvaporationWithReservoir(domain,self.evap_cond_conditions)
             interfaces.append(kLA_kH_evap_cond)
-            
+
         react = rms.Reactor(domain, y0, (0.0, self.tf), interfaces, p=p)
         return react, domain, interfaces, p
 
+class ConstantTPIdealGasReactor(Reactor):
```
hwpang, line 471: I wonder what is the best way to set up different reactors, is it better to make this reactor to support interfaces, or is it better to write new reactors with interfaces

rmgpy/rmg/input.py
```diff
@@ -1153,7 +1153,8 @@ def model(toleranceMoveToCore=None, toleranceMoveEdgeReactionToCore=np.inf, tole
           ignoreOverallFluxCriterion=False,
           maxNumSpecies=None, maxNumObjsPerIter=1, terminateAtMaxObjects=False,
           toleranceThermoKeepSpeciesInEdge=np.inf, dynamicsTimeScale=(0.0, 'sec'),
-          toleranceBranchReactionToCore=0.0, branchingIndex=0.5, branchingRatioMax=1.0):
+          toleranceBranchReactionToCore=0.0, branchingIndex=0.5, branchingRatioMax=1.0,
+          toleranceTransitoryDict={}):
```
hwpang: What is this for again? I'm confused, why do we need this?

examples/rmg/nox_transitory_edge/input.py
```diff
@@ -0,0 +1,85 @@
...
+model(
+    toleranceMoveToCore=0.3,
+    toleranceInterruptSimulation=0.3,
+    maxNumObjsPerIter=1,
+    terminateAtMaxObjects=True,
+    filterReactions=False,
+    #toleranceBranchReactionToCore=0.001,
+    #branchingIndex=0.5,
+    #branchingRatioMax=1.0,
+    toleranceTransitoryDict={'NO':0.2},
```
hwpang, line 63: Is 0.2 tolerance? Is it normalized?

rmgpy/rmg/reactors.py
mjohnson541: When I designed the system I had in mind that each domain+interfaces combination would be its own reactor. I think that keeps the input a bit simpler for the user. However, on the other hand if we start including a lot of these reactors enabling users to specify interfaces within a reactor instead of having multiple reactors might simplify the code a lot long term and make the system more flexible.

rmgpy/rmg/input.py
mjohnson541: toleranceTransitoryDict has the transitory sensitivity tolerance for each species of interest.

examples/rmg/nox_transitory_edge/input.py
mjohnson541: Yes, this is against the normalized transitory edge sensitivity as described in the paper.

mjohnson541:
@hwpang https://github.com/ReactionMechanismGenerator/ReactionMechanismSimulator.jl/pull/196 is merged now. I think we're ready here, can you approve?

rmgpy/rmg/reactors.py
```diff
@@ -463,10 +463,23 @@ def generate_reactor(self, phase_system):
...
+    def __init__(self, core_phase_system, edge_phase_system, initial_conditions, terminations, constant_species=[]):
+        super().__init__(core_phase_system, edge_phase_system, initial_conditions, terminations, constant_species=[])
```
hwpang, line 473: Can you add a line after this function? I also realize that having a mutable default argument isn’t the best, even though we are not mutating it later in the function.

mjohnson541: Fixed

hwpang (approved):
Looks good to me

PR #2319, opened 2022-06-23
Fix neg freq qchem parser

kspieks:
This PR updates our QChem parser to only return the negative frequency from the last frequency block. Currently, the parser returns the negative frequency from the first block, but as described below, this is not always the expected behavior and sometimes it returns the incorrect negative frequency.

As shown in Example 9.208 in the QChem [documentation](https://manual.q-chem.com/5.0/sect-equi.html), sometimes we want to run a frequency job first to obtain an analytical hessian that can be used for the subsequent optimization. During dataset generation, especially for TSs, I often chain 3 jobs together: frequency block to get analytical hessian, TS optimization, another frequency block to verify the optimized structure has a negative frequency. Thus, our QChem parser must be able to handle multiple frequency blocks and return the expected result.

Unit tests were added to test the updated method. Note that `ts004630_2_neg_freqs.log` is identical to `ts004630.log` except I manually changed one of the frequencies to be -999.99 so we could test that an error is raised if the logfile has 2 negative frequencies.


arkane/ess/qchem.py
```diff
@@ -380,17 +380,27 @@ def load_negative_frequency(self):
...
+                if 'VIBRATIONAL ANALYSIS' in line:
+                    read_freqs = True
+
+        logging.info(f'Identified {num_freq_blocks} frequency block(s)...')
+        logging.info('Only examining frequencies from the last block...')
+        neg_idx = np.where(freqs < 0)[0]
+        if len(neg_idx) == 1:
+            return freqs[neg_idx[0]]
+        elif len(neg_idx) > 1:
+            raise LogError('More than one imaginary frequency in QChem output file {0}.'.format(self.path))
```
alongd, line 402: Do we also raise these errors in other ESS adapters? What if we (rarely) care about a second order TS? Perhaps make this a warning? We could also think of making this behavior uniform throughout the ESS adapters.

alongd (commented):
Thanks for this feature, @kspieks!
I added one small question.

arkane/ess/qchem.py
kspieks: @alongd Thanks for reviewing! Currently, the ESS adapters only raise a `LogError` if there are 0 negative frequencies in the file. I initially pushed a change that allowed all ESS adapters to log a warning if there is >1 negative frequency. Unfortunately logging warnings causes our functional tests to [crash](https://github.com/ReactionMechanismGenerator/RMG-Py/runs/7063181776?check_suite_focus=true#step:11:2604
) so I changed it to `logging.info`. 

When reviewing, I'd recommend just looking at the "Files changed" tab. To recap my other fixup commits, I initially assumed the user would want to know the multiple negative frequencies for convenience. However, returning multiple negative frequencies is problematic since it makes the return object inconsistent and thus incompatible with portions of `arkane/statemech.py` that expect 1 negative frequency.

If needed, discussions about handling more than 1 negative frequency could be moved to a separate PR since the primary goal of this PR is to fix the incorrect parsing from the previous QChemLog in which it would return a frequency from an unoptimized geometry.

alongd:
Thanks @kspieks! Feel free to squash

alongd:
@kspieks, see the minor comment by LGTM?

kspieks:
@alongd Thanks again for reviewing! I squashed the commits so this should be good to go. Can you review and merge when you get a chance?

alongd (approved):
Thanks!

PR #2329, opened 2022-08-19
Fix resonance save order

xiaoruiDong:
### Motivation or Problem
In my previous PR (#2229), I was trying to fix cases where atom map changes even when setting `save_order=True` when generating resonance structures. The PR fixes most of the cases except for aromatic molecules that I recently ran into. The atom map is changed whenever RMG tries to analyze the aromaticity and find out potential aromatic rings; Inside `get_aromatic_rings`, RMG will sort the atom order of the molecule and then convert it to an RDKit Mol and do the analysis based on the RDKit Mol. There is a comment inside `to_rdkit_mol` and `to_ob_mol` saying `# sort the atoms before converting to ensure output is consistent between different runs`, this can be useful in the case not sensitive to atom map and unnecessary to the case where the input molecule atom map is critical and need to be preserved. 

### Description of Changes
I naively add `save_order` argument to the following functions:
- `molecule.molecule.Molecule.get_aromatic_rings`
- `molecule.molecule.Molecule.get_aryl_radicals`
- `molecule.converter.to_rdkit_mol`
- `molecule.converter.to_ob_mol`
- `molecule.resonance.analyze_mol`
- `molecule.resonance.generate_aromatic_resonance_structure`
- `molecule.resonance.generate_clar_structures`
- `molecule.resonance._clar_optimization`
- `molecule.fragment.Fragment.get_aromatic_rings`
- `molecule.fragment.Fragment.get_aryl_radicals`
- `molecule.fragment.Fragment.to_rdkit_mol`

I set the default value of all `save_order` to `False`, so the default behavior shouldn't be changed. 

### Testing
Unit tests are added for to_rdkit_mol, get_aromatic_rings, and generate_resonance_structures

### DIFF
--- a/rmgpy/molecule/converter.pxd
+++ b/rmgpy/molecule/converter.pxd
@@ -29,10 +29,10 @@ cimport rmgpy.molecule.molecule as mm
 cimport rmgpy.molecule.element as elements
 
 
-cpdef to_rdkit_mol(mm.Molecule mol, bint remove_h=*, bint return_mapping=*, bint sanitize=*)
+cpdef to_rdkit_mol(mm.Molecule mol, bint remove_h=*, bint return_mapping=*, bint sanitize=*, bint save_order=?)
 
 cpdef mm.Molecule from_rdkit_mol(mm.Molecule mol, object rdkitmol, bint raise_atomtype_exception=?)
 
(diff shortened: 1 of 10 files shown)


kspieks (approved):
Thanks for this important PR! The commit history and code is very clear. And thanks for adding a unit test as well. I ran it locally and it works fine so we can merge this PR

PR #2330, opened 2022-08-23
Add Arkane unit tests

kspieks:
I noticed that a few files in Arkane could have better unit test coverage. This short PR adds a few extra unit tests, specifically for getting atom and bond corrections as well as our molpro parser. We can work on adding additional unit tests at the next RMG hackathon :) 

Note that I set a very loose tolerance when checking AEC and BAC values since it's possible that these values are re-fit and thus changed in the future. I just wanted to add a test that makes sure this functionality doesn't break (even if the values themselves may change).

### DIFF
--- a/arkane/encorr/corrTest.py
+++ b/arkane/encorr/corrTest.py
@@ -1,0 +1,129 @@
+#!/usr/bin/env python3
+
+###############################################################################
+#                                                                             #
+# RMG - Reaction Mechanism Generator                                          #
+#                                                                             #
+# Copyright (c) 2002-2021 Prof. William H. Green (whgreen@mit.edu),           #
+# Prof. Richard H. West (r.west@neu.edu) and the RMG Team (rmg_dev@mit.edu)   #
+#                                                                             #
+# Permission is hereby granted, free of charge, to any person obtaining a     #
+# copy of this software and associated documentation files (the 'Software'),  #
+# to deal in the Software without restriction, including without limitation   #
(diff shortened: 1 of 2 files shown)


arkane/encorr/corrTest.py
```diff
@@ -0,0 +1,123 @@
...
+                                  ])
+        nums = (1, 1)
+
+        # test Petersson BACs
+        bac_type = 'p'
+        bac = get_bac(level_of_theory=self.freq_lot, bonds=bonds,
+                      coords=CCCBDB_coords, nums=nums,
+                      bac_type=bac_type,
+        )
+        self.assertAlmostEqual(bac, 700, places=None, delta=100)
```
xiaoruiDong, line 90: Can you also add a comment here about updating the value? 
`# The value can be updated by BAC(self.freq_lot, bac_type=bac_type).get_correction(bonds=bonds, coords=CCCBDB_coords, nums=nums).value_si`

Or we can also consider replacing the numeric value with the above function call, since we are basically testing if `get_bac` can reproduce the same results from the above function call. (and the above function call is tested in `bacTest.py`)

xiaoruiDong (commented):
@kspieks Thank you for adding more Arkane unit tests. The PR looks good. I just added a minor comment about providing hints on how to obtain the check value for the unit test.

kspieks:
@xiaoruiDong thanks for your review and helpful suggestions. I added the comments for improved readability

xiaoruiDong (approved):
Looks good!

PR #2340, opened 2022-10-23
Allow RMG to load thermo or kinetic libraries from a file

alongd:
### Motivation or Problem
Users who execute RMG on a server where it is installed in a read-only partition cannot add new libraries for their runs.

### Description of Changes
Added functionality to load thermo and kinetic libraries from a user specified location
Tests added

### DIFF
--- a/rmgpy/data/kinetics/database.py
+++ b/rmgpy/data/kinetics/database.py
@@ -236,10 +236,6 @@ class KineticsDatabase(object):
                     library.load(library_file, self.local_context, self.global_context)
                     self.libraries[library.label] = library
                 else:
-                    if library_name == "KlippensteinH2O2":
-                        logging.info("""\n** Note: The KlippensteinH2O2 library was replaced and is no longer available in RMG.
-For H2 combustion chemistry consider using either the BurkeH2inN2 or BurkeH2inArHe
-library instead, depending on the main bath gas (N2 or Ar/He, respectively)\n""")
                     raise IOError("Couldn't find kinetics library {0}".format(library_file))
 
         else:
--- a/rmgpy/data/kinetics/database.py
+++ b/rmgpy/data/kinetics/database.py
@@ -238,12 +238,12 @@ class KineticsDatabase(object):
                     library.load(library_file, self.local_context, self.global_context)
                     self.libraries[library.label] = library
                 elif os.path.exists(library_file):
-                    logging.info('Loading kinetics library {0} from {1}...'.format(library_name, library_file))
+                    logging.info(f'Loading kinetics library {library_name} from {library_file}...')
                     library = KineticsLibrary(label=library_name)
                     library.load(library_file, self.local_context, self.global_context)
                     self.libraries[library.label] = library
                 else:
-                    raise IOError("Couldn't find kinetics library {0}".format(library_file))
+                    raise IOError(f"Couldn't find kinetics library {library_file}")
 
         else:
             # load all the libraries you can find
@@ -255,7 +255,7 @@ class KineticsDatabase(object):
                     if ext.lower() == '.py':
                         library_file = os.path.join(root, f)
                         label = os.path.dirname(library_file)[len(path) + 1:]
-                        logging.info('Loading kinetics library {0} from {1}...'.format(label, library_file))
+                        logging.info(f'Loading kinetics library {label} from {library_file}...')
                         library = KineticsLibrary(label=label)
                         try:
                             library.load(library_file, self.local_context, self.global_context)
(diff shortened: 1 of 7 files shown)


rmgpy/data/kinetics/database.py
```diff
@@ -230,7 +230,14 @@ def load_libraries(self, path, libraries=None):
...
                 library_file = os.path.join(path, library_name, 'reactions.py')
-                if os.path.exists(library_file):
+                if os.path.exists(library_name):
+                    library_file = os.path.join(library_name, 'reactions.py')
+                    short_library_name = os.path.split(library_name)[-1]
+                    logging.info(f'Loading kinetics library {short_library_name} from {library_name}...')
+                    library = KineticsLibrary(label=short_library_name)
+                    library.load(library_file, self.local_context, self.global_context)
+                    self.libraries[library.label] = library
+                elif os.path.exists(library_file):
```
xiaoruiDong, line 240: Can you check if there is a need to update `library.library_order`?

At `rmgpy/data/rmg.py` [line 165](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/76e68348c0b3879c4f964787fee0f76ab3b4b899/rmgpy/data/rmg.py#L165), items in the `kinetics_libraries` will be assigned as `(library, 'Reaction Library'))`. Therefore, if a file path is provided, this will be `(PATH, 'Reaction Library')`. 
At `rmgpy/data/kinetics/database.py` [line 438](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/76e68348c0b3879c4f964787fee0f76ab3b4b899/rmgpy/data/kinetics/database.py#L438-L442), the script iterates through `library_order` and uses the first elements (the `PATH`) as the label to query the library (`self.libraries[label]`), but the change here redefines `label` as `short_library_name`. So, it is possible to cause a KeyError.

xiaoruiDong (commented):
@alongd Thank you for adding this feature, and I agree that this feature is handy! I just had one comment for you to address. Otherwise, it looks great.

rmgpy/data/kinetics/database.py
alongd: Thanks for bringing up these important points, I did not think of them originally.

I think we can either take an approach of converting an external library name (which is a path at first) to the actual label when loading the object, and only working with labels and not paths downstream, or we can retain the full path to an external library as its name within RMG.
The first approach is cleaner or at least has shorter names, but the second might be more robust if users make an external library that has the same name as an internal one (e.g., testing an updated existing library).

If we take the second approach, which I believe is currently implemented in this PR, then I think that the current implementation is OK as is. Here's the train of calls, as I see it, I list it in detail just to help me sort it out :)
 
**1.** We load the database, e.g., as in `rmgpy/rmg/main.py` [line 374](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/main/rmgpy/rmg/main.py#L374)
**2.** We call `load_kinetics()` in `rmgpy/data/rmg.py` in [line 99](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/main/rmgpy/data/rmg.py#L99), passing the database path for the `kinetics` folder, and a list of library names as `reaction_libraries`.
**3.** We append `(library, 'Reaction Library')` tuples in `rmgpy/data/rmg.py` [line 165](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/main/rmgpy/data/rmg.py#L165)
**4.** Using this correct library order, we then turn to load the libraries on `rmgpt/data/kinetics/database.py` [line 230](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/main/rmgpy/data/kinetics/database.py#L230), which is the block this PR modifies.

I think that the modifications to process an external library we did here are downstream to setting the library order, hence it is OK and the order set by the user will still be preserved.

I see the concern that we carry both the database `PATH` as well as the specific path of an external library as the library name. The `load_libraries()` function modified here is in charge of loading the specific library correctly if it represents a complete external path, and in this case the db `PATH` is ignored.

This branch runs as expected: I copied an existing RMG library (e.g., FFCM1) into my run folder, here's part of the RMG.log file:
```
Loading kinetics library GRI-Mech3.0-N from /Data/Code/RMG-database/input/kinetics/libraries/GRI-Mech3.0-N/reactions.py...
Loading kinetics library FFCM1(-) from /Data/Code/runs/debug/RMG_ext_lib/FFCM1(-)...
```

I tried loading an external seed mechanism, and it also works as expected.

I'll add another commit of an example that shows potential users how to use this feature.

xiaoruiDong: Thanks for the detailed explanation and the addition of the example. I double check with the workflow, and it makes sense with the second approach.

xiaoruiDong (approved):
Looks good. Please rebase the branch and I will merge it!

xiaoruiDong (approved):

PR #2345, opened 2022-11-13
Save the actual reaction label when creating a reaction library in Arkane

alongd:
When Arkane saves a kinetics library it currently passes `rxn.label` for the reaction label.
However, in an RMG run PDep networks generated automatically have simple labels such as `reaction1` instead of a chemically meaningful label. Then, running such `network_x.py` files through Arkane will generate an ill-formatted RMG library.
Here we fix the reaction label in the RMG libraries generated by an Arkane kinetics job by using a reasonable reaction label.


kspieks (approved):
This change looks great. Thanks for the PR!

PR #2348, opened 2022-11-28
Make update

kfir4444:
### Motivation or Problem
I added a command to make for automating the updates for users that installed rmg from source. This is due to people having issues with forgetting to recompile after pulling. I think this process is a bit more user friendly, especially for the python users that sometimes forget to compile (myself included).

### Description of Changes
I added a command to the Makefile for automatic updates, and added an appropriate explanation to the documentations.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -145,6 +145,29 @@ scoop: all
 	@ echo "Running minimal example with SCOOP"
 	python -m scoop -n 2 rmg.py -v testing/scoop/input.py
 
+.ONESHELL:
+update:
+	@ echo "Updating RMG-py"
+	@ echo "Please remember to stash unwanted changes."
+	git status
+	@ read -p "Are there any changes that are not staged for committed? [y|n] " user_input
+	if [ "$$user_input" = "N" ] || [ "$$user_input" = "n" ]; then
+		git checkout main || true
+		git pull git@github.com:ReactionMechanismGenerator/RMG-Py.git main || git pull https://github.com/ReactionMechanismGenerator/RMG-Py.git main || true
+		conda activate rmg_env
+		make
+		conda deactivate
+		exit
+	elif [ "$$user_input" = "Y" ] || [ "$$user_input" = "y" ]; then
+		@ echo "Please stash using 'git stash' and run this again"
+		@ echo "After the operation is done, use 'git stash pop'"
+		@ echo "Aborting..."
+		exit
+	fi
+	@ echo Wrong input: $$user_input
+	@ echo "Aborting..."
+	exit
+
 ######### 
 # Section for setting up MOPAC calculations on the Travis-CI.org server
 ifeq ($(TRAVIS),true)
--- a/documentation/source/users/rmg/installation/updatingSourceCode.rst
+++ b/documentation/source/users/rmg/installation/updatingSourceCode.rst
@@ -7,7 +7,14 @@ Updating the RMG-Py Source Code
 It is recommended to keep yourself up to date with the latest patches and bug fixes by RMG developers,
 which is maintained on the official repository at https://github.com/ReactionMechanismGenerator/RMG-Py/ 
 You can view the latest changes by viewing the commits tab on the repository.  
-To update your source code, you can "pull" the latest changes from the official repo by typing the following command in the
+
+To update your local RMG-Py source code automatically, you can use the following commands::
+    
+    cd /path/to/RMG-Py
+    conda activate rmg_env
+    make update
+
+To update your source code manually, you can "pull" the latest changes from the official repo by typing the following command in the
 Command Prompt ::
 
     cd RMG-Py


Makefile
```diff
@@ -145,6 +145,13 @@ scoop: all
 	@ echo "Running minimal example with SCOOP"
 	python -m scoop -n 2 rmg.py -v testing/scoop/input.py
 
+update: all
+	@ echo "Updating RMG-py"
+	git stash || true
+	git pull git@github.com:ReactionMechanismGenerator/RMG-Py.git main || git pull https://github.com/ReactionMechanismGenerator/RMG-Py.git main || true
```
alongd, line 151: Can you explain why pulling twice?

Makefile
```diff
@@ -145,6 +145,13 @@ scoop: all
...
+	make
```
alongd, line 152: Do we assume the Fonda env is activated?

Makefile
```diff
@@ -145,6 +145,13 @@ scoop: all
...
+	git stash pop || true
```
alongd, line 153: Users should know that this could potentially lead to conflicts if the stashed changes change lines that were updated

Makefile
```diff
@@ -145,6 +145,13 @@ scoop: all
 	@ echo "Running minimal example with SCOOP"
 	python -m scoop -n 2 rmg.py -v testing/scoop/input.py
 
+update: all
+	@ echo "Updating RMG-py"
+	git stash || true
```
alongd, line 150: If this is made for people who forget, we also don’t know they didn’t forget to checkout main before updating. Don’t know what to do about it, just bringing this up

documentation/source/users/rmg/installation/updatingSourceCode.rst
```diff
@@ -7,7 +7,14 @@ Updating the RMG-Py Source Code
 It is recommended to keep yourself up to date with the latest patches and bug fixes by RMG developers,
 which is maintained on the official repository at https://github.com/ReactionMechanismGenerator/RMG-Py/ 
 You can view the latest changes by viewing the commits tab on the repository.  
-To update your source code, you can "pull" the latest changes from the official repo by typing the following command in the
+
+To update your local RMG-Py source code automatically, you can use the following commands::
+    
+    cd RMG-Py
```
alongd, line 13: To we want to say that they should consider the location where RMG is installed?

Makefile
```diff
@@ -145,6 +145,13 @@ scoop: all
 	@ echo "Running minimal example with SCOOP"
 	python -m scoop -n 2 rmg.py -v testing/scoop/input.py
 
+update: all
```
alongd, line 148: What does all mean?

alongd (commented):
@kfir4444 , thanks for the contribution!
I added a few points for discussion (some where sent as individual emails, sort)

Makefile
kfir4444: sure, either using `https` or `ssh`.

Makefile
kfir4444: It has to be active if the user used `make` to activate the `make update`

Makefile
alongd: How was this comment resolved?

documentation/source/users/rmg/installation/updatingSourceCode.rst
alongd: Should add something like `path/to` before

Makefile
```diff
@@ -145,6 +145,29 @@ scoop: all
 	@ echo "Running minimal example with SCOOP"
 	python -m scoop -n 2 rmg.py -v testing/scoop/input.py
 
+.ONESHELL:
+update:
+	@ echo "Updating RMG-py"
+	@ echo "Please remember to stash unwanted changes."
+	git status
+	@ read -p "Are there any changes that are not staged for committed? [y|n] " user_input
```
ChrisBNEU, line 153: should it be "staged *or* committed" or "staged *for commit*"?

Makefile
```diff
@@ -145,6 +145,29 @@ scoop: all
...
+	if [ "$$user_input" = "N" ] || [ "$$user_input" = "n" ]; then
```
ChrisBNEU, line 154: does this work on your system? because I have the following error when I select y or n: 
```
(rmg_env) blais.ch@FACC02D75Y4MD6T RMG-Py % make update
Updating RMG-py
Please remember to stash unwanted changes.
git status
On branch make_update
Your branch is up to date with 'origin/make_update'.

nothing to commit, working tree clean
Are there any changes that are not staged for committed? [y|n] y
if [ "$user_input" = "N" ] || [ "$user_input" = "n" ]; then
/bin/sh: -c: line 1: syntax error: unexpected end of file
make: *** [update] Error 2
```
this message appears in both zsh and bash. I assume it has to do with mixing make and bash?

Makefile
```diff
@@ -145,6 +145,29 @@ scoop: all
...
+		git checkout main || true
+		git pull git@github.com:ReactionMechanismGenerator/RMG-Py.git main || git pull https://github.com/ReactionMechanismGenerator/RMG-Py.git main || true
+		conda activate rmg_env
```
rwest, line 157: Would be nice if the environment name weren't hard coded. Mine has a different name, and sometimes I have more than one. 
Perhaps instead have a check that RMG (or its dependencies) are available in the current environment, and a warning to activate the right environment if it is not.

You could probably just add `check` as a dependency for this target? (I think that checks all the python dependencies)

PR #2349, opened 2022-12-02
Added Arkane examples

alongd:
Added thermo and rate Arkane examples


alongd:
This PR enriches our examples for running Arkane.
The addition here show how to define a modelChemistry which is not CBS-QB3 (where one has to actually use the `LevelOfTheory` class and also provide the name of the software that was used), there are no other current thermo examples like that. Also, the existing kinetics example (`NH2 + N2H3 = NH + N2H4`) is based on YAML files, which is not a typical use case (one only gets these YAML files after running Arkane for the first time). All the examples added here mimic "field use cases" where users ask Arkane to parse directly from their QM log files.

kspieks (approved):
Thanks for adding this PR with valuable examples for running Arkane! It looks good to me so I'll merge now

sthakris:
@alongd @kspieks : It it possible to calculate  the rate constant of barrierless reaction with arkane.

PR #2359, opened 2023-01-20
add helpful error message for failed julia dependency imports

JacksonBurns:
### Motivation or Problem
Currently when RMG tries to import the Julia dependencies, a failed import is simply [ignored](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/main/rmgpy/rmg/reactors.py#L38) due to testing limitations.

### Description of Changes
Issues a warning if the import of the Julia dependencies fails, but only when not in the testing mode.

### Testing
I built an environment without the Julia dependencies properly built to verify that the warning would be issued.

### Reviewer Tips
The message should be more helpful, maybe pointing toward the install directions or some relevant GitHub issues.


JacksonBurns:
I had intended this to catch the exception but specified the wrong type, it should be catching a `julia.core.UnsupportedPythonError` instead of an `ImportError`. I'm not sure if we will be able to get to the former since the import breaks inside nose tests, leaving us with the option of making the except bare? Not great practice but maybe all that's possible here.

JacksonBurns:
I welcome input on whether this should raise a warning or an error, as well as what the content of the message should be.

I have also added a test file that bypasses the issue with nosetests and rms being incompatible. The python docs provide a great example on [patching `sys.modules`](https://docs.python.org/3/library/unittest.mock.html#:~:text=Keywords%20can%20be,...%0A%27fish%27) that I have used here.

JacksonBurns:
...sigh, this was working on my local install. Will keep looking at it.

JacksonBurns:
This will be resolved in #2443

PR #2366, opened 2023-01-30
Switch CI & Documentation Build to Micromamba from Conda for Reduced Execution Time

### LINKED ISSUE #2365: Switch CI + Documentation Build to use Mamba instead of Conda
### Motivation or Problem
CI and Documentation GitHub workflows take 1 hour+ to run in some cases - about half of this time is spent configuring the conda environment, solving the dependencies especially.

### Desired Solution
Switch the package resolved to Mamba in the workflows, [as shown in here](https://github.com/mamba-org/provision-with-micromamba).

### Potential Alternatives
Build times on CI aren't the biggest problem in the world but this is a "nice to have".

JacksonBurns:
### Motivation or Problem
Resolves #2365 

### Description of Changes
Changes the GitHub CI to use micromamba instead of conda, which resolves the RMG environment in ~2 mins instead of 30+

### Testing
Successfully ran on [this fork](https://github.com/JacksonBurns/RMG-Py/actions/runs/4046494335/jobs/6959351285).

### Reviewer Tips
The fact that it runs is perhaps sufficient evidence. Might want to do a line-by-line comparison of how the depdencies were resolved to ensure that versions are consistent with conda.

The changes, if good, will also need to be ported into the documentation builder.


JacksonBurns:
The debug level logging in the call to preovision-micromamba should also be removed once the rest of this looks operational.

JacksonBurns:
I have added the mamba build to the documentation action, as well as the runner schedule to the CI action.

I did a comparison between the environment resolved by micromamba and conda (see it for yourself [here](https://www.diffchecker.com/3Njctgy0/) and it resolves almost everything to be the same, with the exception of some operating system level packages that I assume are a function of the action being run on a different machine + VM at GitHub every time. All of our 'major' dependencies (those that we maintain or use heavily) are resolved to the correct major versions.

Tests are still passing and this cuts ~30 minutes off CI and Documentation build.

.github/workflows/CI.yml
```diff
@@ -24,7 +24,7 @@ jobs:
           environment-name: rmg_env
           channels: defaults,rmg,conda-forge
           channel-priority: flexible
-          log-level: debug
+          log-level: warning
```
hwpang, line 30: Why do we reset log level to `warning` here?

hwpang: Can you squash this with the commit that set it to debug?

.github/workflows/CI.yml
```diff
@@ -3,7 +3,7 @@ name: CI Tests
 on:
   schedule:
     # * is a special character in YAML so you have to quote this string
-    - cron:  '0 0 8 ? * MON,TUE,WED,THU,FRI *'
+    - cron:  '0 8 * * 1-5'
```
hwpang, line 6: Could you squash this commit with the previous one?

hwpang (changes requested):
Thanks for the PR! Overall seems good with a few comments. Several of the commits can be squashed together. I check [the fork](https://github.com/JacksonBurns/RMG-Py/actions/runs/4046494335/jobs/6959351285), and it looks like it fails `Trigger RMG-tests`?

.github/workflows/CI.yml
JacksonBurns: Debug writes more to stdout than is useful for any future test running (see [this run with debug](https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/4097452798/jobs/7065836940#step:3:82)) -- I only turned it on to deal with the issue of channel-priority. The default log level is warning

JacksonBurns:
> Thanks for the PR! Overall seems good with a few comments. Several of the commits can be squashed together. I check [the fork](https://github.com/JacksonBurns/RMG-Py/actions/runs/4046494335/jobs/6959351285), and it looks like it fails `Trigger RMG-tests`?

I left a quick comment about the logging level. Will squash after it is resolved.

That action fails on forked repositories regardless of content. See [here](https://github.com/JacksonBurns/RMG-Py/actions/runs/4067828613/jobs/7005648462), where I changed nothing in the source code and it still fails at the same location.

JacksonBurns:
Moved to #2386

PR #2374, opened 2023-02-24
Small fix for explorer tool

mjohnson541:
Skip adding energy_correction to  barriers in explorer tool if it isn't defined yet.

### DIFF
--- a/arkane/explorer.py
+++ b/arkane/explorer.py
@@ -250,9 +250,15 @@ class ExplorerJob(object):
             for rxn in network.path_reactions:
                 if rxn.transition_state is None:
                     if rxn.network_kinetics is None:
-                        E0 = sum([spec.conformer.E0.value_si for spec in rxn.reactants]) + rxn.kinetics.Ea.value_si + network.energy_correction
+                        if network.energy_correction:
+                            E0 = sum([spec.conformer.E0.value_si for spec in rxn.reactants]) + rxn.kinetics.Ea.value_si + network.energy_correction
+                        else:
+                            E0 = sum([spec.conformer.E0.value_si for spec in rxn.reactants]) + rxn.kinetics.Ea.value_si
                     else:
-                        E0 = sum([spec.conformer.E0.value_si for spec in rxn.reactants]) + rxn.network_kinetics.Ea.value_si + network.energy_correction
+                        if network.energy_correction:
+                            E0 = sum([spec.conformer.E0.value_si for spec in rxn.reactants]) + rxn.network_kinetics.Ea.value_si + network.energy_correction
+                        else:
+                            E0 = sum([spec.conformer.E0.value_si for spec in rxn.reactants]) + rxn.network_kinetics.Ea.value_si
                     rxn.transition_state = rmgpy.species.TransitionState(conformer=Conformer(E0=(E0 * 0.001, "kJ/mol")))
                 
         for network in self.networks:
--- a/arkane/explorer.py
+++ b/arkane/explorer.py
@@ -249,16 +249,13 @@ class ExplorerJob(object):
         for network in self.networks:
             for rxn in network.path_reactions:
                 if rxn.transition_state is None:
+                    E0 = sum([spec.conformer.E0.value_si for spec in rxn.reactants])
                     if rxn.network_kinetics is None:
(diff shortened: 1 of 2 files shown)


xiaoruiDong (approved):
Looks good

xiaoruiDong:
I have a more general question.

Is there a reason why we don't automatically return `E0` as `self.to_wilhoit().E0` in `NASA` and `ThermoData` by defining setter and getter functions if `E0` is not provided in `NASA` and `ThermoData` initialization?

For the second commit, I prefer modifying `set_e0_with_thermo` to include updating `thermo.E0`, which only affect statmech related jobs. Something like

```
    def set_e0_with_thermo(self):
        """
        Helper method that sets species' E0 using the species' thermo data
        """
        if self.get_thermo_data().E0 is not None:
            self.conformer.E0 = self.get_thermo_data().E0
        else:
            if not self.thermo.Cp0 or not self.thermo.CpInf:
                # set Cp0 and CpInf
                from rmgpy.data.thermo import find_cp0_and_cpinf
                find_cp0_and_cpinf(self, self.thermo)
            E0 = self.get_thermo_data().to_wilhoit().E0  # modification
            self.thermo.E0 = E0   # modification.
            self.conformer.E0 = E0   # modification
```

mjohnson541:
It absolutely would be better to overhaul how Species handles E0. However, that affects a lot more systems and has a lot more likelihood of introducing new bugs than the commit I use here. I've recently found pdep plays some games with shifting E0 values and I'm unsure how introducing a setter might interact with that and other systems. It might not be that hard...but I don't really have time to do all the testing I think would be necessary right now.

xiaoruiDong:
> It absolutely would be better to overhaul how Species handles E0. However, that affects a lot more systems and has a lot more likelihood of introducing new bugs than the commit I use here. I've recently found pdep plays some games with shifting E0 values and I'm unsure how introducing a setter might interact with that and other systems. It might not be that hard...but I don't really have time to do all the testing I think would be necessary right now.

I agree that tailoring getter and setter functions are beyond the scope and have some potential risks to other systems.

For the second commit, I suggest doing
```
 H0 = sum([spec.get_thermo_data().E0.value_si if spec.get_thermo_data().E0 is not None else spec.conformer.E0.value_si for spec in self.products]) \
             - sum([spec.get_thermo_data().E0.value_si if spec.get_thermo_data().E0 is not None else spec.conformer.E0.value_si for spec in self.reactants])
```
to avoid repeatedly converting thermodata object to the Wilhoit object. I double-checked that all species in both Arkane and RMG runs will have `spec.conformer.E0` existing. Unit tests on the change I suggested can be found at (https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/4269016340/jobs/7431896195)

xiaoruiDong:
Please also add a functional test for the arkane runs using `proxyl+O2-1` and `PMEoxy-1`. We can check if the values match what you reported in the paper, but if you have any concerns writing the test this way, at least we should ensure these jobs can run without any issues in future RMG releases.

xiaoruiDong:
Just for future reference, this PR tries to solve the issues raised while running `arkane.py` with the following as inputs
[PMEoxy-1.py.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/10842854/PMEoxy-1.py.txt)
[PMEoxy-full.py.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/10842857/PMEoxy-full.py.txt)
[propyl+O2-1.py.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/10842858/propyl%2BO2-1.py.txt)
[propyl+O2-full.py.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/10842862/propyl%2BO2-full.py.txt)

rwest:
See if you like my suggested commit. Yours look good to me.

rwest:
Was the suggestion to use spec.conformer.E0 instead of to_wilhoit() a good one? Or should we just merge this and move on?

mjohnson541:
I'm pretty sure I tried to do that at one point and spec.conformer.E0 isn't assigned there or at least not reliably assigned. I think it best that we just merge this and have the bugs fixed.

PR #2381, opened 2023-03-02
Creating and utilising a custom system image of Julia with RMS

Laxzal:
I have changed the channel for pyjulia now, so it will look for it in conda-forge. Addtionally, I have added in an `if exists` statement in `reactors.py` to determine whether `rms.so` exists. Finally, I have also updated the instructions on the website of RMG to explain to the user to run the system image create command.


JacksonBurns:
Rebasing with fixed CI

rwest:
@calvinp0 please could you update the description at the top of this pull request explaining the motivation, some background, and how this helps? Many of us (well, me at least!) are not as familiar with Julia (and PyJulia) and are trying to learn. Thanks.


Also, I note that the environment file currently in this PR still installs `julia` itself from the RMG channel (even though `pyjulia` comes from conda-forge). Can we move `julia` to the conda-forge channel too?

calvinp0:
> Thanks. I have been trying to follow along with the instructions (or some form of them). On an intel mac, with a conda environment containing (among other things)
> 
> ```
> julia                     1.7.2                h132cb31_0    conda-forge
> pyjulia                   0.6.1              pyhd8ed1ab_0    conda-forge
> diffeqpy                  1.1.0                     rmg_2    rmg
> pyrms                     0.1.2                no_rmg_dep    rmg
> ```
> 
> I get (buried in a big error message)
> 
> ```
> DifferentialEquations [0c46a032-eb83-5123-abaf-570d42b7fbaa]
> 
> Error: Missing source file for DifferentialEquations [0c46a032-eb83-5123-abaf-570d42b7fbaa]
> DiffEqBase [2b5f629d-d688-5b77-993f-72d75c75574e]
> 
> Error: Missing source file for DiffEqBase [2b5f629d-d688-5b77-993f-72d75c75574e]
> ```
> 
> during the `python -m julia.sysimage rms.so` step

I don't have access to an intel mac but I could try on my M1 using Rosetta and see what happens.

I assume you have already done the following command:

```
python -c "import julia; julia.install(); import diffeqpy; diffeqpy.install()"

julia -e 'using Pkg; Pkg.add(PackageSpec(name="ReactionMechanismSimulator",rev="main")); using ReactionMechanismSimulator;'
```

If so, was it recent? Because usually when I get Julia errors like this, I would rebuild rmg_env from scratch (although not delete the  RMG-Py folder) and then re-run those two commands. If you are getting the same error, I can see if I can find an image of MacOS and install it on a virtual machine

rwest:
Checking chapter 5 of the [pyJulia documentation](https://pyjulia.readthedocs.io/_/downloads/en/latest/pdf/) suggests it might avoid the need to use `python-jl` to launch python?
Also from those doc
> ## 3.1 Your Python interpreter is statically linked to libpython
>  If you use Python installed with Debian-based Linux distribution such as Ubuntu or install Python by conda, you might have noticed that PyJulia cannot be initialized properly out-of-the-box. This is because those Python executables are statically linked to libpython
> ### 3.1.2 Create a custom system image
>  New in version 0.4.
> A very powerful way to avoid this the issue due to precompilation cache is to create a custom system image. This also has  an additional benefit that initializing PyJulia becomes instant. See Custom Julia system image for how to create and use a  custom system image.
> ### 3.1.3 python-jl: an easy workaround New in version 0.2.
> Another easy workaround is to use the python-jl command bundled in PyJulia

If we're not needing `python-jl` then we wouldn't need the customized version of it (https://github.com/ReactionMechanismGenerator/conda-recipes/commit/b3f6136e9b0ae6493696f0c2753cb2f9ebf245cb ) that we were hosting on the rmg conda channel. I think this could address some of what @JacksonBurns  was dealing with on https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2415

The apparent time-to-launch benefit is attractive too.
@calvinp0 do you think this approach could help? Can we re-open the PR?

PR #2382, opened 2023-03-02
Add initial species in core before seed mechanism 

hwpang:
### Motivation or Problem
The current initialization would fail if one uses a pyrms-based reactor with seed mechanism and the seed mechanism contains species that are the initial species. It would fail with an error related to the initial species not in the `phase.names`.

### Description of Changes
I moved the code block for initializing the bath gases and the initial species in front of the seed mechanism code block.

### DIFF
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -614,6 +614,25 @@ class RMG(util.Subject):
 
         # Initialize reaction model
 
+        # Also always add in a few bath gases (since RMG-Java does)
+        for label, smiles in [('Ar', '[Ar]'), ('He', '[He]'), ('Ne', '[Ne]'), ('N2', 'N#N')]:
+            molecule = Molecule().from_smiles(smiles)
+            spec, is_new = self.reaction_model.make_new_species(molecule, label=label, reactive=False)
+            if is_new:
+                self.initial_species.append(spec)
+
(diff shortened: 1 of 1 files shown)


hwpang:
A bit more detailed explanation for future reference. When RMG tries to add pressure dependent reactions in the seed mechanism to core that are pressure dependent on the initial species, the code crashes when trying to add it into the core phase system. The phase needs the index of that initial species and crashes if the initial species is not in `phase.names`.

mjohnson541 (dismissed):

JacksonBurns:
Rebasing with updated CI

JacksonBurns:
@hwpang this looks simple enough - want to try and revive this one?

hwpang:
I will close this PR for now as I don't have time to work on it.

JacksonBurns:
Blocked by #2514

sevyharris:
This issue is still causing problems for me, so I'm gonna take another stab at it.

sevyharris:
Okay, this is interesting. I spent some time debugging and it looks like the problem I was running into was actually the HTMLWriter changing some of the species labels in the edge. If you just set generateOutputHTML=False in the input file, it doesn't hit this error. I will see if I can reproduce the error Hao-Wei was getting specific to pressure dependence. Otherwise, I will add a correction for the HTMLWriter and call it a day.

PR #2383, opened 2023-03-02
Added Boron into element and atom types

ssun30:
### Motivation or Problem
Add new atom types for Boron to RMG to avoid atomtype error when generating mechanism for boron-related reactions. 

### Description of Changes
Added B and Val3 atomtypes and BDE numbers for Boron.


JacksonBurns:
The linked issue would be resolved by this PR, since use is attempting to generate a mechanism with Boron.

Issue #2392, opened 2023-03-03
Nosetests on Mac fail


ChrisBNEU:
### Bug Description
After PR #2364, on Mac os exclusively, running `make test` produces a segmentation fault: 
```
mkdir -p testing/coverage
rm -rf testing/coverage/*
nosetests --nocapture --nologcapture --all-modules -A 'not functional' --verbose --with-coverage --cover-inclusive --cover-erase --cover-html --cover-html-dir=testing/coverage --exe rmgpy arkane
ERROR: make: *** [test] Segmentation fault: 11
```

### How To Reproduce
install RMG as specified in our [install guide](http://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/installation/anacondaDeveloper.html) following mac os instructions, then run `make test`

### Expected Behavior
Expect it to run the unit tests

### Installation Information
Describe your installation method and system information.
 - mac os catalina v10.15.7
   - Apple clang version 12.0.0 (clang-1200.0.32.2)
 - RMG version
   - RMG-Py commit: d46a43 (Mar 2)
   - RMG-database commit: f058b7 (Mar 2)

### Additional Context
When I comment out the code below from "reactors.py" I am able to run the unit tests: 
```
# if __debug__:
#     try:
#         from os.path import dirname, abspath, join, exists
#         path_rms = dirname(dirname(dirname(abspath(__file__))))
#         from julia.api import Julia
#         jl = Julia(sysimage=join(path_rms,"rms.so")) if exists(join(path_rms,"rms.so")) else Julia(compiled_modules=False)
#         from pyrms import rms
#         from diffeqpy import de
#         from julia import Main 
#     except:
#         pass
# else:
try:
    from pyrms import rms
    from diffeqpy import de
    from julia import Main
except:
    pass
```

kfir4444:
hey @JacksonBurns, @ChrisBNEU
Maybe the root cause of that issue is nosetests? Nosetests is deprecated, and has been so for a long time (can't find specific deprecation time, but we moved from it in ARC around a year and a half ago). I suspect that more and more issues will rise while using nose, and suggest that you use unittest instead. Here is a merged [PR ](https://github.com/ReactionMechanismGenerator/ARC/pull/483) on ARC.

ChrisBNEU:
[...]

PR #2394, opened 2023-03-03
CI Updates: Run on Mac (and Ubuntu)

JacksonBurns:
There are issues running the unittests on mac, so we should try and reproduce the error in the CI. See #2392 and an alternate 'bandaid' solution #2393


JacksonBurns:
@ChrisBNEU I have tried to run this twice now, and then Julia install step takes so long on mac-os that the GitHub kills the runner. I am going to try breaking the install across multiple steps to get it to run at all, but I am not optimistic.

JacksonBurns:
Closing and migrating to a branch on RMG-Py

PR #2396, opened 2023-03-05
Added 'r' at the start of regex pattern strings where it was missing

alongd:
Added the `r` char at the start of regex pattern strings where it was missing. It designates a python "raw" string which passes through backslashes without change. It's recommend to always write Python regex pattern strings with the `r` char. (see [here](https://developers.google.com/edu/python/regular-expressions))

Practically, it causes many warning messages when debugging (I think Python can do without it, but it rightfully complaints)


kspieks (approved):
Thanks for this PR! LGTM

PR #2398, opened 2023-03-14
Correct Troe kinetics formula

xiaoruiDong:
There seems to be a missing term in the Troe kinetics formula in RMG. This PR aims to correct it.

Reference for the formula
- https://cantera.org/science/kinetics.html
- https://cefrc.princeton.edu/sites/g/files/toruqf1071/files/Files/2014%20Lecture%20Notes/Green/Combustion-Chemistry-part-3a.pdf


agi-jaguar (dismissed):
Very good! I missed this C factor before!

rwest:
A bit embarrassing this has been wrong for [over 10 years](https://github.com/ReactionMechanismGenerator/RMG-Py/commit/710fb2850aa0361899f3848f593b4db3bff7dca5) without anyone noticing. Could we have had a unit test?  Not sure what testing method would have helped here. Compare with a "blessed" answer from Cantera? (trying to think in general how to catch this *type* of mistake).

agi-jaguar:
@rwest I think one reason why it has been wrong for over 10 years is that this C factor doesn't significantly change the plot. All this sudden motivation for me to look through the database comes from my current project at JPL.

**Background:** Astrochemists have been using reaction kinetic modeling as a part of 1-D radiative transfer atmospheric photochemical modeling of exoplanet atmospheres. However, the way they are building their own reaction kinetic networks is literally manually taking all those rate coefficients from existing literature. So I am trying to implement RMG into existing various atmospheric photochemical model codes.)

The first step would be comparing _manually taken reaction rate coefficients by astrochemists_ to _RMG-generated rate coefficients_. I found out that usually RMG-generated rate coefficients are better, but sometimes I found huge discrepancies between these two, and that is when I realize some errors in the RMG database. 

For now, I am not sure how to systematically improve the database (as well as RMG-Py codes), but one way would be spreading RMG to researchers in other fields and making them try to implement RMG in their usages. I don't think this can be done by a couple of folks in the RMG group.

JacksonBurns:
@xiaoruiDong this looks good to go, except that [this unit test](https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/4418235321/jobs/7797631653?pr=2398#step:10:6454) needs to be updated

JacksonBurns (approved):
This passes regression testing and the updates to the unit tests match the expected output from the corrected formula. LGTM

PR #2411, opened 2023-04-13
simplify condition checking for scheduled vs pushes CI

JacksonBurns:
Small PR to simplify the scheduled CI condition and fix the syntax (which is currently fine, but leading to unnecessary work by the scheduled runner).

This PR also serves as a test of the new regression testing system


xiaoruiDong (approved):
LGTM, Thanks

PR #2412, opened 2023-04-13
CI: don't make destination directory for reference results when generating reference results

JacksonBurns:
This PR adds a line to the CI that will cause it to skip the `mkdir stable_regression_results` step when it is running on a schedule or a push to main. Under such circumstances, the run _is_ the stable result and will therefore never download anything to this folder making it not needed.


hwpang (approved):
Looks good to me

Issue #2355, opened 2023-01-13
RMG test run error on MacOS 


thethegriddy:
Hello all, I tried to perform the test run for RMG by typing in rmg.py input.py into the terminal. I was given this response:
ImportError: dlopen(/Users/georgio/opt/anaconda3/envs/rmg_env/lib/python3.7/site-packages/cantera/_cantera.cpython-37m-darwin.so, 0x0002): Library not loaded: @rpath/libopenblas.0.dylib

I did not realize I needed to install cantera.
So I did that, and the matlab extension, and I get the same importerror.

Does anyone have any idea how to resolve this? I am a novice at coding so the problem is probably a simple one and likely blew over my head. 
Also, I run a Mac with an intel chip (if that is of any help)

mjohnson541:
What version of Cantera was installed?  You should be able to see this within the rmg_env by running `conda list`.

PR #2415, opened 2023-04-16
`CI.yml` and `environment.yml` updates to run CI on MacOS and Ubuntu

JacksonBurns:
This PR runs the Continuous Integration on a mac runner in addition to the Ubuntu runner, the idea being that we can catch issues that might crop up on the mac platform independently of debian linux. The results from the ubuntu runner are still considered to be the "baseline" for regression testing.

See #2355 and #2392 for issues related to running on Macs

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -94,6 +94,11 @@ jobs:
         with:
           version: "1.8.5"
 
+      - name: Install PyCall on MacOS Global Julia
+        if: runner.os == 'macos'
+        run: |
+          julia -e 'using Pkg; Pkg.add("PyCall"); using PyCall' || true 
+
       # RMS installation and linking to Julia
       # Allow these installs to 'fail' (as they do in RMG-Tests) with the command || True trick
       - name: Install and link Julia dependencies
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -235,7 +235,7 @@ jobs:
   # 
   # taken from https://github.com/orgs/community/discussions/4324#discussioncomment-3477871
   ci-report-status:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     needs: build-and-test-unix
     if: always()
     steps:
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -76,10 +76,12 @@ jobs:
           git clone -b $RMG_DATABASE_BRANCH https://github.com/ReactionMechanismGenerator/RMG-database.git
           
       # modify env variables as directed in the RMG installation instructions
-      - name: Set PYTHONPATH and PATH
+      - name: Set Environment Variables
         run: |
           RUNNER_CWD=$(pwd)
           echo "PYTHONPATH=$RUNNER_CWD/RMG-Py:$PYTHONPATH" >> $GITHUB_ENV
+          echo "JULIA_DEPOT_PATH=/usr/share/miniconda3/envs/rmg_env/share/julia/site" >> $GITHUB_ENV
+          echo "LD_LIBRARY_PATH=/usr/share/miniconda3/envs/rmg_env/lib" >> $GITHUB_ENV
           echo "$RUNNER_CWD/RMG-Py" >> $GITHUB_PATH
 
       # RMG build step
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -88,6 +88,12 @@ jobs:
           make clean
           make
 
+      - name: Add Global Julia Install on MacOS
+        if: runner.os == 'macos'
+        uses: julia-actions/setup-julia@v1
+        with:
+          version: "1.8.5"
+
       # RMS installation and linking to Julia
       # Allow these installs to 'fail' (as they do in RMG-Tests) with the command || True trick
       - name: Install and link Julia dependencies
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -117,7 +117,7 @@ jobs:
               export FAILED=Yes
             fi
           done
-          if [[ -v FAILED ]]; then
+          if [[ ${FAILED} ]]; then
             echo "One or more regression tests could not be executed."
             echo "Please download the failed results or check the above log to see why."
             exit 1
@@ -219,7 +219,7 @@ jobs:
               fi
             fi
           done
-          if [[ -v FAILED ]]; then
+          if [[ ${FAILED} ]]; then
             echo "One or more regression tests failed."
             echo "Please download the failed results and run the tests locally or check the above log to see why."
             exit 1
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -40,7 +40,7 @@ jobs:
   build-and-test-linux:
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-11]
+        os: [ubuntu-latest, macos-latest]
     runs-on: ${{ matrix.os }}
     # skip scheduled runs on main from forks
     if: ${{ !( github.repository != 'ReactionMechanismGenerator/RMG-Py' && github.event_name == 'schedule' ) }}
@@ -51,6 +51,12 @@ jobs:
         shell: bash -l {0}
     steps:
       - uses: actions/checkout@v3
+      
+      - name: Update Xcode on MacOS
+        if: runner.os == 'macos'
+        uses: maxim-lobanov/setup-xcode@v1
+        with:
+          xcode-version: latest-stable
 
       # configures the mamba environment manager and builds the environment
       - name: Setup Mambaforge Python 3.7
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -80,8 +80,6 @@ jobs:
         run: |
           RUNNER_CWD=$(pwd)
           echo "PYTHONPATH=$RUNNER_CWD/RMG-Py:$PYTHONPATH" >> $GITHUB_ENV
-          echo "JULIA_DEPOT_PATH=/usr/share/miniconda3/envs/rmg_env/share/julia/site" >> $GITHUB_ENV
-          echo "LD_LIBRARY_PATH=/usr/share/miniconda3/envs/rmg_env/lib" >> $GITHUB_ENV
           echo "$RUNNER_CWD/RMG-Py" >> $GITHUB_PATH
 
       # RMG build step
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -37,7 +37,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build-and-test-linux:
+  build-and-test-unix:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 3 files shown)


JacksonBurns:
Previous CI failed due to a limitation of macos runners, see https://github.com/byrnedo/docker-reg-tool/issues/9. Updating now

JacksonBurns:
Switching to official Julia binaries, which may resolve the code signing problem, is blocked by the same libgfortran update that is impacting the upgrade to OpenMOPAC (i.e. potential fix is blocked by #2417). See [failed CI run](https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/4910518427/jobs/8767760049?pr=2415) when attempting to use them.

JacksonBurns:
Potential bug in PATH setup for setup-julia action: https://github.com/julia-actions/setup-julia/issues/89

.github/workflows/CI.yml
```diff
@@ -104,10 +107,11 @@ jobs:
 
       # Regression Testing - Test Execution
       - name: Regression Tests - Execution
+        timeout-minutes: 60
```
hwpang, line 102: What is the unit of this?

JacksonBurns: `timeout-minutes: 60`
minutes

JacksonBurns:
Merging this PR also:
 - resolves #2392 
 - resolves #2355 
 - resolves #2290

environment.yml
```diff
@@ -36,8 +36,9 @@ dependencies:
   - psutil
   - rmg::pydas >=1.0.3
   - pydot
+  - conda-forge::julia=1.8.5
+  - conda-forge::suitesparse
   - rmg::pydqed >=1.0.3
-  - rmg::pyjulia
```
calvinp0, line 40: Is pyjulia installed through `conda-forge::julia=1.8.5`?

documentation/source/users/rmg/installation/anacondaDeveloper.rst
```diff
@@ -84,6 +73,16 @@ Installation by Source Using Anaconda Environment for Unix-based Systems: Linux
     cd RMG-Py
     conda env create -f environment.yml
 
+.. warning:: Apple Silicon (M1+) Users
+   Execute the following commands instead of ``conda env create -f environment.yml``: ::
+    
+    conda create -n rmg_env
+    conda activate rmg_env
+    conda config --env --set subdir osx-64
+    conda install python=3.7
```
calvinp0, line 82: A bit pedantic and not really a concern or issue, but I figure we could remove the install of python=3.7 here as it will be install in the `conda env update` part of the script

environment.yml
JacksonBurns: It's not, and the rmg:pyjulia does not either. The conda-recipes repo has the build script and you can see that it only provides Julia (as far as I can tell, I'm new to building conda packages). I think pyjulia is an implicit dependency of something else in the environments file? So it is added when the environment is solved?

documentation/source/users/rmg/installation/anacondaDeveloper.rst
JacksonBurns: Yes that seems like a wasted instruction. We should try the install from scratch on a mac without it just to be sure, though.

environment.yml
calvinp0: So, correct if I am misunderstanding what you are saying, but we will need to keep pyjulia as it is the connection between Julia and Python. So I guess we can either keep getting from the rmg channel or from conda-forge channel.

However, the CI appears to be passing? I am a bit confused though how it is doing that when pyjulia isn't installed in the rmg_env (I am assuming here or really misunderstanding something)

JacksonBurns: > So, correct if I am misunderstanding what you are saying, but we will need to keep pyjulia as it is the connection between Julia and Python. So I guess we can either keep getting from the rmg channel or from conda-forge channel.
> 
> However, the CI appears to be passing? I am a bit confused though how it is doing that when pyjulia isn't installed in the rmg_env (I am assuming here or really misunderstanding something)

My mistake - conda still ends up retrieving pyjulia from the rmg channel. See [this line in the latest CI run](https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/4953296609/jobs/8860637395#step:4:321).

CC'ing @mjohnson - pyjulia, how does it work? The conda-recipe doesn't make sense to us.

calvinp0: So my understanding of PyJulia is that, assuming you have Julia installed and set in the PATH (or set in our conda env path for our case when rmg_env is active), it is imported in python like so:

```python
import julia
```

Then we use PyJulia for the `julia.install()` which does the whole PyCall adding (if not present) and building (unless it is M1, then it appears to have an issue with that

We can also use PyJulia to create our system images for precompiling

calvinp0: Apologies if that was already understood

JacksonBurns:
I have updated `environment.yml` to make it explicit which packages which are resolving and still depend on, and I have also added **extensive** inline documentation. I have also removed the redundant line from the developer installation instructions. The CI should pass overnight and then I will request new reviews.

.github/workflows/CI.yml
hwpang: Lol can't believe I didn't see that on the first glance, thanks!

JacksonBurns:
Tagging a lot of people for review since this PR now contains (1) a different "required" check, which requires admin access and (2) changes to the environment that will change how RMG is run on MacOS.

@hwpang and @rwest please provide general approval for this updated CI approach

@calvinp0 please check the documentation changes and ensure they match the comments from the review

@mjohnson541 please review to make sure that I did not describe any dependencies in the documented `environment.yml`

.github/workflows/CI.yml
```diff
@@ -131,7 +135,7 @@ jobs:
       # Upload Regression Results as Stable if Scheduled or Push to Main
       - name: Upload Results as Reference
         # upload the results for scheduled CI (on main) and pushes to main
-        if: github.ref == 'refs/heads/main'
+        if: github.ref == 'refs/heads/main' && runner.os =='ubuntu'
```
hwpang, line 138: I guess this means only to upload the results from the ubuntu tests since they should be the same on either platform?

JacksonBurns: Yes - we treat the results on Linux as 'correct' and compare MacOS to that. Since we distribute on a Linux image for docker, I have taken that as the absolute "ground truth".

The results should never differ in the first place (hopefully).

.github/workflows/CI.yml
```diff
@@ -227,14 +231,29 @@ jobs:
           mamba install -y -c conda-forge codecov
           codecov
 
+  # This allows us to have a branch protection rule for tests and deploys with matrix
+  # 
+  # taken from https://github.com/orgs/community/discussions/4324#discussioncomment-3477871
+  ci-report-status:
+    runs-on: ubuntu-20.04
```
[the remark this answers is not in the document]
JacksonBurns: I will also add a clarification for something that confused myself, this branch is called `add-macos11-ci` but it actually runs on `macos-latest` which as of writing (May 16, 2023) is MacOS 12 Monterey

documentation/source/users/rmg/installation/anacondaDeveloper.rst
```diff
@@ -84,6 +73,15 @@ Installation by Source Using Anaconda Environment for Unix-based Systems: Linux
     cd RMG-Py
     conda env create -f environment.yml
 
+.. warning:: Apple Silicon (M1+) Users
```
hwpang, line 76: Is this the same suggestion for Intel Mac?

JacksonBurns: No, it only applies to M1 users (cc @calvinp0 who figured this out, thanks!)

environment.yml
```diff
@@ -1,54 +1,97 @@
...
+  - coolprop
   - cantera::cantera=2.6
+  - conda-forge::mopac
   - conda-forge::cclib >=1.6.3
-  - rmg::chemprop
-  - coolprop
+  - conda-forge::openbabel >= 3
+
+# general-purpose external software tools
+  - conda-forge::julia=1.8.5
```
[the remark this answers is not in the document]
rwest: on second thoughts, that could be a separate pull request if you prefer (and don't mind the churn)

rwest (commented):
Some suggestions that could be separate pull requests (eg. trying to change channel for pyjulia) but some fixes to the user instructions that should probably be addressed before merging.
And a question about the benefit of adding a changelog to the environment file (which is already in git with - I hope - helpful commit messages).

environment.yml
```diff
@@ -1,54 +1,97 @@
...
+# There is a mixture of the following
+# - packlages would could be installed at OS level, but we install here
+#   for better version control
+# - python tools
+# - external software tools specific to chemistry
+# - other software we maintain which RMG depends on
+# + some other categories (see below)
+#
+# Changelog:
+# - May 15, 2023 Added this changelog, added inline documentation,
```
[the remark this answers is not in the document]
JacksonBurns: The way I see changelogs is that they provide a "high-level" perspective on how the file has changed over time. The git log is useful for fixing technical issues and precise problems with specific lines, but I think the changelog is better for answering questions like "why do we have some packages from this channel but other from this one" or the like.

`environment.yml` in particular has been going through substantial changes lately, and as we (hopefully) move toward using fewer and fewer in-house packages I think it will be useful to have this high level history of that activity. This is also a file that community developers will interact with a lot - such info would be particularly useful for them.

environment.yml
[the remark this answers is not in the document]
JacksonBurns: > Dare we try `>=` and let the CI alert us if it breaks?

Move fast and break things, I say. Will try now.

> Can we put ` - conda-forge::pyjulia` here instead of the ` - rmg::pyjulia` down below? My impression from #2381 was that the answer is "yes" but maybe we can test it here?
> on second thoughts, that could be a separate pull request if you prefer (and don't mind the churn)

Every package we don't have to maintain is welcome - will try this as well.

rwest:
Sorry - my edits introduced errors to the documentation.
Screenshot 2023-05-16 at 3 18 27 PM

rwest:
Pushed a fix for my documentation error. If/when you rebase do it with `-i --autosquash` to make it a fixup.

environment.yml
mjohnson541: It is a dependency of pyrms so including pyrms in the environment.yml file will cause it to be installed as well.

rwest:
"Job execution time - Each job in a workflow can run for up to 6 hours of execution time. If a job reaches this limit, the job is terminated and fails to complete"
"Constant Integration / build-and-test-unix (macos-latest) (pull_request) Cancelled after 360m"
Install and link julia dependencies:  5h 37m 20s
😫

rwest:
Your earlier run it only took about half an hour:
https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/4987565589/jobs/8929446027

rwest:
I couldn't see anything wrong, so restarted the CI job. Seems to be running fine now 🤷

JacksonBurns:
Thanks for doing that, was going to suggest the same thing. Anecodtally, I have noticed the runners are always slower during work hours (the scheduled CI is always faster than those run during the day). I suspect that all of the containers GitHub provides are running on VMs on the same hardware, and during the day they end up fighting with one another over resources and slowing down.

rwest:
Sorry. The "Switch to conda-forge::pyjulia instead of the rmg channel" commit was meant to be on another branch (that's not working yet).  Will drop it on rebasing when I get a chance.

JacksonBurns:
> Looking good to me. Additional improvements can come in subsequent PRs.

:+1: @rwest could you update the branch protection rules to look for `ci-report-status` instead of `build-and-test-linux`, then merge?

rwest:
I set it to this instead:
Screenshot 2023-05-18 at 9 33 43 AM
which if it does  what I expect, will mean we can remove the "ci-report-status" step.
Do you think that'll work, or have I misunderstood something?

JacksonBurns:
@rwest that works as well - the `ci-report-status` is not needed since it is set that way, can be removed. I will roll that into a larger dead code removal PR.

PR #2418, opened 2023-04-22
Added Arkane citation info

alongd:
Added the citation info for Arkane into the docs and the module README

Also updated RMG's citations list.

Note that RMG's credits docs page has "developers" and "previous developers" lists. The current developers list is outdated, perhaps someone from Green Group and West group can complete it? Alternatively, we can just have one list of credits to avoid the need to update it every couple of years.

(docs were built locally and seem OK)


JacksonBurns:
@alongd can you update the `CITATIONS.bib` file as well?

kspieks (approved):
Thanks for this PR and thanks to Jackson for reviewing as well. LGTM. Merging now

PR #2420, opened 2023-05-02
Modified assertion in thermodata.pyx 

NellyMitnik:
### Motivation or Problem
While trying to create a new thermo library, we got an assertion error since Cp data started in 298K.

### Description of Changes
We modified: `assert Tdata[0] > 298` into `assert  Tdata[0] >= 298`

Also modified gitignore file to consider files of PyCharm and VScode in subfolders

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,9 +40,11 @@ nbproject/*
 .settings/*
 
 # PyCharm project files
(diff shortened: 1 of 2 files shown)


alongd (approved):
Thanks!

PR #2421, opened 2023-05-02
use conda-forge ncurses instead of default

JacksonBurns:
the default does not provide version information which raises an annoying warning in the docker image


hwpang (approved):
Looks fine to me

PR #2422, opened 2023-05-03
fix typo in reactors.py

mjohnson541:
This typo prevents RMG from running if a surface species is in the mechanism.


alongd (approved):
Thanks!

PR #2424, opened 2023-05-03
[Move to Docker]: Add Dockerfile, update installation instructions, add automatic image builds to CI

JacksonBurns:
### Motivation or Problem
Primary end user deployment has moved to Docker.

### Description of Changes
This PR updates the documentation, adds the Dockerfile, and adds a step in the CI to automatically rebuild a 'nightly' image.

### Testing
If someone could try out the installation instructions as added here that would be good.

### Reviewer Tips
The CI step that pushes the docker container will not run until we actually merge into main, which is intended. If it breaks for some reason I will fix it in a follow up PR.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -226,3 +226,32 @@ jobs:
         run: |
           mamba install -y -c conda-forge codecov
           codecov
+
+  build-and-push-docker:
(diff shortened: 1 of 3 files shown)


hwpang (approved):
Looks good to me. Waiting for the tests to pass before merging.

PR #2425, opened 2023-05-04
Calculate x only when semiclassical corrections are needed

kfir4444:
### Motivation or Problem
`frequency` and `x` in get_partition_function are only needed in the end of the function if semiclassical correction is needed. I suggest moving them to the end of the function, after checking `self.semiclassical`

### Description of Changes
Moved `frequency` and `x` to the end of the function.


JacksonBurns (approved):
Nice catch! Should save us some wasted computation. LGTM

alongd:
Looks great! Could you remove the 2nd (empty?) commit?

alongd (approved):
Thanks!

PR #2438, opened 2023-05-16
Change variables from cpdef to cdef in cython.

rwest:
### Motivation or Problem
We were getting warnings
> warning: rmgpy/quantity.pxd:33:6: cpdef variables will not be supported in Cython 3; currently they are no different from cdef variables" because for variables (or constants) there is no difference between cpdef and cdef. A future version of cython will turn these warnings into errors.

### Description of Changes
I found the few instances where we tried to cpdef a variable (as opposed to a function or class) and changed them to cdef.

### Testing
I will let the test suite run.

### Reviewer Tips
See https://github.com/cython/cython/pull/3963 for context.
The CI tests should reveal any problems.

### DIFF
--- a/rmgpy/molecule/translator.pxd
+++ b/rmgpy/molecule/translator.pxd
@@ -28,12 +28,12 @@
 cimport rmgpy.molecule.molecule as mm
 
 
-cpdef list BACKENDS
-cpdef dict INCHI_LOOKUPS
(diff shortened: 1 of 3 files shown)


JacksonBurns (approved):
Build log looks much better - thanks for taking care of this! Next step will be getting rid of that other warning, but this LGTM.

PR #2444, opened 2023-05-19
Fix our use of pyjulia so we can use the conda-forge version.

rwest:
This is a subset of what was in https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2443 but I'm breaking it off into a smaller PR that can be merged first. 

### Motivation or Problem
- We have an old and customized version of pyjulia on the rmg conda channel that we fail to maintain.
- Debugging things with pyjulia involved can be a bit tricky.

### Description of Changes
- switches to the conda-forge version of pyjulia, instead of the customized (and old) rmg channel one, so we are more standardized and have less to maintain.
- fixes our CI workflow so that we can use the new pyjulia (remove the symbolic linking step that made `python-jl` masquerade as `python`).
- adds developer instructions for how to debug (in VSCode)

### Testing

- Runs on my intel Mac, following instructions with a fresh anaconda environment and no system-wide Julia.
- Runs on the CI test suite.

### Reviewer Tips
Most of this has already been reviewed by @JacksonBurns when it was https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2443  so should be a quick review, but I'd like @mjohnson541 's input.

## Possible concern.
The [readme for pyrms](https://github.com/ReactionMechanismGenerator/pyrms#installation-instructions-from-anaconda) says  "to work with conda python efficiently currently requires relinking the python executable to the python-jl executable"
so perhaps @mjohnson541 will tell us that skipping this step (as proposed here) is a bad idea? 
I can't see anything in the commit messages when those instructions were added to explain the details.
I would note that
1. our developer instructions **do not** recommend making the symlink (commit https://github.com/ReactionMechanismGenerator/RMG-Py/commit/5d208269ce5dabc73d3152170ac75f1dfbce4bd1 almost added it? but didn't)
2. our Docker container **does** currently make the symlink (probably that should be removed in this PR, if this change is approved)

## Environment alert ⚠️ 
This updates the conda environment file again.
So to be consistent, you'll need
1. a fresh conda environment (using `pyjulia` from `conda-forge` instead of from the `rmg` channel)
4. you might need to re-do the whole step of installing julia in python and python in julia etc. (following the instructions)
5. if you have symlinked `python` to point to `python-jl`, ether in your normal environment or in some other workflows, revert this.

### DIFF
--- a/utilities.py
+++ b/utilities.py
@@ -310,7 +310,7 @@ def update_headers():
     start of each file, be sure to double-check the results to make sure
     important lines aren't accidentally overwritten.
     """
-    shebang = """#!/usr/bin/env python3
+    shebang = """#!/usr/bin/env python-jl
 
 """
 


JacksonBurns (changes requested):
This LGTM but I have a question that may impact changes (thus "Request changes") - do we now have to use "python-jl rmg.py ..." without a system image and then "python rmg.py ..." otherwise? The docs should clarify this, unless I am just misunderstanding. We should also replace #2310 with this PR, I think.

rwest:
After this, we should always use `python-jl` to launch `rmg.py`. Thanks for reminding me that the "running" instructions may be out of date. (nb. the previous instructions wouldn't have worked for folks installing via the "source code for developers" install instructions). 
I've now
 - updated the running instructions (telling people to use `python-jl`)
 - updated the Docker build file (removing the symlink step)
 - cherry-picked https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2310

rwest:
> do we now have to use "python-jl rmg.py ..." without a system image and then "python rmg.py ..." otherwise? 

Dealing with system images is happening on https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2443 and instructions for what to do when you have one will be added there. (FWIW, I currently can't get it it work).

mjohnson541:
Relinking is mostly a convenience thing. All it does is makes it so when you use `python` you're using `python-jl` if you use `python-jl` when needed there's no reason to do the relinking operation.

rwest:
Great. Thanks for the confrontation @mjohnson541. Do you think this PR is Ok then?

mjohnson541 (approved):

mjohnson541:
LGTM!

Issue #2442, opened 2023-05-17
None of our regression tests use RMS reactors


rwest:
The testing framework runs the examples in `RMG-Py/test/regression` 
which use the following reactor types:
```
aromatics/input.py:simpleReactor(
liquid_oxidation/input.py:liquidReactor(
nitrogen/input.py:simpleReactor(
oxidation/input.py:simpleReactor(
sulfur/input.py:simpleReactor(
sulfur/input.py:simpleReactor(
superminimal/input.py:simpleReactor(
```
so the RMS reactor types are not being tested in this CI framework.

I think _at least_ one of the regression tests should use an RMS reactor, so that a "full run" of RMG that uses RMS is tested via CI.

PR #2446, opened 2023-05-24
Add regression test for RMS reactor

hwpang:
### Motivation or Problem
Currently we don't have any regression test for RMS based reactors. See https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2442.

### Description of Changes
I add `input.py` and `regression_input.py` to test RMS's Constant V Ideal Gas Reactor and RMS's CSTR. Once this regression runs normally on CI, I will add more for other RMS reactors.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -119,7 +119,7 @@ jobs:
       - name: Regression Tests - Execution
         timeout-minutes: 60
         run: |
-          for regr_test in aromatics liquid_oxidation nitrogen oxidation sulfur superminimal;
+          for regr_test in aromatics liquid_oxidation nitrogen oxidation sulfur superminimal RMS_constantVIdealGasReactor_superminimal RMS_CSTR_liquid_oxidation RMS_liquidSurface_ch4o2cat;
           do
             if python-jl rmg.py test/regression/"$regr_test"/input.py; then
               echo "$regr_test" "Executed Successfully"
@@ -190,7 +190,7 @@ jobs:
         env:
           REFERENCE: stable_regression_results
         run: |
-          for regr_test in aromatics liquid_oxidation nitrogen oxidation sulfur superminimal;
+          for regr_test in aromatics liquid_oxidation nitrogen oxidation sulfur superminimal RMS_constantVIdealGasReactor_superminimal RMS_CSTR_liquid_oxidation RMS_liquidSurface_ch4o2cat;
           do
             # Memory Usage and Execution Time
             echo 'Execution time for Reference:'
--- a/rmgpy/rmg/input.py
+++ b/rmgpy/rmg/input.py
@@ -836,7 +836,7 @@ def constant_T_V_liquid_reactor(temperature,
     termination = []
     if terminationConversion is not None:
         for spec, conv in terminationConversion.items():
-            termination.append(TerminationConversion(species_dict[spec], conv))
+            termination.append((species_dict[spec], conv))
     if terminationTime is not None:
         termination.append(TerminationTime(Quantity(terminationTime)))
     if terminationRateRatio is not None:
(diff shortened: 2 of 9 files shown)


JacksonBurns:
This PR resolves #2442

hwpang:
I added regression tests for three RMS reactors: constant V ideal gas reactor, CSTR through constant T V liquid reactor with inlet and outlet, and liquid surface reactor.

The constant V ideal gas reactor regression test uses an input file based on superminial example.
The CSTR regression test uses an input file based on pentane liquid oxidation example, where I remove the constant assumption for oxygen, and instead flow oxygen into the reactor at a constant composition.
The liquid surface regression test uses an input file based on the `catalysis/ch4o2cat` example, while I use `pentane` as a placeholder for solvation correction. I am not familiar with what kind of solvent are more commonly used in the catalysis system. I would greatly appreciate any more realistic example to use for the test.

rwest:
> regardless of what reactors are used to generate the mechanism, we are using Cantera's IdealGasReactor to simulate in the regression step

This does sound curious. Hard to change?

hwpang:
> > regardless of what reactors are used to generate the mechanism, we are using Cantera's IdealGasReactor to simulate in the regression step
> 
> This does sound curious. Hard to change?

Currently, we hard coded to use [Cantera class](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/main/rmgpy/tools/canteramodel.py#L195) for simulation [here](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/main/rmgpy/tools/observablesregression.py#L124). 

Switching to use the reactor that we generate the mechanism shouldn't be too hard. We just need to add a class to simulate RMG-based reactor and a class to simulate for RMS-based reactor and the downstream process to return the organized data same as the [Cantera class's simulate function](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/main/rmgpy/tools/canteramodel.py#L561)

- To simulate RMG-based reactor, you could probably borrow the code in the [simulation script](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/main/rmgpy/tools/simulate.py#L44).
- To simulate RMS-based reactors, you really just need to make the reactor (for example, for constant V ideal gas reactor, you can use the readily available code [here](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/main/rmgpy/rmg/reactors.py#L413) to make the reactor easily) and just call `solve`.

hwpang:
The regression test passed for the constant V ideal gas reactor and the CSTR ones, but it [failed](https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/5082150835/jobs/9131516496?pr=2446#step:12:25454) for the liquid-surface reactor. It failed because we currently don't have a `updatefilterthresholds!` function for `SystemSimulation`, which handles multi-domain case. We currently only have [one](https://github.com/ReactionMechanismGenerator/ReactionMechanismSimulator.jl/blob/main/src/EdgeAnalysis.jl#L609) for `Simulation`, which handles single-domain case. 

We could add one for the multi-domain case in the future. Although I take a quick look at the version for `Simulation`, I don't see anything specific for `Simulation`. Any comments on this, Matt? @mjohnson541 

I don't think this is something needs to be resolved with this PR. I'm going to remove the regression test for the explicit two-phase liquid-surface reactor.

hwpang:
I rebased and removed the regression test for RMS's liquidSurfaceReactor explicitly

mjohnson541:
You can turn off the filter and that should let that RMG job run properly. The barrier to turning the filter on for SystemSimulations is that filter rate constants are only defined between liquid species and between gas phase species. I don't believe we have defined filter rate coefficients on surfaces or on inter-phase reactions. We should be able to turn it on for SystemSimulations that involve the same phase...which the code right now would prevent.

hwpang:
@mjohnson541 Good point. I turned off the filtering for the regression test case for the liquid surface reactor in the new push.

JacksonBurns (commented):
The CI fail is to be expected here as @hwpang stated. Commenting on this but not approving since I can't critique the new examples.

mjohnson541 (commented):
These changes look good to me...although some commits should be squashed.

hwpang:
I squashed relevant comments and rebased! I will force merge once the tests proceed to the expected failure point.

JacksonBurns:
This is blocked by an issue with the CI (#2455) and the attempted resolution #2456

hwpang:
I rebased this PR with `main`.

hwpang:
The CI test failed at the expected point as there's no "stable" version of the regression test I just added. I think this PR is ready to merge.

JacksonBurns (approved):
Huge thanks @hwpang ! LGTM

PR #2447, opened 2023-05-25
Allow auxilary keywords in an Arkane species YAML file

alongd:
### Motivation or Problem
Different use cases require additional information to be stored in an Arkane YAML file that is generated using the `RMGObject`. Thy may include atom order in a molecule.

### Description of Changes
Allow RMG to read Arkane YAML files through the `RMGObject` while ignoring ` mol` and `aux` keywords to store molecular and auxiliary information.

### Testing
A test was added


kfir4444 (approved):
LGTM :)

Issue #2445, opened 2023-05-23
Python 3.7 End-of-Life and Upgrading to Python 3.11


JacksonBurns:
Python 3.7 is reaching end of life [this summer](https://devguide.python.org/versions/) and will no longer receive security updates. Ultimately because we distribute on Docker now we can continue to ship the outdated Python version for a while before things actually _stop_ working, but it will atrophy.

We will need to increment our minor version of Python, preferably to 3.11 for the speed improvements and longest support window. I say this rather than just shooting for 3.8, since I believe the additional effort of incrementing 4 minor versions will not be much worse than 1 (see below).

I anticipate that this will be an absolute nightmare, but a (forced) opportunity for some cleanup.
 - `pydas` and `pydqed` were originally written in Python 2.7, hacked to get into Python 3.7, and will likely go kicking and screaming into Python 3.11
 - `pyjulia` and `diffeqpy` are both presumably locked to Python 3.7
 - Our _remarkably_ outdated build of chemprop (0.0.1) only supports 3.7, though chemprop v2 will support 3.11 and newer.
 - many of the packages we built and 'maintain' on the conda rmg channel expect Python 3.7 and will need to be rebuilt (just especially the ones mentioned above)
 - Cython will need to be incremented to 0.29 (from 0.27)

This might take long enough that we end up targeting 3.12, which promises some very nice parallelism improvements so maybe that isn't such a bad thing.

Some things we should get done (wrt clearing tech debt) before attempting this:
 - release v3.2.0 via #2311 
 - #975 delete what we aren't using so don't have to try and upgrade it for no purpose
 - #2380 as broken as nose is now, it will be even more broken-er in Python 3.11
 - environment updates, including #2375, #2379 (and its PR #2443), and #2322 which should hopefully make it easier to upgrade by reducing # of dependencies and their complexity
 - Fix #2442 (and #2429 by consequence) so that we can validate results with RMS reactors in 3.11 versions

JacksonBurns:
#2429 was resolved by @JacksonBurns in the process of assisting @hwpang, who is resolving #2442 via #2446

rwest:
Regarding chemprop. You wrote above and in the environment.yml file:
- "Our remarkably outdated build of chemprop (0.0.1) only supports 3.7, though chemprop v2 will support 3.11 and newer."
- "Our build of this is version 0.0.1 (!!) and we are using parts of the API that are now gone. Need a serious PR to fix this."

[...]

PR #2450, opened 2023-05-31
Remove `chemprop`-dependent Code

JacksonBurns:
### Motivation or Problem
As discussed in #2445 our build of `chemprop` is very old and does not support Python 3.11, which will prevent us from doing so as well.

### Description of Changes
Right now the only change is to switch the channel in `environment.yml` from `rmg` to `conda-forge` for `chemprop`. This will likely break other code in `RMG-Py` and require additional changes afterward.

### Testing
I have done a `--dry-run` of resolving this environment file so that at least _works_, but there will be other things to fix as revealed by the CI. 

#### Notes
This PR will later need to be followed by an `upgrade` PR (as part of #2445) to move to chemprop v2, which will support Python 3.11

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -15,6 +15,8 @@
 # - May 15, 2023 Added this changelog, added inline documentation,
 #   made depdency list more explicit (@JacksonBurns).
 #
+# - May 31, 2023 Switch chemprop to official builds
+#
 name: rmg_env
 channels:
   - defaults
@@ -40,6 +42,7 @@ dependencies:
 
 # external software tools for chemistry
   - coolprop
+  - conda-forge::chemprop
   - cantera::cantera=2.6
   - conda-forge::mopac
   - conda-forge::cclib >=1.6.3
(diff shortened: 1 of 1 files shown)


JacksonBurns:
Note from the RMG developers meeting - the `chemprop` model included in RMG-Py right now may not survive the upgrade to Python 3.11. The timeline of the upgrade to Python 3.11 and the difficulty of training the model and then fixing the API are challenging.

JacksonBurns:
For the transition to Python 3.11, features requiring `chemprop` will be removed. They can be added back in later, but that will require:
 - retraining the ML models
 - re-writing the ML estimator to use a more updated version of `chemprop`
 - validating that the new models are better than the current capabilities of RMG

There is also discussion of build a more flexible interface to ML-based estimators, such that the estimator itself would not be distributed with RMG but would just need to inherit from some ML Estimator class in RMG that defines a `get_enthalpy`-esque set of methods.

JacksonBurns:
`chemprop` version 2 is upcoming and will add support for Python 3.11 and 3.12 but will face the same issues as above.

PR #1648, opened 2019-07-11
Expand species length

goldmanm:
### Motivation or Problem
Anyone who uses RMG likely has encountered the frustrating `S(14567)` species label. 

### Description of Changes
This commit tries to make species labels more readable and intuitive. It does this in Arkane by using the species label given in the input file for the species. In RMG it expands the species length from 10 characters to 16 characters, almost entirely eliminating the dreaded `S` label.

This PR is compatible with the cantera and modern versions of chemkin. Unmodified versions of chemkin 2 will not be able to run some models because of a limitation on the length of reaction. chemkin 2, however, can be modified by the users to be compatible with the more widely used chemkin specifications. 

The commits in this PR were moved from #1607 since they appear to pertain to different problems. They have already been code reviewed in that commit. The main question holding this PR back is whether we should eliminate the ambiguous species label `S` or stay completely compatible with unmodified versions of chemkin 2.

PR #1923, opened 2020-04-16
Allow colons and periods in chemkin species names

rwest:
### Motivation or Problem
When trying to import a chemkin file from NIST, we choked on a species called `CHF:CF2` because RMG does not expect chemkin species labels to contain colons.

### Description of Changes
Added `:` to the regular expression used to check valid species labels before returning them.
Also found a commit from ages ago where we did the same thing for periods, for some other chemkin file. I added that commit to this PR. Then I found the model also had `[` and `]` in the species names, so I added those.

### Testing
I will cross my fingers and let travis run the unit tests.

### Reviewer Tips
Hopefully any other parts of code that parse strings using regular expressions (or worse, `split()`) are well written and tested.

### Note
We are still, in fact, too restrictive to parse all valid Chemkin names. For example see some of the pathological names that Cantera checks for in https://github.com/Cantera/cantera/blob/master/test/data/species-names.inp
which contains `(Parens)  @#$%^-2  co:lons: [xy2]*{.}  plus+ eq=uals plus trans_butene co`

PR #2454, opened 2023-06-05
Increase allowed length of species labels for chemkin output. 

rwest:
### Motivation or Problem
Anyone who uses RMG likely has encountered the frustrating `S(14567)` species label.

### Description of Changes
This has a few commits by @goldmanm cherry-picked from https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1648 which was recently closed as abandoned.

He wrote:
> In RMG it expands the species length from 10 characters to 16 characters, almost entirely eliminating the dreaded S label.
> This PR is compatible with the cantera and modern versions of chemkin. Unmodified versions of chemkin 2 will not be able to run some models because of a limitation on the length of reaction. chemkin 2, however, can be modified by the users to be compatible with the more widely used chemkin specifications.

I skipped the commits that alter Arkane.

There's also a commit that changes the default chemkin output style for a reaction from 
`A+B=C+D` to `A+B<=>C+D` (notice the arrowheads).
This change I'm not so sure about, because it's not strictly necessary and some folks may have downstream scripts that search for or parse reactions in a certain way.  Anyone have opinions?

### DIFF
--- a/rmgpy/chemkin.pyx
+++ b/rmgpy/chemkin.pyx
@@ -1732,7 +1732,7 @@ def write_reaction_string(reaction, java_library=False):
 
     if kinetics is None:
         reaction_string = ' + '.join([get_species_identifier(reactant) for reactant in reaction.reactants])
-        reaction_string += ' => ' if not reaction.reversible else ' = '
+        reaction_string += ' <=> ' if reaction.reversible else ' => '
         reaction_string += ' + '.join([get_species_identifier(product) for product in reaction.products])
         return reaction_string
 
@@ -1777,7 +1777,7 @@ def write_reaction_string(reaction, java_library=False):
 
         reaction_string = '+'.join([get_species_identifier(reactant) for reactant in reaction.reactants])
         reaction_string += third_body
-        reaction_string += '=' if reaction.reversible else '=>'
+        reaction_string += '<=>' if reaction.reversible else '=>'
         reaction_string += '+'.join([get_species_identifier(product) for product in reaction.products])
         reaction_string += third_body
 
--- a/rmgpy/chemkin.pyx
+++ b/rmgpy/chemkin.pyx
@@ -1779,7 +1779,7 @@ def write_reaction_string(reaction, java_library=False):
         reaction_string += third_body
 
     if len(reaction_string) > 52:
-        logging.warning("Chemkin reaction string {0!r} is too long for Chemkin 2!".format(reaction_string))
+        logging.debug("Chemkin reaction string '%s' is too long for Chemkin 2!", reaction_string)
     return reaction_string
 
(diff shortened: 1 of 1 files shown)


alongd:
Thanks for reviving these commits!
Intuitively I'm OK with the addition of arrowheads to the chemkin files (given that it is supported by Chemkin). There's a small downside as mentioned. Is there an additional upside other than better readability? I think I'm neutral.

Nora-Khalil:
Hi, so I just tried this on one of the RMG examples (c3h4) in RMG-Py. No S labels to be seen :) .One problem I was running into that we might want to be careful about is that sometimes the species labels will include a "#". This becomes a problem when you create a solution array in Cantera, but then try to reload the solution array as an initial guess. For example: 

if you make a Solution object
`gas=ct.Solution(<path_to_yaml_file>)`

Create a flame object and solve the flame: 
`mole_frac_dict = {'CH4(16)': 1, 'O2(8)':1, 'N2':3.76} 
results = {}

To = 298
Po = 1e5 # ct.one_atm
gas.TPX = To, Po, mole_frac_dict

width = 0.08
flame = ct.FreeFlame(gas, width=width)
flame.set_refine_criteria(ratio=3, slope=0.1, curve=0.1) 
flame.max_time_step_count = 900
loglevel = 1 

flame.solve(loglevel=loglevel, auto=False)
Su = flame.velocity[0]
results[1] = Su
sltn = flame.to_solution_array()
df1 = sltn.to_pandas()
df1.to_csv(f'data/initial.csv', index=False) `

but then try to create an initial guess with that same solution array: 

`d = './data/initial.csv'
arr2 = ct.SolutionArray(gas)
arr2.read_csv(d)
flame.set_initial_guess(data=arr2)`

You get the following error: 
`ValueError: Some errors were detected !
    Line #2 (got 68 columns instead of 7)
    Line #3 (got 68 columns instead of 7)
    Line #4 (got 68 columns instead of 7)
    Line #5 (got 68 columns instead of 7)
    Line #6 (got 68 columns instead of 7)
    Line #7 (got 68 columns instead of 7)
    Line #8 (got 68 columns instead of 7)
...`

which happens when one of the field name in the data of the csv file contains an invalid character (in this case, a # in the species i.e. C#CC#C(963)). Is there a way we can make sure that the species labels don't include invalid characters `~!@#$%^&*()-=+~\|]}[{';: /?.>,<.` ?

rwest:
Great work - thanks for the thorough testing. That sounds like a Cantera bug - it should be able to cope with # in species names.

rwest:
See the note on https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1923 which mentions some of the valid Cantera names

PR #2456, opened 2023-06-05
Trying to fix CI see issue #2455

rwest:
Trying to debug #2455. 

We were getting errors in the CI runs like
```
ERROR: Unable to dlopen(cxxpath) in parent!
Message: /usr/share/miniconda3/envs/rmg_env/lib/python3.7/site-packages/rdkit/../../../libstdc++.so.6nts a live Julia runtime.
: cannot open shared object file: File name too long
```

It's hard to debug without a pull request, so here's a pull request. 

Feel free to push or force-push to this `cifix` branch to test out ideas - I don't feel I "own" this PR.

PR #2458, opened 2023-06-07
[Docker]: add linux patch step for `libstdcxx-ng<13`

JacksonBurns:
This small patch adds the environment file patch needed for linux systems (see #2456).

The current user Docker image is unaffected, this will just fix the nightly image (and of course later the user image when we release 3.2.0).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -58,8 +58,12 @@ RUN make
 RUN python -c "import julia; julia.install(); import diffeqpy; diffeqpy.install()" || true
 RUN julia -e 'using Pkg; Pkg.add(PackageSpec(name="ReactionMechanismSimulator",rev="main")); using ReactionMechanismSimulator' || true 
 
-# RMG-Py should now be installed and ready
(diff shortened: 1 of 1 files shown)


JacksonBurns:
> As discussed in [#2456 (comment)](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2456#issuecomment-1580765785), we recently pushed a broken docker build, because we didn't notice it was broken. You noted:
> 
> > The theory was that if the tests pass, nothing in the docker build should go wrong (and --help is just there to trigger precompilation). Now that the patch is in for libstdcxx-ng<13 that should be the case again.
> 
> ...but one could argue that the theory has already been disproven. We should expect the unexpected - especially when we have so recently been given an example. It would only take a few minutes of build time (not user run time) to check that RMG _actually_ runs before pushing the docker image. Is there a drawback I'm not seeing?

A good point - just added in 942cfec.

rwest (approved):
LGTM. Thanks for addressing my concern.

PR #2459, opened 2023-06-07
Regression testing updates

rwest:
### Motivation or Problem
I suspect regression tests are often not looked at because the results are buried in the logs.

### Description of Changes
The biggest change is that regression tests actually report a failure if a model changes significantly - either the core or edge changes or an observable in the simulation changes significantly.
Many changes to RMG are deliberate improvements and ought to change things, so sometimes we want these "regression" tests to pass - but it should probably be a deliberate act of an admin approving the pull request despite the regression, not just everyone ignoring the regression tests because nobody checks the logs.  If there are too many failures and so we have a bottle-neck of approvals, or admins over-riding tests too much, we can revisit the policy. At least these changes now give us the framework to detect the regressions during the testing.

Other changes include
- the results from the matrix build jobs are stored separately
- failures of regression tests are reported in the job step summary
- only the official main ubuntu run is "blessed" as the stable job. All others are dynamic for comparison (including the 'main' branch on forks)
- other small changes to the CI file.

### Testing
Tested it in the CI workflow. 

### Reviewer Tips
I tried to keep each commit self-contained with commit messages, and rebased a few times merging fixups.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -193,13 +193,13 @@ jobs:
           for regr_test in aromatics liquid_oxidation nitrogen oxidation sulfur superminimal RMS_constantVIdealGasReactor_superminimal RMS_CSTR_liquid_oxidation RMS_liquidSurface_ch4o2cat;
           do
             # Memory Usage and Execution Time
-            echo 'Execution time for Reference:'
+            echo -n 'Reference: '
             grep "Execution time" $REFERENCE/"$regr_test"/RMG.log | tail -1
-            echo 'Execution time for Current:'
+            echo -n 'Current:   '
             grep "Execution time" test/regression/"$regr_test"/RMG.log | tail -1
-            echo 'Memory used for Reference:'
+            echo -n 'Reference: '
             grep "Memory used:" $REFERENCE/"$regr_test"/RMG.log | tail -1
-            echo 'Memory used for Current:'
+            echo -n 'Current:   '
             grep "Memory used:" test/regression/"$regr_test"/RMG.log | tail -1
 
             # Compare the edge and core
@@ -222,7 +222,8 @@ jobs:
               if python-jl rmgpy/tools/regression.py \
                 test/regression/"$regr_test"/regression_input.py \
                 $REFERENCE/"$regr_test"/chemkin \
-                test/regression/"$regr_test"/chemkin; then
+                test/regression/"$regr_test"/chemkin
+              then
                 echo "$regr_test" "Passed Regression Testing"
               else
                 echo "$regr_test" "Failed Regression Testing"
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -121,7 +121,7 @@ jobs:
 
       # Regression Testing - Test Execution
       - name: Regression Tests - Execution
-        id: regression
+        id: regression-execution
         timeout-minutes: 60
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/CI.yml
```diff
@@ -157,21 +159,21 @@ jobs:
 
       # Upload Regression Results as Dynamic if Push to non-main Branch
       - name: Upload Results as Dynamic
-        if: github.ref != 'refs/heads/main'
+        if: !( env.REFERENCE_JOB )
```
hwpang: This change will make it only upload results as dynamic if it's pushed to `ReactionMechanismGenerator/RMG-Py`, but previously it should upload for a forked branch as well. I think we should allow upload as dynamic for forked ones to encourage people to develop with forked repos and keep the branches on `ReactionMechanismGenerator/RMG-Py` clean.

rwest: I think the opposite. This change _increases_ the number of things that are now uploaded as dynamic. Previously, `main` on any fork would upload as its own "reference" job. Because we define above `REFERENCE_JOB: ${{ github.ref == 'refs/heads/main' && matrix.os =='ubuntu-latest' && github.repository == 'ReactionMechanismGenerator/RMG-Py' }}` now the "reference" job is only the official main ubuntu, all other jobs (including forks) are "dynamic". 

I'm get to see if forks can successfully download the reference results (or if they need a github access token)

hwpang: Yep you're right. I missed the `!` in the statement.

rwest:
I switched to the patched download-artifact action you suggested and turned on `ensure_latest` but it still found one from 2023-05-18T08:20:16Z. I think that might actually _be_ the most recent, because if you check lot of [recent runs on the main branch](https://github.com/ReactionMechanismGenerator/RMG-Py/actions?query=branch%3Amain+workflow%3A*Integration) have been skipping the "Upload Results as Reference" step. 

I suggest that, once someone approves this, we merge it to main and see if it uploads a new regression result as reference - including the new RMS tests that were just merged?

JacksonBurns:
> I switched to the patched download-artifact action you suggested and turned on `ensure_latest` but it still found one from 2023-05-18T08:20:16Z. I think that might actually _be_ the most recent, because if you check lot of [recent runs on the main branch](https://github.com/ReactionMechanismGenerator/RMG-Py/actions?query=branch%3Amain+workflow%3A*Integration) have been skipping the "Upload Results as Reference" step.
> 

Ah, the upload results step has a syntax error in the if statement. This PR fixes that.

> I suggest that, once someone approves this, we merge it to main and see if it uploads a new regression result as reference - including the new RMS tests that were just merged?

Agreed.

JacksonBurns (approved):
Thanks for these updates - LGTM.

JacksonBurns:
> The overnight scheduled job uploaded a new reference result partying_face , and the CI test on matt's PR fetched them for comparison partying_face .

:tada:

> Unfortunately there seems to be a bug causing a crash when comparing some of the results (see [#2446 (comment)](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2446#issuecomment-1584572690) ) and quite a few of the regression tests have different core and edges -- is this to be expected by random fluctuations?

I can't comment on the science behind this, but I will say that a 'failure' in the old testing scheme was never established either. The output from the regression tests was just left to the user to read. The change you made in 1a424ef makes sense, but I think its up to us now to decide what constitutes failure.

PR #2466, opened 2023-06-11
Don't replace PDep kinetics of a library reaction

alongd:
### Motivation or Problem
Library reactions marked with an ``elementary_high_p`` flag set to ``True`` can either have PDep kinetics or Arrhenius kinetics. The high-pressure-limit kinetics is used in PDep networks to improve them instead of using RMG's estimations. However, if the original library reaction already has PDep kinetics, then this library rate should end up in the model rather than a PDep estimation. Of course, if the library reaction only has Arrhenius (non-PDep) kinetics, and the reaction is pressure-dependent, the kinetics must be replaced with the PDep estimation.

### Description of Changes
Only omit a library reaction from the core if an isomorphic  PDep reaction was created and if the library reaction has no PDep kinetics (changes done to model.py).
Also, and complementary, don't add a network reaction to the core if the core already has an isomorphic library reaction with PDep kinetics marked as `elementary_high_p = True`.

### DIFF
--- a/rmgpy/rmg/model.py
+++ b/rmgpy/rmg/model.py
@@ -864,9 +864,9 @@ class CoreEdgeReactionModel:
                 # Since PDepReactions are created as irreversible, not doing so
                 # would cause you to miss the reverse reactions!
                 self.add_reaction_to_unimolecular_networks(rxn, new_species=new_species, network=pdep_network)
-                if isinstance(rxn, LibraryReaction):
-                    # If reaction came from a reaction library, omit it from the core and edge so that it does
-                    # not get double-counted with the pdep network
+                if isinstance(rxn, LibraryReaction) and not rxn.kinetics.is_pressure_dependent():
+                    # If the reaction came from a library, and it does not have PDep kinetics,
+                    # omit it from the core and edge so that it does not get double-counted with the pdep network
                     if rxn in self.core.reactions:
                         self.core.reactions.remove(rxn)
                     if rxn in self.edge.reactions:
--- a/rmgpy/rmg/pdep.py
+++ b/rmgpy/rmg/pdep.py
@@ -898,7 +898,7 @@ class PDepNetwork(rmgpy.pdep.network.Network):
                         reactants=configurations[j],
                         products=configurations[i],
                         network=self,
-                        kinetics=None
+                        kinetics=None,
                     )
                     net_reaction = reaction_model.make_new_pdep_reaction(net_reaction)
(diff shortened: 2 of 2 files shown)


mjohnson541 (commented):
Implementation looks fine to me. 

However, suppose there is a network with channels A+B, C, D+E, F, and G+H and we have library calculations, but the library calculations only included channels A+B, C and D+E. I believe what this PR will do is cause all rates between channels in that library network to be library reactions and reactions involving a channel not considered in that library will be from a separate RMG generated network. 

1) Does this separate RMG generated network use the high_p_limit rate coefficients from the library or estimates from RMG for those rate coefficients?
2) We're at least guaranteed thermodynamic consistency in the combined network...but do we expect to have any concerns about how the pdep rate coefficients calculated using different network definitions will mesh in these cases?

alongd:
Thanks for the discussion, @mjohnson541, there are good points.

The intention of the `elementary_high_p` flag is to use the high-pressure kinetics information of a library reaction in RMG's networks, but the current behavior of replacing the library reaction in the final model with on-the-fly values is unexpected to users that create/use these libraries. It's also not the original intention of this flag.

Regarding point 1 of network separation, I think that most users would prefer that rate coefficients someone worked on for a while will be used as is. The (high-pressure limit) kinetic information will still be transmitted to the RMG network, hopefully making this separation less of an issue. It might be too complicated, but we could add another flag for library creators to decide whether to overwrite a PDep reaction in the final model with RMG's rates or not.

Regarding point 2 of meshing the two approaches, I think we'll never have a definite answer since there are many cases one can think of in terms of the quality of the data. My current approach, which I hope makes sense, is that RMG values, which are not based on 3D/QM calculations (again, for many cases, but probably not true for some other cases), are regarded as the best estimations we can make on-the-fly. In later stages, we refine some of the rates, and if they happened to be PDep then the library developer should take care to include a consistent set of PDep rates that captures all important pathways for a PES. "Important" could change by application (and by entry channel), and it's indeed not a simple issue. But library developers don't expect their carefully-put rates (hopefully...) to be overwritten. This PR intends to bring back the intended behavior of this flag instead of an unexpected behavior.

mjohnson541:
Okay sounds good. Point (1) was more about whether we're sure it's still using the high-p-limit properly for that with this implementation rather than what the right choice is.

alongd:
@mjohnson541, I rebased the branch, please let me know if there are any other comments, or if something wasn't fully addressed.

mjohnson541 (approved):

Issue #2455, opened 2023-06-05
During CI testing: Unable to dlopen(cxxpath) in parent! cannot open shared object file: File name too long


rwest:
### Bug Description

The build-documentation continuous integration run at https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/5180345336 failed when running sphinx with the log: 

```
Running Sphinx v5.0.2
making output directory... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 206 source files that are out of date
updating environment: [new config] 206 added, 0 changed, 0 removed
reading sources... [  0%] contents
reading sources... [  0%] latex-api
reading sources... [  1%] latex-rmg
reading sources... [  1%] reference/arkane/common
ERROR: Unable to dlopen(cxxpath) in parent!
Message: /usr/share/miniconda3/envs/rmg_env/lib/python3.7/site-packages/rdkit/../../../libstdc++.so.6nts a live Julia runtime.

        Note: Use `LibJulia` to fully control the initialization of
        the Julia runtime.

        Parameters
        ==========

        init_julia : bool
            If True, try to initialize the Julia runtime. If this code is
            being called from inside an already running Julia, the flag should
            be passed as False so the interpreter isn't re-initialized.

            Note that it is safe to call this class constructor twice in the
            same process with `init_julia` set to True, as a global reference
            is kept to avoid re-initializing it. The purpose of the flag is
            only to manage situations when Julia was initialized from outside
            this code.

        runtime : str
            Custom Julia binary, e.g. "/usr/local/bin/julia" or "julia-1.0.0".

        debug : bool
            If True, print some debugging information to STDERR
: cannot open shared object file: File name too long
Error: Process completed with exit code 1.
```

I can't see what has changed to cause it, so I am attempting to just restart the job and see what happens.

and shortly after that, this build-and-test job failed in a similar way
https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/5180282876/jobs/9334261409#step:9:39

```
run make test-unittests
  make test-unittests
  shell: /usr/bin/bash -l {0}
  env:
    RMG_DATABASE_BRANCH: main
    INPUT_RUN_POST: true
    CONDA: /usr/share/miniconda3
    CONDA_PKGS_DIR: /home/runner/conda_pkgs_dir
    PYTHONPATH: /home/runner/work/RMG-Py/RMG-Py/RMG-Py:
[...]

PR #2469, opened 2023-06-15
Revert `libstdcxx-ng` Patches in CI, Docs, and Dockerfile

JacksonBurns:
a new version of Julia has been released that fixes the issue we were facing, so these changes are no longer needed (see issue #2455 and PR #2456).

I have added that version of Julia (1.9.0) to the "do not allow" list for the environment file. We could probably find a way to specifically disallow the problematic _build_ instead, but this is easier.

With this CI update in place, we can look to figure out the issue with regression testing and then finally get the updated CI into RMG-database.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -43,6 +43,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest]
     runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.os == 'macos-latest' }}
     # skip scheduled runs from forks
     if: ${{ !( github.repository != 'ReactionMechanismGenerator/RMG-Py' && github.event_name == 'schedule' ) }}
     env: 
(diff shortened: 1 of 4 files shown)


.github/workflows/CI.yml
```diff
@@ -41,16 +42,7 @@ jobs:
...
-        libstdcxx-ng: [null, 12, 13]
-        exclude:
-          - os: macos-latest
-            libstdcxx-ng: 12
-          - os: macos-latest
-            libstdcxx-ng: 13
-          - os: ubuntu-latest
-            libstdcxx-ng: null
     runs-on: ${{ matrix.os }}
-    continue-on-error: ${{ ( matrix.libstdcxx-ng == 13 || matrix.os == 'macos-latest' ) }} # allow (expect) this to fail
```
rwest, line 53: Does the Mac one now pass often enough that we want to require that it passes? I'd say if it still times out for Not Our Fault 20% of the time then maybe we want to keep the continue-on-error

rwest (dismissed):
If this runs ok then it looks good to me (with one question/comment about requiring the Mac tests to pass)

rwest (approved):

JacksonBurns:
@rwest this will require changing the required checks back to just `ubuntu-latest` from `ubuntu-latest, 12`

rwest:
We forgot to do the documentation building CI job

PR #2470, opened 2023-06-15
Don't specify reactive for regression input

hwpang:
### Motivation or Problem
The current RMS regression test [failed](https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/5278941741/jobs/9548955163#step:18:266) due to the reactive flag in the regression input.

### Description of Changes
I removed the reactive flag from the regression input.


JacksonBurns:
@hwpang thanks for getting this! I think the other error is that `RMS_liquidSurface_ch4o2cat` is not generating a `chemkin` directory in the output? See [this failed regression test comparison](https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/5278941741/jobs/9548955163#step:18:239) on a PR where no source code changed.

hwpang:
@JacksonBurns Yeah, as you said, the chemkin input files generated have the `-gas` and `-surface` postfix, while we hard-coded the file name [here](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/main/.github/workflows/CI.yml#L204) for the core and edge comparison in the CI. I will fix this soon.

The input file for cantera was actually generated without any problem (at the path of `cantera/chem_annotated.yaml`)

rwest:
The `FileNotFoundError: [Errno 2] No such file or directory: 'stable_regression_results/RMS_liquidSurface_ch4o2cat/chemkin/chem_annotated-liquid.inp'` on Jackson's last commit makes sense, because the `stable_regression_results` are from the main branch which doesn't yet have this `chem_annotated-liquid.inp` stuff on it.  

But the failure on Hao-Wei's previous commit makes less sense to me 
```
aromatics Passed Core Comparison
Error: Process completed with exit code 1.
```

I've reverted Jackson's temporary commit and added another which should try to save the diff files, though this won't necessarily help debug the current issue.

rwest:
I think I figured it out.
`comparison_pass=$( python-jl scripts/checkModels.py args )`
doesn't do what you might think.
It will store the STDOUT from the command into the variable `$comparison_pass` but not the exit code.
So then `if $comparison_pass` will be quite different from what we want.
Hence it was reporting a pass when it failed, etc.

Let's see how this latest version does...

rwest:
The tests now run correctly (the CI instructions and bash scripts) but there's a problem with that new RMS liquid surface one:

Unable to calculate enthalpy for species 'CH4': no thermo or statmech data available.

@hwpang please could you take another look? (See the latest CI run log files)

hwpang:
I rebased with the newest main, allowed thermo files as inputs, and supplied the liquid Chemkin file as the thermo file for both core and edge comparison.

rwest:
The two-file output is part of Chemkin's syntax, and not something we can change. We might have to change the model comparison tool. I think our "load chemkin" methods should be able to read an entire model (surface and bulk-phase files simultaneously).  But perhaps this is all on https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2335 which has not yet been merged. Tagging @ChrisBNEU to see if he recalls the current state of things.

ChrisBNEU:
@rwest the current [load_chemkin_file](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/694c13ecc7a0f0042ed23db12706792f6e71a18b/rmgpy/chemkin.pyx#L963) method does not allow you to load the gas and surface files simultaneously into a reaction model, but my pr adds an extra arg for the surface input file. I also updated the [compare tool ](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/694c13ecc7a0f0042ed23db12706792f6e71a18b/rmgpy/tools/diffmodels.py#L70) to be able to analyze combined gas/surface models.

rwest:
Thanks Chris. 
I suggest we "put a pin in" this PR until #2335 is rebased and merged.

In the mean time I suggest we just skip the problematic case, and have made a PR to do so:
https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2473

hwpang:
@rwest @ChrisBNEU Thanks Richard and Chris! I agree that we can put a pin in this PR until https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2335 is rebased and merged. Feel free to tag me to review that PR if you need a reviewer.

JacksonBurns:
Tracking, blocked by:
 - [ ] #2335 
Closing this PR and deleting `fix_regression`. Once above is merged, please restore the branch, re-open PR, and resume work.

PR #2471, opened 2023-06-15
Fix docker instruction

hwpang:
### Motivation or Problem
I notice that we point the users to the latest at one point in the installation instruction, while we specify 3.1.1 in the other parts.

### Description of Changes
I change from latest to 3.1.1.


JacksonBurns (approved):
Great catch, thanks! LGTM

hwpang:
The CI failed at the regression test that I'm currently fixing in `https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2470`. I think we can merge this regardless.

JacksonBurns:
> The CI failed at the regression test that I'm currently fixing in #2470. I think we can merge this regardless.

Agreed, please do!

PR #2473, opened 2023-06-20
Getting the regression tests to work

rwest:
### Motivation or Problem
The regression tests are not working.
Pull request #2470  shows the difficulty in getting the RMS_liquidSurface_ch4o2cat test to work.
It'll be easier once #2335 is merged, which enables loading and comparison of heterogeneous catalysis models.
But in the mean time, let's just skip that RMS_liquidSurface_ch4o2cat test.
We'll still run the test case, so we know it doesn't crash, but we won't try
comparing the output for now, because trying to fix it is holding back our testing.
Also, we weren't saving the results of the model comparisons.

### Description of Changes

In this PR we don't try comparing the results of the RMS_liquidSurface_ch4o2cat with the "benchmark" run. We still run it, to check that it doesn't crash, but don't check the results.

This PR also enhances the regression testing by saving the results of the comparison in the cases that differences are found in the models, and uploads them as artifacts for later inspection.

### Testing
Parts of it have been tested on #2470, but let's wait for the CI to run.

### DIFF
--- a/scripts/checkModels.py
+++ b/scripts/checkModels.py
@@ -285,7 +285,7 @@ def initialize_log(verbose, log_file_name='checkModels.log'):
     logging.basicConfig(
         filename=log_file_name,
         filemode='w',
-        format='%(name)s=%(message)s',
+        format='%(message)s',
         level=verbose
     )
 
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -204,7 +204,7 @@ jobs:
 
             # Compare the edge and core
             if python-jl scripts/checkModels.py \
-                "$regr_test" \
+                "$regr_test-core" \
                 $REFERENCE/"$regr_test"/chemkin/chem_annotated.inp \
                 $REFERENCE/"$regr_test"/chemkin/species_dictionary.txt \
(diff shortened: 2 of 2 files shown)


scripts/checkModels.py
```diff
@@ -285,7 +285,7 @@ def initialize_log(verbose, log_file_name='checkModels.log'):
     logging.basicConfig(
         filename=log_file_name,
         filemode='w',
-        format='%(name)s=%(message)s',
```
hwpang, line 288: What does this change mean? What does it look like before?

hwpang (approved):
Just one question; otherwise LGTM.

rwest:
It works! (at running the tests). The artifacts section of the [job summary](https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/5326113227?pr=2473) has a zip of the [failed regression comparison results ubuntu-latest](https://github.com/ReactionMechanismGenerator/RMG-Py/suites/13740426838/artifacts/760838457) 

But we still don't have the tests _passing_.

scripts/checkModels.py
rwest: It used to print something like `checkModels=` at the start of every line of the log file. Not helpful and just made it harder to read.

PR #2316, opened 2022-06-19
RMG-Electrochem

mjohnson541:
Enables RMG to operate on solid electrolyte interface conditions in Li-ion batteries. Adds Li atoms, adds electrochemical reactions, adds a liquid-catalysis reactor and upgrades tree generation adding corrections for kinetic solvent effect.

PR #2491, opened 2023-06-23
A collection of Matt and David's commits from #2316

rwest:
### Motivation or Problem
Pull request #2316 which adds electrochemistry, adds lithium, adds solvation to surfaces, adds charge transfer reactions, is rather large and hard to review, and also has some necessarry but unrelated things, that can be helpful elsewhere.

### Description of Changes
I've pulled out a bunch (but not all) of those things into this PR so they can be reviewed and merged, then the electrochem PR can be rebased onto it.   I also reordered things so the fixups commits are mostly right after the things they're fixing.

One feature that was added was the ability to forbid species and functional groups in input files. As I was trying to add examples to the input files and documentation, I found it was already there, since this feature was added in https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2185 - authored around the same time in October 2021. Rather than add a second syntax for the same thing, I reverted the commits on this branch.

### Testing
I read the commits as I was cherry-picking, got the test suite to pass, and also added some more unit tests.

### DIFF
--- a/arkane/common.py
+++ b/arkane/common.py
@@ -390,7 +390,11 @@ def get_element_mass(input_element, isotope=None):
         number = input_element
     elif isinstance(input_element, str):
         symbol = input_element
-        number = next(key for key, value in symbol_by_number.items() if value == input_element)
+        try:
+            number = next(key for key, value in symbol_by_number.items() if value == input_element)
+        except:
+            symbol = input_element[0] + input_element[1].lower()
+            number = [key for key, value in symbol_by_number.items() if value == symbol][0]
 
     if symbol is None or number is None:
         raise ValueError('Could not identify element {0}'.format(input_element))
--- a/arkane/common.py
+++ b/arkane/common.py
@@ -391,10 +391,10 @@ def get_element_mass(input_element, isotope=None):
     elif isinstance(input_element, str):
         symbol = input_element
         try:
-            number = next(key for key, value in symbol_by_number.items() if value == input_element)
-        except:
-            symbol = input_element[0] + input_element[1].lower()
-            number = [key for key, value in symbol_by_number.items() if value == symbol][0]
+            number = number_by_symbol[symbol]
+        except KeyError:
+            symbol = input_element.capitalize()
+            number = number_by_symbol[symbol]
 
     if symbol is None or number is None:
         raise ValueError('Could not identify element {0}'.format(input_element))
@@ -438,6 +438,7 @@ symbol_by_number = {1: 'H', 2: 'He', 3: 'Li', 4: 'Be', 5: 'B', 6: 'C', 7: 'N', 8
                     92: 'U', 93: 'Np', 94: 'Pu', 95: 'Am', 96: 'Cm', 97: 'Bk', 98: 'Cf', 99: 'Es', 100: 'Fm', 101: 'Md',
                     102: 'No', 103: 'Lr', 104: 'Rf', 105: 'Db', 106: 'Sg', 107: 'Bh', 108: 'Hs', 109: 'Mt', 110: 'Ds',
                     111: 'Rg', 112: 'Cn', 113: 'Nh', 114: 'Fl', 115: 'Mc', 116: 'Lv', 117: 'Ts', 118: 'Og'}
+number_by_symbol = {value: key for key, value in symbol_by_number.items()}
 
 # Structure of mass_by_symbol items: list(list(isotope1, mass1, weight1), list(isotope2, mass2, weight2), ...)
 mass_by_symbol = {
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -3621,9 +3621,12 @@ class KineticsFamily(Database):
         inds = inds.tolist()
         revinds = [inds.index(x) for x in np.arange(len(inputs))]
 
-        pool = mp.Pool(nprocs)
+        if nprocs > 1:
+            pool = mp.Pool(nprocs)
+            kinetics_list = np.array(pool.map(_make_rule, inputs[inds]))
+        else:
+            kinetics_list = np.array(list(map(_make_rule, inputs[inds])))
 
-        kinetics_list = np.array(pool.map(_make_rule, inputs[inds]))
         kinetics_list = kinetics_list[revinds]  # fix order
 
         for i, kinetics in enumerate(kinetics_list):
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -4683,16 +4683,6 @@ def average_kinetics(kinetics_list):
     logA = 0.0
     n = 0.0
     Ea = 0.0
-    alpha = 0.5
-    electrons = None
-    if isinstance(kinetics_list[0], SurfaceChargeTransfer) or isinstance(kinetics_list[0], ArrheniusChargeTransfer):
-        if electrons is None:
-            electrons = kinetics_list[0].electrons.value_si
-        if not all(np.abs(k.V0.value_si) < 0.0001 for k in kinetics_list):
-            raise ValueError(f"Trying to average charge transfer rates with non-zero V0 values: {[k.V0.value_si for k in kinetics_list]}")
-        if not all(np.abs(k.alpha.value_si - 0.5) < 0.001 for k in kinetics_list):
-            raise ValueError(f"Trying to average charge transfer rates with alpha values not equal to 0.5: {[k.alpha for k in kinetics_list]}")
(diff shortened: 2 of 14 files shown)


rwest:
The only test failures are expected failures of the regression tests because of the random run-to-run variations.

rmgpy/data/kinetics/family.py
```diff
@@ -4670,3 +4673,80 @@ def _child_make_tree_nodes(family, child_conn, template_rxn_map, obj, T, nprocs,
...
+    elif Aunits in ['m^5/(mol^2*s)', 'cm^5/(mol^2*s)', 'm^5/(molecule^2*s)', 'cm^5/(molecule^2*s)']:
+        # surface: dissociative adsorption
+        Aunits = 'm^5/(mol^2*s)'
+    elif Aunits == '':
+        # surface: sticking coefficient
+        pass
+    else:
+        raise Exception('Invalid units {0} for averaging kinetics.'.format(Aunits))
+
+    if type(kinetics) not in [Arrhenius,SurfaceChargeTransfer,ArrheniusChargeTransfer]:
```
rwest, line 4725: Oops. These `ChargeTransfer` reaction classes not yet defined on this branch (a result of cherry-picking out of order)

rmgpy/rmg/input.py
```diff
@@ -1526,7 +1528,8 @@ def read_input_file(path, rmg0):
         'SMILES': smiles,
         'InChI': inchi,
         'adjacencyList': adjacency_list,
-        'adjacencyListGroup': adjacency_list_group,
+        'adjacencyListGroup': group_adjacency_list,
```
rwest, line 1531: We should add to one of the example input files, and documentation.

rmgpy/data/kinetics/family.py
```diff
@@ -4670,3 +4673,80 @@ def _child_make_tree_nodes(family, child_conn, template_rxn_map, obj, T, nprocs,
...
+    logA = 0.0
+    n = 0.0
+    Ea = 0.0
+    alpha = 0.5
+    electrons = None
+    if isinstance(kinetics_list[0], SurfaceChargeTransfer) or isinstance(kinetics_list[0], ArrheniusChargeTransfer):
+        if electrons is None:
+            electrons = kinetics_list[0].electrons.value_si
+        assert all(np.abs(k.V0.value_si) < 0.0001 for k in kinetics_list), [k.V0.value_si for k in kinetics_list]
+        assert all(np.abs(k.alpha.value_si - 0.5) < 0.001 for k in kinetics_list), [k.alpha for k in kinetics_list]
```
JacksonBurns, line 4692: Could these `assert` statements include a message, i.e. `assert all(...), "Expected all V0 to be..." - right now this will raise odd errors to the user, should these asserts not pass.

rmgpy/data/kinetics/family.py
rwest: I removed those (in a commit that can be reverted when needed)

rmgpy/data/kinetics/family.py
rwest: I replaced the Asserts with ValueErrors

rwest (commented):
@mjohnson541 isn't this "make it possible to explicitly forbid molecules and groups in the input file"  feature already here thanks to https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2185 ? 

Besides adding a new, undocumented, name for it, not sure what this commit adds? Can you use the existing mechanism in your input files and we drop this commit?

I see you authored this commit a couple of weeks before PR #2185 was merged, so you're forgiven for not using it back then :)

rmgpy/rmg/input.py
rwest: Instead I removed this feature, because it already exists in https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2185

rmgpy/data/kinetics/family.py
```diff
@@ -4670,3 +4673,55 @@ def _child_make_tree_nodes(family, child_conn, template_rxn_map, obj, T, nprocs,
...
+        count += 1
+        logA += np.log10(kinetics.A.value_si)
+        n += kinetics.n.value_si
+        Ea += kinetics.Ea.value_si
+
+    logA /= count
+    n /= count
+    Ea /= count
+    Aunits = kinetics_list[0].A.units
+    if Aunits == 'cm^3/(mol*s)' or Aunits == 'cm^3/(molecule*s)' or Aunits == 'm^3/(molecule*s)':
```
JacksonBurns, line 4697: All of these statements should look like `if Aunits in ("cm^3/(mol*s), (cm^3/(molecule*s)), ...)` rather than the big or piles.

JacksonBurns: Note use of tuple, not list.

JacksonBurns (changes requested):
Looks good overall, have requested some small changes for good code style and clarity.

rmgpy/data/kinetics/family.py
rwest: I was curious...
Screenshot 2023-07-14 at 2 09 45 PM

JacksonBurns: This isn't a speed thing as much as an overly strict linting rule - see [pylint's R1714](https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/consider-using-in.html) - that I just happen to really like. I was actually wrong though, we should be using a set which should run even faster.

rmgpy/data/kinetics/family.py
```diff
@@ -4670,3 +4673,55 @@ def _child_make_tree_nodes(family, child_conn, template_rxn_map, obj, T, nprocs,
...
+    count = 0
+    for kinetics in kinetics_list:
+        count += 1
+        logA += np.log10(kinetics.A.value_si)
+        n += kinetics.n.value_si
+        Ea += kinetics.Ea.value_si
+
+    logA /= count
+    n /= count
+    Ea /= count
```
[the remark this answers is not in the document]
rwest: But we'd have to iterate through the list multiple times (once for each attribute) and create a bunch of lists. Is this a verbosity thing? an efficiency thing? a less-likely-to-make-mistakes thing?
I don't like playing code golf for the sake of it.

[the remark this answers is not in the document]
JacksonBurns: > But we'd have to iterate through the list multiple times (once for each attribute) and create a bunch of lists. 

You can do it very efficiently in one line like this:
```python
from types import SimpleNamespace
import numpy as np

# mimic the kinetics objects
kinetics_list = [SimpleNamespace(a=1, b=2, c=3) for _ in range(3)]

a, b, c = np.mean([[np.log10(i.a), i.b, i.c] for i in kinetics_list], axis=1)
```

> Is this a verbosity thing? an efficiency thing? a less-likely-to-make-mistakes thing? I don't like playing code golf for the sake of it.

This is an antipattern, which we benefit from removing for all of these reasons. Numpy will be faster, less code is easier to maintain and read, etc.

rmgpy/data/kinetics/family.py
rwest: You're right that `in set` is faster still.

rmgpy/data/kinetics/family.py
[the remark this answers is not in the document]
rwest: A hole in one!

I quite like it. Although, having now _taught_ programming a few times, I am learning to appreciate the benefits of code that a novice can understand at a glance. More participation, fewer mistakes, less time spent explaining or parsing, etc. etc.

Anyway, the main reason to not do it your way here and now is that we soon need to revert this commit 
https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2491/commits/21801750f0396821f5700f0a6a344adecf3c927f
and the merge conflicts would be horrible to resolve.

We can revisit optimizing this code after that happens. 
(Though I think this code is seldom called)

arkane/common.py
```diff
@@ -390,7 +390,11 @@ def get_element_mass(input_element, isotope=None):
         number = input_element
     elif isinstance(input_element, str):
         symbol = input_element
-        number = next(key for key, value in symbol_by_number.items() if value == input_element)
+        try:
```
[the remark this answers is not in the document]
rwest: Done in cfe970f4bcaf4037696c7625f1439811c3db678f

rwest:
Think I've addressed all your comments. I pushed back on one suggestion - though I like it - because it'll cause horrible merge conflicts soon because that method needs to be more complicated than it currently looks. But the rest I adopted.

JacksonBurns (approved):
LGTM!

PR #2494, opened 2023-06-27
minor changes to CI to fix reusable workflow

JacksonBurns:
- `workflow_call` now accepts a branch name from RMG-database so we can pull the right branch
 - `actions/checkout` now explicitly checks out RMG-Py to avoid accidentally cloning RMG-database when re-using the workflow

Also updated docs in `CI.yml`.

These changes do not affect the CI in RMG-Py at all.


JacksonBurns:
Minor error, the default value of "main" is only read when calling from RMG-database, going to figure out a way to provide a default value when running from RMG-Py

mjohnson541 (approved):
Looks good to me!

Issue #2453, opened 2023-06-03
CI for Mac frequently timing out


rwest:
The continuous integration test runners for MacOS keeps timing out on the installing Julia dependencies step. 

This PR for example is marked to auto merge once the tests pass, but twice in a row the MacOS test has hung for 6 hours then timed out:
https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/5158881342

If we keep the MacOS test as a requirement for merging, it is going to hold everything up, unless we can figure out why that step is hanging and fix it.

JacksonBurns:
From what I have seen there is no consensus on why this happens, and all the solutions people have are random (see [here](https://github.com/actions/runner/issues/2063) and [here](https://github.com/actions/runner/issues/1326), for example). I'm fine with removing the requirement that the macos CI passes for PRs to be merged, but leaving it on as a "nice to have" for PRs and something that will at least run successfully overnight to hopefully catch any serious problems.

JacksonBurns:
Resolved in #2456 which now allows the Mac Runner to fail. If possible, re-evaluate requiring this runner to pass in the future if the GitHub mac runners become more consistent (see https://github.com/actions/runner/issues/2063).

PR #2500, opened 2023-07-17
Make the Continuous Integration Pass by Default

JacksonBurns:
Right now the regression tests are failing by default - this PR implements the following to fix that:
 - MacOS runners just can't build Julia dependencies, so only the ability to solve the environment and build RMG, then exit
 - Regression tests are considered 'passing' if they execute at all, the actual comparison will be written to the PR by a bot for manual review by humans

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -52,20 +52,14 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build-and-test-unix:
-    strategy:
-      matrix:
-        os: [ubuntu-latest, macos-latest]
-    runs-on: ${{ matrix.os }}
+  build-osx:
+    runs-on: macos-latest
     # skip scheduled runs from forks
     if: ${{ !( github.repository != 'ReactionMechanismGenerator/RMG-Py' && github.event_name == 'schedule' ) }}
     env: 
       # if running on RMG-Py but requiring changes on an un-merged branch of RMG-database, replace
       # main with the name of the branch
       RMG_DATABASE_BRANCH: main
-
-      # This is true only if this is a reference case for the regression testing:
-      REFERENCE_JOB: ${{ github.ref == 'refs/heads/main' && matrix.os =='ubuntu-latest' && github.repository == 'ReactionMechanismGenerator/RMG-Py' }}
     defaults:
       run:
         shell: bash -l {0}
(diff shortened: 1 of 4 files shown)


ChrisBNEU:
This looks good, really like the bot. I might have missed it but do we know why the MacOS runners can't build the julia dependencies?

JacksonBurns:
> This looks good, really like the bot. I might have missed it but do we know why the MacOS runners can't build the julia dependencies?

Thanks! See https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2453 TL;DR: the GitHub runners just don't seem up to the task.

ChrisBNEU (dismissed):
This looks good to me! Thanks putting the PR together.

rwest:
The `exit 0` for the MacOS doesn't seem to do what's intended.

In this run https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/5590577981/jobs/10220474560 the mac runner carried on to the julia step, timed out, and failed (thus killing the ubuntu run).

And in this run https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/5590578639/jobs/10220475578 the mac runner is still going.

JacksonBurns:
Ah, I had done this differently on a previous commit [like this](https://github.com/ReactionMechanismGenerator/RMG-Py/commit/5ddc6521156c05a63265877fafe6c909c74ff929) and then changed it to this hoping it would work. Fixing now.

mjohnson541:
I like it! Looks good to me!

rwest:
I merged this, so that we can get all the other pull requests to use it, but I can think of a couple of further improvemenst.
1. the Regression Test results are split between the "Regression Tests - Compare to Baseline" output in the job runner log (eg. [here](https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/5592025019/jobs/10223920519)) that I think nobody will now ever look at, and the current annotation which in fact has a bunch more detail than is often needed. Would be nicer if the type of summary that's currently in the job logs ended up in the annotation.
2. the mac runner *sometimes* succeeds at doing julia step. It might be nice to distinguish between a time-out (github computers not up to the task) and a failure (because something has changed in our install stack and things are actually broken for MacOS). i.e. we could let the MacOS runner *try* the julia steps, but accept a time-out as a "pass".

PR #2501, opened 2023-07-20
Expand the regression test annotations.

rwest:
### Motivation or Problem
The results of the regression tests were largely buried in the logs of the continuous integration action workflow, rather than being nicely summarized on the annotation to the pull request.

### Description of Changes
The summaries and comparisons that are generated by the CI running the regression tests are captured, and used to make the annotation.
The summary is stored as a text file, with the pull request number stored on the first line. This is then uploaded as an artifact. A new workflow runs whenever the CI workflow completes, then downloads the artifact, extracts the summary, and posts a comment. This complicated runaround is needed because workflows run by pull requests from forks do not have permission to write comments directly.

### Testing
For the markdown formatting, I have run it a bunch, and checked things render correctly. (See some of the run logs, where things are duplicated into the workflow summary). 
For the artifact and annotation scripts, I debugged and tested it on my own fork. It'll only function properly once the annotation.yml workflow ends up on the main branch - but I tested it there :)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -269,6 +269,7 @@ jobs:
         env:
           REFERENCE: stable_regression_results
         run: |
+          exec 2> >(tee -a regression.stderr >&2) 1> >(tee -a regression.stdout)
           mkdir -p "test/regression-diff"
           for regr_test in aromatics liquid_oxidation nitrogen oxidation sulfur superminimal RMS_constantVIdealGasReactor_superminimal RMS_CSTR_liquid_oxidation;
           do
@@ -284,6 +285,7 @@ jobs:
             echo -n 'Current:   '
             grep "Memory used:" test/regression/"$regr_test"/RMG.log | tail -1
 
+            echo "<details><summary>"
             # Compare the edge and core
             if python-jl scripts/checkModels.py \
                 "$regr_test-core" \
@@ -292,12 +294,15 @@ jobs:
                 test/regression/"$regr_test"/chemkin/chem_annotated.inp \
                 test/regression/"$regr_test"/chemkin/species_dictionary.txt
             then
-              echo "$regr_test Passed Core Comparison"
+              echo "$regr_test Passed Core Comparison ✅"
             else
-              echo "$regr_test Failed Core Comparison" | tee -a $GITHUB_STEP_SUMMARY
+              echo "$regr_test Failed Core Comparison ❌" | tee -a $GITHUB_STEP_SUMMARY
               cp "$regr_test-core.log" test/regression-diff/
               export FAILED=Yes
             fi
+            echo "</summary>"
+            cat "$regr_test-core.log"
+            echo "</details>\n<details><summary>"
             if python-jl scripts/checkModels.py \
                 "$regr_test-edge" \
                 $REFERENCE/"$regr_test"/chemkin/chem_edge_annotated.inp \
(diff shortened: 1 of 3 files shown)


.github/workflows/CI.yml
```diff
@@ -264,100 +264,111 @@ jobs:
...
-          echo "<summary>Regression Testing Results</summary>" >> summary.txt
-          echo "" >> summary.txt
-          tail -n +1 test/regression-diff/* >> summary.txt
-          echo ""
+          echo "## Regression Testing Results" > summary.txt
+          echo "${{ steps.regression-execution.outputs.stderr }}" >> summary.txt
+          echo "<details>" >> summary.txt
+          echo "<summary>Detailed regression test results.</summary>" >> summary.txt
+          echo "${{ steps.regression-comparison.outputs.stdout }}" >> summary.txt
           echo "</details>" >> summary.txt
```
JacksonBurns, line 355: Unless stdout is guaranteed to have a blank line at the end, you will need to add `echo ""` since `</details>` won't work without a blank line before it

JacksonBurns (changes requested):
Small note

.github/workflows/CI.yml
JacksonBurns: That's so much nicer - I will be doing that from now on

rwest: <details>
You can put the summary at the end (or even in the middle!). Helpful when your script spits out things as it goes and doesn't know the conclusion until it is complete
<summary>Something I discovered yesterday when trying to get this to work...</summary>

```
<details>
You can put the summary at the end! helpful when your script spits out things as it goes 
and doesn't know the pass/fail conclusion until it is complete
<summary>Something I discovered yesterday when trying to get this to work...</summary>
</details>
```
</details>

rwest:
I think this annotation idea is not going to work for pull requests from forked repositories. Only pull requests made from branches of the official repository.

Check the "Maximum access for pull requests from public forked repositories" column [here](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github\_token) 
This is a safety thing, to stop people doing nefarious things by opening a pull request that modifies the github action recipes.

Options seem to be...
* continue having PRs from the official repo rather than forks, so we keep annotations.
* give up on this annotations idea
* do something complicated with a `[pull_request_target`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) workflow, which runs in the context of the target (official main) branch and has extra permissions. But getting the results into that workflow might be complicated.

rwest:
Here is a workaround https://github.com/nyurik/auto_pr_comments_from_forks. The `pull_request` target saves the comment markdown as an artifact, that another workflow (in their case running on a cron job) posts as a comment.

JacksonBurns:
I will start a separate PR to implement this workaround to keep this one in scope.

rwest:
I have an idea I'm going to try

rwest:
Sorry for pushing a couple of commits to master. If we undo (reset the head to before I did that and force push) then those commits will become part of this pull request, which can then be merged.

Anyway, I got it working (with a bunch of testing done in relative privacy of my own fork).
This is a bit simpler than the workaround hack I linked to (and that you started in #2502).

rwest:
Dammit. Is #PR_NUMBER defined in the CI workflow?

rwest:
Dammit. Is #PR_NUMBER defined in the CI workflow?

rwest:
The "Continuous Integration" workflow runs on a push to master (eg merging a PR) and on a cron job, both of which will trigger the Annotation workflow, but neither of which will have uploaded a regression test summary file, so the Annotation workflow will fail.

PR #2502, opened 2023-07-21
hack to allow comments on PRs to RMG-Py from forks

JacksonBurns:
See https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2501


rwest:
I think the approach now in #2501 is simpler and perhaps more reliable (now that it's finally working).

JacksonBurns:
Closing in favor of #2501

PR #2505, opened 2023-07-23
Fix a bug in the CI workflow

rwest:
### Motivation or Problem
There are a couple of problems with the new regression test result annotations workflow introduced in #2501
1. I stupidly left out a couple of lines. We need to define the pull request number in an environment variable in the step where we write it to the file. Sorry. (I had done the testing on a much accelerated workflow, and then didn't correctly copy it across to the real workflow).
2. The "Continuous Integration" workflow runs on a push to master (eg merging a PR) and on a cron job, both of which will trigger the Annotation workflow, but neither of which will have uploaded a regression test summary file, so the Annotation workflow will fail. The only problem with this is folks getting emails and notifications about workflow failures. Ideally the workflow just wouldn't run in those cases, but I can't see how to filter a [`workflow_run`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) trigger this way.

### Description of Changes
Fix problem 1.

If someone has ideas about problem 2, contribute them before merging. Else let's just get this in to fix problem 1 - assuming that it does so (let's let the slow tests run this time)

### DIFF
--- a/.github/workflows/annotate.yml
+++ b/.github/workflows/annotate.yml
@@ -41,7 +41,7 @@ jobs:
             fs.writeFileSync(`${process.env.GITHUB_WORKSPACE}/regression_summary.zip`, Buffer.from(download.data));
 
       - name: Report the github.event.workflow_run as json
-        if: ${{ error() }}
+        if: ${{ failure() }}
         run: echo '${{ toJson(github.event.workflow_run) }}'
 
       - name: 'Unzip artifact'
--- a/.github/workflows/annotate.yml
+++ b/.github/workflows/annotate.yml
@@ -19,6 +19,7 @@ jobs:
     steps:
       - run: echo 'The triggering workflow passed'
       - name: 'Download regression_summary artifact'
+        continue-on-error: true
         uses: actions/github-script@v6
         with:
           script: |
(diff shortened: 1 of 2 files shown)


JacksonBurns:
@rwest for problem number 2 - put a `continue-on-error: true` for the `on-sucess` job so it passes regardless?

rwest:
I've cherry-picked 12fa87b77666c18fb61ad78bcaaea9cca65d9d14 onto my own main, which might allow us to test it (it'll only function when on a main branch).

rwest:
I think I likely did it wrong.

.github/workflows/annotate.yml
```diff
@@ -39,16 +40,22 @@ jobs:
             let fs = require('fs');
             fs.writeFileSync(`${process.env.GITHUB_WORKSPACE}/regression_summary.zip`, Buffer.from(download.data));
 
+      - name: Report the github.event.workflow_run as json
+        if: ${{ error() }}
```
rwest, line 44:
```suggestion
        if: ${{ failure() }}
```

rwest (commented):
D'oh. I blame copilot!

JacksonBurns (approved):
LGTM! Thanks for the quick fix

PR #2506, opened 2023-07-24
More github workflow improvements

rwest:
### Motivation or Problem
More improvements to the new github workflows

### Description of Changes
* the regression summary is prepared and added to the workflow job summary for everything but the reference job (the push to main and nightly build).
* the annotation workflow is only run for pull requests, not for pushes and scheduled runs.
* if the annotation job fails for some reason, it reports the event that triggered it, for debugging..

### Testing
I ran things on my main branch. I think they're working - but this workflow stuff has been remarkably hard to get right first time.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -346,9 +346,9 @@ jobs:
           fi
 
       - name: Prepare Results for PR Comment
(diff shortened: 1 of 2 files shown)


JacksonBurns (approved):
LGTM, though as you said this has been vexing - let's hope this is the fix that gets it.

Issue #2499, opened 2023-07-10
yaml writer slows down execution


ChrisBNEU:
### Bug Description
It appears that the yaml writer uses a lot of overhead during the execution of an rmg run. I ran a modified ethane pyrolysis input file with cProfile. The only thing I changed in rmg was I did not write a yaml file for rms every execution. I did that by commenting out [this line](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/ef83a1c0e33854aaf749484755cc02a943c1ab01/rmgpy/rmg/main.py#L688C54-L688C54) in main.py: 

```python3
self.attach(RMSWriter(self.output_directory))
```

the difference in execution times were: 
- with RMSWriter listener: 15 min 15 sec
- without RMSWriter listener: 7min 35 sec

I attached the profiles and the rmg logs below. It is possible this becomes insignificant at longer execution times, I have not tried it. The Profiler graph shows it executing every time the model is enlarged, so it is also possible it gets worse with larger mechanisms.

### How To Reproduce
Comment out the line specified above, then run the profiler on the input file I attached below (`python-jl rmg.py -p input.py`)

### Installation Information
Describe your installation method and system information.
 - macOS 10.15.7
 - Installed from source via conda
 - RMG version information:
   - RMG-Py: 3.0.0-1437-gef83a1c0e
   - RMG-database: 3.1.0-616-g75fabcb6c

### Additional Context
input file used: 
[input.py.zip](https://github.com/ReactionMechanismGenerator/RMG-Py/files/11996169/input.py.zip)

profiles: 
[profile_with_listener.pdf](https://github.com/ReactionMechanismGenerator/RMG-Py/files/11996153/profile_with_listener.pdf)
[profile_without_listener.pdf](https://github.com/ReactionMechanismGenerator/RMG-Py/files/11996145/profile_without_listener.pdf)

rmg logs: 
[with_listener.log](https://github.com/ReactionMechanismGenerator/RMG-Py/files/11996157/with_listener.log)
[without_listener.log](https://github.com/ReactionMechanismGenerator/RMG-Py/files/11996158/without_listener.log)

JacksonBurns:
Thanks for the thorough report @ChrisBNEU. I think a short term solution would be to add a way to optionally disable this writer from the input, but long term would be to use multiprocessing to spawn a task that runs the yaml writing parallel to the simulation itself. We could also look at a different output format that writes faster, like JSON.

[...]

PR #2508, opened 2023-07-26
Allow disabling of RMS yaml writer for speedup

ChrisBNEU:
### Motivation or Problem
See issue #2499, profiling and rmg execution showed that we spend a significant amount of time writing rms yaml output files. this pr adds an option that allows the user to generate a rms yaml file each iteration (like we do with chemkin files), or just one yaml file at the end of the run (like cantera)

### Description of Changes
Added `generateRMSEachIter` input file arg for the `options` block. `True` makes an rms input file every iter as before. `False` generates only one. The default is `False`. 

### Testing
Ran the Minimal example with `True` set and `False` set. They performed as expected, and the final output files were identical. I think in #2499 we decided that this would be a temporary fix, so I am not sure what level of additional unit testing is appropriate. There do not seem to be existing tests that test some of these miscellaneous options (although I could be looking in the wrong places). 

### Reviewer Tips
I'd like a second opinion on what built in unit testing/functional testing might be required. I couldn't find one that seemed appropriate in `inputTest.py` or `mainTest.py`.

### DIFF
--- a/documentation/source/users/rmg/input.rst
+++ b/documentation/source/users/rmg/input.rst
@@ -956,6 +956,7 @@ Miscellaneous options::
         saveSeedToDatabase=True,
         units='si',
         generateOutputHTML=True,
+        generateRMSEachIter=True,
         generatePlots=False,
         saveSimulationProfiles=True,
         verboseComments=False,
@@ -975,6 +976,8 @@ The ``units`` field is set to ``si``.  Currently there are no other unit options
 
 Setting ``generateOutputHTML`` to ``True`` will let RMG know that you want to save 2-D images (png files in the local ``species`` folder) of all species in the generated core model.  It will save a visualized
 HTML file for your model containing all the species and reactions.  Turning this feature off by setting it to ``False`` may save memory if running large jobs.
+
+Setting ``generateRMSEachIter`` to ``True`` will have rmg generate an output yaml file containing the core mechanism for the current iteration, with the number of core species appended to the name of the file (e.g. for 28 core species, the file will be named ``chem28.rms``). If it is ``False`` then only one file will be generated at the end of the run. A value of ``False`` may result in a faster run, as this feature can take up a lot of cpu time. 
 
 Setting ``generatePlots`` to ``True`` will generate a number of plots describing the statistics of the RMG job, including the reaction model core and edge size and memory use versus  execution time. These will be placed in the output directory in the plot/ folder.
 
--- a/rmgpy/rmg/input.py
+++ b/rmgpy/rmg/input.py
@@ -1337,8 +1337,8 @@ def pressure_dependence(
 
 def options(name='Seed', generateSeedEachIteration=True, saveSeedToDatabase=False, units='si', saveRestartPeriod=None,
             generateOutputHTML=False, generatePlots=False, saveSimulationProfiles=False, verboseComments=False,
-            saveEdgeSpecies=False, keepIrreversible=False, trimolecularProductReversible=True, wallTime='00:00:00:00',
-            saveSeedModulus=-1):
+            saveEdgeSpecies=False, keepIrreversible=False, trimolecularProductReversible=True, generateRMSEachIter=True, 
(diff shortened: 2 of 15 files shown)


mjohnson541:
I'm fine with the option, but I think it should be `True` by default. There are outputs that Chemkin can't properly represent and it's quite common now to need to take mechanisms from runs that don't complete.

JacksonBurns:
FWIW all of the regression tests executed faster (might be spurious)

ChrisBNEU:
@JacksonBurns I think `functools.cache` might only be available in python 3.9, and I think our tests/env uses 3.7.

JacksonBurns:
> @JacksonBurns I think `functools.cache` might only be available in python 3.9, and I think our tests/env uses 3.7.
> 
> [link](https://docs.python.org/3/library/functools.html#functools.cache)

Yeah that's what I get for editing from the browser

JacksonBurns:
Unfortunately caching will not work, so this last two commits can be dropped. We could make caching work if we re-wrote the `obj_to_dict` method to operate on some hashable type, but at that point it would just be faster to speed up the yaml writing instead.

JacksonBurns:
I have added a simple unit test to check that we can override the default behavior of RMG with respect to the new `generateRMSEachIter` option, and have disabled the writer in all of the regression tests and input test for speed's sake (and because we weren't doing anything with those files anyway).

JacksonBurns (dismissed):
I think this is ready to merge. Thanks for moving on this quickly @ChrisBNEU.

JacksonBurns:
@ChrisBNEU after #2516 is in we should rebase and get this PR through, it's a straightforward and quite helpful change.

JacksonBurns:
@ChrisBNEU I think we need to drop 8cbdf1e (bad change) and then just fix that unit test to access the property correctly, then this should be good to go.

JacksonBurns:
#2539 is overlapping with this PR. Over there, I have RMG automatically detect if RMS' reactors are actually being used (or if Julia is even installed), and then it disable the writer appropriately. This PR adds the option to turn it off, which I think would make sense if users are actually running RMS reactors. How do we want these two PRs to overlap?

JacksonBurns:
We should leave the RMS yaml writer on for one of the regression tests to check that it works (suggestion from @hwpang).

ChrisBNEU:
@JacksonBurns Sorry for the lack of communication on this. 

> #2539 is overlapping with this PR. Over there, I have RMG automatically detect if RMS' reactors are actually being used (or if Julia is even installed), and then it disable the writer appropriately. This PR adds the option to turn it off, which I think would make sense if users are actually running RMS reactors. How do we want these two PRs to overlap?

We could make it an argument for the rms reactors in the input file, instead of an option. This way the non-rms ones would be unaffected. 

> We should leave the RMS yaml writer on for one of the regression tests to check that it works (suggestion from @hwpang).

I can turn it on for the liq catalysis example? That example probably has the most variety in terms of species and phases written to the output yaml, so it would probably be better for testing purposes.

JacksonBurns:
> @JacksonBurns Sorry for the lack of communication on this.

My apologies actually - as you can see, I'm a bit behind :sweat_smile: 

> We could make it an argument for the rms reactors in the input file, instead of an option. This way the non-rms ones would be unaffected.

We have had some dicssussion about that PR in the intervening months, and decided to move the 'optional RMS reactors' part to a new PR. The same point you made here though, still important. Are you thinking we would add an option like 'rms_reactors: True/false' to the input file, that way I don't have to try and deduce if they are needed in the backend?

> I can turn it on for the liq catalysis example? That example probably has the most variety in terms of species and phases written to the output yaml, so it would probably be better for testing purposes.

That's a great idea. If you can turn that on here and rebase this PR, I think we can get this merged.

JacksonBurns:
@ChrisBNEU I have resolved the merge conflicts for this PR and I think it is close. The stuff about adding `rms_reactors` to the input file is proceeding slowly in #2631 so we can go ahead and merge this as soon as it's ready instead.

PR #2511, opened 2023-07-31
adding resonance for adsorbates

bjkreitz:
### Motivation or Problem
Adsorbate molecules can exhibit resonance similar to gas-phase molecules. For example, the bidentate adsorbate XCXC (where X is a surface site) has the following three configurations:
[X]C#C[X] <-> [X]#CC#[X] <-> [X]=C=C=[X]
Currently, RMG cannot identify resonance structures for adsorbates. 
In this pull request, I made the first step and added functions for the generation of resonance structures for adsorbates. 

### Description of Changes
Resonance occurs only for adsorbates that have at least two binding sites, also known as bidentate adsorbates. 
Therefore, a function was added to molecule.py to check if an adsorbate has two binding sites. 
The pathfinder.py was updated with functions to find the structures of adsorbates which can eventually have resonance structures. The structures that are searched for in the adsorbates are X-C-C-X and X-C-C-C-X. 
Five types of resonance for adsorbates were added to resonance.py. The identified and implemented resonance types are as follows:

- Shift 2 electrons from a C=/#C bond to the X-C bond (generate_adsorbate_shift_down_resonance_structures)
- Shift 2 electrons from a X=/#C bond to a C-C bond (generate_adsorbate_shift_up_resonance_structures)
- Shift 4 electrons from a C#C bond to a X-C bond (generate_adsorbate_double_shift_down_resonance_structures)
- Shift 4 electrons from a X#C bond to a C-C bond (generate_adsorbate_double_shift_up_resonance_structures)
- Shift 2 electrons in a conjugate pi system for bridged X-C-C-C-X adsorbates (generate_adsorbate_conjugate_resonance_structures)

Apologies for the funky names of the resonance types. Resonance has not really been discussed in catalysis yet. 
This pull request is work in progress and requires more changes to the source code. This is meant to start the discussion on resonance for adsorbates. 

### Testing
I have tested the functionalities of the added resonance functions by providing various adsorbates with and without resonance and confirmed that the functions work properly. 
An RMG run for an catalytic system completed also without any errors. 

### Reviewer Tips
Use the functions in a standalone Jupyter notebook and see if RMG can identify the resonance structures correctly. A selection of bidentate adsorbates with resonance is: "XCXC, XCHXCH, XCXCH, XCXCCH2, XCHXCCH2, XCHXCXC, XCHXCXCH, XCHCXC, XCHCHXC, XCHCHXCH" where X is a surface atom. 

### Open Tasks

A function is required to generate the resonance structures for bidentate format (XOC(H)XO) and bicarbonate (XOC(OH)XO). Adding resonance generation for the formate case requires more changes to RMG to handle 'van der Waals' bonds. These changes are going to be part of another pull request.

### DIFF
--- a/rmgpy/molecule/molecule.py
+++ b/rmgpy/molecule/molecule.py
@@ -2864,6 +2864,16 @@ class Molecule(Graph):
         cython.declare(atom=Atom)
         return [atom for atom in self.atoms if atom.is_surface_site()]
 
+    def is_multidentate(self):
+        """
+        Return ``True`` if the adsorbate contains at least two binding sites,
+        or ``False`` otherwise.
+        """
+        cython.declare(atom=Atom)
+        if len([atom for atom in self.atoms if atom.is_surface_site()])>=2:
+            return True
+        return False
+
     def get_adatoms(self):
         """
         Get a list of adatoms in the molecule.
--- a/rmgpy/molecule/molecule.py
+++ b/rmgpy/molecule/molecule.py
@@ -883,6 +883,13 @@ class Bond(Edge):
         """
         return self.is_order(4)
 
+    def is_double_or_triple(self):
+        """
+        Return ``True`` if the bond represents a double or triple bond or ``False``
+        if not.
+        """
+        return self.is_order(2) or self.is_order(3)
+
     def is_benzene(self):
         """
         Return ``True`` if the bond represents a benzene bond or ``False`` if
--- a/rmgpy/molecule/pathfinder.py
+++ b/rmgpy/molecule/pathfinder.py
@@ -480,3 +480,58 @@ def is_atom_able_to_lose_lone_pair(atom):
     return (((atom.is_nitrogen() or atom.is_sulfur()) and atom.lone_pairs in [1, 2, 3])
             or (atom.is_oxygen() and atom.lone_pairs in [2, 3])
             or atom.is_carbon() and atom.lone_pairs == 1)
+
+
+def find_adsorbate_delocalization_paths(atom1):
+    """
+    Find all multidentate adsorbates which have a bonding configuration X-C-C-X.
+    Examples:
+
+    - XCXC, XCHXCH, XCXCH, where X is the surface site. The adsorption site X
+      is always placed on the left-hand side of the adatom and every adatom
+      is bonded to only one surface site X.
+
+    In this transition atom1 and atom4 are surface sites while atom2 and atom3
+    are carbon or nitrogen atoms.
+    """
+    cython.declare(paths=list, atom2=Vertex, atom3=Vertex, atom4=Vertex, bond12=Edge, bond23=Edge, bond34=Edge)
+
+    paths = []
+    if atom1.is_surface_site():
+        for atom2, bond12 in atom1.edges.items():
+            if atom2.is_carbon():
+                for atom3, bond23 in atom2.edges.items():
+                    if atom3.is_carbon():
+                        for atom4, bond34 in atom3.edges.items():
+                            if atom4.is_surface_site():
+                                paths.append([atom1, atom2, atom3, atom4, bond12, bond23, bond34])
+    return paths
+
+
+def find_adsorbate_conjugate_delocalization_paths(atom1):
+    """
+    Find all multidentate adsorbates which have a bonding configuration X-C-C-C-X.
+    Examples:
+
+    - XCHCHXCH/XCHCHXC, where X is the surface site. The adsorption site X
+      is always placed on the left-hand side of the adatom and every adatom
+      is bonded to only one surface site X.
+
+    In this transition atom1 and atom5 are surface sites while atom2, atom3,
+    and atom4 are carbon or nitrogen atoms.
+    """
+
+    cython.declare(paths=list, atom2=Vertex, atom3=Vertex, atom4=Vertex, atom5=Vertex, bond12=Edge, bond23=Edge, bond34=Edge, bond45=Edge)
+
+    paths = []
+    if atom1.is_surface_site():
+        for atom2, bond12 in atom1.edges.items():
+            if atom2.is_carbon():
+                for atom3, bond23 in atom2.edges.items():
+                    if atom3.is_carbon():
+                        for atom4, bond34 in atom3.edges.items():
+                            if atom2 is not atom4 and atom4.is_carbon():
+                                for atom5, bond45 in atom4.edges.items():
+                                    if atom5.is_surface_site():
+                                        paths.append([atom1, atom2, atom3, atom4, atom5, bond12, bond23, bond34, bond45])
+    return paths
--- a/rmgpy/molecule/pathfinder.py
+++ b/rmgpy/molecule/pathfinder.py
@@ -499,9 +499,9 @@ def find_adsorbate_delocalization_paths(atom1):
     paths = []
     if atom1.is_surface_site():
         for atom2, bond12 in atom1.edges.items():
-            if atom2.is_carbon():
+            if atom2.is_carbon() or atom2.is_nitrogen():
                 for atom3, bond23 in atom2.edges.items():
-                    if atom3.is_carbon():
+                    if atom3.is_carbon() or atom3.is_nitrogen():
                         for atom4, bond34 in atom3.edges.items():
                             if atom4.is_surface_site():
                                 paths.append([atom1, atom2, atom3, atom4, bond12, bond23, bond34])
@@ -526,11 +526,11 @@ def find_adsorbate_conjugate_delocalization_paths(atom1):
     paths = []
     if atom1.is_surface_site():
         for atom2, bond12 in atom1.edges.items():
-            if atom2.is_carbon():
+            if atom2.is_carbon() or atom2.is_nitrogen():
                 for atom3, bond23 in atom2.edges.items():
-                    if atom3.is_carbon():
+                    if atom3.is_carbon() or atom3.is_nitrogen():
                         for atom4, bond34 in atom3.edges.items():
-                            if atom2 is not atom4 and atom4.is_carbon():
+                            if atom2 is not atom4 and (atom4.is_carbon() or atom4.is_nitrogen()):
                                 for atom5, bond45 in atom4.edges.items():
                                     if atom5.is_surface_site():
                                         paths.append([atom1, atom2, atom3, atom4, atom5, bond12, bond23, bond34, bond45])
--- a/rmgpy/molecule/resonance.py
+++ b/rmgpy/molecule/resonance.py
@@ -1180,20 +1180,21 @@ def generate_adsorbate_shift_up_resonance_structures(mol):
         for atom in mol.vertices:
             paths = pathfinder.find_adsorbate_delocalization_paths(atom)
             for atom1, atom2, atom3, atom4, bond12, bond23, bond34 in paths:
-                if (bond12.is_double() and bond23.is_single() and bond34.is_double()) or (bond23.is_double() and bond12.is_double() and bond34.is_double()) or (bond12.is_triple() and bond23.is_single() and bond34.is_triple()) or (bond12.is_triple() and bond23.is_single() and bond34.is_double()):
-                    bond12.decrement_order()
-                    bond23.increment_order()
-                    bond34.decrement_order()
-                    structure = mol.copy(deep=True)
-                    bond12.increment_order()
-                    bond23.decrement_order()
-                    bond34.increment_order()
-                    try:
-                        structure.update_atomtypes(log_species=False)
-                    except AtomTypeError:
-                        pass
-                    else:
-                        structures.append(structure)
+                if ((bond12.is_double_or_triple() and bond23.is_single() and bond34.is_double_or_triple()) or
+                    (bond12.is_double() and bond23.is_double() and bond34.is_double())):
+                        bond12.decrement_order()
+                        bond23.increment_order()
+                        bond34.decrement_order()
+                        structure = mol.copy(deep=True)
+                        bond12.increment_order()
+                        bond23.decrement_order()
+                        bond34.increment_order()
+                        try:
+                            structure.update_atomtypes(log_species=False)
+                        except AtomTypeError:
+                            pass
+                        else:
+                            structures.append(structure)
     return structures
 
 
@@ -1215,9 +1216,9 @@ def generate_adsorbate_conjugate_resonance_structures(mol):
         for atom in mol.vertices:
             paths = pathfinder.find_adsorbate_conjugate_delocalization_paths(atom)
             for atom1, atom2, atom3, atom4, atom45, bond12, bond23, bond34, bond45 in paths:
-                    if ((bond12.is_double() or bond12.is_triple()) and
+                    if (bond12.is_double_or_triple() and
                         (bond23.is_single() or bond23.is_double()) and
-                        (bond34.is_double() or bond34.is_triple()) and
+                        bond34.is_double_or_triple() and
                         (bond45.is_single() or bond45.is_double())):
                                 bond12.decrement_order()
                                 bond23.increment_order()
--- a/rmgpy/molecule/resonance.py
+++ b/rmgpy/molecule/resonance.py
@@ -49,6 +49,10 @@ Currently supported resonance types:
     - ``generate_kekule_structure``: generate a single Kekule structure for an aromatic compound (single/double bond form)
     - ``generate_opposite_kekule_structure``: for monocyclic aromatic species, rotate the double bond assignment
     - ``generate_clar_structures``: generate all structures with the maximum number of pi-sextet assignments
+- Multidentate adsorbates only
+    - ``generate_adsorbate_shift_down_resonance_structures``: shift 2 electrons from a C=/#C bond to the X-C bond
+    - ``generate_adsorbate_shift_up_resonance_structures``: shift 2 electrons from a X=/#C bond to a C-C bond
+    - ``generate_adsorbate_conjugate_resonance_structures``: shift 2 electrons in a conjugate pi system for bridged X-C-C-C-X adsorbates
 """
 
 import logging
@@ -87,6 +91,9 @@ def populate_resonance_algorithms(features=None):
             generate_aryne_resonance_structures,
             generate_kekule_structure,
             generate_clar_structures,
+            generate_adsorbate_shift_down_resonance_structures,
+            generate_adsorbate_shift_up_resonance_structures,
+            generate_adsorbate_conjugate_resonance_structures
         ]
     else:
         # If the molecule is aromatic, then radical resonance has already been considered
@@ -110,7 +117,10 @@ def populate_resonance_algorithms(features=None):
                 # solution. A more holistic approach would be to identify these cases in generate_resonance_structures,
                 # and pass a list of forbidden atom ID's to find_lone_pair_multiple_bond_paths.
                 method_list.append(generate_lone_pair_multiple_bond_resonance_structures)
-
+        if features['is_multidentate']:
+            method_list.append(generate_adsorbate_shift_down_resonance_structures)
+            method_list.append(generate_adsorbate_shift_up_resonance_structures)
+            method_list.append(generate_adsorbate_conjugate_resonance_structures)
     return method_list
 
 
@@ -130,6 +140,7 @@ def analyze_molecule(mol, save_order=False):
                 'is_aryl_radical': False,
                 'hasNitrogenVal5': False,
                 'hasLonePairs': False,
+                'is_multidentate': mol.is_multidentate(),
                 }
 
     if features['is_cyclic']:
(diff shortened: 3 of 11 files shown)


rmgpy/molecule/pathfinder.py
```diff
@@ -480,3 +480,51 @@ def is_atom_able_to_lose_lone_pair(atom):
     return (((atom.is_nitrogen() or atom.is_sulfur()) and atom.lone_pairs in [1, 2, 3])
             or (atom.is_oxygen() and atom.lone_pairs in [2, 3])
             or atom.is_carbon() and atom.lone_pairs == 1)
+
```
alongd, line 483: Do you know whether we actively "eradicate" radical electrons in an adsorbed molecule, e.g., by turning them into a lone pair, and same for partial charges? I guess not, since this will add a net charge to the molecule and RMG does not support that yet, right? Let's discuss this offline

rmgpy/molecule/pathfinder.py
```diff
@@ -480,3 +480,51 @@ def is_atom_able_to_lose_lone_pair(atom):
...
+def find_adsorbate_delocalization_paths(atom1):
+    """
+    Find all bidentate adsorbates which have a bonding configuration X-C-C-X.
+    Examples:
+
+    - XCXC, XCHXCH, XCXCH, where X is the surface site
```
alongd, line 489: What is the meaning of, e.g., `XCXC`, is the second X connected to the 1st or 2nd carbon? Or maybe it doesn't matter?

alongd (commented):
Thanks!! I added some questions

rmgpy/molecule/resonance.py
```diff
@@ -1116,3 +1133,183 @@ def _clar_transformation(mol, aromatic_ring):
...
+                    bond34.increment_order()
+                    try:
+                        structure.update_atomtypes(log_species=False)
+                    except AtomTypeError:
+                        pass
+                    else:
+                        structures.append(structure)
+    return structures
+
+def generate_adsorbate_double_shift_up_resonance_structures(mol):
```
alongd, line 1204: As we discussed, I wonder if RMG will capture this pathway on its own by applying the `generate_adsorbate_shift_up_resonance_structures` template twice

rmgpy/molecule/molecule.py
```diff
@@ -2759,6 +2759,16 @@ def get_surface_sites(self):
         cython.declare(atom=Atom)
         return [atom for atom in self.atoms if atom.is_surface_site()]
 
+    def is_bidentate(self):
+        """
+        Return ``True`` if the adsorbate contains at least two binding sites,
+        or ``False`` otherwise.
+        """
+        cython.declare(atom=Atom)
+        if len([atom for atom in self.atoms if atom.is_surface_site()])>=2:
```
alongd, line 2768: we also return `True` if the molecule is tridentate. Would we like to reflect that in the docstring (if not in the function name)?

rmgpy/molecule/resonance.py
```diff
@@ -1116,3 +1133,183 @@ def _clar_transformation(mol, aromatic_ring):
...
+                        pass
+                    else:
+                        structures.append(structure)
+    return structures
+
+def generate_adsorbate_conjugate_resonance_structures(mol):
+    """
+    Generate all of the resonance structures formed by the shift of two electrons in a conjugated pi bond system of
+    a bidentate adsorbate with a bridging C atom in between.
+    Example XCHCHXC: [X]#CC=C[X] <=> [X]=C=CC=[X]
```
alongd, line 1284: A base case example would help understanding (and good to also keep the preset example as well)

rmgpy/molecule/pathfinder.py
```diff
@@ -480,3 +480,51 @@ def is_atom_able_to_lose_lone_pair(atom):
...
+                                path.append([atom1, atom2, atom3, atom4, bond12, bond23, bond34])
+                                return path
+    return path
+
+def find_adsorbate_conjugate_delocalization_paths(atom1):
+    """
+    Find all bidentate adsorbates which have a bonding configuration X-C-C-C-X.
+    Examples:
+
+    - XCHCHXCH/XCHCHXC, where X is the surface site
```
alongd, line 512: It's a bit confusing for me that `X` is not always listed at the end. Is this customary?

rmgpy/molecule/resonance.py
```diff
@@ -1116,3 +1133,183 @@ def _clar_transformation(mol, aromatic_ring):
...
+
+    structures = []
+    if mol.is_bidentate():
+        for atom in mol.vertices:
+            paths = pathfinder.find_adsorbate_conjugate_delocalization_paths(atom)
+            for atom1, atom2, atom3, atom4, atom45, bond12, bond23, bond34, bond45 in paths:
+                if bond23.is_single():
+                    if (bond12.is_double() or bond12.is_triple()):
+                        if (bond34.is_double() or bond34.is_triple()):
+                            if (bond45.is_single() or bond45.is_double()):
```
alongd, line 1299: Consider turning these four nested `if` conditions into a single `if` condition with `and` in between.
Not a strong recommendation though, if you think the `and` version will be less readable, then OK to keep the nested version

rmgpy/molecule/resonance.py
```diff
@@ -1116,3 +1133,183 @@ def _clar_transformation(mol, aromatic_ring):
...
+    cython.declare(structures=list, paths=list, index=cython.int, structure=Graph)
+    cython.declare(atom=Vertex, atom1=Vertex, atom2=Vertex, atom3=Vertex, atom4=Vertex, atom5=Vertex, bond12=Edge, bond23=Edge, bond34=Edge, bond45=Edge)
+    cython.declare(v1=Vertex, v2=Vertex)
+
+    structures = []
+    if mol.is_bidentate():
+        for atom in mol.vertices:
+            paths = pathfinder.find_adsorbate_conjugate_delocalization_paths(atom)
+            for atom1, atom2, atom3, atom4, atom45, bond12, bond23, bond34, bond45 in paths:
+                if bond23.is_single():
```
alongd, line 1296: Do you think this pathway could also be valid if `bond_23` is double?
If I apply the recipe for
`X1 {D/T} C2 {S} C3 {D/T} C4 {S/D} X5`
I get
`X1 {S/D} C2 {D} C3 {S/D} C4 {D/T} X5`
reversing the atom labels this is:
`X1 {D/T} C2 {S/D} C3 {D} C4 {S/D} X5`
I expected that the 3rd line will be identical to the 1st line, since the recipe could be equally applied for it as well.

So maybe we should check for these bonds to get a mirror image:
`X1 {D/T} C2 {S/D} C3 {D/T} C4 {S/D} X5`

alongd (commented):
Thanks for this excellent addition, @bjkreitz !
I added some comments, some very minor and cosmetic ones and some for discussion

rmgpy/molecule/pathfinder.py
bjkreitz: The notation for (bidentate) adsorbates is a little tricky and there is no real consensus in the literature. I always place the adsorption site (the 'X') on the left side of the adatom, e.g. XC means that a carbon atom is bonded to the surface and XCXC means that the first and the second carbon atom have a bond to the surface. That said, CXCX or XCCX would be perfectly fine, too. I add some text to explain my notation.

rmgpy/molecule/pathfinder.py
```diff
@@ -480,3 +480,51 @@ def is_atom_able_to_lose_lone_pair(atom):
...
+    In this transition atom1 and atom4 are surface sites while atom2 and atom3 are carbon atoms.
+    """
+    cython.declare(paths=list, atom2=Vertex, atom3=Vertex, atom4=Vertex, bond12=Edge, bond23=Edge, bond34=Edge)
+
+    path=[]
+    if atom1.is_surface_site():
+        for atom2, bond12 in atom1.edges.items():
+            if atom2.is_carbon():
+                for atom3, bond23 in atom2.edges.items():
+                    if atom1 is not atom3 and atom3.is_carbon():
```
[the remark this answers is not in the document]
bjkreitz: I agree and removed the check

rmgpy/molecule/pathfinder.py
```diff
@@ -480,3 +480,51 @@ def is_atom_able_to_lose_lone_pair(atom):
...
+                        for atom4, bond34 in atom3.edges.items():
+                            if atom2 is not atom4 and atom4.is_surface_site():
+                                path.append([atom1, atom2, atom3, atom4, bond12, bond23, bond34])
+                                return path
```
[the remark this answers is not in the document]
bjkreitz: Thanks for the hint. I think it is fine to use return only at the end.

rmgpy/molecule/pathfinder.py
```diff
@@ -480,3 +480,51 @@ def is_atom_able_to_lose_lone_pair(atom):
...
+    Find all bidentate adsorbates which have a bonding configuration X-C-C-X.
+    Examples:
+
+    - XCXC, XCHXCH, XCXCH, where X is the surface site
+
+    In this transition atom1 and atom4 are surface sites while atom2 and atom3 are carbon atoms.
+    """
+    cython.declare(paths=list, atom2=Vertex, atom3=Vertex, atom4=Vertex, bond12=Edge, bond23=Edge, bond34=Edge)
+
+    path=[]
```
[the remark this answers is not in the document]
bjkreitz: Done

rmgpy/molecule/pathfinder.py
```diff
@@ -480,3 +480,51 @@ def is_atom_able_to_lose_lone_pair(atom):
...
+
+    cython.declare(paths=list, atom2=Vertex, atom3=Vertex, atom4=Vertex, atom5=Vertex, bond12=Edge, bond23=Edge, bond34=Edge, bond45=Edge)
+
+    path=[]
+    if atom1.is_surface_site():
+        for atom2, bond12 in atom1.edges.items():
+            for atom3, bond23 in atom2.edges.items():
+                if atom1 is not atom3 and atom3.is_carbon():
+                    for atom4, bond34 in atom3.edges.items():
+                        if atom2 is not atom4 and atom4.is_carbon():
```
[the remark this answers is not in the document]
bjkreitz: I removed the checks

rmgpy/molecule/resonance.py
bjkreitz: I think the nested notation is easier to understand, at least for me. I can also change it if preferred.

bjkreitz: Done

rmgpy/molecule/molecule.py
bjkreitz: Good catch. I renamed the function to is_multidentate since all multidentate species can have resonance.

rmgpy/molecule/resonance.py
bjkreitz: Thanks, Alon. You are right, `bond_23` could also be a double bond. I changed the if statement.

bjkreitz:
Thank you for the thorough review, Alon. I incorporated all your comments. Also, the functions `generate_adsorbate_double_shift_down_resonance_structures` and `generate_adsorbate_double_shift_up_resonance_structures` are actually redundant, so I removed them. Moreover, the unit tests failed because the `fragment.py` file was missing a check for multidentate species. After adding this check, I was able to generate the resonance structures for the examples in `fragmentTest.py` without errors. I haven't really grasped why this check is necessary for the fragments, though.

rmgpy/molecule/pathfinder.py
bjkreitz: Unfortunately, the notation for adsorbates is a little confusing since it is difficult to represent the adsorbate structure as a string of letters. This is the current notation that we and some other groups use.

rmgpy/molecule/resonance.py
bjkreitz: What do you mean? I thought `XCHCHXC: [X]#CC=C[X] <=> [X]=C=CC=[X]` was the example.

bjkreitz:
Just making a comment so that it is no longer a stale PR. I'll pick up the work on resonance again soon.

bjkreitz:
@alongd I re-requested your review for this PR. I revised and cleaned up the code based on your comments. It would be great, if you could take another look at this when you have time.

rwest (dismissed):
Looks good to me, but while rebasing and reviewing I added some more commits and made changes, so now I request @bjkreitz reviews it again :-)

I paste here edited parts of one of my commit messages, for discussion.

> If you have a Species with multiple Molecules (resonance structures)
then the list of Molecules now gets re-ordered when you estimate the thermo
so that the most stable (lowest H298) goes first.
This will then be used for drawing pictures, printing SMILES strings, etc.
which might mean it's not the one that you started with.
This may surprise some people in some situations, but I think is the
way the gas phase works, and it has some benefits, especially for
machine-generated species.

> One thing we don't yet do in adsorbates (but I think happens for the gas phase) is prioritize resonance structures coming
from thermo libraries, over those that are merely estimated.
Not sure of the consequences of this.

rwest:
> One thing we don't yet do in adsorbates (but I think happens for the gas phase) is prioritize resonance structures coming
> from thermo libraries, over those that are merely estimated.
> Not sure of the consequences of this.

Is this hypothetical situation possible?:

There is some bidentate adsorbate with two resonance structures, A and B. RMG is estimating that it adsorbs too strongly.
So the user adds it to a library, choosing to represent it at resonance structure A, because that is most stable. The new library value is higher in energy than the RMG estimates for either A or B.
When discovering this species in an RMG run, RMG generates both resonance structures, finds A in the library and estimates B. It then decides on the basis of these values that  B is more stable than A, and so uses B to represent the structure, with the estimated thermo for B, ignoring the library value for A.

This would be a problem. 


I think to avoid this the gas phase species code prioritizes resonance structures with thermo from a library in preference to those with estimated thermo, before then sorting based on H298 to pick the preferred structure. Maybe we should do the same here.  But so far this is speculation.

bjkreitz:
Thank you for the changes, Richard and for cleaning up the git history. 

Regarding your question: 
RMG prioritizes the thermo values from a library for resonance structures of an adsorbate over the estimation routine similar to the gas-phase procedure. The first call in `get_thermo_data()` is `thermo0 = self.get_thermo_data_from_libraries(species)`, which calls`get_thermo_data_from_library()`. This function searches the thermo libraries for every resonance structure in `species.molecules`. Only if it doesn't find a library entry for any of the resonance structures, it applies the estimation routine, where it searches for the estimate with the lowest enthalpy of formation.

rwest (dismissed):
Great. Thanks Bjarne. So is this ready to merge?

bjkreitz:
I haven't addressed the linear scaling of resonance structures and I think it makes sense to integrate it into this PR. Also, I should probably add some more tests for the thermo estimation routine.

rwest:
Ah - you mean if one structure is X=CH-CH=X it would scale very differently from X-CH=CH-X when we apply LSR rules to change metal? Yes, I see. Is it even clear what is the right thing to do here?

rmgpy/molecule/resonance.py
bjkreitz: I think this is resolved

rmgpy/molecule/resonance.py
```diff
@@ -1116,3 +1133,183 @@ def _clar_transformation(mol, aromatic_ring):
...
+    cython.declare(structures=list, paths=list, index=cython.int, structure=Graph)
+    cython.declare(atom=Vertex, atom1=Vertex, atom2=Vertex, atom3=Vertex, atom4=Vertex, bond12=Edge, bond23=Edge, bond34=Edge)
+    cython.declare(v1=Vertex, v2=Vertex)
+
+    structures = []
+    if mol.is_bidentate():
+        for atom in mol.vertices:
+            paths = pathfinder.find_adsorbate_delocalization_paths(atom)
+            for atom1, atom2, atom3, atom4, bond12, bond23, bond34 in paths:
+                if (bond12.is_double() and bond23.is_single() and bond34.is_double()) or (bond23.is_double() and bond12.is_double() and bond34.is_double()) or (bond12.is_triple() and bond23.is_single() and bond34.is_triple()) or (bond12.is_triple() and bond23.is_single() and bond34.is_double()):
```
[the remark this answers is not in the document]
bjkreitz: Thank you for the suggestion. I added this feature and simplified the resonance generation recipes.

bjkreitz:
I followed Alon's suggestion and added a is_double_or_triple() function to the Bond class in molecule.py to simplify the resonance structure generation function.

rmgpy/molecule/resonance.py
```diff
@@ -129,6 +145,7 @@ def analyze_molecule(mol, save_order=False):
                 'is_aryl_radical': False,
                 'hasNitrogenVal5': False,
                 'hasLonePairs': False,
+                'is_bidentate':mol.is_bidentate(),
```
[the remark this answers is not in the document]
alongd: There are some unaddressed minor suggestions of formatting (e.g., here ` 'is_bidentate': mol.is_bidentate(),` instead of ` 'is_bidentate':mol.is_bidentate(),`)
See if you agree with them to use consistent formatting

rmgpy/molecule/resonance.py
```diff
@@ -1116,3 +1133,183 @@ def _clar_transformation(mol, aromatic_ring):
...
+    (where '=' denotes a double bond, '#' denotes a triple bond)
+    """
+    cython.declare(structures=list, paths=list, index=cython.int, structure=Graph)
+    cython.declare(atom=Vertex, atom1=Vertex, atom2=Vertex, atom3=Vertex, atom4=Vertex, bond12=Edge, bond23=Edge, bond34=Edge)
+    cython.declare(v1=Vertex, v2=Vertex)
+
+    structures = []
+    if mol.is_bidentate():
+        for atom in mol.vertices:
+            # print(atom)
```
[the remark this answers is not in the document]
alongd: please delete the debug message before merging this PR

rmgpy/molecule/pathfinder.py
```diff
@@ -480,3 +480,51 @@ def is_atom_able_to_lose_lone_pair(atom):
...
+
+    In this transition atom1 and atom5 are surface sites while atom2, atom3, and atom4 are carbon atoms.
+    """
+
+    cython.declare(paths=list, atom2=Vertex, atom3=Vertex, atom4=Vertex, atom5=Vertex, bond12=Edge, bond23=Edge, bond34=Edge, bond45=Edge)
+
+    path=[]
+    if atom1.is_surface_site():
+        for atom2, bond12 in atom1.edges.items():
+            for atom3, bond23 in atom2.edges.items():
```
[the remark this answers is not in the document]
bjkreitz: This check was added in a later commit

rmgpy/molecule/pathfinder.py
```diff
@@ -480,3 +480,51 @@ def is_atom_able_to_lose_lone_pair(atom):
...
+                if atom1 is not atom3 and atom3.is_carbon():
+                    for atom4, bond34 in atom3.edges.items():
+                        if atom2 is not atom4 and atom4.is_carbon():
+                            for atom5, bond45 in atom4.edges.items():
+                                if atom3 is not atom5 and atom5.is_surface_site():
+                                    path.append([atom1, atom2, atom3, atom4, atom5, bond12, bond23, bond34, bond45])
+                                    return path
+    return path
```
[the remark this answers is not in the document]
bjkreitz: resolved

rmgpy/molecule/resonance.py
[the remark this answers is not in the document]
bjkreitz: Resolved

rmgpy/molecule/resonance.py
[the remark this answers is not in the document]
bjkreitz: This has been resolved

rmgpy/molecule/resonance.py
```diff
@@ -1116,3 +1133,183 @@ def _clar_transformation(mol, aromatic_ring):
...
+            paths = pathfinder.find_adsorbate_delocalization_paths(atom)
+            for atom1, atom2, atom3, atom4, bond12, bond23, bond34 in paths:
+                if bond23.is_single():
+                    pass
```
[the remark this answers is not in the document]
bjkreitz: Resolved in a later commit

rmgpy/molecule/resonance.py
bjkreitz: Yes it does. I removed this additional function

test/rmgpy/molecule/pathfinderTest.py
```diff
@@ -530,3 +532,41 @@ def test_hnnoo(self):
...
+1 X u0 p0 c0 {3,D}
+2 X u0 p0 c0 {4,S}
+3 C u0 p0 c0 {1,D} {4,D}
+4 C u0 p0 c0 {2,S} {3,D} {5,S}
+5 H u0 p0 c0 {4,S}  
+        """
+
+        mol = Molecule().from_adjacency_list(adjlist)
+        paths = find_adsorbate_delocalization_paths(mol.atoms[0])
+        assert paths
```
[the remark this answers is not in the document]
bjkreitz: I added additional tests

bjkreitz:
I resolved @alongd comments and cleaned up the commit history by squashing a few commits. @alongd and @rwest, please take another look

alongd (approved):
Thanks, looking good!

PR #2510, opened 2023-07-28
fix ci reusable workflow to checkout the correct rmg-py

JacksonBurns:
See https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2504#issuecomment-1656372161 and nearby comments - fixes an issue with the CI not correctly checking out RMG-Py from forks.

PR #2512, opened 2023-07-31
remove reusable workflow logic, make RMG_DATABSE_BRANCH file-scoped

JacksonBurns:
See https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2510#issuecomment-1658870085

Re-using the RMG-Py workflow removes the ability to (easily) run the tests from forks or to target forks, so we will instead copy the testing action and test files to RMG-database. The workflow will be similar but (as mentioned below) can skip some steps.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -38,13 +38,6 @@ on:
   push:
   # runs on PRs against RMG-Py (and anywhere else, but we add this for RMG-Py)
   pull_request:
-  # allow calling from other repos in the ReactionMechanismGenerator organization
-  workflow_call:
-    inputs:
-      rmg-db-branch:
-        # if calling from RMG-database, must provide a branch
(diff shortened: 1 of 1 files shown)


rwest:
What's the idea here? (perhaps you can updated the [summary](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2512#issue-1829737999) to answer)

We stop trying to make a re-usable workflow, and just have a similar-ish workflow action in the RMG-database repository? That sounds reasonable. Downside is we'd have to remember to maintain it independently when we make tweaks and improvements to this workflow. Upsides are: it's simpler to get running, and it could be tailored to just database testing (eg. skip the unit test - just do database tests and regression tests).   Am I understanding correctly?

rwest (approved):
Looks like a reasonable plan.

PR #2513, opened 2023-08-01
When loading all reaction libraries, only process "reactions.py" files.

rwest:
### Motivation or Problem
I had a reaction library folder with another python file in it (`renumber.py`)that had been used to renumber the reactions
at some point. RMG was trying to load this as if it were a `reactions.py`, and failing the database tests. 

### Description of Changes

Since named reaction libraries are forced to have the file be called "reactions.py", I think it is OK that the "load all libraries" method makes the same assumption.

So now it only searches for libraries with a file called `reactions.py` (though it is case insensitive).

### Testing
On my computer, `make test-database` now passes (on the new pytest system) whereas it didn't before this.


JacksonBurns (approved):
LGTM

PR #2515, opened 2023-08-02
Fix CI workflow to function on forked pushes.

rwest:
This is what we did in the (abandoned) PR #2510 and should have been done in PR #2512.

It means that when the CI runs on pushes to a forked repository, it checks out that forked repository, instead of the official one.

Let's check it actually does the right thing before merging this time!


JacksonBurns (approved):
Good catch

rwest:
[It did it! ](https://github.com/rwest/RMG-Py/actions/runs/5742429367)
Updated the comments and going to merge.

Issue #2518, opened 2023-08-03
Species with "R" in name gets mistaken as Fragment


hwpang:
### Bug Description
If a species has "R" in the name in species dictionary file, it gets mistaken as Fragment.

### How To Reproduce
Create a species dictionary file with the following species, run `load_species_dictionary` on the created file, and check that the type of the created `species.molecule[0]` is mistaken to be Fragment. If you remove the "R" in the name, the type becomes the correct Molecule type.

```
RAD1(14)
multiplicity 2
1  C u0 p0 c0 {2,S} {9,S} {21,S} {22,S}
2  C u0 p0 c0 {1,S} {3,S} {23,S} {24,S}
3  C u0 p0 c0 {2,S} {4,S} {25,S} {26,S}
4  C u0 p0 c0 {3,S} {5,S} {27,S} {28,S}
5  C u0 p0 c0 {4,S} {6,S} {29,S} {30,S}
6  C u0 p0 c0 {5,S} {7,S} {31,S} {32,S}
7  C u0 p0 c0 {6,S} {8,S} {33,S} {34,S}
8  C u0 p0 c0 {7,S} {10,S} {35,S} {36,S}
9  C u0 p0 c0 {1,S} {11,S} {19,S} {20,S}
10 C u0 p0 c0 {8,S} {13,S} {37,S} {38,S}
11 C u0 p0 c0 {9,S} {39,S} {40,S} {41,S}
12 C u0 p0 c0 {13,S} {14,B} {15,B}
13 C u1 p0 c0 {10,S} {12,S} {42,S}
14 C u0 p0 c0 {12,B} {16,B} {43,S}
15 C u0 p0 c0 {12,B} {18,B} {47,S}
16 C u0 p0 c0 {14,B} {17,B} {44,S}
17 C u0 p0 c0 {16,B} {18,B} {45,S}
18 C u0 p0 c0 {15,B} {17,B} {46,S}
19 H u0 p0 c0 {9,S}
20 H u0 p0 c0 {9,S}
21 H u0 p0 c0 {1,S}
22 H u0 p0 c0 {1,S}
23 H u0 p0 c0 {2,S}
24 H u0 p0 c0 {2,S}
25 H u0 p0 c0 {3,S}
26 H u0 p0 c0 {3,S}
27 H u0 p0 c0 {4,S}
28 H u0 p0 c0 {4,S}
29 H u0 p0 c0 {5,S}
[...]

PR #2519, opened 2023-08-03
Remove label in adjlist before detecting cutting label

hwpang:
### Motivation or Problem
See https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2518.

### Description of Changes
I remove the label before detecting cutting label.

### Testing
I test it on the example in https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2518, and it works.

### DIFF
--- a/rmgpy/species.py
+++ b/rmgpy/species.py
@@ -365,8 +365,17 @@ class Species(object):
         list in the `molecule` attribute. Does not generate resonance isomers
         of the loaded molecule.
         """
+        lines = adjlist.splitlines()
+
+        if len(lines[0].split()) == 1:
+            label = lines.pop(0) # remove the first line if it is a label before detecting cutting label
+            adjlist_no_label = '\n'.join(lines)
+        else:
+            adjlist_no_label = adjlist
+
         # detect if it contains cutting label
-        _ , cutting_label_list = Fragment().detect_cutting_label(adjlist)
+        _ , cutting_label_list = Fragment().detect_cutting_label(adjlist_no_label)
+        
         if cutting_label_list == []:
             self.molecule = [Molecule().from_adjacency_list(adjlist, saturate_h=False,
                                                             raise_atomtype_exception=raise_atomtype_exception,


lily90502 (approved):
This looks good to me.

rmgpy/species.py
```diff
@@ -365,8 +365,16 @@ def from_adjacency_list(self, adjlist, raise_atomtype_exception=True, raise_char
...
+        lines = adjlist.splitlines()
+
+        if len(lines[0].split()) == 1:
+            label = lines.pop(0) # remove the first line if it is a label before detecting cutting label
+
+        adjlist_no_label = '\n'.join(lines)
+
         # detect if it contains cutting label
-        _ , cutting_label_list = Fragment().detect_cutting_label(adjlist)
+        _ , cutting_label_list = Fragment().detect_cutting_label(adjlist_no_label)
```
rwest, line 376: `detect_cutting_label` is designed to work on SMILES-like strings, and looks quite complex.
what does a cutting label in an adjacency list look like? there might be a much faster way to just see if it's present in the adjacency list (which seems to be all you need to know here?)

rmgpy/species.py
```diff
@@ -365,8 +365,16 @@ def from_adjacency_list(self, adjlist, raise_atomtype_exception=True, raise_char
         list in the `molecule` attribute. Does not generate resonance isomers
         of the loaded molecule.
         """
+        lines = adjlist.splitlines()
+
+        if len(lines[0].split()) == 1:
+            label = lines.pop(0) # remove the first line if it is a label before detecting cutting label
+
+        adjlist_no_label = '\n'.join(lines)
```
rwest, line 373: if the first line *isn't* a label, then you don't need to do the whole `'\n'.join(lines)` operation. That could be inside the `if` block, and an `else` block is just `adjlist_no_label = adjlist`

rmgpy/species.py
lily90502: An example of fragment `RCC` adjacency list:
`1 C u0 p0 c0 {2,S} {3,S} {4,S} {5,S}
2 C u0 p0 c0 {1,S} {6,S} {7,S} {8,S}
3 R u0 p0 c0 {1,S}
4 H u0 p0 c0 {1,S}
5 H u0 p0 c0 {1,S}
6 H u0 p0 c0 {2,S}
7 H u0 p0 c0 {2,S}
8 H u0 p0 c0 {2,S}`

I think another way is to check the atom type to see if it is `Atom` or `Cuttinglabel`. But I guess this will be slower?

hwpang: Yeah, there are definitely better ways to do this. I'm not familiar with the `detect_cutting_label` function, so this PR is a quick fix, but someone should work on improving `detect_cutting_label` to specialize in adjacency list.

hwpang:
I rebased the PR

lily90502 (approved):
The fix looks ok to me

hwpang:
Most tests passed. The regression tests failed due to reasons unrelated to shi PR. With @lily90502's approval, I will merge this.

PR #2538, opened 2023-09-11
Update RMSE/MAE average formula for BAC cross validation

hwpang:
### Motivation or Problem
When using cross validation, the average RMSE and average MAE are currently calculated as RMSE_avg = sum^{n_fold}_i(RMSE_i)/n_fold and MAE_avg = sum^{n_fold}_i(MAE_i)/n_fold. 

However, during leave-one-out cross validation, this method returns the same number for RMSE_avg and MAE_avg. This is because when there is only one testing data, RMSE_i = MAE_i. 

Instead, I think this average formula should be used, where 
RMSE_avg = sqrt( sum^{n_fold}_i (RMSE_i^2 * Ndata_i) / sum^{n_fold}_i (Ndata_i) )
MAE_avg = sum^{n_fold}_i (MAE_i * Ndata_i) / sum^{n_fold}_i (Ndata_i)

### Description of Changes
I calculate the RMSE_avg and MAE_avg for leave-one-out cross validation case.

### DIFF
--- a/arkane/encorr/bac.py
+++ b/arkane/encorr/bac.py
@@ -1047,20 +1047,21 @@ class CrossVal:
             logging.info(f'RMSE/MAE before fitting: {stats_before.rmse:.2f}/{stats_before.mae:.2f} kcal/mol')
             logging.info(f'RMSE/MAE after fitting: {stats_after.rmse:.2f}/{stats_after.mae:.2f} kcal/mol')
 
-        rmse_before = [test_data.calculate_stats().rmse for test_data in test_data_results]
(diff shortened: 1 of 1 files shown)


hwpang:
I had an offline discussion with @oscarwumit and he thinks it's better to keep it consistent, so I change the RMSE_avg and MAE_avg for not only leave-one-out cross validation but also other leave-one-out cross validation cases.

oscarwumit (approved):
Looks good. Thanks for identifying and fixing the bug.

PR #2539, opened 2023-09-12
Build a `rmgmolecule` Subpackage in Place, fix `rmg` conda recipe, initial `SimulatorInterface` Efforts

JacksonBurns:
# Summary
This PR is attempting to do 3 things:
 1. Build a subpackage from RMG-Py called `rmgmolecule`, which contains just the code contained in `rmgpy/molecule`, so that other softwares can use it without having to install all of RMG-Py.
 2. Fix the RMG conda recipe, since doing 1. required setting up custom actions runners anyway and it makes organizational sense to pair these efforts.
 3. Separate the pure-python and mixed-python components of RMG (namely RMG and RMS) via a `SimulatorInterface`, which will allow users to install RMG without installing RMS and its additional requirements.
    - this became necessary to build the conda package (2.), since Julia/PyJulia/RMS could not work
    - see ReactionMechanismGenerator/RMG-Py#2547 for an example of why having RMS included by default is expensive
    - this opens the door to eventually writing a `SimulatorInterface` class which will allow users to 'plug-in' whatever simulators they want, but for the time being the solution is going to look more like "if you want RMS, follow these additional installation instructions after installing RMG" (which are going to be the dev installation instructions for RMS as given in RMG)

This was originally small in scope but these sub-goals have required one another to proceed.

## TODO
 - [x] clean up the git history, which currently contains a number of experimental commits needed to test the self-hosted runner and CI
 - [x] actually fix the conda build for RMG, which is currently broken because the testing environment resolves Julia, which is broken
 - [ ] check if `rmgmolecule` can be built on windows, and or more version of python/numpy/etc., since it now has lighter dependencies
 - [x] also lighten the dependencies for rmgmolecule
 - [ ] change `reactors.py` to `julia_reactors.py`, bury the import statement and add a runtime warning when attempting to import it that says RMS must be installed, and then update the docs to explain how to install RMS
 - [ ] update the CI to perform the default RMG install and then the RMS installation steps to be described in the above point, and then continue executing tests as we are right now (covering RMS anyway)


JacksonBurns:
TODO: consolidate the two `setup.py` scripts

JacksonBurns:
Also TODO: consolidate the package build action into the CI workflow (for the auto-canceller)

JacksonBurns:
Also switch the builder to use libmamba solver for speed up, may also fix the memory problem

.rmgmolecule_conda/conda_build_config.yaml
rwest, line 1: end file with a line ending?
(and [when] can we try adding more versions of numpy?)

rmgpy/molecule/draw.py
```diff
@@ -49,6 +49,7 @@
 import math
 import os.path
 import re
+import warnings
```
rwest, line 52: does this do anything? I can't see what

rwest (commented):
This looks promising.

When would we publish new conda binaries of the molecule package?
I suggest we try building it on every push to detect errors, but presumably only want to publish a new "release" whenever something significant changes that RMS/pynta/etc. or another package is depending on?  Would this be done by just incrementing a minor version number in a yml file or something?  We might wish to do it more often than a full RMG release.

JacksonBurns:
TODO: try and remove some of the deps

JacksonBurns:
also TODO: test and see if, with deps removed, the build can work on windows

.rmgmolecule_conda/conda_build_config.yaml
JacksonBurns: This file has since moved to a private repo so that it can call self-hosted runners

As for the multiple versions, I'll put that into a matrix build and try.

We can look over this in greater detail offline.

rmgpy/molecule/draw.py
JacksonBurns: I had intended for [this line](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2539/files/5ad2f2484df06d5910f3c09ad94ccf3ccdee9c3d#diff-8571db88680406d006088de6253f48278b66c7b3f6fd15f711878f822333eed3R70) to use `warnings.warn` but since I changed it I will remove this.

JacksonBurns:
Building RMG binary with Julia included is proving to be annoying. Currently attempting two fixes: (1) relax version requirements on RMS to try and get the environment to solve and precompile and (2) pin packages that are compatible. 

Alternative to these alternatives is to ship the binary with only the basic RMG python functionality, and then add installation steps for installing RMS from the command line (this might be a really good idea). Would need to change tests to use Python instead of python-jl.

JacksonBurns:
Julia is attempting to install the `rmg` conda package... while building the `rmg` conda package. This is disastrous for obvious reasons.

JacksonBurns:
I am removing (or rather not adding, because it was not added when the RMS changes were implemented) RMS to the conda recipe for RMG because integrating Julia is a complete and total nightmare. An incomplete list of issues it causes:
 - if anything is missing from conda environment, Julia will attempt to install its own copy in its own local environment with its own conda
 - Julia's conda package states that the intermediate build environment used by the real conda to build the package is not a valid conda environment, and crashes the run
 - 'Precompilation' fails for unknown reasons on random dependencies
 - Julia is incredibly picky about environment variables pointing to the proper Python binary and path for RMG-Py, both of which during the installation process because conda re-installs the built binary as part of the testing procedure (breaking those paths)
 - The mere _existence_ of PyCall

Assuming I can get just the plain build to work, I am going to update the user installation instructions with how to install RMS after installing the RMG conda package, which should work. This, as far as I see it, is the first step in establishing more clear separation between RMG and the simulators that it uses. Users install a "just rmg" binary, and then optionally some others. Trying to ship it all at once is complete murder, an unholy union that god himself would frown upon.

JacksonBurns:
bd8f7c3 is fine. Commits 7f3ba46 through eb7dd50 (and later) can be squashed, since they are addressing points 1. and 2. of this PR.

JacksonBurns:
Have to use `conda-build==3.24.0` for some reason, see https://github.com/conda/conda-build/issues/4993#issuecomment-1735589444

JacksonBurns:
`libmamba` must be set as the solver inside the solving environment to force `conda build` to actually solve.

JacksonBurns:
squashed debugging commits and rebased onto main

JacksonBurns:
This PR should fix the linked issue when cleaning the environment file.

JacksonBurns:
Update docs to clarify that Arkane-only users only need to install the base rmg package

rmgpy/rmg/model.py
```diff
@@ -1382,35 +1379,36 @@ def prune(self, reaction_systems, tol_keep_in_edge, tol_move_to_core, maximum_ed
...
         self.edge.species.remove(spec)
         self.index_species_dict.pop(spec.index)
-        self.edge.phase_system.remove_species(spec)
+        if not NO_JULIA and requires_rms:
+            self.edge.phase_system.remove_species(spec)
 
         # clean up species references in reaction_systems
         for reaction_system in reaction_systems:
-            if not isinstance(reaction_system, Reactor):
+            if NO_JULIA or not requires_rms or not isinstance(reaction_system, Reactor):
```
hwpang, line 1411: Should be RMSReactor?

rmgpy/rmg/model.py
```diff
@@ -1163,29 +1165,24 @@ def add_species_to_core(self, spec):
             for rxn in rxn_list:
                 self.add_reaction_to_core(rxn)
```
hwpang, line 1166: Need to add require_rms=require_rms

JacksonBurns:
- RMS is now using the `ReactionMechanismGenerator/molecule` package, which is larger in scope than `rmgmolecule`. Without more specific demand for it, it probably isn't worth keeping.

JacksonBurns:
#2564 would be resolved by this PR. The error over there has to do with attempting to pass a solvent to RMS with incomplete data that fails, even though we aren't using an RMS reactor and really have no reason to do the passing.

JacksonBurns:
WIP

JacksonBurns:
Splitting this into three PRs:
 1. The `NO_JULIA` internal tracking variable
 2. Fix the RMG conda recipe
 3. Add the subpackage for molecule (still might drop this one)

Issue #2536, opened 2023-08-31
SMILES strings are wrong for some gas-phase species


bjkreitz:
### Bug Description
There seems to be an issue with the conversion of adjacency lists to SMILES strings. Some SMILES strings do not represent the correct adjacency list. 

### How To Reproduce
Generate the SMILES string from the following adjacency list: 

```
multiplicity 3
1 *3 C u1 p1 c0 {2,S}
2 *3 C u1 p1 c0 {1,S}
```

using the RMG website Molecule Search tool, results in [CH2][CH2]. Feeding this SMILES string back to RMG yields the following adjacency list:

```
multiplicity 3
1 C u1 p0 c0 {2,S} {3,S} {4,S}
2 C u1 p0 c0 {1,S} {5,S} {6,S}
3 H u0 p0 c0 {1,S}
4 H u0 p0 c0 {1,S}
5 H u0 p0 c0 {2,S}
6 H u0 p0 c0 {2,S}
```

The correct smiles string would be [C][C]. This issue occurred for several gas-phase species, which have many free electrons.  

### Expected Behavior
RMG should return the correct SMILES strings.

JacksonBurns:
@bjkreitz could you include some example code to demonstrate how you are feeding this string back to RMG?

bjkreitz:
You can test this using the RMG website. I'm not feeding the SMILES string back to RMG. I was working on an unrelated PR, where I was looking at `get_desorbed_molecules` function in `molecule.py`. The function printed out completely wrong SMILES strings, although the adjacency list are correct, which made me think that the function was wrong. Using the website tool reproduced the error.

JacksonBurns:
Ok, when I run on RMG-website the SMILES string `[C][C]` is converted into this:
```
multiplicity 7
1 C u3 p0 c0 {2,S}
2 C u3 p0 c0 {1,S}
```
and `[CH2][CH2]` is converted into this:
```
multiplicity 3
1 C u1 p0 c0 {2,S} {3,S} {4,S}
2 C u1 p0 c0 {1,S} {5,S} {6,S}
3 H u0 p0 c0 {1,S}
4 H u0 p0 c0 {1,S}
5 H u0 p0 c0 {2,S}
6 H u0 p0 c0 {2,S}
```

Now when running in RMG-Py from the docker container, I get this:
```
(rmg_env) root@8e8b9955bf56:/rmg/RMG-Py# python-jl
[...]

PR #2545, opened 2023-09-19
Fix CC smiles

xiaoruiDong:
### Motivation or Problem
This PR tries to fix issue #2536, where the SMILES of bicarbon (CC) is not correctly generated.

### Description of Changes
This PR forces the setting of no implicit Hs when converting the RMG Molecule to an RDKit Mol object. This is viable since all hydrogens in the RMG Molecules are defined explicitly.

### Testing
Add a few unit tests to test the molecule of CC with various multiplicities. I comment out the case with a quadruple bond, as it is very likely the installed RDKit version doesn't have support for output smiles of quadruple bond (which is introduced in sometime 2022)

<del> There is an InChI error raised due to this change, and I am investigating it.
https://github.com/xiaoruiDong/RMG-Py/actions/runs/6238513699/job/16934497099

### DIFF
--- a/rmgpy/molecule/converter.py
+++ b/rmgpy/molecule/converter.py
@@ -116,6 +116,8 @@ def to_rdkit_mol(mol, remove_h=True, return_mapping=False, sanitize=True, save_o
         for label, ind_list in label_dict.items():
             for ind in ind_list:
                 Chem.SetSupplementalSmilesLabel(rdkitmol.GetAtomWithIdx(ind), label)
+    [atom.SetNoImplicit(True) for atom in rdkitmol.GetAtoms()
+     if atom.GetAtomicNum() > 1]
     if sanitize:
         Chem.SanitizeMol(rdkitmol)
     if remove_h:
--- a/rmgpy/molecule/converter.py
+++ b/rmgpy/molecule/converter.py
@@ -116,8 +116,9 @@ def to_rdkit_mol(mol, remove_h=True, return_mapping=False, sanitize=True, save_o
         for label, ind_list in label_dict.items():
             for ind in ind_list:
                 Chem.SetSupplementalSmilesLabel(rdkitmol.GetAtomWithIdx(ind), label)
-    [atom.SetNoImplicit(True) for atom in rdkitmol.GetAtoms()
-     if atom.GetAtomicNum() > 1]
+    for atom in rdkitmol.GetAtoms():
+        if atom.GetAtomicNum() > 1:
+            atom.SetNoImplicit(True)
     if sanitize:
         Chem.SanitizeMol(rdkitmol)
     if remove_h:
--- a/rmgpy/molecule/inchi.py
+++ b/rmgpy/molecule/inchi.py
@@ -602,12 +602,12 @@ def create_augmented_layers(mol):
     else:
         molcopy = mol.copy(deep=True)
 
+        rdkitmol = to_rdkit_mol(molcopy, remove_h=True)
+        _, auxinfo = Chem.MolToInchiAndAuxInfo(rdkitmol, options='-SNon')  # suppress stereo warnings
+
         hydrogens = [at for at in molcopy.atoms if at.number == 1]
         for h in hydrogens:
             molcopy.remove_atom(h)
-
-        rdkitmol = to_rdkit_mol(molcopy)
-        _, auxinfo = Chem.MolToInchiAndAuxInfo(rdkitmol, options='-SNon')  # suppress stereo warnings
 
         # extract the atom numbers from N-layer of auxiliary info:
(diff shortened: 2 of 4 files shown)


xiaoruiDong:
This seems to be an RMG inchi parser problem. The following results are from the current `main`:

```
Molecule().from_augmented_inchi('InChI=1S/C5H6O/c6-5-3-1-2-4-5/h1-3,5H,4H2/u1,6')
```
Got InvalidAdjacencyListError

```
Molecule().from_augmented_inchi('InChI=1S/C5H6O/c6-5-3-1-2-4-5/h1-3,5H,4H2/u2,6').to_augmented_inchi()
>>> 'InChI=1S/C5H6O/c6-5-3-1-2-4-5/h1-3,5H,4H2/u1,6'
```

```
Molecule().from_augmented_inchi('InChI=1S/C5H6O/c6-5-3-1-2-4-5/h1-3,5H,4H2/u2,6').to_smiles()
>>> '[O]C1C=C[CH]C1'
Molecule().from_smiles(smi).to_augmented_inchi()
>>> 'InChI=1S/C5H6O/c6-5-3-1-2-4-5/h1-3,5H,4H2/u1,6'
[mol.to_augmented_inchi() for mol in Molecule().from_smiles(smi).generate_resonance_structures()]
>>> ['InChI=1S/C5H6O/c6-5-3-1-2-4-5/h1-3,5H,4H2/u1,6',
     'InChI=1S/C5H6O/c6-5-3-1-2-4-5/h1-3,5H,4H2/u3,6']

**Review by JacksonBurns — APPROVED** [review]
Thanks for the quick patch @xiaoruiDong! This looks thorough and quite reasonable to me, and the regression tests indicate that this isn't impacting anything. I think we should merge, but let's get additional comments from @hwpang as well.

File: rmgpy/molecule/inchi.py [hunk 1]
```diff
@@ -602,13 +602,13 @@ def create_augmented_layers(mol):
     else:
         molcopy = mol.copy(deep=True)
 
+        rdkitmol = to_rdkit_mol(molcopy, remove_h=True)
+        _, auxinfo = Chem.MolToInchiAndAuxInfo(rdkitmol, options='-SNon')  # suppress stereo warnings
+
```
rwest, line 607: Is Codecov correct that lines #L605 - L606 were not covered by tests ? Does that mean this whole method is not tested? (or is codecov not right?)

rwest (commented):
Made a couple comments. In general looks like a good fix. Thanks
I'm not sure how to interpret the Codecov annotations - I assume it's being tested?

bjkreitz:
Thanks, this seems to work properly. I tested it for the different CC structures and also some other species with unpaired electrons at the carbon atom such as CCH3. RMG returned the correct SMILES string.

bjkreitz (approved):

rmgpy/molecule/converter.py
```diff
@@ -116,6 +116,8 @@ def to_rdkit_mol(mol, remove_h=True, return_mapping=False, sanitize=True, save_o
         for label, ind_list in label_dict.items():
             for ind in ind_list:
                 Chem.SetSupplementalSmilesLabel(rdkitmol.GetAtomWithIdx(ind), label)
+    [atom.SetNoImplicit(True) for atom in rdkitmol.GetAtoms()
```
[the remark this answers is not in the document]
xiaoruiDong: Thanks for the suggestion! With some Google searches, the use of list comprehension here is less Pythonic, as my purpose is not to create a list. I've changed this as suggested.

rmgpy/molecule/inchi.py
xiaoruiDong: This is something weird. The failures in the unit tests testing this function inspired my change here. I looked at the [codecov coverage profile](https://app.codecov.io/gh/ReactionMechanismGenerator/RMG-Py/pull/2545/blob/rmgpy/molecule/inchi.py) and found it reported 0% coverage of the whole inchi.py, but RMG actually has [unit tests](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/main/test/rmgpy/molecule/inchiTest.py) for it. @JacksonBurns any thoughts?

JacksonBurns: So this test is definitely running (you can search for it in the CI logs) and the new code (and this whole file) are being run, so CodeCov is just wrong. From brief googling I can't find a solution, so I will debug this separately, but this PR should be fine.

JacksonBurns (approved):
The new unit tests cover this edge case well, and the solution is solid. CodeCov is incorrectly reporting the result, which is a problem for another time. This PR LGTM

bjkreitz (approved):

PR #2550, opened 2023-09-26
Allow user to plot correlation matrix for AECs

hwpang:
### Motivation or Problem
This is one of the changes for the BAC paper. Per Bill's request, we add a discussion regarding the correlation for both AEC and p-BAC.

### Description of Changes
I add a plotting function and allow the user to plot and save correlation matrix for AECs.

### DIFF
--- a/arkane/encorr/ae.py
+++ b/arkane/encorr/ae.py
@@ -32,6 +32,7 @@ This module provides classes for fitting atom energies based on a very
 small, predetermined set of molecules.
 """
 
+import os
 import importlib
 import json
 import logging
@@ -47,6 +48,7 @@ from rmgpy.molecule import get_element, Molecule
 import arkane.encorr.data as data
 from arkane.encorr.reference import ReferenceDatabase
 from arkane.modelchem import LevelOfTheory, CompositeLevelOfTheory
+from arkane.encorr.bac import _covariance_to_correlation
 
 # List of species labels that will be used for fitting (labels should match reference database)
 SPECIES_LABELS = [
@@ -103,12 +105,13 @@ class AEJob:
         self.overwrite = overwrite
         self.ae = AE(species_energies)
 
-    def execute(self, output_file: str = None):
+    def execute(self, output_directory: str = None, plot: bool = False):
         """
         Execute the atom energy job.
 
         Args:
-            output_file: Write the fitted energies to this file.
+            output_directory: Write the fitted energies to this directory.
+            plot: Save plots of results.
         """
         if self.level_of_theory is None:
             logging.info('Fitting atom energies')
@@ -116,7 +119,10 @@ class AEJob:
             logging.info(f'Fitting atom energies for {self.level_of_theory}')
         self.ae.fit()
 
-        if output_file is not None:
+        if output_directory is not None:
+            model_chemistry_formatted = self.level_of_theory.to_model_chem().replace('//', '__').replace('/', '_')
+            output_file = os.path.join(output_directory, f'AEC_{model_chemistry_formatted}.out')
+
             with open(output_file, 'a') as f:
                 if self.level_of_theory is not None:
                     f.write(f'#  {self.level_of_theory}\n')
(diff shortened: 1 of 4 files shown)


arkane/encorr/ae.py
```diff
@@ -134,6 +144,49 @@ def execute(self, output_file: str = None):
...
+                c = self.ae.correlation[j, i]
+                ax.text(i, j, f'{c: .2f}', va='center', ha='center', fontsize=8)
+
+        # Save lims because they get changed when modifying labels
+        xlim = ax.get_xlim()
+        ylim = ax.get_ylim()
+
+        ax.set_xticks(list(range(len(self.ae.correlation))))
+        ax.set_yticks(list(range(len(self.ae.correlation))))
+        ax.set_xticklabels(labels, fontsize=14)
```
JacksonBurns, line 181: make the font sizes passable as a dictionary, or don't set them (since they will be hard to make generally look correct)

arkane/encorr/ae.py
```diff
@@ -125,6 +131,10 @@ def execute(self, output_file: str = None):
                 f.writelines(self.ae.format_atom_energies(
                     'atom_energies' if self.level_of_theory is None else self.level_of_theory))
 
+            if plot:
+                correlation_path = os.path.join(output_directory, f'AEC_{model_chemistry_formatted}_correlation.pdf')
```
JacksonBurns, line 135: add a unique timestamp to the end so you don't overwrite previous files by accident

arkane/encorr/ae.py
```diff
@@ -134,6 +144,49 @@ def execute(self, output_file: str = None):
...
+        Save a visual representation of the parameter correlation matrix.
+
+        Args:
+            path: Path to save figure to.
+            labels: Parameter labels.
+        """
+        try:
+            import matplotlib.pyplot as plt
+        except ImportError:
+            return
```
JacksonBurns, line 158: please add `logging.error('cannot import matplotlib, original error'+str(e)`

JacksonBurns: And the same in BACs

arkane/encorr/ae.py
```diff
@@ -134,6 +144,49 @@ def execute(self, output_file: str = None):
...
+    def save_correlation_mat(self, path: str, labels: List[str] = None):
+        """
+        Save a visual representation of the parameter correlation matrix.
+
+        Args:
+            path: Path to save figure to.
+            labels: Parameter labels.
+        """
+        try:
+            import matplotlib.pyplot as plt
```
hwpang, line 156: Unbury this in both AEC and BAC

arkane/encorr/ae.py
hwpang: I don't think this is necessary though because we put the level of theory name in the save path (`model_chemistry_formatted` ). Say if you fit for the same level of theory and plot it for multiple times because something wasn't quite right previously, I think as a user you would only care about the last one.

arkane/encorr/ae.py
hwpang: I look into this part of code again, and my new thought is that plotting functions are not necessary for the Arkane job execution to fail, so it makes sense to do this

hwpang:
> @hwpang can you add some documentation describing how to use this?

Thanks for the comment! Arkane's default behavior is to use all the plots, so this function is already being used in the default settings, unless the user passes in a flag indicating it to skip the plotting step (see [documentation](https://reactionmechanismgenerator.github.io/RMG-Py/users/arkane/running.html#suppressing-plot-creation)). What kind of documentation do you think these need? Happy to add a documentation but not sure what to add.

arkane/encorr/ae.py
hwpang: Added!

arkane/encorr/ae.py
hwpang: Added!

hwpang:
@JacksonBurns The new push rebased and addressed most of the comments! I also have some thoughts/questions for some of your comments (please see them above).

arkane/encorr/ae.py
JacksonBurns: RE in-person discussion: make the import look like this: https://github.com/ReactionMechanismGenerator/RMG-Py/blob/main/rmgpy/ml/estimator.py#L121

arkane/encorr/ae.py
JacksonBurns: RE in-person discussion: update arkane docs to state that previous files will be overwritten on subsequent runs

JacksonBurns (changes requested):
additional comments from in-person meeting

JacksonBurns (approved):
LGTM! Thanks for addressing the requested edits.

PR #2523, opened 2023-08-09
Instruct Docker, Github Actions, and users to use 'for_rmg' branch of RMS

rwest:
### Motivation or Problem
The "main" branch of ReactionMechanismSimulator.jl is temporarily incompatible with RMG, causing all builds to fail.

### Description of Changes
For now, there is a dedicated "for_rmg" branch, which we can use, which is currently where main was just before the most recent breaking PR was merged.  Hopefully things can be resolved and a main branch will work for both.

### Testing
I plan to let the CI run its course.

### Other context

The failure looks like this in the CI logs:

```
Wed, 09 Aug 2023 15:50:00 GMT   ....installing Julia stuff...
Wed, 09 Aug 2023 16:20:00 GMT
└  
Wed, 09 Aug 2023 16:21:43 GMT
[ Info: Installing molecule.molecule via the Conda rmgmolecule package...
Wed, 09 Aug 2023 16:21:44 GMT
[ Info: Running `conda config --add channels hwpang --file /usr/share/miniconda3/envs/rmg_env/condarc-julia.yml --force` in root environment
Wed, 09 Aug 2023 16:21:44 GMT
[ Info: Running `conda install -q -y rmgmolecule` in root environment
Wed, 09 Aug 2023 16:22:13 GMT
Collecting package metadata (current_repodata.json): ...working... done
Wed, 09 Aug 2023 16:26:58 GMT
Solving environment: ...working... failed with initial frozen solve. Retrying with flexible solve.
Wed, 09 Aug 2023 16:32:06 GMT
Solving environment: ...working... failed with repodata from current_repodata.json, will retry with next repodata source.
Wed, 09 Aug 2023 16:33:57 GMT
Collecting package metadata (repodata.json): ...working... done
Wed, 09 Aug 2023 17:11:31 GMT
Solving environment: ...working... failed with initial frozen solve. Retrying with flexible solve.
Wed, 09 Aug 2023 17:50:35 GMT
Error: The action has timed out.
```

@mjohnson541 writes in Slack " One of the major improvements was switching RMS’s dependency from rmg to molecule…which is great for RMS, but it is probably causing RMG to automatically install molecule when RMS tries to import it…the simplest solution is probably to try importing rmgpy first and import from there if it finds it…I’m trying to think if that will cause confusion…sorry I’m at a conference right now…"

PR #2552, opened 2023-10-01
Switch back to RMS main in CI

mjohnson541:


JacksonBurns (changes requested):
Thanks for getting the patch working! Could you also change the docs and Dockerfile, i.e undo the changes in https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2523

mjohnson541:
Good call, I've swapped it out in those files as well now.

JacksonBurns (approved):
Thanks for the updates (here and on RMS) - LGTM

PR #2553, opened 2023-10-03
`environment.yml` Unblocking Python 3.11 - rdkit to conda-forge channel, remove chemprop for now

### LINKED ISSUE #2462: RDKit should be upgraded, potentially requiring API changes
We currently use an RMG channel build of RDKit, which we should replace with one of the official builds from conda-forge.

Unfortunately this transition will require first getting everything _else_ to Python 3.11 so that we can then resolve an updated version of RDKit, and then we can see where the API breaks.

JacksonBurns:
Resolves #2462

### DIFF
--- a/rmgpy/molecule/molecule.py
+++ b/rmgpy/molecule/molecule.py
@@ -1768,9 +1768,13 @@ class Molecule(Graph):
         os.unlink(temp_file_name)
         return png
 
-    def from_inchi(self, inchistr, backend='try-all', raise_atomtype_exception=True):
+    def from_inchi(self, inchistr, backend='openbabel-first', raise_atomtype_exception=True):
         """
         Convert an InChI string `inchistr` to a molecular structure.
+
+        RDKit and Open Babel are the two backends used in RMG. It is possible to use a
+        single backend or try different backends in sequence. The available options for the ``backend``
+        argument: 'openbabel-first'(default), 'rdkit-first', 'rdkit', or 'openbabel'.
         """
         translator.from_inchi(self, inchistr, backend, raise_atomtype_exception=raise_atomtype_exception)
         return self
@@ -1782,9 +1786,13 @@ class Molecule(Graph):
         translator.from_augmented_inchi(self, aug_inchi, raise_atomtype_exception=raise_atomtype_exception)
         return self
 
-    def from_smiles(self, smilesstr, backend='try-all', raise_atomtype_exception=True):
+    def from_smiles(self, smilesstr, backend='openbabel-first', raise_atomtype_exception=True):
         """
         Convert a SMILES string `smilesstr` to a molecular structure.
+
+        RDKit and Open Babel are the two backends used in RMG. It is possible to use a
+        single backend or try different backends in sequence. The available options for the ``backend``
+        argument: 'openbabel-first'(default), 'rdkit-first', 'rdkit', or 'openbabel'.
         """
         translator.from_smiles(self, smilesstr, backend, raise_atomtype_exception=raise_atomtype_exception)
         return self
@@ -1874,8 +1882,9 @@ class Molecule(Graph):
         Convert a molecular structure to an InChI string. Uses
         `OpenBabel <http://openbabel.org/>`_ to perform the conversion.
 
-        Available options for InChI backend: 'rdkit-first' (default),
-        'try-all', 'rdkit', or 'openbabel'.
+        It is possible to use a single backend or try different backends in sequence.
+        The available options for the ``backend`` argument: 'rdkit-first'(default),
+        'openbabel-first', 'rdkit', or 'openbabel'.
         """
         try:
             return translator.to_inchi(self, backend=backend)
@@ -1890,8 +1899,9 @@ class Molecule(Graph):
 
         Separate layer with a forward slash character.
 
-        Available options for InChI backend: 'rdkit-first' (default),
-        'try-all', 'rdkit', or 'openbabel'.
+        RDKit and Open Babel are the two backends used in RMG. It is possible to use a
+        single backend or try different backends in sequence. The available options for the ``backend``
+        argument: 'rdkit-first'(default), 'openbabel-first', 'rdkit', or 'openbabel'.
         """
         try:
             return translator.to_inchi(self, backend=backend, aug_level=2)
@@ -1909,8 +1919,9 @@ class Molecule(Graph):
         Convert a molecular structure to an InChI Key string. Uses
         `RDKit <http://rdkit.org/>`_ to perform the conversion.
 
-        Available options for InChI backend: 'rdkit-first' (default),
-        'try-all', 'rdkit', or 'openbabel'.
+        It is possible to use a single backend or try different backends in sequence.
+        The available options for the ``backend`` argument: 'rdkit-first'(default),
+        'openbabel-first', 'rdkit', or 'openbabel'.
         """
         try:
             return translator.to_inchi_key(self, backend=backend)
@@ -1926,8 +1937,9 @@ class Molecule(Graph):
         Simply append the multiplicity string, do not separate by a
         character like forward slash.
 
-        Available options for InChI backend: 'rdkit-first' (default),
-        'try-all', 'rdkit', or 'openbabel'.
+        RDKit and Open Babel are the two backends used in RMG. It is possible to use a
+        single backend or try different backends in sequence. The available options for the ``backend``
+        argument: 'rdkit-first'(default), 'openbabel-first', 'rdkit', or 'openbabel'.
         """
         try:
             return translator.to_inchi_key(self, backend=backend, aug_level=2)
--- a/arkane/common.py
+++ b/arkane/common.py
@@ -191,11 +191,11 @@ class ArkaneSpecies(RMGObject):
             self.multiplicity = species.molecule[0].multiplicity
             self.formula = species.molecule[0].get_formula()
             try:
-                inchi = to_inchi(species.molecule[0], backend='try-all', aug_level=0)
+                inchi = to_inchi(species.molecule[0], backend='openbabel-first', aug_level=0)
             except ValueError:
                 inchi = ''
             try:
-                inchi_key = to_inchi_key(species.molecule[0], backend='try-all', aug_level=0)
+                inchi_key = to_inchi_key(species.molecule[0], backend='openbabel-first', aug_level=0)
             except ValueError:
                 inchi_key = ''
             self.inchi = inchi
--- a/rmgpy/molecule/translator.py
+++ b/rmgpy/molecule/translator.py
@@ -169,7 +169,7 @@ def to_inchi(mol, backend='rdkit-first', aug_level=0):
     Uses RDKit or OpenBabel for conversion.
 
     Args:
-        backend     choice of backend, 'rdkit-first' (default), 'try-all', 'rdkit', or 'openbabel'
+        backend     choice of backend, 'rdkit-first' (default), 'openbabel-first', 'rdkit', or 'openbabel'
         aug_level   level of augmentation, 0, 1, or 2
     """
     cython.declare(inchi=str, ulayer=str, player=str, mlayer=str)
@@ -205,7 +205,7 @@ def to_inchi_key(mol, backend='rdkit-first', aug_level=0):
     Uses RDKit or OpenBabel for conversion.
 
     Args:
-        backend     choice of backend, 'rdkit-first' (default), 'try-all', 'rdkit', or 'openbabel'
+        backend     choice of backend, 'rdkit-first' (default), 'openbabel-first', 'rdkit', or 'openbabel'
         aug_level   level of augmentation, 0, 1, or 2
     """
     cython.declare(key=str, ulayer=str, player=str, mlayer=str)
@@ -274,10 +274,10 @@ def to_smiles(mol, backend='default'):
         return output
 
 
-def from_inchi(mol, inchistr, backend='try-all', raise_atomtype_exception=True):
+def from_inchi(mol, inchistr, backend='openbabel-first', raise_atomtype_exception=True):
     """
     Convert an InChI string `inchistr` to a molecular structure. Uses
-    a user-specified backend for conversion, currently supporting 'try-all' (default), rdkit-first,
+    a user-specified backend for conversion, currently supporting 'openbabel-first' (default), rdkit-first,
     rdkit, and openbabel.
     """
     if inchiutil.INCHI_PREFIX in inchistr:
@@ -325,11 +325,11 @@ def from_smarts(mol, smartsstr, backend='rdkit', raise_atomtype_exception=True):
     return _read(mol, smartsstr, 'sma', backend, raise_atomtype_exception=raise_atomtype_exception)
 
 
-def from_smiles(mol, smilesstr, backend='try-all', raise_atomtype_exception=True):
+def from_smiles(mol, smilesstr, backend='openbabel-first', raise_atomtype_exception=True):
     """
     Convert a SMILES string `smilesstr` to a molecular structure. Uses
-    a user-specified backend for conversion, currently supporting try-all (default), rdkit-first,
-    rdkit (default) and openbabel.
+    a user-specified backend for conversion, currently supporting openbabel-first (default), rdkit-first,
+    rdkit and openbabel.
     """
     return _read(mol, smilesstr, 'smi', backend, raise_atomtype_exception=raise_atomtype_exception)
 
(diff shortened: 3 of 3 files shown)


JacksonBurns:
@xiaoruiDong there are 3 failures. Two of them are inchi translation related - rdkit gives a dofferent adjacency list than we expect. The other is different - we expect rdkit to fail to translate a molecule, but it suceeds. 

For the first two errors, we need to check and see if the inchi are equivalent. For the last one, I think we need to just find a new un-translatable molecule. Thoughts?

JacksonBurns:
I will also add that we have moved about 2 years of RDKit updates: https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/6395007324/job/17357683191#step:4:354

xiaoruiDong:
Thanks. I will look into those InChI Generation.

I hope to reproduce it on my M2 Mac, with the conda trick to install osx-64 deps. However, surprisingly, I keep getting dep conflict errors, though the CI for MacOS installation is error free....

JacksonBurns:
@xiaoruiDong thanks for the quick update!

For the first two failues - glad to see that the translation are equivalent. But this presents a more interesting problem. The unit test calls `self.compare`, which does this:

https://github.com/ReactionMechanismGenerator/RMG-Py/blob/c9cb854307632830e4b06c7db77c4484d25e71c4/test/rmgpy/molecule/translatorTest.py#L72-L80

I think RMG is calling (inside of the `get_augmented_inchi`) OpenBabel, which is returning the original test target.

That would mean that this failure is related to the second one. We do not need to test for RDKit failing to translate to InChI here, since it seems to do better now.

Here is my proposal:
 1. Change the environment file to specify a version of RDKit which does not have the bug the previous version had.
    - we could easily just do `>=2022.9.1` which is the version the CI picked
 2. Add a `pytest.skip` to the third failing unit test, with `reason=` this unit test checks for a bug which has been patched in version of RDKit of the above version or newer
 3. Change the code for `Molecule.get_augmented_inchi` to call RDKit instead of OpenBabel

Could you implement this, and/or let me know what you think would be better?

xiaoruiDong:
> @xiaoruiDong thanks for the quick update!
> 
> For the first two failues - glad to see that the translation are equivalent. But this presents a more interesting problem. The unit test calls `self.compare`, which does this:
> 
> https://github.com/ReactionMechanismGenerator/RMG-Py/blob/c9cb854307632830e4b06c7db77c4484d25e71c4/test/rmgpy/molecule/translatorTest.py#L72-L80
> 
> I think RMG is calling (inside of the `get_augmented_inchi`) OpenBabel, which is returning the original test target.

As a record, 

Inside of the `get_augmented_inchi`, it is `translator.to_inchi` that is being called. For radical molecules, the normal inchi is created by `_write`, with the backend being `'rdkit-first'`, and the u layer is created by `inchiutil`'s method `create_augmented_layers` and `compose_aug_inchi`. In `_writer`, RMG will convert an RMG molecule to an `rdkitmol` and use `Chem.inchi.MolToInchi` on it to obtain the inchi.

For the example here, RDKit with a version before 2020.03.01 returns the `InChI=1/CH2O2/c2-1-3/h1H,(H,2,3)`, while some later version (I don't know exactly which) returns `InChI=1/CH2O2/c2-1-3/h1-2H`, the u indices are created based on the obtained normal inchi and the molecule's radical sites.

> 
> That would mean that this failure is related to the second one. We do not need to test for RDKit failing to translate to InChI here since it seems to do better now.
> 
> Here is my proposal:
> 
> 1. Change the environment file to specify a version of RDKit that does not have the bug the previous version had.
>    
>    * we could easily just do `>=2022.9.1` which is the version the CI picked

Sounds good 

> 2. Add a `pytest.skip` to the third failing unit test, with `reason=` this unit test checks for a bug which has been patched in version of RDKit of the above version or newer

Sounds good 

> 3. Change the code for `Molecule.get_augmented_inchi` to call RDKit instead of OpenBabel
> 


See my comments above. I also checked the inchi generated by Openbabel, and it is `InChI=1/CH2O2/c2-1-3/h1-2H`, the one generated the by later version of RDKit. 

`translator.to_inchi` accepts an argument of `backend` while `Molecule.get_inchi` and `Molecule.get_augmented_inchi` don't. I may prefer to expose the `backend` to them for the user to choose which backend to use.


> Could you implement this, and/or let me know what you think would be better?

JacksonBurns:
@xiaoruiDong the tests are failing and I don't know why. When running the regression tests a file called `aromatic-core.log` should be written, but for some reason it isn't ([link to failure](https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/6412195184/job/17439693795#step:17:108)) - do you know what in this PR could be causing that?

JacksonBurns:
I'm grasping at straws here - I downloaded the stable regression results from the action runs on RMG-Py, the dynamic results from one of the above runs, and then manually executed `checkModels.py` to see if the file would not be written. But it was. So I'm really not sure what's wrong.

JacksonBurns:
Apparently a bad build ? https://github.com/conda-forge/rdkit-feedstock/issues/93

JacksonBurns:
Thanks to @rwest this PR can actually move forward - I'm going to copy paste the solution to the weird problem (see the hidden comments from me) for reference:
```python
# since RDKit 2022.03.1, logging is done using the Python logger instead of the
# Cout streams. This does not affect running RMG normally, but this testing file
# only works properly if it is the only logger
# see https://github.com/rdkit/rdkit/pull/4846 for the changes in RDKit

# clear all other existing loggers
# https://stackoverflow.com/a/12158233
for handler in logging.root.handlers[:]:
    logging.root.removeHandler(handler)

# once moved to a more recent python (at least 3.8), just add force=true to this statement
# and remove the above
```

I'm going to let the CI go, and this should now pass, or at least run 😅

JacksonBurns:
The patch required from switching RDKit versions is also now causing [tons of useless warnings](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2553#issuecomment-1765047352) to be printed in the regression testing output. This was resolved in later versions of descripatastorus (the package raising the errors, see [here](https://github.com/chemprop/chemprop/issues/113)), so I am switching to `conda-forge` to try and get a later version.

rwest:
Sounds [here](https://github.com/bp-kelley/descriptastorus/issues/7) like that could put the warnings into a different logger namespace, so may need additional tweaking after the switch. Good luck!

JacksonBurns:
Fix? Put log purge at very top of file, after other imports?

JacksonBurns:
Stuck between a rock and a hard place.

Upgrading to the latest RDKit also allowed upgrading descriptastorus, a dependency of chemprop, that we have our own custom build of (which we should stop using in favor of conda-forge). The latest version of descriptastorus that we can use (that supports Python 3.7) raises obnoxious warnings on import time that we can't (easily) avoid. We can't upgrade to the actual latest version that doesn't do this, because it requires SciPy 1.9 or newer, which is Python 3.8+ only.

Temporary patch to get this PR through is to pull the even older version of descriptastorus on the RMG channel, for now. The new version of chemprop no longer uses it at all, so once that transition is made this will be a moot point. Actually transitioning to the latest chemprop (which supports python 3.11/12 will be difficult too).

JacksonBurns:
If above CI fails, may have potential incompatibilities of PyJulia/DiffEqPy/etc. with Julia 1.9

JacksonBurns:
Error looks as expected, now need to import `admonition` from the ML estimator file and `pytest.skip` the failing test, with that admonition as the reason.

JacksonBurns:
@xiaoruiDong we got RDKit working, but doing that required increasing the version of another dependency (descriptastorus) that was part of chemprop and this change broke the way we do logging (because descriptastorus is bad). To resolve this, we temporarily removed Chemprop to also removed that other dependency. I think this was OK because we were going to need to remove Chemprop anyway since it does not support modern python.

Your thoughts on this and on the way forward?

xiaoruiDong:
> @xiaoruiDong we got RDKit working, but doing that required increasing the version of another dependency (descriptastorus) that was part of chemprop and this change broke the way we do logging (because descriptastorus is bad). To resolve this, we temporarily removed Chemprop to also removed that other dependency. I think this was OK because we were going to need to remove Chemprop anyway since it does not support modern python.
> 
> Your thoughts on this and on the way forward?

I agree with the plan. With the modifications you mentioned, are we able to utilize a newer version of python now?

Another thing potentially relevant. I was looking at if we can get rid of the dependency of `lpsolve55` today. The conclusion is yes. There are two places using lpsolve, one at `resonance`, the other at `isodesmic`, and I think I know how to do it. The only issue is the relevant API requires Python >= 3.8, and RMG is still locked at Python 3.7 now (?) I installed a fresh environment on my Mac today, and it is still python 3.7.12. Once RMG is able to upgrade to Python 3.8 or higher, I can work on it.

xiaoruiDong:
@JacksonBurns, I saw some failures and errors in the CI related to the changes of removing chemprop.

Can you skip or comment out the tests related to ML estimator setup and prediction?
1. https://github.com/ReactionMechanismGenerator/RMG-Py/blob/04559705742d3948712526427cfd60106d21fae6/test/rmgpy/ml/estimatorTest.py#L37-L62
2. https://github.com/ReactionMechanismGenerator/RMG-Py/blob/04559705742d3948712526427cfd60106d21fae6/test/rmgpy/data/thermoTest.py#L125C1-L130C65
3. https://github.com/ReactionMechanismGenerator/RMG-Py/blob/04559705742d3948712526427cfd60106d21fae6/test/rmgpy/data/thermoTest.py#L605-L653
4. https://github.com/ReactionMechanismGenerator/RMG-Py/blob/04559705742d3948712526427cfd60106d21fae6/test/rmgpy/data/thermoTest.py#L655-L728

JacksonBurns:
After this is merged, we need to rebuild pydas and pydqed for the newer version of Python, then we should be able to upgrade to the latest version on Python.

I would very much appreciative getting rid of lpsolve55!

I will skip the failing tests so we can get these merged.

JacksonBurns:
@xiaoruiDong this PR should be good to go in soon. Then we should meet up offline and go over the plan for upgrading to Python 3.11 etc.

xiaoruiDong:
@JacksonBurns 
```
==================================== ERRORS ====================================
________________ ERROR collecting test/rmgpy/data/thermoTest.py ________________
test/rmgpy/data/thermoTest.py:106: in <module>
    class TestThermoDatabase:
test/rmgpy/data/thermoTest.py:605: in TestThermoDatabase
    @pytest.mark.skip(reason=ADMONITION)
E   NameError: name 'ADMONITION' is not defined
```

xiaoruiDong:
> @xiaoruiDong this PR should be good to go in soon. Then we should meet up offline and go over the plan for upgrading to Python 3.11 etc.

Sure. We can discuss it during the RMG subgroup meeting.

JacksonBurns:
Argh forgot to import ADMONITION, will fix this later.

JacksonBurns:
@xiaoruiDong should work now!

xiaoruiDong:
@JacksonBurns Unfortunately, there are still a few more tests involving the ml estimator module, can you take a look? Thanks!

JacksonBurns:
Ok this time I think I got it for good...

rmgpy/molecule/molecule.py
```diff
@@ -1863,62 +1863,74 @@ def to_single_bonds(self, raise_atomtype_exception=True):
...
-        
+
         or
-        
+
         Convert a molecular structure to an InChI string. Uses
         `OpenBabel <http://openbabel.org/>`_ to perform the conversion.
+
+        Available options for InChI backend: 'rdkit-first' (default),
+        'try-all', 'rdkit', or 'openbabel'.
```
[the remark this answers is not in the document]
jonwzheng: In this respect, following off of the previous comment, it also feels awkward to need to repeat the arguments for the `translator` function in the docstring here. If the supported keyword arguments change in `translator` we have to remember to update it here.

jonwzheng (commented):
The changes look good to me and I see the tests are passing. Please see some small comments above.

Philosophically, what is the idea behind exposing the translator functions to `Molecule`  at all? Looking in the codebase, it seems like there is an awkward mixture of importing the `to_inchi` functions from `translator` and calling `Molecule.to_inchi` in the unit tests. To me, it feels like it would make more sense to just use the `translator` functions in all cases (like the paradigm of using `Chem.MolToSmiles` rather than `Mol.ToSmiles`). Including these functions in `Molecule` risks unnecessarily duplicating code and docstrings that could easily become out-of-date. However, maybe people find it more convenient. Let me know what you think.

xiaoruiDong:
> The changes look good to me and I see the tests are passing. Please see some small comments below.
> 
> Philosophically, what is the idea behind exposing the translator functions to `Molecule` at all? Looking in the codebase, it seems like there is an awkward mixture of importing the `to_inchi` functions from `translator` and calling `Molecule.to_inchi` in the unit tests. It's not clear what the "canonical" implementation is, especially if there is the chance for the default arguments to diverge.
> 
> To me, it feels like it would make more sense to just use the `translator` functions in all cases (like the paradigm of using `Chem.MolToSmiles` rather than `mol.ToSmiles`). Including these functions in `Molecule` risks unnecessarily duplicating code and docstrings that could easily become out-of-date. However, maybe people find it more convenient. Let me know what you think.

Good question. I guess Jackson may have a more valuable insight from a more professional point of view. To me, the motivation is really about "convenience for users" when implementing something like this in python. Like You don't have to import different modules and usually it ends up with a slightly shorter code. It is also very prevalent, like in PyTorch, there are `torch.sum(tensor)` and `Tensor.sum()`. From a developer's point of view, this indeed increase some extra work, but still manageable.

xiaoruiDong:
@JacksonBurns Any thoughts about @jonwzheng 's question?
@jonwzheng The docstrings in the translator module are corrected.

JacksonBurns:
I see both perspectives here in terms of convenience for users and ease of development. I think we should probably clean up the code internally so that we don't have copies of defaults anywhere that could possibly go out of date (and at the same time ensure that the class methods are calling those functions and pointing to their docs), but that's a later problem and I believe out of scope of this PR.

jonwzheng (approved):
Changes look good to me and I agree any internal changes would be outside the scope of this PR. Nice work and thank you for working on this issue. Setting this to approved.

JacksonBurns:
This PR is very close to ready.

@rwest left a comment here (https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2628#discussion_r1518076887) approving of the changes to the inchi strings, which is probably the only big 'theory' question in this PR.

Since this will be a semi-large change (removing a feature), I think we should add one more review to be safe. @hwpang please take a look over this PR and let us know if anything needs changes.

environment.yml
```diff
@@ -54,7 +57,7 @@ dependencies:
   - coverage
   - cython >=0.25.2
   - scikit-learn
-  - scipy
+  - scipy <1.11
```
hwpang, line 60: Why do we add this?

rmgpy/molecule/molecule.py
hwpang, line 1878: Could you add an explanation in the docstring on what packages `try-all` uses and their sequence?

test/rmgpy/molecule/translatorTest.py
```diff
@@ -238,7 +239,7 @@ def test_ch2o2(self):
 3 O 1 {1,S}
 """
 
-        aug_inchi = "InChI=1/CH2O2/c2-1-3/h1H,(H,2,3)/u1,2"
+        aug_inchi = "InChI=1/CH2O2/c2-1-3/h1-2H/u1,3"
```
hwpang, line 242: Why is the inchi different? Is it due to RDKit update?

hwpang (commented):
Thanks for the PR! Mostly LGTM with some questions

hwpang:
I also see that the title suggests we get diffeqpy from pip now, but I don't see relevant changes, could you update the title to reflect what has been changed?

JacksonBurns:
> I also see that the title suggests we get diffeqpy from pip now, but I don't see relevant code changes, could you update the title to reflect what has been changed?

The docs build is fixed here: https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2628 which we can merge after this PR

rmgpy/molecule/molecule.py
JacksonBurns: @xiaoruiDong can you do this?

environment.yml
JacksonBurns: Thanks for catching this. I believe I originally added this because of descriptastorus (which was incompatible with later versions of scipy) but it might not be required. I will remove it and we can see what happens in the CI.

JacksonBurns:
> I also see that the title suggests we get diffeqpy from pip now, but I don't see relevant code changes, could you update the title to reflect what has been changed?

I think this might have got wiped out in a force push. I will add it back and see what happens in the CI.

JacksonBurns:
@hwpang turns out that one of our other dependencies requires diffeqpy during the conda install, so we end up with two copies of it if we ask for it from pip. We will have to deal with this in a separate PR.

hwpang (approved):
LGTM

hwpang:
Approved as agreed upon in RMG subgroup

Issue #2549, opened 2023-09-22
Incorrect equilibrium constant for reactions with multidentate species


bjkreitz:
### Discussed in https://github.com/ReactionMechanismGenerator/RMG-Py/discussions/2534

<div type='discussions-op-text'>

<sup>Originally posted by **bjkreitz** August 29, 2023</sup>
**Issue**

I think that the calculation of the concentration equilibrium constant is wrong for reactions involving multidentate species. 
The  concentration equilibrium constant is calculated for surface reactions according to the following equation: 

$$K_{ci}=\frac{p}{RT}^{\sum{\nu_{gi}}} \Gamma^{\sum{\nu_{si}}}K_{pi} $$

where the first sum is over all gas-phase species \$g\$ in reaction \$i\$ and the second is over all surface species \$s\$. According to the Chemkin manual (https://github.com/ReactionMechanismGenerator/RMG-Py/files/12370978/chemkin11surfacechemkin.pdf, page 64), the calculation of the equilibrium constant is different when adsorbates with more than one binding site are involved. The correct equation should be:

$$K_{ci}=\frac{p}{RT}^{\sum{\nu_{gi}}} \Gamma^{\sum{\nu_{si}}}\prod{\sigma_{s}^{-\nu_{si}}} K_{pi} $$ 

If a reaction involves only monodentate species, the term \$\prod{\sigma_{s}^{-\nu_{si}}}\$ is 1. Otherwise a small factor is used to multiply \$K_{pi}\$.

The question is whether the second equation is correct. I haven't found a detailed derivation of this equation yet. 
If the second equation is correct, I can start a PR since I have already written the code to fix this.</div>

After making some tests in Cantera and pen/paper calculations, I figured out that the second equation is used in Cantera for reactions involving multidentate species. I'm going to make a PR to fix the equation for the calculation of the equilibrium constant in RMG.

PR #2554, opened 2023-10-04
fix Kc for bidentate rxns

bjkreitz:
### Motivation or Problem
The current calculation of the concentration equilibrium constant is wrong for reactions involving multidentate species. According to the Chemkin manual, there is a term that accounts for the number of binding sites as described in issue #2549 

### Description of Changes
I have updated the calculation of the equilibrium constant to account for this term. 

### Testing
An example reaction for which this term is going to be relevant is:  *C*C <=> 2 *
In this case, the factor should be $\prod{\sigma_{s}^{-\nu_{si}}}=2$, which is also reproduced by RMG. 
I ran RMG and there were no issues. 

### Reviewer Tips
Since this term is only relevant for reactions involving multidentate species, should we implement a check before we calculate $\prod{\sigma_{s}^{-\nu_{si}}}$ for every reaction to save time? But maybe it doesn't matter.


bjkreitz:
There was an issue with a gas-phase reaction test in the `TestReaction` class for  `test_equilibrium_constant_kc`
In the test, the molecule object is removed from the reactant via `rxn2_copy.reactants[0].molecule = []`
Since the molecule list is now empty, the check `contains_surface_site()` gave an error. I added a check to to see if the molecule list is empty, which fixes the problem. I don't understand the purpose of the test, though.

rmgpy/reaction.py
```diff
@@ -636,13 +636,31 @@ def get_equilibrium_constant(self, T, type='Kc', surface_site_density=2.5e-05):
...
+                if product.contains_surface_site() and product not in unique_species:
+                    unique_species.append(product)
+                    sigma_nu *= product.get_binding_sites() ** (-self.get_stoichiometric_coefficient(product))
+
+        for reactant in self.reactants:
+            if len(reactant.molecule) != 0:
+                if reactant.contains_surface_site() and reactant not in unique_species:
+                    unique_species.append(reactant)
+                    sigma_nu *= reactant.get_binding_sites() ** (-self.get_stoichiometric_coefficient(reactant))
+
```
rwest, line 654: I think this might add ineficciency and a bug.
If we have `A + A <=> B + A`.
we loop through each product (B then A)
if it's the first time we've seen it (not in unique_species) then we multiply by $\sigma ^ \nu$.
Works for B, then works for A.
then we loop the reactants (A then A).
First the bug - the unique_species list isn't reset, so it's skipped because we saw A as a product.
That's easily fixed: reset the list.
But then the inefficiency: 
Rather than just do the multiplication twice (once for each instance of A) we check if it's the first time we see A, then ask for get_stoichiometric_coefficient, which ..... aaaah. I take it back... the inefficiency cancels the bug, because of the way get_stoichiometric_coefficient returns the net value.
We already took care of all the A's.
OK. So I now think it's probably bug-free, but I think perhaps still a bit inefficient.
`get_stoichiometric_coefficient` loops through all the reactants and products every time it is called, counting how many times the requested species shows up.  Also, raising things to exponentials with `**` isn't fast. 

But correct is better than fast and wrong (probably, usually), so I don't mind you merging this and we speed it up later. I don't have time now.

And maybe what I think is faster, isn't much.

rmgpy/species.py
```diff
@@ -488,6 +488,16 @@ def is_surface_site(self):
         """Return ``True`` if the species is a vacant surface site."""
         return self.molecule[0].is_surface_site()
 
+    def get_binding_sites(self):
```
rwest, line 491: Number of binding sites is in fact a property of the molecule. Perhaps the method should live in that Class.
You're just assuming here that the first molecule is the representative one.
(Once we get surface resonance working, it might make a difference...?)

rmgpy/reaction.py
```diff
@@ -636,13 +636,31 @@ def get_equilibrium_constant(self, T, type='Kc', surface_site_density=2.5e-05):
         dN_surf = number_of_surface_products - number_of_surface_reactants # change in mols of surface spcs
         dN_gas = number_of_gas_products - number_of_gas_reactants # change in mols of gas spcs
 
+        # Determine the multiplication factor of the binding site^(-stoichiometric coefficient)
+        # (only relevant for reactions involving multidentate adsorbates)
+        unique_species = []
+        sigma_nu = 1
+        for product in self.products:
+            if len(product.molecule) != 0:
```
rwest, line 644: I'd probably wrap it in a try/catch instead of putting this check. It's (during a 'typical' RMG run) always going to pass this check, so checking it is a waste of time.

rwest (commented):
Shared some thoughts. Don't have time for more right this second.

rmgpy/reaction.py
JacksonBurns: I don't think this will be much slower than the alternative, and this is clear in what it does. I vote that this goes in as-is.

rmgpy/reaction.py
JacksonBurns: Do we need a try/catch, even? Are there cases where a reaction would have either no products or no reactants?

JacksonBurns (changes requested):
One minor clarifying question about the `len(product.molecule)` statement, but otherwise this looks great! Nearly ready - going to 'Request Changes' but I just want to make sure we discuss the aforementioned.

rmgpy/reaction.py
bjkreitz: Why do we need this case for the `test_equilibrium_constant_kc` where the molecule is removed or what is the purpose of it? If we get rid of this part of the test, we could get rid of the check.

rmgpy/species.py
bjkreitz: That's a good suggestion. I can move the method. Surface resonance does not change the number of binding sites, so we can use the first molecule to check it.

rwest:
Bjarne's question is discussed a bit in https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1945 which added that bit of the test. I think there are some corners of other scripts that use parts of RMG that might want to evaluate a reaction rate without molecules being designed (eg. when importing a chemkin file with no dictionary). Hence we check that it's possible.  **During** an RMG run, the `molecule[0]` will always exist.

Anyway, I then got carried away adding commits. Sorry to trigger a new round of reviewing!

One thing I noticed is that we don't actually have a unit test for this, so it's quite possible that my implementation doubles when it should halve, or vice versa. I tried to get it right, but would feel happier if this was tested.

rmgpy/molecule/molecule.py
```diff
@@ -1155,6 +1155,19 @@ def contains_surface_site(self):
...
+    def number_of_surface_sites(self):
+        """
+        Returns the number of surface sites in the molecule.
+        e.g. 2 for a bidentate adsorbate
+        """
+        cython.declare(atom=Atom)
+        cython.declare(count=cython.int)
+        count = 0
+        for atom in self.atoms:
+            if atom.symbol == 'X':
```
JacksonBurns, line 1167: The class `atom` implements a `is_surface_site()` function that we should use. It does the same thing that you have here (avoiding code duplication) but also if we ever add other symbols for sites we will be more future-proof:
```suggestion
            if atom.is_surface_site():
```

rmgpy/reaction.py
```diff
@@ -637,12 +642,34 @@ def get_equilibrium_constant(self, T, type='Kc', surface_site_density=2.5e-05):
         dN_gas = number_of_gas_products - number_of_gas_reactants # change in mols of gas spcs
 
         if type == 'Kc':
+            # Determine the multiplication factor of the binding site^(-stoichiometric coefficient)
+            # (only relevant for reactions involving multidentate adsorbates)
```
JacksonBurns, line 646: Thanks for adding these explanations inline - very helpful.

JacksonBurns (dismissed):
This LGTM, with one small note about checking for surface sites - I think we should get it into main to enable development elsewhere and find any issues.

bjkreitz:
Thanks, Richard. Your code is much more efficient, but less straightforward. It took me a while to figure out that it is correct. I think a unit test would be helpful and I'm going to take a first stab at making one that checks if the right multiplication factor is applied.

rwest:
I'm sorry Bjarne. I thought it was simpler, having spent a while figuring out that *your* code was correct! 
I guess I've learned that "straightforward" is subjective, if people conceptualize  a phenomenon or algorithm differently! Sorry for the extra time. But *maybe* it's worth it - this may get called quite often 🤷

bjkreitz:
I added a test to check if Kc is correctly calculated for a reaction with a bidentate adsorbate.

JacksonBurns (changes requested):
A note on the checking for surface sites, but otherwise looking good.

test/rmgpy/reactionTest.py
```diff
@@ -372,14 +470,20 @@ def test_get_rate_coefficient_units_from_reaction_order(self):
...
+        Test the Reaction.get_equilibrium_constant() method for Kc of a surface reaction
+        and a reaction involving bidentate adsorbates
         """
         Tlist = numpy.arange(400.0, 1600.0, 200.0, numpy.float64)
         Kclist0 = [15375.20186, 1.566753, 0.017772, 0.0013485, 0.000263180, 8.73504e-05]
         Kclist = self.rxn1s.get_equilibrium_constants(Tlist, type="Kc")
         for i in range(len(Tlist)):
             assert round(abs(Kclist[i] / Kclist0[i] - 1.0), 4) == 0
 
+        Kclist0_rxn3 = [8.898588e+07, 3.591402e+03, 2.279520e+01, 1.097370, 1.454379e-01, 3.436572e-02]
```
rwest, line 482: I'm going to take your word for it 😄 
I'm sure you're right. And this will at least detect regressions - i.e. if something changes it.

In some ways its good to use real species, but if I were designing the test, I'd have probably used all zeros for the NASA polynomials, so ∆Hºrxn = ∆Sºrxn = 0 and the Kc is just 2 (or 0.5, lol).   Anyway this looks fine to me. Thanks for adding the test.

rwest (approved):
Thanks Bjarne! Looks good to me

JacksonBurns (approved):
LGTM - with @rwest already approving, I will go ahead and rebase and enable auto-merge. Thanks @bjkreitz!

Issue #2493, opened 2023-06-26
RMG documentation (12.1.2 Forbidden Groups) includes incorrect instructions about forbidding molecules and species


Nora-Khalil:
### Bug Description
RMG documentation ([12.1.2 Forbidden Groups](http://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/database/introduction.html?highlight=forbidden)) reads as if you can specify a forbidden species or molecule **inside the groups.py file** inside a specific kinetics family's folders. It mentions that you can "simply replace the 'group' keyword with 'molecule' or 'species'". In reality, this cannot be done in groups.py, and raises an error that 'species' or 'molecule' is not recognized:

`TypeError: load_forbidden() got an unexpected keyword argument 'species'`

I believe forbidding species and molecules can be done in the global forbidden structure database only.  The documentation should probably be updated to make this clear. 

### How To Reproduce
1. Add in a forbidden species in any of the groups.py. I did something kind of like: 
```

forbidden(
    label = "C_quintet",
    species =
"""
multiplicity 5
1 C u4 p0
""",
    shortDesc = u"""""",
    longDesc =
u"""

""",
)

```
in the groups.py file in the halocarbene_CO_dimerization kinetics family. 

### Expected Behavior
Expected no error, and expected that the database could be loaded properly at the beginning of the RMG job. 

### Installation Information
OS: macOS 12.1
Python: Python 3.7.11
RMG version information: RMG-Py v3.1.0

JacksonBurns:
@Nora-Khalil this seems like a bug, but only in the docs which should be updated to reflect the proper usage. Could you make a PR to clarify?

Nora-Khalil:
Sounds good.

xiaoruiDong:
According to the line history, it seems that the support of "molecule and species" is never implemented.

Nora-Khalil:
I think the documentation should be edited instead of adding support for passing `molecule` and `species`. Looking back to when I made this issue, one of the families was creating multiple similar-looking species that violated the octet rule. Instead of naming each one as forbidden, the best solution was to change the nodes in group.py so these were not produced at all. So I can make a PR as originally planned to update the documentation so it reflects the proper usage, if everyone else agrees.

JacksonBurns:
[...]

PR #2561, opened 2023-10-24
changing forbidden groups documentation 

Nora-Khalil:
### Motivation or Problem
This is the PR in response to [issue #2493](https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2493). It updates the documentations and makes it clear that you can only specify forbidden molecules or species in the forbiddenStructures.py file and not in groups.py. 

### Description of Changes
Added in two sentences to `introduction.rst` that remind the reader that forbidden molecules/species cannot be specified in groups.py.

### Testing
Checked what the new updates look like in the documentation. looked fine.


JacksonBurns (approved):
LGTM! Thanks for the patch

PR #2514, opened 2023-08-02
Add surface combustion example for regression testing

sevyharris:
moved over from https://github.com/ReactionMechanismGenerator/RMG-tests/pull/85

### Motivation or Problem
We want to make sure in our testing that surface examples work. 

### Description of Changes
Adding a surface test input file.

This also adds functionality to rmgpy.tools.canteramodel.py to load, simulate, and compare surface mechanisms. Then it also adds hooks in the upstream callers, rmgpy.tools.regression.py and rmgpy.tools.observablesregression.py to pass the surface-specific information into the cantera model

### Testing
I ran the regression test on an example surface file and then also on the superminimal and nitrogen examples.

### Reviewer Tips
This PR can be broken down into two parts: adding the surface functionality to the regression tool, and then adding the surface testing files to the CI workflow.

PR #2563, opened 2023-10-26
Add surface mechanisms to regression tool

sevyharris:
### Motivation or Problem
We would like to add surface regression tests to the testing suite, but the regression tool cannot yet handle surface mechanisms, so this adds functionality for surface mechanism testing

### Description of Changes
This adds functionality to rmgpy.tools.canteramodel.py to load, simulate, and compare surface mechanisms. Then it also adds hooks in the upstream callers, rmgpy.tools.regression.py and rmgpy.tools.observablesregression.py to pass the surface-specific information into the cantera model

### Testing
I've tested this on the superminimal (non-surface mechanism), and a minimal surface example. Here is the minimal surface [input.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/13180867/input.txt) and the minimal surface  [regression_input.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/13180868/regression_input.txt). 
I'll try testing this on a few other regression examples to confirm I haven't broken anything.

### Reviewer Tips
rmgpy.tools.regression.py calls rmgpy.tools.observablesregression.py which calls rmgpy.tools.canteramodel.py
The regression tool can be called with the following command: 
`python-jl /path/to/RMG-Py/rmgpy/tools/regression.py /path/to/regression_input.py /path/to/old/mechanism/chemkin /path/to/new/mechanism/chemkin`


sevyharris:
Deleting this PR and moving the changes to https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2514 because I think it makes more sense to combine them.

Issue #2559, opened 2023-10-17
Status Tracking: `chemprop` and ML Thermochemistry Prediction


JacksonBurns:
Summary:
 - RMG currently uses `chemprop` version 0.0.1 to perform machine learning prediction of molecular thermochemistry
 - This version only supports Python version 3.7 and 3.8, so it is blocking the upgrade of RMG to Python 3.11/12
 - `chemprop` will be temporarily removed from RMG to allow upgrading Python versions

[This](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2450) was the original PR where removing `chemprop` was started, and was eventually folded into #2553.

The eventual re-integration of `chemprop` will be dependent on the first stable release of v2 (and other steps mentioned in the first linked PR) detailed in this issue:
 - [ ] https://github.com/chemprop/chemprop/issues/517

JacksonBurns:
@KnathanM has agreed to start some basic feasability checks for the re-addition of Chemprop for thermochemistry prediction

PR #2565, opened 2023-10-27
exempt announcements from being closed as stale

JacksonBurns:
Stale label action currently closes old issues and PRs if they have had no interaction in 90 days, _except_ when they are labeled with `bug`. This PR also excepts those labeled with `Announcement`, especially meant for #2559 which will be open for a while but is not a bug.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -24,7 +24,7 @@ jobs:
           stale-issue-label: stale
           stale-pr-label: stale
           operations-per-run: 1000
-          exempt-issue-labels: bug
+          exempt-issue-labels: bug,Announcement
           exempt-pr-labels: bug
           close-issue-label: abandoned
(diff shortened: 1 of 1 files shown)


xiaoruiDong (approved):
LGTM. 

Minor: There are a few issues labeled as python 3.11 transition. Do you think it is better to include this label as `exempt-issue-labels` and remove it once we finish the transition or add the bug/announcement label to them to avoid them being automatically closed?

JacksonBurns:
> Minor: There are a few issues labeled as python 3.11 transition. Do you think it is better to include this label as `exempt-issue-labels` and remove it once we finish the transition or add the bug/announcement label to them to avoid them being automatically closed?

Easier to just add the announcement label to them, otherwise we would need another PR in the future to remove their label from this action. I also don't mind those issues getting closed - it reminds me to get back to work on them 😅

PR #2568, opened 2023-10-30
RMG-database Twin PR: Update to CI

JacksonBurns:
### Motivation or Problem
CI on RMG-database is currently failing because it is out of date, so it is being added over there requiring some changes over here.

### Description of Changes
This PR removes the database tests from CI being run on RMG-Py PRs, and does some minor cleanup.

### Testing
CI run.

### Reviewer Tips
Also see RMG-database PR https://github.com/ReactionMechanismGenerator/RMG-database/pull/635

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -131,7 +131,6 @@ jobs:
 
       # Clone RMG-database
       - name: Clone RMG-database
-        if: github.repository != 'ReactionMechanismGenerator/RMG-database'
         run: |
           cd ..
           git clone -b $RMG_DATABASE_BRANCH https://github.com/ReactionMechanismGenerator/RMG-database.git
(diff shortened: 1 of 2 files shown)


hwpang:
Thanks for the PR! This makes me wonder if we would miss bugs in the code that generates the tree, since the code is in `RMG-Py`?

hwpang:
I gave this another think, and I actually think this might be ok. The error in the tree generation code will be detected by the tests [here](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/main/test/rmgpy/data/kinetics/familyTest.py#L716). If we can get a second opinion that also ok this then I'm happy to merge. I also just rebased the branch with main.

JacksonBurns:
@hwpang something is missing from the regression testing (like, one of the files appears to be missing), can you tell what it is? It only fails the observable testing.

hwpang:
I rebased this

PR #2413, opened 2023-04-16
`fragment.py`: refactoring, patches, and compatibility with RMS

JacksonBurns:
`fragment.py` should inherit from `molecule.py` rather than copy methods. This PR will attempt to implement this.

PR #2570, opened 2023-11-01
Getting structure information correctly for fragment

hwpang:
### Motivation or Problem
This PR contains changes made to https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2413. See this [comment](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2413#discussion_r1362981840) for details. I put it into a separate PR rather than push commits directly PR as suggested by @JacksonBurns.

### Description of Changes
I use a representative molecule for a fragment to calculate the correct molecular weight, number of atoms, and number of bonds. The number of atoms for the representative molecule is used to estimate radius for a fragment, which is used to estimate the stokes diffusivity. I also pass the kLA and kH into the RMS Species for fragments.

### DIFF
--- a/rmgpy/molecule/fragment.py
+++ b/rmgpy/molecule/fragment.py
@@ -322,15 +322,6 @@ class Fragment(Molecule):
         result = Graph.is_subgraph_isomorphic(self.mol_repr, other, new_initial_map)
         return result
 
-    def get_molecular_weight(self):
-        """
-        Return the fragmental weight of the fragment in kg/mol.
-        """
-        mass = 0
-        for vertex in self.vertices:
-            mass += vertex.element.mass
-        return mass
-
     def calculate_cp0(self):
         """
         Return the value of the heat capacity at zero temperature in J/mol*K.
--- a/rmgpy/rmg/reactors.py
+++ b/rmgpy/rmg/reactors.py
@@ -680,38 +680,20 @@ def to_rms(obj, species_names=None, rms_species_list=None, rmg_species=None):
     elif isinstance(obj, NASA):
         return rms.NASA([to_rms(poly) for poly in obj.polynomials], rms.EmptyThermoUncertainty())
     elif isinstance(obj, Species):
-        atomnums = dict()
-        for atm in obj.molecule[0].atoms:
-            try:
-                if atomnums.get(atm.element.symbol):
-                    atomnums[atm.element.symbol] += 1
-                else:
-                    atomnums[atm.element.symbol] = 1
-            except AttributeError:
-                # means it is fragment's cutting label
-                pass
-        bondnum = len(obj.molecule[0].get_all_edges())
-        
+
         if isinstance(obj.molecule[0], Fragment):
-            th = obj.get_thermo_data()
-            thermo = to_rms(th)
-            return rms.Species(
-                name=obj.label,
-                index=obj.index,
-                inchi="",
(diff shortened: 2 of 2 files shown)


JacksonBurns:
@hwpang change you change the base branch to the `afm_refactor` branch?

hwpang:
@JacksonBurns Just did!

rmgpy/rmg/reactors.py
```diff
@@ -680,38 +680,23 @@ def to_rms(obj, species_names=None, rms_species_list=None, rmg_species=None):
...
-                bondnum=bondnum,
-                diffusion=rms.EmptyDiffusivity(),
-                radius=0.0,
-                radicalelectrons=obj.molecule[0].multiplicity - 1,
-                molecularweight=0.0,
-                henrylawconstant=rms.EmptyHenryLawConstant(),
-                liquidvolumetricmasstransfercoefficient=rms.EmptyLiquidVolumetricMassTransferCoefficient(),
-                comment=obj.thermo.comment,
-            )
+            atomnums = dict()
```
JacksonBurns, line 685: I don't think this is needed here, since we set `atomnums = dict()` outside this if/else block (line 693)

rmgpy/molecule/fragment.py
```diff
@@ -326,8 +326,9 @@ def get_molecular_weight(self):
         """
         Return the fragmental weight of the fragment in kg/mol.
         """
+        self.assign_representative_molecule()
```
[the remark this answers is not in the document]
hwpang: I realize that we have access to the `get_molecular_weight` directly from `self.mol_repr`, so I did that in the new push.

rmgpy/rmg/reactors.py
hwpang: Good catch!

hwpang:
I realize that there are unit tests that test the molecular weight of a fragment is calculated based on the fragment size. This makes me wonder whether using the fragment structure or the representative molecule structure is correct. I will revert my changes regarding the molecular weight as I am not sure how it might change the downstream application. For example, is the mass balance of a reaction checked somewhere? If that's checked, then using the fragment structure is best.

hwpang:
> This looks good! Should fix some of the bugs we were having. A few small comments, but nothing major.
> 
> One overall comment that we should think about - I think that we might want to 'flip' the logic of Fragment. Right now we store the cutting-label containing 'molecule' inside the same attribute that the parent class Molecule stores its _actual_ molecule in, which is what is requiring us to override so many of the methods. Would it make more sense to store the representative molecule in the molecule attribute, and then store the sliced molecule in a new attribute? We would then just need to change the Fragment-specific methods (like `sliceitup_etc`) to operate on this new attribute, but I think that we could inherit all of the other methods from Molecule without any changes.

I think this will also involve changing many codes used by [`react_species`](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/dc2da84b1a6a637ba2af6b1f33d01f2c34d60af4/rmgpy/rmg/react.py#L82) and we would need to change many following parts of code to point them to the new attribute containing the sliced molecule.

Since the reacting part of code is much more core function of RMG and much more involved to change, I think it makes more sense to keep it as it is right now, as functions like calculating molecular weight are easy to work around.

JacksonBurns (approved):
LGTM! Merging now

PR #2571, opened 2023-11-02
Add regression input file for fragment + simple reactor and regression test for fragment + RMS reactor

hwpang:
### Motivation or Problem
The current fragment regression test only tests that the mechanism can be generated, but not testing that the results actually stay consistent with previous runs. Additionally, currently we are not testing fragment is compatible with RMS reactor.

### Description of Changes
I add a `regression_input.py` for the fragment + simple reactor regression test. I also add a new regression test for fragment + RMS reactor.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -168,7 +168,7 @@ jobs:
         id: regression-execution
         timeout-minutes: 60
         run: |
-          for regr_test in aromatics liquid_oxidation nitrogen oxidation sulfur superminimal RMS_constantVIdealGasReactor_superminimal RMS_CSTR_liquid_oxidation RMS_liquidSurface_ch4o2cat fragment;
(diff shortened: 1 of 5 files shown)


JacksonBurns:
@hwpang I added the fragment and rms fragment regression to the other line in the CI file makes it run the regression_input comparison

hwpang:
Thanks!

JacksonBurns:
@hwpang so the CI is failing for two reasons:
1. The new RMS+fragment regression test has never been run on main, so there are no regression results to do the baseline comparison.
2. The fragment regression test is raising an actual error about a missing file in the reference results, see https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/6737504658/job/18315100139#step:17:482

Can you take a look at 2. and help me figure out what went wrong?

JacksonBurns (approved):
This LGTM and will need to be merged with failing CI.

hwpang:
I got @JacksonBurns's offline approval to merge this in. The failure is because the edge was not saved in the stable run due to not setting `saveEdgeSpecies` to True.

PR #2574, opened 2023-11-07
Compatibility of `regression.py` with fragment

hwpang:
### Motivation or Problem
We merged in the fragment regression tests in https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2571. We thought the failed test was because there was no stable version for comparison. However, it was actually caused by current code not able to read regression input for fragments.

### Description of Changes
I modified the `regression.py` to read fragment structures by importing necessary functions from `rmgpy.rmg.input`. I also replaced some duplicate functions.

### Testing
Let's see if the regression test for fragment works in the CI.

### DIFF
--- a/rmgpy/tools/regression.py
+++ b/rmgpy/tools/regression.py
@@ -75,9 +75,9 @@ def read_input_file(path):
         'True': True,
         'False': False,
         'observable': observable,
-        'SMILES': SMILES,
+        'SMILES': smiles,
         'species': species,
-        'adjacencyList': adjacencyList,
+        'adjacencyList': adjacency_list,
         'reactorSetups': reactorSetups,
         'options': options,
     }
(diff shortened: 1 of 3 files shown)


JacksonBurns:
@hwpang I think you need to take this line out of your regression input for RMS+fragment:
https://github.com/hwpang/RMG-Py/blob/fix_fragment_regression/test/regression/RMS_constantVIdealGasReactor_fragment/regression_input.py#L51

hwpang:
@JacksonBurns Thanks for catching this! The new push removes the unnecessary reactive flag. Hopefully this would work now.

hwpang:
@JacksonBurns Both fragment regression tests have passed!

JacksonBurns (approved):
LGTM - thanks for the quick patch!

Issue #2562, opened 2023-10-24
`get_deterministic_sssr` is not really deterministic


xiaoruiDong:
### 1. Bug Description
The bug was identified by the witness of @jonwzheng and @hwpang when searching thermo for [`C1=CC2C=CC=1C=C2`](https://rmg.mit.edu/database/thermo/molecule/1%20%20C%20u0%20p0%20c0%20%7B2,D%7D%20%7B6,D%7D%0A2%20%20C%20u0%20p0%20c0%20%7B1,D%7D%20%7B3,S%7D%20%7B9,S%7D%0A3%20%20C%20u0%20p0%20c0%20%7B2,S%7D%20%7B4,S%7D%20%7B8,S%7D%20%7B10,S%7D%0A4%20%20C%20u0%20p0%20c0%20%7B3,S%7D%20%7B5,D%7D%20%7B11,S%7D%0A5%20%20C%20u0%20p0%20c0%20%7B4,D%7D%20%7B6,S%7D%20%7B12,S%7D%0A6%20%20C%20u0%20p0%20c0%20%7B1,D%7D%20%7B5,S%7D%20%7B7,S%7D%0A7%20%20C%20u0%20p0%20c0%20%7B6,S%7D%20%7B8,D%7D%20%7B13,S%7D%0A8%20%20C%20u0%20p0%20c0%20%7B3,S%7D%20%7B7,D%7D%20%7B14,S%7D%0A9%20%20H%20u0%20p0%20c0%20%7B2,S%7D%0A10%20H%20u0%20p0%20c0%20%7B3,S%7D%0A11%20H%20u0%20p0%20c0%20%7B4,S%7D%0A12%20H%20u0%20p0%20c0%20%7B5,S%7D%0A13%20H%20u0%20p0%20c0%20%7B7,S%7D%0A14%20H%20u0%20p0%20c0%20%7B8,S%7D%0A), the result was non-deterministic (try refreshing the page). The issue was first mentioned in #2490. However, this is not actually a decision tree issue, so I decided to make it a separate PR.

### 2. Reproducing the issue with RMG APIs
```
import os.path as osp
import rmgpy
from rmgpy.molecule.molecule import Molecule
from rmgpy.data.thermo import ThermoDatabase

tmdb = ThermoDatabase()
tmdb.load(osp.join(rmgpy.settings["database.directory"], "thermo"))

smi = "C1=CC2C=CC=1C=C2"
[...]

PR #2589, opened 2023-12-19
Update uncertainty SI when averaging ThermoData

hwpang:
### Motivation or Problem
We should update `uncertainty_si`, instead of `uncertainty` to be consistent with units when averaging `ThermoData`.

### DIFF
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -292,7 +292,7 @@ def average_thermo_data(thermo_data_list=None):
                 averaged_thermo_data.Cpdata.value_si[i] /= num_values
 
                 cp_data = [thermo_data.Cpdata.value_si[i] for thermo_data in thermo_data_list]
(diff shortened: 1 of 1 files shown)


JacksonBurns (commented):
@hwpang this seems correct - could you take a look at the regression testing results, especially the aromatics, and see if the changes introduced here could be responsible for the (pretty significant) deviations?

hwpang:
@JacksonBurns Thanks for the comment! The regression testing results show changes in the thermochemistry values, but not the uncertainty values. Since this PR only changes the calculation on the uncertainty values, I don't think that is caused by this PR. They seem to be caused by the known nondeterministic thermo estimation for ring species mentioned in https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2562.

JacksonBurns:
@hwpang great, thanks for explaining - this LGTM. Thanks for catching this.

JacksonBurns (approved):

PR #2599, opened 2024-02-01
Change Documentation Build Action to use github-actions bot instead of RMG_dev

JacksonBurns:
GitHub is now requiring 2FA on all accounts - this means our helpful bot RMG_dev, which is used to push documentation to the repo, is 'banished' (or will at least break soon). This PR changes the docs building action to use the GitHub actions bot account to push the changes to the docs instead.

I'm not sure how to test this besides running it on GHA, just to see if it works?


rwest (approved):
Reading this, it looks like it should do the right thing (but that's the extent of my testing).  The proof is in the pudding.

PR #2602, opened 2024-02-05
fix checkout command for docs building action

JacksonBurns:
As luck would have it #2599 did _not_ in fact work, so this PR is attempting (again) to fix it.

Seems like after using the default git checkout github actions step, we can't use other git commands because of some nonsense with the ssh key. Rather than deal with figuring that out, I have just moved all of the weirdness we were doing in the git command to the original checkout command.

I have also removed the `rm -rf build/html` command. That directory does not exist before the clone (I think this was carried over from a previous script, because it doesn't actually break anything).

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -16,6 +16,9 @@ jobs:
         shell: bash -l {0}
     steps:
       - uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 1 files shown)


hwpang:
Force merge this before the tests pass as per request to fix documentation

Issue #2603, opened 2024-02-05
Can't construct surface rmgpy.molecule.Molecule()


sevyharris:
### Bug Description
I'm trying to create a rmgpy.molecule.Molecule() using smiles
```
rmgpy.molecule.Molecule(smiles='O=C=[X]')
```

But it fails from an RDKit error:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/tmp/ipykernel_18284/1886713532.py in <module>
----> 1 rmgpy.molecule.Molecule(smiles='O=C=[X]')
~/rmg/RMG-Py/rmgpy/molecule/molecule.cpython-37m-x86_64-linux-gnu.so in rmgpy.molecule.molecule.Molecule.__init__()
~/rmg/RMG-Py/rmgpy/molecule/molecule.cpython-37m-x86_64-linux-gnu.so in rmgpy.molecule.molecule.Molecule.from_smiles()
~/rmg/RMG-Py/rmgpy/molecule/translator.cpython-37m-x86_64-linux-gnu.so in rmgpy.molecule.translator.from_smiles()
~/rmg/RMG-Py/rmgpy/molecule/translator.cpython-37m-x86_64-linux-gnu.so in rmgpy.molecule.translator.from_smiles()
~/rmg/RMG-Py/rmgpy/molecule/translator.cpython-37m-x86_64-linux-gnu.so in rmgpy.molecule.translator._read()
~/rmg/RMG-Py/rmgpy/molecule/translator.cpython-37m-x86_64-linux-gnu.so in rmgpy.molecule.translator._rdkit_translator()
ValueError: Could not interpret the identifier 'O=C=[X]'

```
### Possible Solution
I'm wondering if we can get around RDKit's limitations with the following procedure:
- replace the surface sites X with a nonreactive species like Ar
- generate that molecule's adjacency list
- replace Ar with X and delete extra pairs of electrons on the X

It would look something like adding this to the _rdkit_translator() function in RMG-Py/rmgpy/molecule/translator.py:
```
from rdkit import Chem
import rmgpy.molecule

COX_smiles = 'O=C=[X]'

rdkitmol = Chem.MolFromSmiles(COX_smiles.replace('X', 'Ar'))
mol = rmgpy.molecule.molecule.Molecule()
output = rmgpy.molecule.converter.from_rdkit_mol(mol, rdkitmol)

lines = output.to_adjacency_list().split('\n')

[...]

PR #2607, opened 2024-02-07
Make surface molecules from smiles using Argon workaround

sevyharris:
### Motivation or Problem
The issue is described here, https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2603, but basically RDKit complains when you try to make a surface molecule using smiles.

### Description of Changes
This PR adds a workaround where you initialize the molecule using Ar as a replacement for X to get the adjacency list and then remove the extra pairs of electrons and change Ar back to X to get the originally intended molecule.

### Testing
Run this code to test out instantiating all the species in the Surface family training reactions:
**NOTE TO SELF -- DON'T FORGET TO REBUILD!**

```
import os
import glob
import rmgpy
import rmgpy.chemkin
import rmgpy.molecule

# Get all of the surface training reactions
search_str = os.path.join(rmgpy.settings['database.directory'], 'kinetics', 'families', 'Surface_*', 'training', 'dictionary.txt')
species_dict_files = glob.glob(search_str)

# Test on all the species in surface training reactions
for dict_file in species_dict_files:
    species_dict = rmgpy.chemkin.load_species_dictionary(dict_file)
    for key in species_dict.keys():
        sp = species_dict[key]
        
        test_smiles = sp.smiles.replace('Pt', 'X')
        test_sp = rmgpy.molecule.Molecule(smiles=test_smiles)
        
        assert(test_sp.is_isomorphic(sp.molecule[0]))    
```

### DIFF
--- a/rmgpy/molecule/adjlist.py
+++ b/rmgpy/molecule/adjlist.py
@@ -92,7 +92,9 @@ class ConsistencyChecker(object):
         the theoretical one:
 
         """
-        if atom.symbol in {'X','L','R'} or 'X' in [z.label for z in get_atomtype(atom, atom.bonds).generic]:
+        # if atom.symbol in {'X','L','R'} or 'X' in [z.label for z in get_atomtype(atom, atom.bonds).generic]:
+        # TODO handle this in a more generic way so we don't have to add more metals here
+        if atom.symbol in {'X','L','R','Pt'}:
             return  # because we can't check it.
 
         valence = PeriodicSystem.valence_electrons[atom.symbol]
--- a/rmgpy/molecule/adjlist.py
+++ b/rmgpy/molecule/adjlist.py
@@ -92,7 +92,7 @@ class ConsistencyChecker(object):
         the theoretical one:
 
         """
-        if atom.symbol in {'X','L','R'}:
+        if atom.symbol in {'X','L','R'} or 'X' in [z.label for z in get_atomtype(atom, atom.bonds).generic]:
             return  # because we can't check it.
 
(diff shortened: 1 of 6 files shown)


rwest (commented):
This looks good. Could we have a couple unit tests? I think there are some templates to copy.  Can we do a round-trip (i.e. generate SMILES-like strings with X in, that reproduce the original molecule on import)?

sevyharris:
This is working with the test code described above, so now I just need to put that into an actual unit test and it'll be ready to go.

sevyharris:
Unit tests added, but don't work yet because the SMILES string from the rdkit translator's to_smiles() uses 'Pt' instead of 'X'. Need to figure out where is the most appropriate place to make this final translation, but then it should be ready

sevyharris:
Some mission creep: RMG molecule had no way to distinguish between specific Pt atom and more general surface X atom.

I wanted `rmgpy.molecule.Molecule(smiles="[Pt]").to_smiles()` to return '[Pt]' and `rmgpy.molecule.Molecule(smiles="[X]").to_smiles()` to return '[X]', but there was no good way to do this since the internal RMG molecule adjacency list always converted Pt to X and then converted all X's back to Pt in the to_smiles() function.

So, I added a specific Pt atom type with the intention of being able to add other metals later on.

It passes my simple test cases, but now I need to clean up the PR and look for other places this might cause problems for RMG-cat.

sevyharris:
Richard's working on a related PR: https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2701. The rdkit * might be a more elegant way of doing this than the Ar workaround

PR #2608, opened 2024-02-08
Continuous Integration, Documentation Build, and Dockerfile Fixes

JacksonBurns:
These changes are unrelated but small, so I have combined them with reckless abandon.

 - The CI has been failing because of an upstream problem with Julia dependencies. To try and resolve this @mjohnson541 suggested pinning the Julia version to 1.8.5 - this PR does that in the CI files with `sed`. (note: ended up using 1.9.1 because 1.8.5 did not work)
 - In a continuing effort to address #2582 (at which #2599 and #2602 were only _partially_ successful) this changes some minor syntax in `git` commands inside the doc building action.
 - Dockerfile would not run properly on windows, a small patch has been added (see commit message for [51924d9](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2608/commits/51924d98ed65bac069cbbde0ab7c894979c567a5))

If the CI passes this should be good to go in.

PR #2609, opened 2024-02-11
Julia restriction

calvinp0:
### Motivation or Problem
Julia has been updated to 1.10.0 on Conda-Forge and therefore RMS does not install. https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2608 rectified this issue by patching in a version restriction on Julia to 1.9.1. This PR attempts to restrict the version in the environment.yml file so that we will not need to maintain the patches in the CI/Dockerfile/Doc

### Description of Changes
Restricted Julia to 1.9.1 in the environment.yml file

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -62,8 +62,6 @@ jobs:
         uses: actions/checkout@v3
 
       # configures the mamba environment manager and builds the environment
-      - name: Patch Environment File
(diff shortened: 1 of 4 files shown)


JacksonBurns (approved):
LGTM! Thanks for the quick patch. In the future when we try to upgrade to Python 3.11+ we may need to add back some flexibility to this version (though I don't think so), so I will keep this PR in mind.

PR #2610, opened 2024-02-12
#2602 part 2, Electric Boogaloo

JacksonBurns:
_Still_ trying to fix the docs building action. This time, it seems like omitting `--global` from the git sign in has meant that, for some reason, it just didn't work. See:
https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/7873601921/job/21481327559#step:12:38

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -59,8 +59,8 @@ jobs:
       - name: Make documentation - to publish
         if: ${{  github.event_name == 'push' && github.repository == 'ReactionMechanismGenerator/RMG-Py' }}
         run: |
-          git config user.name "github-actions"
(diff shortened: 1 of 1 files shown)


JacksonBurns:
> looks like this fixed the publish documentation issue!

Unfortunately we can't actually tell if the publish step works until merged into main, since it is skipped on PRs (thus the increasingly long line of PRs trying to fix the same thing).

> I saw that some of the check documentation links failed. Is that expected?

Some yes, others no. It's a mix of private links that always fail, some which have actually broken that we need to fix, and some that are broken but are in deprecated parts of our docs. Another PR could look through them in the future, but no users have complained about the docs so we can safely ignore for the time being.

Thanks for the quick review!

Issue #2605, opened 2024-02-06
Removed kinetic group additivity methods are still referenced in codebase


jonwzheng:
### Bug Description

A commit from last August (https://github.com/ReactionMechanismGenerator/RMG-Py/commit/9f293bec90a093cf23b42089bb7bb0b14e05a2a7) removed various deprecated code, including functions in `rmgpy/data/kinetics/groups` and `rmgpy/data/kinetics/family` related to computing rate coefficients using group additivity methods. Specifically, the method named `estimate_kinetics_using_group_additivity` in both of those files.

However, references to these methods are still available in the codebase. For example, in `rmgpy/data/kinetics/family`: `get_kinetics_for_template` includes the option to use group additivity methods. This template method is called in `get_kinetics` and uses group additivity if `kinetics_estimator` is not provided, or if `group_additivity` is provided as the estimator.

In `rmgpy/rmg/main.py`, the docstring mentions that `group_additivity` is a valid kinetic estimator method even though this does not seem to be the case anymore. And `rmgpy/rmg/model.py` considers the case if `kinetics_estimator` is group additivity.

### Expected Behavior
Code should run without calling a removed method that leads to an AttributeError.

### Additional Context
We could either revert these changes, or we could commit to removing all references to group additivity. Previous discussion (see https://github.com/ReactionMechanismGenerator/RMG-Py/issues/975#issuecomment-400842324 and codebase pre-commit) indicated that group additivity should be deprecated for kinetics, however it has been available and a default method on the RMG Website. It might be helpful to figure out if the group additivity methods are so bad that we should just remove them altogether, or if they might be potentially useful for users on the website.

JacksonBurns:
I think Bill would want these to stay - I'm pretty sure he had us use RMG website to do group additivity estimates during his kinetics class.

jonwzheng:
Addressed by #2612

PR #2612, opened 2024-02-14
Remove kinetic group additivity methods from code and documentation

jonwzheng:
### Motivation or Problem
See #2605. We have decided to remove group additivity as a valid estimator for kinetics. Now, only rate rules will be supported moving forward.
A previous commit (9f293bec90a093cf23b42089bb7bb0b14e05a2a7) removed the main group additivity estimator methods. However, there were still a few references to kinetic group additivity, including as the default method for `get_kinetics` if `kinetics_estimator` is not provided. There were also some mentions in the documentation and docstrings that suggest group additivity is supported.

### Description of Changes
This PR removes most mentions of kinetic group additivity from the code, documentation, and docstrings. 

However, this commit does _not_ remove the `KineticsGroup` class (see `rmgpy/data/kinetics/groups.py`), which is used extensively in the definition of a `KineticsFamily`. I would welcome some discussion on whether we should remove those and rewrite the kinetics family code as well, or if we want to keep it for compatibility / future reference/use etc.

### Testing
A local version of RMG-website using this commit hash seems to return the same kinetics as is currently returned on RMG-Py.


JacksonBurns:
Re-running CI in the hope that the failures were spurious Julia dependency stuff.

JacksonBurns:
Branch was outdated - rebased.

JacksonBurns:
@jonwzheng I think there is just a simple formatting error in this table: https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/7932965819/job/21660635035?pr=2612#step:9:335

jonwzheng:
> @jonwzheng I think there is just a simple formatting error in this table: https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/7932965819/job/21660635035?pr=2612#step:9:335

fixed this in the latest commit!

rmgpy/data/kinetics/family.py
```diff
@@ -2396,23 +2396,20 @@ def get_reaction_template(self, reaction):
...
-        RMG-Java behavior, and default RMG-Py behavior).
+        `template` and reaction-path `degeneracy`. There is currently only one method to use:
+        'rate rules' (old RMG-Java behavior, and default RMG-Py behavior). Group additivity was removed in August 2023.
         
         Returns a tuple (kinetics, entry):
         If it's estimated via 'rate rules' and an exact match is found in the tree,
         then the entry is returned as the second element of the tuple.
-        But if an average is used, or the 'group additivity' method, then the tuple
-        returned is (kinetics, None).
+        But if an average is used,then the tuple returned is (kinetics, None).
```
JacksonBurns, line 2405: Sorry I didn't catch this before - missing a space between the comma and `then`

JacksonBurns (approved):
Thanks for the patch! Enabling auto-merge

JacksonBurns:
@jonwzheng could you check the live docs pages online and see if the changes here are reflected there?

jonwzheng:
The live docs don't show this yet; see [this](https://reactionmechanismgenerator.github.io/RMG-Py/reference/data/templatereaction.html?highlight=rate%20rules) and [this](https://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/kinetics.html?highlight=rate+rules).

JacksonBurns:
Ugh, the latest doc building action fixes still aren't working. @jonwzheng could you put another pair of eyes on this workflow file and tell me if you see what is wrong? https://github.com/ReactionMechanismGenerator/RMG-Py/blob/main/.github/workflows/docs.yml

Here is the workflow run from merging this PR, it seems like the building worked but then the push failed?? https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/7935329144/job/21668223608

jonwzheng:
I'm looking into this and it looks like documentation build workflow has had this issue as early as last week: see [this for example](https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/7876110961/job/21489442930).

Yeah, it looks like it's something specifically with the push step even though the build works. I'm not too familiar with how the workflow automation works, so I don't see anything that strikes me as particularly wrong.
The error `nothing to commit, working tree clean` is really weird. Is there any chance that the `/build/html` in the `Publish documentation` action is different from the `build/html/` compiled during `Make documentation - to publish`? I'm not sure how else to explain that there's nothing to commit unless we're operating in the original (rather than re-created) directory.

rwest:
I think there could be a difference between `documentation/build/html` and `build/html`  ? 
Seem to be doing the git actions (checking out, committing, etc.) in the latter, but building the documentation in the former?
Could add some path printing stuff to debug, or could try something like https://github.com/mxschmitt/action-tmate to poke around and see what's happening.

Issue #2604, opened 2024-02-06
Issue with species_list: NoneType' object is not iterable (while generating collision_rate_violator log)


PoojaNem:
Hello Everyone,

**Problem Description:**

I am trying to generate a kinetic model for n-propylcyclohexane oxidation in RMG. However, on running every time RMG throws a type error that says: "'**NoneType' object is not iterable**."  
<img width="359" alt="image" src="https://github.com/ReactionMechanismGenerator/RMG-Py/assets/131012151/4ef1ebb4-c3e3-4b79-a8f1-8847e85dfdb0">

There are a few points to notice here:
1. Error happens after adding **JetSurf2.0** thermo library to the input file (Current simulation utilizes JetSurf2.0_Modified, with thermodynamic data for species C5H5 from AramcoMech2.0).
2. The issue only pops up at the simulation's end while generating the **collision_rate_violators log**.
3. RMG successfully generates Cantera and Chemkin files that can be used for further processing.
4. Following the error traceback shows that in **line 244 of the reaction.py file**, the **species_list** argument (to the function **write_kinetic_entry** inside function **to_chemkin**) is an **empty list** which consequently results in an error while processing **line 1890 (for species in species_list:) in chemkin.pyx** file. 
5. Based on the definition of the function **to_chemkin** (provided in the file reaction.py line 234), I also tried setting **kinetics to False**. But, the error remained the same.  
6. Related to #2096. 

The zip folder containing an input file and database for reproducing a model is attached below. 

**Installation Information:**
 - OS: Windows 11, WSL2, Ubuntu 18.04 
 - Installation method: Installation by Source Using Anaconda Environment for Unix-based Systems: Linux and Mac OSX
 - RMG version information: 
   - RMG-Py: Latest version
   - RMG-database: Latest database

[ZipFolder.zip](https://github.com/ReactionMechanismGenerator/RMG-Py/files/14177953/ZipFolder.zip)

PoojaNem:
Hello Everyone,

[...]

PR #2614, opened 2024-02-26
Modified main.py script to include TROE reactions in the violators log

PoojaNem:
### Motivation or Problem
Interruption in simulation during the final model check. 

### Description of Changes
Added Troe kinetic (along with thirdbody) to the main.py file of the RMG. 
image
The present RMG code checks if violator[0].kinetics is the ThirdBody and, if True, enters into the if condition where self.reaction_model.core.species (species_list) is supplied as an argument to the function to_chemkin. For all the other types of kinetics, the else loop is followed, which does not have any argument. However, Troe kinetics also requires species_list as an argument to identify collider species, which is considered None as per the current code. Therefore, adding Troe to the if statement is essential to include reactions with Troe kinetics in the violators' log.
For more details, refer to the RMG issue #2604.

### Testing
The model generation was completed successfully without any interruption.

### DIFF
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -65,6 +65,7 @@ from rmgpy.exceptions import ForbiddenStructureException, DatabaseError, CoreErr
 from rmgpy.kinetics.diffusionLimited import diffusion_limiter
 from rmgpy.data.vaporLiquidMassTransfer import vapor_liquid_mass_transfer
 from rmgpy.kinetics import ThirdBody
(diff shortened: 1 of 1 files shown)


JacksonBurns:
@PoojaNem thank you for submitting this Pull Request! This change looks excellent. I will review now and approve. After the automated tests are completed, this should be merged into the `main` branch of RMG. We will do a formal release of RMG in the future, but in the meantime you can then use this code by updating your local installation to `main` or running the `latest` tag of the Docker image.

Issue #2070, opened 2021-02-11
Currently not working when DASPK is not available


mark-pfeifle:
There is currently a bug in rmgpy/solver/base.pyx: 

"    from pydas.daspk import DASSLError as DASxError"

in one of the first code lines. This should read 

"    from pydas.dassl import DASSLError as DASxError"

Otherwise, if daspk is not available, the program will still try to access pydas.daspk. 
This applies to the latest commit on master as of Feb 11 2021.

Best regards, 
Mark

mark-pfeifle:
I've been automatically notified that this is a stale issue now. As far as I can see, the bug is still present in the code.

JacksonBurns:
Hi @mark-pfeifle thanks for the followup! I am going to add the bug label to this so that it won't be bothered by the bot, and we will try and get toward resolving this.

Julian-Ufert:
I changed the respective code line. Please pull the changes into the upstream.

PR #2615, opened 2024-02-27
changed typo from pydas.daspk to pydas.dassl #2070

Julian-Ufert:
### Motivation or Problem
There was a typo in the rmgpy/solver/base.pyx described in issue #2070.

### Description of Changes
I fixed the typo such that the file accesses pydas.dassl if pydas.daspk is not available.

### Testing
No tests performed.


JacksonBurns (approved):
Hi @Julian-Ufert thank you for the fix! This looks correct to me.

I will set this to auto-merge after the continuous integration testing passes. I don't think the tests actually evaluate this statement, but the patch is correct by inspection and the user report in the linked issue is very clear.

PR #2619, opened 2024-02-28
Attempt to fix docs building (#2602 part 3)

jonwzheng:
See #2602, #2610, and #2612. Doc building has been having issues. 

This PR implements a few of debugging steps that should be **reverted** before merging in: 

1. Swaps the test and publish actions so that we can test using this PR.
2. Add `--dry-run to `git push` args so that any test changes from here don't get actively pushed.
3. `ls -l` and `pwd` during the suspect stages of the build stages.


JacksonBurns:
@jonwzheng the build logs showed that the docs were indeed being updated, and were even in the correct place. I think this may have to do with the `checkout` command not cloning with the entire history of the branch, so the copy that the runner has does not know that it is ahead.

jonwzheng:
That's weird. After the `make` step, files in `/build/html` are timestamped at 17:16. But in the `build` step they're timestamped `17:14`. So the files are different even though the directory _appears_ to be the same...
Sorry it's a bit hard to see because the sphinx build was super verbose. 
Could it be that it's somehow running `git checkout` and stashing or reverting the changes? Or is it deleting the files or something?

JacksonBurns:
You are correct, the checkout option is running clean and reset before cloning, see https://github.com/actions/checkout (Search for clean) - editing now.

JacksonBurns:
Maybe build/html cannot see the correct upstream branch?

JacksonBurns:
The good news is that this is apparently known to be hugely annoying: https://github.com/actions/checkout/issues/578

JacksonBurns:
Note if the above does not work - might need to also checkout the complete history of the whole repo as well, i.e. fetch-depth: 0 when we checkout RMG-Py too?

JacksonBurns:
Moving the conclusions drawn from this investigation into https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2621 and closing this, leaving it as an artifact for future archaeologists and/or ChatGPT.

PR #2620, opened 2024-03-01
Add libxrender1 as a system-level dependency in Unix-based installation instructions

ntietje1:
### Motivation or Problem
The "Installation by Source Using Anaconda Environment for Unix-based Systems" page does not mention the system-level dependency of libxrender1, which is not installed by default on systems not intended to render images such as Windows Subsystem for Linux.

### Description of Changes
Added a sentence stating that libxrender1 is a system-level dependency next to the other listed system level dependencies. Also modified the installation commands to include libxrender1. 

### Testing
N/A

### Reviewer Tips
I'm not sure if modifying the installation commands is necessary, and it may be more concise/sufficient to just have the single sentence noting the dependency. However, for future WSL users, mentioning this dependency would save hours of debugging as it is not immediately obvious that libxrender1 is the cause of the issues when it's missing.

### DIFF
--- a/documentation/source/users/rmg/installation/anacondaDeveloper.rst
+++ b/documentation/source/users/rmg/installation/anacondaDeveloper.rst
@@ -26,7 +26,7 @@ Installation by Source Using Anaconda Environment for Unix-based Systems: Linux 
       Note that you should reinitialize or restart your terminal in order for the changes to take effect, as the installer will tell you.
 
 #. There are a few system-level dependencies which are required and should not be installed via Conda. These include
-   `Git <https://git-scm.com/>`_ for version control, `GNU Make <https://www.gnu.org/software/make/>`_, and the C and C++ compilers from the `GNU Compiler Collection (GCC) <https://gcc.gnu.org/>`_ for compiling RMG.
+   `Git <https://git-scm.com/>`_ for version control, `GNU Make <https://www.gnu.org/software/make/>`_, and the C and C++ compilers from the `GNU Compiler Collection (GCC) <https://gcc.gnu.org/>`_ for compiling RMG. For WSL users, `libxrender1 <https://packages.debian.org/sid/libxrender1>`_ may also need to be installed.
 
    For Linux users, you can check whether these are already installed by simply calling them via the command line, which
    will let you know if they are missing. To install any missing packages, you should use the appropriate package manager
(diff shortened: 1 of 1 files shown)


rwest (commented):
This looks like an improvement. Thanks.

Have the Fedora, openSUSE etc. been tested, or did you just guess them?
eg. on Fedora perhaps it's called [libXrender](https://packages.fedoraproject.org/pkgs/libXrender/libXrender/) instead of libxrender1 ? (though I only did a quick search).

I notice we have a whole page of instructions for linuxSubsystem
https://github.com/ReactionMechanismGenerator/RMG-Py/blob/main/documentation/source/users/rmg/installation/linuxSubsystem.rst
and that includes:
> `sudo apt install gcc g++ make graphviz`
Note: installing graphviz as a system package makes sure that rendering libraries that are not included in the WSL version of Ubuntu are installed. Unfortunately, the graphviz conda package does not install these dependencies in the RMG conda environment (later in the installation), instead relying on system libraries even in native Linux installations.

Does this take care of it?

Perhaps we should instead update and refresh that page, and direct people towards it if they're using WSL ?

JacksonBurns:
@rwest installation via WSL is considered an "unsupported" method, see [this part of the docs](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/0419c44f52ef054e10d46cb317aea6f1e9fa19dd/documentation/source/users/rmg/installation/index.rst?plain=1#L86) - I think the scope of changes that @ntietje1 has here is probably what we want, rather than restarting first-party support for WSL.

ntietje1:
@rwest I did not test on other distributions. I'm not sure how to go about doing that either.

rwest:
I suggest just include the one we know works.

rwest:
It now looks like you edited every single line in the file. Did the line endings change?

ntietje1:
I just went through the install instructions on a different machine (still using WSL), and I didn't run into the same issue as above. I believe it's not worth changing the documentation for what was probably an issue on my part, so I'm closing this PR.

ntietje1:
Sorry for the close and reopen, when installing on another machine, everything seemed to be working with no missing dependencies even after running 'make', however attempting to actually run RMG resulted in an error related to missing libxrender.

documentation/source/users/rmg/installation/anacondaDeveloper.rst
```diff
@@ -26,15 +26,15 @@ Installation by Source Using Anaconda Environment for Unix-based Systems: Linux
...
+   `Git <https://git-scm.com/>`_ for version control, `GNU Make <https://www.gnu.org/software/make/>`_, and the C and C++ compilers from the `GNU Compiler Collection (GCC) <https://gcc.gnu.org/>`_ for compiling RMG. For WSL users, `libxrender1 <https://packages.debian.org/sid/libxrender1>`_ may also need to be installed.
 
    For Linux users, you can check whether these are already installed by simply calling them via the command line, which
    will let you know if they are missing. To install any missing packages, you should use the appropriate package manager
    for your system.
 
    a. On Ubuntu and Debian the package manager is ``apt`` ::
 
-       sudo apt install git gcc g++ make
+       sudo apt install git gcc g++ make libxrender1
```
JacksonBurns, line 37: I think this is still worth getting into the docs!

ntietje1: I cleaned up the commit history of this branch and updated it. Should be good to merge now?

JacksonBurns: This LGTM! I'll approve, and let's get a review from Richard as well.

JacksonBurns (approved):

PR #2621, opened 2024-03-03
documentation builder bug fixes and general overhaul

JacksonBurns:
Following extensive investigation in https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2619 this PR finally fixes the documentation build and modernizes the workflow a bit:
 - runs the workflow on a schedule so that we can catch future errors once a week instead of whenever we update the docs
 - fixes the incorrect installation order by copying from the CI setup steps
 - fixes the path error in the docs building/committing
 - reorders the steps somewhat to improve clarity of the actual doc build

This PR will be successful if we get all the way to the last step in the docs build and it is skipped, and if in the "Build Documentation" step all of the changes are shown in the output from the `git add` command.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -5,17 +5,27 @@ on:
     branches:
       - main
   pull_request:
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    - cron: "0 8 * * 1"
 
 jobs:
   build-documentation:
     runs-on: ubuntu-latest
-    strategy:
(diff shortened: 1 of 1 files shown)


jonwzheng (approved):
LGTM. Thanks for figuring this out!

jonwzheng:
There are some commits in #2619 on the `fix_docs_attempt` branch. Were those intentional?
If not, we should remember to delete that branch once this is merged in.

Issue #2611, opened 2024-02-13
Change CI to use download-artifact@v4


JacksonBurns:
The GitHub official upload/download action now supports retrieving from previous runs and cross-repository downloads (for RMG-database), would be good to switch the CI to use this instead of the third party action we use now.

JacksonBurns:
Also see: https://github.blog/2024-02-12-get-started-with-v4-of-github-actions-artifacts/

PR #2628, opened 2024-03-07
GitHub Actions Maintenance

JacksonBurns:
#2621 fixed the docs building but was not tested from forks or with PRs that do not update the docs, and in both cases the action now _looks_ like it fails even though it's fine. This PR removes the test push (which failed from forks) and allow the commit step to fail, which it would do if the PR does not contain docs updates.

#2512 and #2608 left some lint in the CI which this PR removes.

Blocked by #2553 since I accidentally made this branch locally off of that one and am too lazy to rebase + this isn't urgent.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -81,19 +81,13 @@ jobs:
           cd build/html
           touch .nojekyll
           git add -A --force .
-          git commit -m "Automated documentation rebuild"
+          git commit -m "Automated documentation rebuild" || true  # allow this to fail if no changes were made
 
       - name: Check documentation links
         continue-on-error: true
         run: |
           cd documentation
           sphinx-build -b linkcheck -d build/doctrees/ source/ build/linkcheck | grep -e broken -e redirect | grep -v -e 'redirect  https://doi.org/' -e 'broken    https://doi.org/.* 403 Client Error: Forbidden'
-
-      - name: Test Publish Documentation
-        if: ${{  github.event_name != 'push' && github.repository == 'ReactionMechanismGenerator/RMG-Py' }}
-        run: |
-          cd documentation/build/html
-          git push --dry-run origin gh-pages
 
       - name: Publish Updated Documentation
         if: ${{  github.event_name == 'push' && github.repository == 'ReactionMechanismGenerator/RMG-Py' }}
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -81,7 +81,7 @@ jobs:
           cd build/html
           touch .nojekyll
           git add -A --force .
-          git commit -m "Automated documentation rebuild" || true  # allow this to fail if no changes were made
+          git diff --exit-code --quiet || git commit -m "Automated documentation rebuild"  # only commit if changes were made
 
       - name: Check documentation links
         continue-on-error: true
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -81,7 +81,7 @@ jobs:
           cd build/html
           touch .nojekyll
           git add -A --force .
-          git diff --exit-code --quiet || git commit -m "Automated documentation rebuild"  # only commit if changes were made
+          git diff --staged --exit-code --quiet || git commit -m "Automated documentation rebuild"  # only commit if changes were made
 
       - name: Check documentation links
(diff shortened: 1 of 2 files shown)


JacksonBurns:
This issue also resolves #2611

JacksonBurns:
Nevermind, misread my own issue. That issue is updating the download-artifacts step to v4, this PR just update checkout to v4

test/rmgpy/molecule/translatorTest.py
```diff
@@ -238,7 +239,7 @@ def test_ch2o2(self):
 3 O 1 {1,S}
 """
 
-        aug_inchi = "InChI=1/CH2O2/c2-1-3/h1H,(H,2,3)/u1,2"
```
rwest, line 241: I think I'm not currently using InChIs for anything, but isn't an InChI changing kind of a big deal? Thought the whole point was they persisted!

JacksonBurns: This change is actually caused by #2553 - could you copy this over there and tag @xiaoruiDong about this?

rwest: I re-read the inchi discussion in that thread from October 2023, and don't want to follow up. Sounds like RDKit changed something but now match OpenBabel, so I'm gonna assume the change is a bug fix and let it go :)

.github/workflows/docs.yml
```diff
@@ -81,20 +81,14 @@ jobs:
           cd build/html
           touch .nojekyll
           git add -A --force .
-          git commit -m "Automated documentation rebuild"
+          git commit -m "Automated documentation rebuild" || true  # allow this to fail if no changes were made
```
jonwzheng, line 84: I see that this will resolve the issue of if there are no changes to commit, but what if there is an issue with the `git commit` step that's due to some separate issue? I think this is how we saw there was a problem with the documentation builder in the first place right?
Maybe this can be modified to something like `git commit -m "Automated documentation rebuild" || <some code that will return true if there are no changes to commit> `
Or before `git add -A --force .`, can check to see if there are any files to commit with something like `git diff --exit-code` or `git diff --quiet`.

.github/workflows/docs.yml
```diff
@@ -81,20 +81,14 @@ jobs:
...
 
       - name: Check documentation links
         continue-on-error: true
         run: |
           cd documentation
           sphinx-build -b linkcheck -d build/doctrees/ source/ build/linkcheck | grep -e broken -e redirect | grep -v -e 'redirect  https://doi.org/' -e 'broken    https://doi.org/.* 403 Client Error: Forbidden'
 
-      - name: Test Publish Documentation
```
jonwzheng, line 92: are you suggesting this as a temporary change to allow the CI workflow to pass for forks? it seems undesirable to eliminate the test publish step completely. 
based on discussion [here](https://github.com/orgs/community/discussions/25217) it looks like we could also filter this out for repos that are forks (i.e. execute only `if: "! github.event.pull_request.head.repo.fork`)

.github/workflows/docs.yml
JacksonBurns: This is a good point - let me add this now.

.github/workflows/docs.yml
JacksonBurns: I suggest this as a permanent change. I can't think of many cases (besides deleting `gh-pages` branch or revoking actions push permission) that could cause this to fail (since by this point we have already committed without problems). I don't think we need to worry about it.

jonwzheng: ohh i see, this is just for for the `git push` step. In my initial comment I thought this was for building the docs but now I remember we changed in a recent PR.

.github/workflows/docs.yml
```diff
@@ -81,20 +81,14 @@ jobs:
           cd build/html
           touch .nojekyll
           git add -A --force .
-          git commit -m "Automated documentation rebuild"
+          git diff --exit-code --quiet || git commit -m "Automated documentation rebuild"  # only commit if changes were made
```
jonwzheng, line 84: It seems like if a new file is **added** rather than modified, we need to add the `--staged` flag to `git diff`

jonwzheng (approved):
LGTM

PR #2631, opened 2024-03-11
Avoid Unnecessary Julia Operations + Optional RMS Installation

### LINKED ISSUE #2547: Arkane takes a long time to run due to loading Julia dependencies
### Bug Description
Running `Arkane.py` takes much longer than expected (i.e. close to 1 hour rather than 1 second on a local machine) for simple TST calculations.

### How To Reproduce
Go to `RMG-Py/examples/arkane/reactions/` and run `Arkane.py` on any of the `input.py` files in any of the examples.
You will see that Arkane starts executing, but slows down as it starts processing the Julia imports.

### Expected Behavior
Arkane should finish running in just a few seconds.

### Installation Information
- Ubuntu 20.04.6 LTS
- install from source with Anaconda,  RMG-Py tag `3.2.0-115-g87b96e1b0`, RMG-database tag `3.2.0`

### Additional Context
The slowdown is from Arkane importing RMG's `CoreEdgeReactionModel` , which imports RMG reactors, which then import the Julia modules (loading the Julia stuff is really slow.)

Since I'm only using Arkane for conventional TST, I was able to get the runtime back to the expected ~1 seconds by commenting out all of the instances where `CoreEdgeReactionModel` or `ExplorerJob` are imported when Arkane.py is called.

JacksonBurns:
Replaces part of #2539 which 1. fixed the conda recipe 2. added a subpackage 3. avoided unnecesary RMS operations and 4. grew totally un-reviewable. This PR contains only 3. - later PRs will implement 1. and 2.

## Description of Changes
Two new internal-only boolean variables are added, `NO_JULIA` and `requires_rms`:
 - `NO_JULIA` is set based on whether or not the Julia dependencies (which are now optional) are installed. This resolves #2547 by allowing for a standalone installation of RMG or Arkane.
 - `requires_rms` is set if the user calling RMG actually has a reactor system that requires an RMS-based reactor. This avoids a lot of work (see which lines can now be skipped in this PR) when running RMG from the CLI inside the Docker container (or a source install with RMS).

I have also renamed the reactors file for clarity, to differentiate it from reactors.py which has the RMG-based pure-python reactors.

## Impact of Changes
As of right now, the CI and Docker image still install Julia and RMS as they do now, so the version of RMG that we send out has all the features RMS offers.
This change is intended for:
 - generally speeding up RMG
 - allowing people to extend RMG more easily
 - allow fixing the conda recipe (see the referenced PR at the top that necessitated this change, or wait for the follow up PR)

## TODO Before Merging
Update user-facing docs to provide details on how to do a julia-less installation and **indicate which features require Julia**.

I am opening and requesting review now so that we can hash out the finer points of the implementation, and then I can write up the docs for the users once that is settled.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,10 +28,6 @@ RUN wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh &
     rm Miniconda3-latest-Linux-x86_64.sh
 ENV PATH="$PATH:/miniconda/bin"
 
-# Set solver backend to mamba for speed
-RUN conda install -n base conda-libmamba-solver && \
-    conda config --set solver libmamba
-
 # Set Bash as the default shell for following commands
 SHELL ["/bin/bash", "-c"]
 
@@ -44,10 +40,7 @@ RUN git clone --single-branch --branch main --depth 1 https://github.com/Reactio
 
 WORKDIR /rmg/RMG-Py
 # build the conda environment
-RUN conda env create --file environment.yml && \
-    conda install -c conda-forge julia=1.9.1 pyjulia>=0.6 && \
-    conda install -c rmg pyrms diffeqpy && \
-    conda clean --all --yes
+RUN conda env create --file environment.yml
 
 # This runs all subsequent commands inside the rmg_env conda environment
 #
@@ -55,6 +48,10 @@ RUN conda env create --file environment.yml && \
 # since that requires running conda init and restarting the shell (not possible
 # in a Dockerfile build script)
 SHELL ["conda", "run", "--no-capture-output", "-n", "rmg_env", "/bin/bash", "-c"]
+
+RUN conda install -c conda-forge julia=1.9.1 pyjulia>=0.6 && \
+    conda install -c rmg pyrms diffeqpy && \
+    conda clean --all --yes
 
 # Set environment variables as directed in the RMG installation instructions
 ENV RUNNER_CWD=/rmg
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,6 +45,8 @@ RUN git clone --single-branch --branch main --depth 1 https://github.com/Reactio
 WORKDIR /rmg/RMG-Py
 # build the conda environment
 RUN conda env create --file environment.yml && \
+    conda install -c conda-forge julia=1.9.1 pyjulia>=0.6 && \
+    conda install -c rmg pyrms diffeqpy && \
     conda clean --all --yes
 
 # This runs all subsequent commands inside the rmg_env conda environment
--- a/environment.yml
+++ b/environment.yml
@@ -16,6 +16,7 @@
 #   made dependency list more explicit (@JacksonBurns).
 # - October 16, 2023 Switched RDKit and descripatastorus to conda-forge,
 #   moved diffeqpy to pip and (temporarily) removed chemprop
+# - Mar 11, 2024 Removed Julia dependencies, now considered optional
 #
 name: rmg_env
 channels:
@@ -47,10 +48,6 @@ dependencies:
   - conda-forge::cclib >=1.6.3,<1.8.0
   - conda-forge::openbabel >= 3
   - conda-forge::rdkit >=2022.09.1
-
-# general-purpose external software tools
-  - conda-forge::julia=1.9.1
-  - conda-forge::pyjulia >=0.6
 
 # Python tools
   - python >=3.7
@@ -88,20 +85,23 @@ dependencies:
   - rmg::pydas >=1.0.3
   - rmg::pydqed >=1.0.3
   - rmg::pyrdl
-  - rmg::pyrms
   - rmg::symmetry
-
-# packages we would like to stop maintaining (and why)
-  - rmg::diffeqpy
-  # we should use the official verison https://github.com/SciML/diffeqpy),
-  # rather than ours (which is only made so that we can get it from conda)
-  # It is only on pip, so we will need to do something like:
-  # https://stackoverflow.com/a/35245610
-  # Note that _some other_ dep. in this list requires diffeqpy in its recipe
-  # which will cause it to be downloaded from the rmg conda channel
 
 # conda mutex metapackage
   - nomkl
+
+# optional dependencies for using ReactionMechanismSimulator
+# remove the leading '#' to install the required dependencies
+  # - conda-forge::julia=1.9.1
+  # - conda-forge::pyjulia >=0.6
+  # - rmg::pyrms
+  # - rmg::diffeqpy
+
+# Note about diffeqpy:
+# we should use the official verison https://github.com/SciML/diffeqpy),
+# rather than ours (which is only made so that we can get it from conda)
+# It is only on pip, so we will need to do something like:
+# https://stackoverflow.com/a/35245610
 
 # additional packages that are required, but not specified here (and why)
   # pydqed, pydas, mopac, and likely others require a fortran compiler (specifically gfortran)
--- a/rmgpy/pdep/sls.py
+++ b/rmgpy/pdep/sls.py
@@ -50,7 +50,7 @@ try:
     from diffeqpy import de
     from julia import Main
 except Exception as e:
-    warnings.warn(
+    logging.info(
         f"Unable to import Julia dependencies, original error: {str(e)}"
         ". Master equation method 'ode' will not be available on this execution."
     )
@@ -163,7 +163,7 @@ def get_rate_coefficients_SLS(network, T, P, method="mexp", neglect_high_energy_
         if NO_JULIA:
             raise RuntimeError(
                 "Required Julia dependencies for method 'ode' are not installed.\n"
-                "Please check your installation (https://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/installation/index.html)."
+                "Please follow the steps to install Julia dependencies at https://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/installation/anacondaDeveloper.html."
             )
         f = Main.eval(
             """
--- a/rmgpy/pdep/sls.py
+++ b/rmgpy/pdep/sls.py
@@ -32,9 +32,8 @@ Contains functionality for directly simulating the master equation
 and implementing the SLS master equation reduction method
 """
 import sys
+import warnings
 
-from diffeqpy import de
-from julia import Main
 import scipy.sparse as sparse
 import numpy as np
 import scipy.linalg
@@ -45,6 +44,17 @@ import rmgpy.constants as constants
 from rmgpy.pdep.reaction import calculate_microcanonical_rate_coefficient
 from rmgpy.pdep.me import generate_full_me_matrix, states_to_configurations
 from rmgpy.statmech.translation import IdealGasTranslation
+
+NO_JULIA = False
+try:
+    from diffeqpy import de
+    from julia import Main
+except Exception as e:
+    warnings.warn(
+        f"Unable to import Julia dependencies, original error: {str(e)}"
+        ". Master equation method 'ode' will not be available on this execution."
+    )
+    NO_JULIA = True
 
 
 def get_initial_condition(network, x0, indices):
@@ -150,6 +160,11 @@ def get_rate_coefficients_SLS(network, T, P, method="mexp", neglect_high_energy_
     tau = np.abs(1.0 / fastest_reaction)
 
     if method == "ode":
+        if NO_JULIA:
+            raise RuntimeError(
+                "Required Julia dependencies for method 'ode' are not installed.\n"
+                "Please check your installation (https://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/installation/index.html)."
+            )
         f = Main.eval(
             """
 function f(u,M,t)
--- a/rmgpy/pdep/sls.py
+++ b/rmgpy/pdep/sls.py
@@ -32,7 +32,7 @@ Contains functionality for directly simulating the master equation
 and implementing the SLS master equation reduction method
 """
 import sys
-import warnings
+import logging
 
 import scipy.sparse as sparse
 import numpy as np
--- a/rmgpy/rmg/input.py
+++ b/rmgpy/rmg/input.py
@@ -1562,7 +1562,7 @@ def read_input_file(path, rmg0):
         if NO_JULIA:
             logging.error(
                 "During runtime, import of Julia dependencies failed. To use phase systems and RMS reactors, install RMG-Py with RMS."
-                " (https://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/installation/index.html)"
+                " (https://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/installation/anacondaDeveloper.html)"
             )
         logging.exception(e)
         raise
--- a/rmgpy/rmg/input.py
+++ b/rmgpy/rmg/input.py
@@ -48,9 +48,10 @@ from rmgpy.solver.simple import SimpleReactor
 from rmgpy.solver.surface import SurfaceReactor
 from rmgpy.util import as_list
 from rmgpy.data.surface import MetalDatabase
-from rmgpy.rmg.reactors import Reactor, ConstantVIdealGasReactor, ConstantTLiquidSurfaceReactor, ConstantTVLiquidReactor, ConstantTPIdealGasReactor
 from rmgpy.data.vaporLiquidMassTransfer import liquidVolumetricMassTransferCoefficientPowerLaw
 from rmgpy.molecule.fragment import Fragment
+from rmgpy.rmg.reactionmechanismsimulator_reactors import Reactor, ConstantVIdealGasReactor, ConstantTLiquidSurfaceReactor, ConstantTVLiquidReactor, ConstantTPIdealGasReactor, NO_JULIA
+ 
 
 ################################################################################
 
@@ -1558,6 +1559,8 @@ def read_input_file(path, rmg0):
         exec(f.read(), global_context, local_context)
     except (NameError, TypeError, SyntaxError) as e:
         logging.error('The input file "{0}" was invalid:'.format(full_path))
+        if NO_JULIA:
+            logging.error("During runtime, import of Julia dependencies failed. To use phase systems and RMS reactors, install RMG-Py with RMS.")
         logging.exception(e)
         raise
     finally:
--- a/rmgpy/rmg/input.py
+++ b/rmgpy/rmg/input.py
@@ -1560,7 +1560,10 @@ def read_input_file(path, rmg0):
     except (NameError, TypeError, SyntaxError) as e:
         logging.error('The input file "{0}" was invalid:'.format(full_path))
         if NO_JULIA:
-            logging.error("During runtime, import of Julia dependencies failed. To use phase systems and RMS reactors, install RMG-Py with RMS.")
+            logging.error(
+                "During runtime, import of Julia dependencies failed. To use phase systems and RMS reactors, install RMG-Py with RMS."
+                " (https://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/installation/index.html)"
+            )
         logging.exception(e)
         raise
     finally:
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -81,7 +81,6 @@ from rmgpy.tools.plot import plot_sensitivity
 from rmgpy.tools.uncertainty import Uncertainty, process_local_results
 from rmgpy.yml import RMSWriter
 from rmgpy.rmg.reactionmechanismsimulator_reactors import Reactor as RMSReactor
-from rmgpy.rmg.reactionmechanismsimulator_reactors import NO_JULIA
 
 ################################################################################
 
@@ -560,7 +559,7 @@ class RMG(util.Subject):
             self.reaction_model.add_species_to_edge(spec, requires_rms=requires_rms)
 
         for reaction_system in self.reaction_systems:
-            if not NO_JULIA and isinstance(reaction_system, RMSReactor):
+            if isinstance(reaction_system, RMSReactor):
                 reaction_system.finish_termination_criteria()
 
         # Load restart seed mechanism (if specified)
@@ -711,7 +710,9 @@ class RMG(util.Subject):
         # advantages to write it here: this is run only once (as species indexes does not change over the generation)
         if self.solvent is not None:
             for index, reaction_system in enumerate(self.reaction_systems):
-                if ((NO_JULIA or not isinstance(reaction_system, RMSReactor)) and reaction_system.const_spc_names is not None):  # if no constant species provided do nothing
+                if (
+                    not isinstance(reaction_system, RMSReactor)
+                ) and reaction_system.const_spc_names is not None:  # if no constant species provided do nothing
                     reaction_system.get_const_spc_indices(self.reaction_model.core.species)  # call the function to identify indices in the solver
 
         self.initialize_reaction_threshold_and_react_flags()
@@ -729,7 +730,7 @@ class RMG(util.Subject):
         """
 
         self.attach(ChemkinWriter(self.output_directory))
-        if not NO_JULIA and requires_rms:
+        if requires_rms:
             self.attach(RMSWriter(self.output_directory))
 
         if self.generate_output_html:
@@ -742,7 +743,7 @@ class RMG(util.Subject):
 
         if self.save_simulation_profiles:
             for index, reaction_system in enumerate(self.reaction_systems):
-                if not NO_JULIA and requires_rms and isinstance(reaction_system, RMSReactor):
+                if requires_rms and isinstance(reaction_system, RMSReactor):
                     typ = type(reaction_system)
                     raise InputError(f"save_simulation_profiles=True not compatible with reactor of type {typ}")
                 reaction_system.attach(SimulationProfileWriter(self.output_directory, index, self.reaction_model.core.species))
@@ -786,7 +787,7 @@ class RMG(util.Subject):
             # Update react flags
             if self.filter_reactions:
                 # Run the reaction system to update threshold and react flags
-                if not NO_JULIA and requires_rms and isinstance(reaction_system, RMSReactor):
+                if requires_rms and isinstance(reaction_system, RMSReactor):
                     self.update_reaction_threshold_and_react_flags(
                         rxn_sys_unimol_threshold=np.zeros((len(self.reaction_model.core.species),), bool),
                         rxn_sys_bimol_threshold=np.zeros((len(self.reaction_model.core.species), len(self.reaction_model.core.species)), bool),
@@ -908,7 +909,7 @@ class RMG(util.Subject):
                             prune = False
 
                         try:
-                            if not NO_JULIA and requires_rms and isinstance(reaction_system, RMSReactor):
+                            if requires_rms and isinstance(reaction_system, RMSReactor):
                                 (
                                     terminated,
                                     resurrected,
@@ -1010,7 +1011,7 @@ class RMG(util.Subject):
                             temp_model_settings.tol_keep_in_edge = 0
                             if not resurrected:
                                 try:
-                                    if not NO_JULIA and requires_rms and isinstance(reaction_system, RMSReactor):
+                                    if requires_rms and isinstance(reaction_system, RMSReactor):
                                         (
                                             terminated,
                                             resurrected,
@@ -2229,7 +2230,7 @@ class RMG_Memory(object):
                 if isinstance(value, list):
                     self.Ranges[key] = [v.value_si for v in value]
 
-        if not NO_JULIA and isinstance(reaction_system, RMSReactor):
+        if isinstance(reaction_system, RMSReactor):
             self.tmax = reaction_system.tf
         else:
             for term in reaction_system.termination:
--- a/rmgpy/rmg/main.py
+++ b/rmgpy/rmg/main.py
@@ -80,7 +80,8 @@ from rmgpy.thermo.thermoengine import submit
 from rmgpy.tools.plot import plot_sensitivity
 from rmgpy.tools.uncertainty import Uncertainty, process_local_results
 from rmgpy.yml import RMSWriter
-from rmgpy.rmg.reactors import Reactor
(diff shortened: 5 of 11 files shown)


JacksonBurns:
Also TODO: use `pytest.mark` to label tests with those that require Julia to allow for running `pytest` without the Julia stuff.

mjohnson541:
I'm all in favor of avoiding the julia start up cost for runs that don't need it. Which should solve issue #2547. Although, as I noted in that issue Arkane is dependent on pyjulia entirely independently of RMS. 

However, RMS is not an optional install or anything you can treat as an optional backend. We have published RMG features that are not available without RMS, RMS is the only sensible way right now to add a new reactor to RMG, RMG electrochemistry is entirely dependent on RMS because none of the RMG reactors can run it. 

Doing so does not save anyone time maintaining the installation because the installation with RMS has to work regardless. The initial install time for RMS is not particularly burdensome. It will even make debugging PRs more complicated because there will be more differences between users' local copies and the tests.

hwpang (commented):
Thanks for the PR! I have some questions and some comments

Dockerfile
```diff
@@ -44,8 +40,7 @@ RUN git clone --single-branch --branch main --depth 1 https://github.com/Reactio
 
 WORKDIR /rmg/RMG-Py
 # build the conda environment
-RUN conda env create --file environment.yml && \
-    conda clean --all --yes
```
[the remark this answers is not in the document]
JacksonBurns: This was added [here](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2437/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R48) in #2437 to reduce the size of the finished `docker` image. As per the [conda docs](https://conda.io/projects/conda/en/latest/commands/clean.html#removal-targets) it just removes some junk files we don't need.

rmgpy/rmg/input.py
```diff
@@ -1558,6 +1559,8 @@ def read_input_file(path, rmg0):
         exec(f.read(), global_context, local_context)
     except (NameError, TypeError, SyntaxError) as e:
         logging.error('The input file "{0}" was invalid:'.format(full_path))
+        if NO_JULIA:
+            logging.error("During runtime, import of Julia dependencies failed. To use phase systems and RMS reactors, install RMG-Py with RMS.")
```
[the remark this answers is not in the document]
JacksonBurns: Good call! I will add this to my TODO comment

environment.yml
```diff
@@ -85,16 +82,9 @@ dependencies:
   - rmg::pydas >=1.0.3
   - rmg::pydqed >=1.0.3
   - rmg::pyrdl
-  - rmg::pyrms
   - rmg::symmetry
 
 # packages we would like to stop maintaining (and why)
-  - rmg::diffeqpy
```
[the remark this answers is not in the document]
JacksonBurns: Thanks for pointing this out! It's good that there are other methods already implemented as well, and informative error catch will go nicely here. A note for when this is added: we should probably only code the 'hey your julia stuff is missing' error string in one place and then import it elsewhere.

JacksonBurns:
> I'm all in favor of avoiding the julia start up cost for runs that don't need it. Which should solve issue #2547. Although, as I noted in that issue Arkane is dependent on pyjulia entirely independently of RMS.

Thanks for pointing that out! I have added a comment on that issue.

> However, RMS is not an optional install or anything you can treat as an optional backend. We have published RMG features that are not available without RMS, RMS is the only sensible way right now to add a new reactor to RMG, RMG electrochemistry is entirely dependent on RMS because none of the RMG reactors can run it.

I see this as less about removing features and more about increasingly the 'extend-ability' of RMG.
The RMS features are great, yet some of our users only want to run basic reactors and 'plug in' RMG to other tools, which is difficult in the current state.

Our primary distribution method (Docker) will still include all of the RMS stuff, so by default everyone will still have access to every feature we have ever promised.
The CI, too, will continue to install the RMS features and ensure forward-compatibility with the inter-operation of the two packages.

I'll also add that this is somewhat being forced as a technical issue, or at least an issue with my technical understanding.
I only alluded to this in my initial comment ("allow fixing the conda recipe") but I'll be more explicit here.
In the process of attempting to fix the conda recipe, I did not find a way to build a new RMG binary with the RMS installation included. When attempting to build RMS within the conda recipe I couldn't get the build to 'find' RMG (so it went out to install the old binary), couldn't get the build to run (would fail for difficult-to-discern reasons), and overall the Julia language seems to only just be adopting conda as a binary distribution method (with 'new' packages like [Conda.jl](https://github.com/JuliaPy/Conda.jl)) picking up steam).

There may very well be a way to get the binary creation to work, but it will certainly be more difficult than the workflow that this change would allow:
`We Build Standalone RMG Binary -> Users Install Standalone RMG Binary -> Users Add RMS on Top of It Locally`

We have also historically 'promised' binaries, and I see this as the only way to provide those.

> Doing so does not save anyone time maintaining the installation because the installation with RMS has to work regardless. The initial install time for RMS is not particularly burdensome. It will even make debugging PRs more complicated because there will be more differences between users' local copies and the tests.

Partly replied to this above - CI will continue to install RMS so it will continue to work in the future.

Second point is a very real challenge, I agree. I can readily imagine circumstances where a user might suggest changes to RMG that pass RMG-only tests locally and then fail RMG+RMS CI on GitHub. This might ultimately be resolved with good practices (i.e. requiring people to run RMG+RMS for development) but I am open to input from the dev team on this.

mjohnson541:
> > I'm all in favor of avoiding the julia start up cost for runs that don't need it. Which should solve issue #2547. Although, as 
> > However, RMS is not an optional install or anything you can treat as an optional backend. We have published RMG features that are not available without RMS, RMS is the only sensible way right now to add a new reactor to RMG, RMG electrochemistry is entirely dependent on RMS because none of the RMG reactors can run it.
> 
> I see this as less about removing features and more about increasingly the 'extend-ability' of RMG. The RMS features are great, yet some of our users only want to run basic reactors and 'plug in' RMG to other tools, which is difficult in the current state.
> 
Once the julia imports are avoided running RMG without using julia features will not have a meaningful distinction between running RMG without RMS installed. So aside from installation, there won't be a meaningful distinction in difficulty for users. 
> Our primary distribution method (Docker) will still include all of the RMS stuff, so by default everyone will still have access to every feature we have ever promised. The CI, too, will continue to install the RMS features and ensure forward-compatibility with the inter-operation of the two packages.
> 
> I'll also add that this is somewhat being forced as a technical issue, or at least an issue with my technical understanding. I only alluded to this in my initial comment ("allow fixing the conda recipe") but I'll be more explicit here. In the process of attempting to fix the conda recipe, I did not find a way to build a new RMG binary with the RMS installation included. When attempting to build RMS within the conda recipe I couldn't get the build to 'find' RMG (so it went out to install the old binary), couldn't get the build to run (would fail for difficult-to-discern reasons), and overall the Julia language seems to only just be adopting conda as a binary distribution method (with 'new' packages like [Conda.jl](https://github.com/JuliaPy/Conda.jl)) picking up steam).
> 
I haven't seen the build script you devised, but this at least sounds resolvable to me. It sounds like RMS is being imported before julia is linked to the correct python (which will cause it to start installing things). Julia-python linking is most easily done inside the python you want to link to by calling `julia.install()` that should automatically identify julia and link it to the python you are in. I would think that should be done in pyjulia's build script, but for the RMG binary it can be done inside RMG's build script. The point of conda is to be universally applicable so we shouldn't be hindered by choice of language.

mjohnson541:
Oh wait, sorry...the issue isn't that...the issue is that the python that you're building with in conda is not the python in use when you run the binary. The linking needs done for the run python.

mjohnson541:
The best way to do that is probably to write values for "CONDA_JL_HOME" (path to the env) and "PYTHON" (path to the env python) into the file that loads variables when you enter the conda environment this should cause it to link properly automatically when julia starts up without any need to figure out someway to make julia.install() occur.

JacksonBurns:
> > I see this as less about removing features and more about increasingly the 'extend-ability' of RMG. The RMS features are great, yet some of our users only want to run basic reactors and 'plug in' RMG to other tools, which is difficult in the current state.
> 
> Once the julia imports are avoided running RMG without using julia features will not have a meaningful distinction between running RMG without RMS installed. So aside from installation, there won't be a meaningful distinction in difficulty for users.

My apologies, I'm not sure I understand what you mean, could you restate this another way?

> > Our primary distribution method (Docker) will still include all of the RMS stuff, so by default everyone will still have access to every feature we have ever promised. The CI, too, will continue to install the RMS features and ensure forward-compatibility with the inter-operation of the two packages.
> > I'll also add that this is somewhat being forced as a technical issue, or at least an issue with my technical understanding. I only alluded to this in my initial comment ("allow fixing the conda recipe") but I'll be more explicit here. In the process of attempting to fix the conda recipe, I did not find a way to build a new RMG binary with the RMS installation included. When attempting to build RMS within the conda recipe I couldn't get the build to 'find' RMG (so it went out to install the old binary), couldn't get the build to run (would fail for difficult-to-discern reasons), and overall the Julia language seems to only just be adopting conda as a binary distribution method (with 'new' packages like [Conda.jl](https://github.com/JuliaPy/Conda.jl)) picking up steam).
> 
> I haven't seen the build script you devised, but this at least sounds resolvable to me. It sounds like RMS is being imported before julia is linked to the correct python (which will cause it to start installing things). Julia-python linking is most easily done inside the python you want to link to by calling `julia.install()` that should automatically identify julia and link it to the python you are in. I would think that should be done in pyjulia's build script, but for the RMG binary it can be done inside RMG's build script. The point of conda is to be universally applicable so we shouldn't be hindered by choice of language.

The `build.sh` looked like this:
```
make install
export PYTHON=$PREFIX/bin/python
export PYTHONPATH=$SRC_DIR:$PYTHONPATH
python -c "import julia; julia.install(); import diffeqpy; diffeqpy.install()"
julia -e 'using Pkg; Pkg.add(PackageSpec(name="Functors",version="0.4.3")); Pkg.pin("Functors"); Pkg.add(PackageSpec(name="ReactionMechanismSimulator",rev="for_rmg")); using ReactionMechanismSimulator'
```
 the `conda_build_config.yaml` looked like this:
```
python:
  - 3.7
numpy:
  - 1.15
 ```
and the `meta.yaml` is here:
<details>
<summary>recipe</summary>

```
# For conda build
package:
  name: rmg
  version: 3.2.0

source:
  path: ../

build:
  number: {{ environ.get('GIT_DESCRIBE_NUMBER', 0) }}

requirements:
  build:
    - {{ compiler('c') }}
  host:
    - cairo
    - cairocffi
    - ffmpeg
    - xlrd
    - xlwt
    - h5py
    - graphviz
    - markupsafe
    - psutil
    - ncurses
    - suitesparse

  # external software tools for chemistry
    - coolprop
    - cantera=2.6
    - mopac
    - cclib >=1.6.3,!=1.8.0
    - openbabel >=3

  # Python tools
    - python >=3.7
    - coverage
    - cython >=0.25.2
    - scikit-learn
    - scipy
    - numpy >=1.10.0
    - pydot
    - jinja2
    - jupyter
    - pymongo
    - pyparsing
    - pyyaml
    - networkx
    - matplotlib >=1.5
    - mpmath
    - pandas

  # packages we maintain
    - rmgdatabase
    - gprof2dot
    - lpsolve55
    - muq2
    - numdifftools
    - pydas >=1.0.3
    - pydqed >=1.0.3
    - pyrdl
    - quantities
    - symmetry
    - chemprop==0.0.1
    - rdkit >=2020.03.3.0
  run:
    - cairo
    - cairocffi
    - ffmpeg
    - xlrd
    - xlwt
    - h5py
    - graphviz
    - markupsafe
    - psutil
    - ncurses
    - suitesparse

  # external software tools for chemistry
    - coolprop
    - cantera=2.6
    - mopac
    - cclib >=1.6.3,!=1.8.0
    - openbabel >=3

  # Python tools
    - python >=3.7
    - coverage
    - cython >=0.25.2
    - scikit-learn
    - scipy
    - numpy >=1.10.0
    - pydot
    - jinja2
    - jupyter
    - pymongo
    - pyparsing
    - pyyaml
    - networkx
    - matplotlib >=1.5
    - mpmath
    - pandas

  # packages we maintain
    - rmgdatabase
    - gprof2dot
    - lpsolve55
    - muq2
    - numdifftools
    - pydas >=1.0.3
    - pydqed >=1.0.3
    - pyrdl
    - quantities
    - symmetry
    - chemprop==0.0.1
    - rdkit >=2020.03.3.0
test:
  requires:
    - cairo
    - cairocffi
    - ffmpeg
    - xlrd
    - xlwt
    - h5py
    - graphviz
    - markupsafe
    - psutil
    - ncurses
    - suitesparse

  # external software tools for chemistry
    - coolprop
    - cantera=2.6
    - mopac
    - cclib >=1.6.3,!=1.8.0
    - openbabel >=3

  # Python tools
    - python >=3.7
    - coverage
    - cython >=0.25.2
    - scikit-learn
    - scipy
    - numpy >=1.10.0
    - pydot
    - jinja2
    - jupyter
    - pymongo
    - pyparsing
    - pyyaml
    - networkx
    - matplotlib >=1.5
    - mpmath
    - pandas

  # packages we maintain
    - rmgdatabase
    - gprof2dot
    - lpsolve55
    - muq2
    - numdifftools
    - pydas >=1.0.3
    - pydqed >=1.0.3
    - pyrdl
    - quantities
    - symmetry
    - chemprop==0.0.1
    - rdkit >=2020.03.3.0
  source_files:
    - 'examples/rmg/superminimal'
    - 'examples/arkane/networks/n-butanol'
  imports:
    - rmgpy
    - arkane
  commands:
    - rmg.py examples/rmg/superminimal/input.py
    - Arkane.py examples/arkane/networks/n-butanol/input.py

about:
  home: https://github.com/ReactionMechanismGenerator/RMG-Py
  license: MIT
  summary: "A program for automatically generating kinetic models of chemical reaction mechanisms."
```

</details>

though this last component would be a little bit simpler now that some of the deps have been removed.

I agree that this **should** be resolvable, but I'm also anxious about the maintainability of the results. Seems like it would be house of cards waiting to crash down, whereas a system that built an RMG binary independently and then added RMS on top of it would allow for greater flexibility in the two steps.

**Comment by mjohnson541** [comment]
Sorry, I felt like the reply stuff was getting too nested. 

On that point, I guess I don't understand why making RMS an optional install makes future development and plug-ability any easier given that after installation the user experience is the same for a user not using those features. 

Super interested about why you needed that Functors line in the build.  

Overall on the issues with the build my understanding is that if RMG starts using an RMS feature, RMS tries to install stuff?
My interpretation of this is
1) Python/pyjulia knows where Julia is
2) The Julia python finds has RMS installed
3) Julia/PyCall finds the wrong python and can't find the rmg stuff so tries to install it (probably using the default python)

So it seems that our problem is simply that while everything we need is installed PyCall doesn't know where the right python/conda env is. This can be set by defining environment variables: PYTHON and maybe also CONDA_JL_HOME (see https://github.com/JuliaPy/PyCall.jl and https://github.com/JuliaPy/Conda.jl). In the build we can set environmental variables that will automatically load when the environment loads so in theory if we do this for PYTHON and CONDA_JL_HOME RMS will grab the appropriate python and find that the packages it needs are already there. An example of doing this from the  build file I wrote for julia binaries is below: 

```
#make julia directory
mkdir -p ${PREFIX}/share/julia/site
mkdir -p ${PREFIX}/bin
#set JULIA_DEPOT_PATH in conda env
export JULIA_DEPOT_PATH="${PREFIX}/share/julia/site"
ACTIVATE_ENV="${PREFIX}/etc/conda/activate.d/env_vars.sh"
DEACTIVATE_ENV="${PREFIX}/etc/conda/deactivate.d/env_vars.sh"

if [ -f "$ACTIVATE_ENV" ]; then
        echo "export JULIA_DEPOT_PATH=\"${PREFIX}/share/julia/site\"" >> $ACTIVATE_ENV
        echo "export JULIA_OLD_LD_LIBRARY_PATH=$LD_LIBRARY_PATH" >> $ACTIVATE_ENV
        echo "export LD_LIBRARY_PATH=\"${PREFIX}/lib\"" >> $ACTIVATE_ENV
else
        mkdir -p ${PREFIX}/etc/conda/activate.d
        touch ${PREFIX}/etc/conda/activate.d/env_vars.sh
        echo '#!/bin/sh' >> $ACTIVATE_ENV
        echo "export JULIA_DEPOT_PATH=\"${PREFIX}/share/julia/site\"" >> $ACTIVATE_ENV
        echo "export JULIA_OLD_LD_LIBRARY_PATH=$LD_LIBRARY_PATH" >> $ACTIVATE_ENV
        echo "export LD_LIBRARY_PATH=\"${PREFIX}/lib\"" >> $ACTIVATE_ENV
fi
if [ -f "$DEACTIVATE_ENV" ]; then
        echo "unset JULIA_DEPOT_PATH" >> $DEACTIVATE_ENV
        echo "export LD_LIBRARY_PATH=$JULIA_OLD_LD_LIBRARY_PATH" >> $DEACTIVATE_ENV
        echo "unset JULIA_OLD_LD_LIBRARY_PATH" >> $DEACTIVATE_ENV
else
        mkdir -p ${PREFIX}/etc/conda/deactivate.d
        touch ${PREFIX}/etc/conda/deactivate.d/env_vars.sh
        echo '#!/bin/sh' >> $DEACTIVATE_ENV
        echo "unset JULIA_DEPOT_PATH" >> $DEACTIVATE_ENV
        echo "export LD_LIBRARY_PATH=$JULIA_OLD_LD_LIBRARY_PATH" >> $DEACTIVATE_ENV
        echo "unset JULIA_OLD_LD_LIBRARY_PATH" >> $DEACTIVATE_ENV
fi
```

**Comment by JacksonBurns** [comment]
At a high level I'm still skeptical that this is possible - the whole point of Conda.jl seems to be building Julia packages that require Python rather than Python Packages that require Julia. I can give this script a shot, but I must emphasize that this is a maintainability concern. As soon as `Conda.jl` changes anything we are going to be stuck unable to build RMG binaries again despite nothing being wrong with the Python side.

There's also just such a _natural_ separation between mechanism generator in Python + reactor simulator in Julia that it seems (to me) unnecessary to try and combine the two. This would decouple our release schedule, allow people to extend RMG without also installing RMS, etc.

To me this is also just occam's razor. Why do this (comparatively) complex installation when we could so easily separate the two?

I don't recall the exact reason the `Functors` thing was there, I think something was wrong with that upstream on whatever day I wrote that build script.

**Comment by mjohnson541** [comment]
I'm happy to explain how the python-julia system works, why what I proposed is a robust solution to your particular problem as described (which honestly is rare to find with conda-build problems), what role PyCall.jl and Conda.jl play in the system and why there aren't any concerns about maintainability related to them in particular. 

I know this part is new for the binaries and you guys aren't as experienced with these systems as I am. I'm surprised you didn't bring this up with me earlier in the process when you ran into this issue. I would hope that I could have helped you avoid the frustrations it seems you've clearly run into. I'm well aware building binaries is not a fun process especially when dealing with systems you aren't intimately familiar with. 

However, I think it would be best to have such a conversation offline.

**Comment by mjohnson541** [comment]
> Sorry, I felt like the reply stuff was getting too nested.
> 
> On that point, I guess I don't understand why making RMS an optional install makes future development and plug-ability any easier given that after installation the user experience is the same for a user not using those features.
> 
> Super interested about why you needed that Functors line in the build.
> 
> Overall on the issues with the build my understanding is that if RMG starts using an RMS feature, RMS tries to install stuff? My interpretation of this is
> 
>     1. Python/pyjulia knows where Julia is
> 
>     2. The Julia python finds has RMS installed
> 
>     3. Julia/PyCall finds the wrong python and can't find the rmg stuff so tries to install it (probably using the default python)
> 
> 
> So it seems that our problem is simply that while everything we need is installed PyCall doesn't know where the right python/conda env is. This can be set by defining environment variables: PYTHON and maybe also CONDA_JL_HOME (see https://github.com/JuliaPy/PyCall.jl and https://github.com/JuliaPy/Conda.jl). In the build we can set environmental variables that will automatically load when the environment loads so in theory if we do this for PYTHON and CONDA_JL_HOME RMS will grab the appropriate python and find that the packages it needs are already there. An example of doing this from the build file I wrote for julia binaries is below:
> 
> ```
> #make julia directory
> mkdir -p ${PREFIX}/share/julia/site
> mkdir -p ${PREFIX}/bin
> #set JULIA_DEPOT_PATH in conda env
> export JULIA_DEPOT_PATH="${PREFIX}/share/julia/site"
> ACTIVATE_ENV="${PREFIX}/etc/conda/activate.d/env_vars.sh"
> DEACTIVATE_ENV="${PREFIX}/etc/conda/deactivate.d/env_vars.sh"
> 
> if [ -f "$ACTIVATE_ENV" ]; then
>         echo "export JULIA_DEPOT_PATH=\"${PREFIX}/share/julia/site\"" >> $ACTIVATE_ENV
>         echo "export JULIA_OLD_LD_LIBRARY_PATH=$LD_LIBRARY_PATH" >> $ACTIVATE_ENV
>         echo "export LD_LIBRARY_PATH=\"${PREFIX}/lib\"" >> $ACTIVATE_ENV
> else
>         mkdir -p ${PREFIX}/etc/conda/activate.d
>         touch ${PREFIX}/etc/conda/activate.d/env_vars.sh
>         echo '#!/bin/sh' >> $ACTIVATE_ENV
>         echo "export JULIA_DEPOT_PATH=\"${PREFIX}/share/julia/site\"" >> $ACTIVATE_ENV
>         echo "export JULIA_OLD_LD_LIBRARY_PATH=$LD_LIBRARY_PATH" >> $ACTIVATE_ENV
>         echo "export LD_LIBRARY_PATH=\"${PREFIX}/lib\"" >> $ACTIVATE_ENV
> fi
> if [ -f "$DEACTIVATE_ENV" ]; then
>         echo "unset JULIA_DEPOT_PATH" >> $DEACTIVATE_ENV
>         echo "export LD_LIBRARY_PATH=$JULIA_OLD_LD_LIBRARY_PATH" >> $DEACTIVATE_ENV
>         echo "unset JULIA_OLD_LD_LIBRARY_PATH" >> $DEACTIVATE_ENV
> else
>         mkdir -p ${PREFIX}/etc/conda/deactivate.d
>         touch ${PREFIX}/etc/conda/deactivate.d/env_vars.sh
>         echo '#!/bin/sh' >> $DEACTIVATE_ENV
>         echo "unset JULIA_DEPOT_PATH" >> $DEACTIVATE_ENV
>         echo "export LD_LIBRARY_PATH=$JULIA_OLD_LD_LIBRARY_PATH" >> $DEACTIVATE_ENV
>         echo "unset JULIA_OLD_LD_LIBRARY_PATH" >> $DEACTIVATE_ENV
> fi
> ```

Actually, in addition to this we also need to run PyCall.build() once in the conda environment after construction...cleanest would be the first time the conda environment activates, although calling julia.install() inside RMG will also do this.

**Comment by JacksonBurns** [comment]
Rebased onto the latest cclib CI fix in #2638

**Comment by JacksonBurns** [comment]
@hwpang please re-review!

**Comment by hwpang** [comment]
@mjohnson541 I'm seeing that there are many discussions above, could you summarize and list out the action items (if there are any) so that we can merge this PR?

**Comment by mjohnson541** [comment]
So I don't have any issues with the "Avoid Unnecessary Julia Operations" part of this PR. However, I think the "Optional RMS Installation" part is very premature: 

1. The "Avoid Unnecessary Julia Operations" removes nearly all annoyance users have with waiting for julia precompilation when they don't need it. 

2. In PR #2636 we showed that Julia/RMS is not the source of the issues building binaries, the issue at stake was shown to be a cython compilation issue that RMS merely catches first because it was the first thing to import from RMG. 

3. We are about to completely overhaul this system in #2640 so even if there were issues that would warrant a design change such as making RMS optional, it doesn't make sense to make that decision, let alone implement it now, based on the current RMG-RMS interface when we are about to overhaul the whole system in what we believe will be a highly positive way with #2640.

**Comment by hwpang** [comment]
@mjohnson541 Regarding the third point, I think the problem is that #2640 is blocked by #2635, and #2635 is blocked by this PR according to my understanding from discussion with @JacksonBurns.

@JacksonBurns Could you elaborate on how #2635 is blocked by this PR? This will help us decide what to do. Could you also comment on Matt's second point?

**Comment by JacksonBurns** [comment]
## Long Comment! Breaking Into Sections

### Continuing Review

@hwpang I have addressed your comments and rebased - please re-review!

I am going to tack on a couple commits to the docs about the installation process.

### Addressing @hwpang's Notes

> @JacksonBurns Could you elaborate on how #2635 is blocked by this PR?

 - The Python version upgrade (#2635) is blocked by the conda binary fix (#2641) because we want to release conda binaries with Python 3.7 before we move to Python 3.11+ (to support older workloads, and because the version upgrade will change the performance of RMG).
 -  The conda binary fix is blocked by this PR, because that fix assumes that RMG can be installed and run _without_ RMS in order to build the conda binary.
 - Your Julia dependency overhaul (#2640) is blocked by the Python version upgrade because PythonCall does not support our version of Python, making that PR very circuitously blocked by this one.

### Addressing @mjohnson541's Notes

> 1. The "Avoid Unnecessary Julia Operations" removes nearly all annoyance users have with waiting for julia precompilation when they don't need it.

:+1: to this benefit - but I'd also like to mention that there are other benefits:
 - would allow developers to run `pytest` via Python instead of python-jl, dramatically cutting down on execution time
 - would make using RMG as a library in others software easier, since it doesn't _require_ this additional set of dependencies

> 2. In PR #2636 we showed that Julia/RMS is not the source of the issues building binaries, the issue at stake was shown to be a cython compilation issue that RMS merely catches first because it was the first thing to import from RMG.

I don't think this is a reason to not merge this PR. #2641 shows that by merging this we could get the build to work.

> 3. We are about to completely overhaul this system in #2640 [...]

See my notes in the above section about why this is blocked.

File: rmgpy/pdep/sls.py [hunk 4]
```diff
@@ -150,6 +160,11 @@ def get_rate_coefficients_SLS(network, T, P, method="mexp", neglect_high_energy_
     tau = np.abs(1.0 / fastest_reaction)
 
     if method == "ode":
+        if NO_JULIA:
+            raise RuntimeError(
+                "Required Julia dependencies for method 'ode' are not installed.\n"
+                "Please check your installation (https://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/installation/index.html)."
```
hwpang, line 166: Since the user will only encounter this if they install from source, we should point them to the developer installation.

```suggestion
                "Please follow the steps to install Julia dependencies at https://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/installation/anacondaDeveloper.html."
```

rmgpy/rmg/input.py
```diff
@@ -1558,6 +1559,11 @@ def read_input_file(path, rmg0):
         exec(f.read(), global_context, local_context)
     except (NameError, TypeError, SyntaxError) as e:
         logging.error('The input file "{0}" was invalid:'.format(full_path))
+        if NO_JULIA:
+            logging.error(
+                "During runtime, import of Julia dependencies failed. To use phase systems and RMS reactors, install RMG-Py with RMS."
+                " (https://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/installation/index.html)"
+            )
```
hwpang, line 1566: Same as above, this only happens to the user who installs from source, so we should guide them to the corresponding installation page.

```suggestion
            logging.error(
                "During runtime, import of Julia dependencies failed. To use phase systems and RMS reactors, install RMG-Py with RMS."
                " (https://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/installation/anacondaDeveloper.html)"
            )
```

rmgpy/rmg/main.py
```diff
@@ -550,10 +557,10 @@ def initialize(self, **kwargs):
         self.load_database()
 
         for spec in self.initial_species:
-            self.reaction_model.add_species_to_edge(spec)
+            self.reaction_model.add_species_to_edge(spec, requires_rms=requires_rms)
 
         for reaction_system in self.reaction_systems:
-            if isinstance(reaction_system, Reactor):
+            if not NO_JULIA and isinstance(reaction_system, RMSReactor):
```
hwpang, line 563: What happens if there is no julia and the user specify to use an RMSReactor? Should we raise an error for that? Or have we already raised an error for that? If we have, I think we don't need to check NO_JULIA here again?

rmgpy/pdep/sls.py
```diff
@@ -46,6 +45,17 @@
...
 
+NO_JULIA = False
+try:
+    from diffeqpy import de
+    from julia import Main
+except Exception as e:
+    warnings.warn(
+        f"Unable to import Julia dependencies, original error: {str(e)}"
+        ". Master equation method 'ode' will not be available on this execution."
+    )
```
hwpang, line 56: Logging.info is used in the rmg.rms_reactor, so I think we should be consistent about the way we show this information

```suggestion
    logging.info(
        f"Unable to import Julia dependencies, original error: {str(e)}"
        ". Master equation method 'ode' will not be available on this execution."
    )
```

hwpang:
@JacksonBurns Thanks for the explanation! I have left some minor comments, and I am happy to merge this (so that we can get other PRs unblocked) after they are addressed.

rmgpy/rmg/main.py
JacksonBurns: You are right - in the lines above I have actually left a comment that we would have already failed. So I will remove this part of the check.

hwpang: I think there are other instances where we check both `not NO_JULIA` and `isinstance(reaction_system, RMSReactor)`, but if `NO_JULIA` and `isinstance(reaction_system, RMSReactor)` that should have already failed, I don't think we need to check them again. Could you go through these cases and remove the unnecessary checks?

JacksonBurns: Good point, just got them all. We can rely only on `requires_rms`, though `NO_JULIA` is still used for better errors and in the initialization of the model (which must stay, since it doesn't know anything the reactors being used).

hwpang (approved):
LGTM

mjohnson541 (changes requested):
Setting this to automerge without resolving discussions is absolutely not appropriate behavior!

hwpang:
@mjohnson541 My bad, sorry, I thought Jackson has replied and resolved your concerns, so I set it to automerge after seeing that he addressed my change requests. Thanks for disabling the automerge.

JacksonBurns:
What else should we discuss?

JacksonBurns:
Ongoing

JacksonBurns:
Porting to #2694

Issue #2630, opened 2024-03-08
Generating reactions for nitrobenzenes is very slow


jonwzheng:
### Bug Description
There are two separate but related issues that I observed. First, an RMG website user mentioned that using RMG kinetics search for nitro-substituted benzenes causes the kinetics search to timeout when calling `KineticsDatabase.generate_reactions`. Here are two examples:

**1. Bimolecular reaction**

Finding the bimolecular reactions of a nitrobenzene with itself, including just monosubstituted nitrobenzene (`[O-][N+](=O)c1ccccc1`), will hang on execution.

_Minimal example:_
```
from rmgpy import settings
from rmgpy.data.rmg import RMGDatabase
from rmgpy import molecule
from rmgpy.data.kinetics import KineticsDatabase, TemplateReaction

database_directory = settings["database.directory"]
database = RMGDatabase()
database.load(database_directory, kinetics_families="all")

reactant = molecule.Molecule(smiles="[O-][N+](=O)c1ccccc1")
reactants = [reactant]
print("Generating unimolecular reaction")
database.kinetics.generate_reactions(
        reactants, None, only_families=None, resonance=True)

print("Generating bimolecular reaction")
reactants2 = [reactants[0], reactants[0]]
database.kinetics.generate_reactions(
        reactants2, None, only_families=None, resonance=True)
```

Following the traceback it looks like it gets stuck on `Species.generate_resonance_structures()` --> `Molecule.is_identical()`.
Even if you set `resonance=False`, this still occurs because the traceback leads to `family.calculate_degeneracy(rxn)
` --> `ensure_independent_atom_ids(reactants, resonance=True)` in which resonance is hardcoded to True.

Nitrobenzenes have several resonance structures, but is that enough to cause this issue? 

**2. Unimolecular reaction for trinitrobenzene**

_Minimal example:_
Searching the unimolecular reactions of trinitrobenzene will hang:
```
reactant = molecule.Molecule(smiles="[O-][N+](=O)c1c(cccc1[N+](=O)[O-])[N+](=O)[O-]")
reactants = [reactant] 
database.kinetics.generate_reactions(reactants, None, 
                                     only_families=None, resonance=True)
```

hanging on the same step as above.

### Other thoughts

I am still investigating this, and would appreciate any input if anyone has run into a similar issue working with the resonance code, or thoughts on where to troubleshoot next.

hwpang:
[...]

PR #2632, opened 2024-03-14
Clean up some code for calculating degeneracy in kinetics codebase

jonwzheng:
### Motivation or Problem
In #2630 it was observed that `kinetics.database.generate_reactions_from_families` and `kinetics.family.calculate_degeneracy` can be modularized and improved.

### Description of Changes
This PR accomplishes the following:
1. Copies the overlapping code in the above functions and re-implements as a new function `kinetics.common.check_for_same_reactants`, which checks to see whether the provided reactants list has duplicate reactants (or not).
2. Allow `resonance` to propagate in `kinetics.family.calculate_degeneracy`. This way, you can specify `resonance=False` when generating families and it will not consider resonance anymore (which is crucial for some species that have many resonance forms, see #2630).
3. Fix up various typos; for example previously `kinetics.database.generate_reactions` does not allow `only_families` to propagate rendering that argument useless.

### Testing
Verified that both of these functions work in terms of allowing `kinetics.database.generate_reactions_from_families` and setting `resonance=False` (see minimal example in #2630 above).

Need to check if any unit tests fail.

### DIFF
--- a/rmgpy/data/kinetics/common.py
+++ b/rmgpy/data/kinetics/common.py
@@ -222,9 +222,51 @@ def ensure_independent_atom_ids(input_species, resonance=True):
         for species in input_species:
             species.generate_resonance_structures(keep_isomorphic=True)
 
+def check_for_same_reactants(reactants):
+    """
+    Given a list reactants, check if the reactants are the same.
+    If they refer to the same memory address, then make a deep copy so they can be manipulated independently.
+    
+    Returns the modified reactants list and an integer containing the number of same reactants in the reactants list. 
+    
+    """
+
+    same_reactants = 0
+    if len(reactants) == 2:
+        if reactants[0] is reactants[1]:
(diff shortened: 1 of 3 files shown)


JacksonBurns (approved):
This LGTM! My only thought is that we might want to raise an error in `check_for_same_reactants` if the user gives a list of `len>3` - right now the function would just silently return the given list. This doesn't matter for the devs (since nothing we consider has 4 reactants) but perhaps users? If you think it's worth adding please do, otherwise go ahead and merge after CI passes (assuming it does)

jonwzheng:
That's a good point, it's definitely undesired behavior for it to just silently return the list without doing anything. I'll make it raise an exception and see if that breaks anything in the unit tests.

jonwzheng:
@JacksonBurns I've added the exception checker, lmk if it looks good and we're ready to merge.

JacksonBurns (approved):
LGTM! Thanks!

PR #2535, opened 2023-08-30
Fix Failing CI because of `cclib`

JacksonBurns:
### Motivation or Problem
The CI has been failing ([see here](https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/6009736707/job/16299748843#step:4:86)) for a couple days because of the latest release of `cclib` (see [here](https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/5973597214/job/16206162069#step:4:86) for reference), breaking unrelated PRs such as #2533 

### Description of Changes
Disallows `1.8.0` of `cclib` in the RMG `environment.yml`. This is intended as a temporary patch so PRs can keep working and I can do some work locally to find out why this version of `cclib` is failing, and potentially file a bug report with the maintainers.

### Testing
Just using the CI here. As mentioned above, need to do some local debugging to try and find out _why_ this is failing.

### Reviewer Tips
Should be quick to merge (famous last words).

PR #2638, opened 2024-03-20
restrict newer versions of cclib

ChrisBNEU:
### Motivation or Problem
this is related to #2535, regression tests are failing because we are using a newer version of cclib. the previous statement did not restrict newer versions of cclib, just versions that are !=1.8.0. 1.8.1 was released recently and broke out regressions tests again. 

### Description of Changes
change environment.yaml line for cclib to disallow versions that are equal to or greater than cclib 1.8.0

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -44,7 +44,7 @@ dependencies:
   - coolprop
   - cantera::cantera=2.6
   - conda-forge::mopac
-  - conda-forge::cclib >=1.6.3,!=1.8.0
+  - conda-forge::cclib >=1.6.3,<1.8.0
   - conda-forge::openbabel >= 3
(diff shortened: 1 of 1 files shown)


rwest (approved):
This seems no worse than what we were doing last week, so I'll approve it, but if there's not already an issue open to make us compatible with the newer CCLIB, please could you open one? (could be a good thing for a keen undergrad to try).  Otherwise we'll get stuck in the past for ever, and I bet using old CCLIB binaries on conda-forge will then stick us to old versions of everything else. So this is more important than it might at first seem.

JacksonBurns:
Commenting to add that a more permanent solution to this problem is being addressed here: https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2639

PR #2639, opened 2024-03-20
updating 'rotconsts' ref to be compatible with cclib v1.8+

ChrisBNEU:
### Motivation or Problem
regression tests fail with the superminimal example, because it uses mopac for qm calcs. the root cause is the parser has since been updated, and the variable name for rotational constants has been changed from "rotcons" to "rotconsts" (see [this](https://github.com/cclib/cclib/commit/ebff98a7d0dadc92b1c6c23a91210586ab2470aa) commit on the cclib repo). 

additionally, in more recent versions, it looks like the "rotconsts" is part of a default cclib parser object for MOPAC output files, so there the workaround we use is not necessary. I have left the "old way" of adding rotational constants in for backwards compatibility. 

### Description of Changes
I added checks for whether the cclib_data object has "rotconsts" specified, and assigns the appropriate values to the qm object. 

### Testing
unit tests and functional tests. If the superminimal passes with both 1.8 and <1.8 then it should be fine I think.

### DIFF
--- a/rmgpy/qm/molecule.py
+++ b/rmgpy/qm/molecule.py
@@ -409,18 +409,25 @@ class QMMolecule(object):
         parser.logger.setLevel(
             logging.ERROR
         )  # cf. http://cclib.sourceforge.net/wiki/index.php/Using_cclib#Additional_information
-        parser.rotcons = (
-            []
-        )  # give it an attribute and it won't delete it, leaving it on the parser object
+
         parser.molmass = None  # give it an attribute and it won't delete it, leaving it on the parser object
         cclib_data = parser.parse()
+
+        # for bacwards compatibility with cclib <1.6
+        if not hasattr(cclib_data, "rotconsts"):
+             # this hack required because rotconsts not part of a default cclib data object
+            parser.rotconsts = (
+                []
+            )  # give it an attribute and it won't delete it, leaving it on the parser object
+
+            cclib_data.rotconsts = (
+                parser.rotconsts
+            ) 
+
         radical_number = self.molecule.get_radical_count()
-        cclib_data.rotcons = (
-            parser.rotcons
-        )  # this hack required because rotcons not part of a default cclib data object
         cclib_data.molmass = (
             parser.molmass
-        )  # this hack required because rotcons not part of a default cclib data object
+        )  # this hack required because molmass is not part of a default cclib data object
         qm_data = parse_cclib_data(
             cclib_data, radical_number + 1
         )  # Should `radical_number+1` be `self.molecule.multiplicity` in the next line of code? It's the electronic ground state degeneracy.
--- a/rmgpy/qm/molecule.py
+++ b/rmgpy/qm/molecule.py
@@ -408,7 +408,7 @@ class QMMolecule(object):
         parser = self.get_parser(self.output_file_path)
         parser.logger.setLevel(
             logging.ERROR
-        )  # cf. http://cclib.sourceforge.net/wiki/index.php/Using_cclib#Additional_information
+        )  # cf. https://cclib.github.io/index.html#how-to-use-cclib
         parser.molmass = None # give it an attribute and it won't delete it, leaving it on the parser object
         parser.rotcons = (  # for cclib < 1.8.0
             []
--- a/rmgpy/qm/molecule.py
+++ b/rmgpy/qm/molecule.py
@@ -409,7 +409,23 @@ class QMMolecule(object):
         parser.logger.setLevel(
             logging.ERROR
         )  # cf. http://cclib.sourceforge.net/wiki/index.php/Using_cclib#Additional_information
-
+parser.molmass = None # give it an attribute and it won't delete it, leaving it on the parser object
+parser.rotcons = (  # for cclib < 1.8.0
+    []
+)
+parser.rotconsts = (  # for cclib >= 1.8.0
+    []
+)
+cclib_data = parser.parse()  # fills in either parser.rotcons or parser.rotconsts but not both
+assert bool(parser.rotconsts) != bool(parser.rotcons)
+if parser.rotcons:  # for cclib < 1.8.0
+    cclib_data.rotcons = (
+        parser.rotcons
+    )
+else:  # for cclib >= 1.8.0
+    cclib_data.rotconsts = (
+        parser.rotconsts
+    )
         parser.molmass = None  # give it an attribute and it won't delete it, leaving it on the parser object
         cclib_data = parser.parse()
 
--- a/rmgpy/qm/molecule.py
+++ b/rmgpy/qm/molecule.py
@@ -409,23 +409,23 @@ class QMMolecule(object):
         parser.logger.setLevel(
             logging.ERROR
         )  # cf. http://cclib.sourceforge.net/wiki/index.php/Using_cclib#Additional_information
-parser.molmass = None # give it an attribute and it won't delete it, leaving it on the parser object
-parser.rotcons = (  # for cclib < 1.8.0
-    []
-)
-parser.rotconsts = (  # for cclib >= 1.8.0
-    []
-)
-cclib_data = parser.parse()  # fills in either parser.rotcons or parser.rotconsts but not both
-assert bool(parser.rotconsts) != bool(parser.rotcons)
-if parser.rotcons:  # for cclib < 1.8.0
-    cclib_data.rotcons = (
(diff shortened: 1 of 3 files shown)


JacksonBurns:
@ChrisBNEU since we are closing in on upgrading to the latest versions of Python, I am keen to just limit the version of `cclib` to `1.8.0` or newer to help simplify the work you have going here. Thoughts? Thanks for getting this underway, as well.

berquist:
Maybe for my own clarification, I was trying to understand what was going on here, because https://github.com/cclib/cclib/commit/ebff98a7d0dadc92b1c6c23a91210586ab2470aa makes it look like I introduced a breaking change in a minor version.  https://github.com/cclib/cclib/pull/1054 introduced `rotconsts`, and a `rotcons` attribute technically didn't exist before: since it wasn't part of the list on the data object, `https://github.com/cclib/cclib/pull/1054/files#diff-a710fa5d121b2ed984e78fcd7ed5a27924b0cad89d40e34e36d65f8039bc2285L176` should have been removed as part of filtering (https://github.com/cclib/cclib/blob/e53eaf471384ca97fabdf699db0e511b55f0d638/cclib/parser/data.py#L311).

The problem is https://github.com/ReactionMechanismGenerator/RMG-Py/blob/a458b1deb3c26f69cff627396e3a635538c1cff6/rmgpy/qm/molecule.py#L418 where the public `cclib.parser.data.ccData` API isn't being followed, which is basically what's said in the comment there and this PR.

Semi-related, but the next release after 1.8.1 is going to be 2.0, and the attribute name isn't going to change, but other API things will, so I suggest pinning to `<2`.  This will happen in a few months.

JacksonBurns:
@berquist thanks for the input!

@ChrisBNEU I have added the suggestion from above and rebased this - can we pick this up again?

ChrisBNEU:
> @berquist thanks for the input!
> 
> @ChrisBNEU I have added the suggestion from above and rebased this - can we pick this up again?

@JacksonBurns sure, I'll give it another look over. if nothing fails I'll grab someone to review.

rmgpy/qm/molecule.py
```diff
@@ -409,18 +409,25 @@ def parse(self):
         parser.logger.setLevel(
             logging.ERROR
         )  # cf. http://cclib.sourceforge.net/wiki/index.php/Using_cclib#Additional_information
-        parser.rotcons = (
-            []
-        )  # give it an attribute and it won't delete it, leaving it on the parser object
+
```
sevyharris, line 412: I think this needs to be left in for backwards compatibility with cclib <1.8.0.

Here is my recommendation for lines 412-426:
```suggestion
parser.molmass = None # give it an attribute and it won't delete it, leaving it on the parser object
parser.rotcons = (  # for cclib < 1.8.0
    []
)
parser.rotconsts = (  # for cclib >= 1.8.0
    []
)
cclib_data = parser.parse()  # fills in either parser.rotcons or parser.rotconsts but not both
assert bool(parser.rotconsts) != bool(parser.rotcons)
if parser.rotcons:  # for cclib < 1.8.0
    cclib_data.rotcons = (
        parser.rotcons
    )
else:  # for cclib >= 1.8.0
    cclib_data.rotconsts = (
        parser.rotconsts
    )
```

sevyharris: So in total, I think the parse function in RMG-Py/rmgpy/qm/molecule.py should look like this:
```
def parse(self):
        """
        Parses the results of the Mopac calculation, and returns a QMData object.
        """
        parser = self.get_parser(self.output_file_path)
        parser.logger.setLevel(
            logging.ERROR
        )  # cf. http://cclib.sourceforge.net/wiki/index.php/Using_cclib#Additional_information

        parser.molmass = None  # give it an attribute and it won't delete it, leaving it on the parser object
        parser.rotcons = (  # for cclib < 1.8.0
            []
        )
        parser.rotconsts = (  # for cclib >= 1.8.0
            []
        )
        cclib_data = parser.parse()  # fills in either parser.rotcons or parser.rotconsts but not both
        assert bool(parser.rotconsts) != bool(parser.rotcons)
        if parser.rotcons:  # for cclib < 1.8.0
            cclib_data.rotcons = (
                parser.rotcons
            )
        else:  # for cclib >= 1.8.0
            cclib_data.rotconsts = (
                parser.rotconsts
            )

        radical_number = self.molecule.get_radical_count()
        cclib_data.molmass = (
            parser.molmass
        )  # this hack required because molmass is not part of a default cclib data object
        qm_data = parse_cclib_data(
            cclib_data, radical_number + 1
        )  # Should `radical_number+1` be `self.molecule.multiplicity` in the next line of code? It's the electronic ground state degeneracy.
        return qm_data
```

rmgpy/qm/molecule.py
```diff
@@ -409,18 +409,41 @@ def parse(self):
...
+        if not hasattr(cclib_data, "rotconsts"):
+             # this hack required because rotconsts not part of a default cclib data object
+            parser.rotconsts = (
+                []
+            )  # give it an attribute and it won't delete it, leaving it on the parser object
+
+            cclib_data.rotconsts = (
+                parser.rotconsts
+            ) 
+
```
sevyharris, line 442: I couldn't figure out how to do this as an inline suggestion earlier, but this section should be deleted. 
Note the `parser.molmass = None` line has been moved up to line 412, which is why I suggest deleting it here.
```suggestion
```

sevyharris (dismissed):
I've tested this with cclib 1.7.2 and 1.8.1 and it fixes the failing mopac test in both. I think it's ready to merge (assuming it passes the usual automated tests)

rmgpy/qm/molecule.py
```diff
@@ -409,18 +409,27 @@ def parse(self):
         parser.logger.setLevel(
             logging.ERROR
         )  # cf. http://cclib.sourceforge.net/wiki/index.php/Using_cclib#Additional_information
```
JacksonBurns, line 411: The link in this comment is dead, could you fix it?

sevyharris, line 411: Updated link to cclib documentation
```suggestion
        )  # cf. https://cclib.github.io/index.html#how-to-use-cclib
```

sevyharris (approved):
cclib documentation link is updated, so it should be ready to go

berquist:
> @berquist, just so I understand correctly, the rotcons attribute never existed, we just made that up in rmg so we'd have a place to put it, right?

The formal attribute never existed.  The class member did exist, only on the MOPAC parser, and it must not have been cleaned out properly by `.parse()`.

rwest:
We have code being added explicitly for cclib >1.8, and Sevy confirms in a comment above that it works for her on cclib 1.8.1.  I also see the note on pinning <2. But is there a reason the conda environment file still specifies <1.8.0 ?
```- conda-forge::cclib >=1.6.3,<1.8.0,<2```
Apologies if I missed this in the discussion.

JacksonBurns:
@sevyharris @ChrisBNEU as we are upgrading to Python 3.9 in #2687 and associated PRs, we have the opportunity to upgrade to cclib 1.8 and force it to be 1.8 or newer. Do you have the time/interest to make this part of that work? Was this one close to working, and would finishing be simplified if we remove the requirement to support 1.7 and 1.8?

berquist:
> @sevyharris @ChrisBNEU as we are upgrading to Python 3.9 in #2687 and associated PRs, we have the opportunity to upgrade to cclib 1.8 and force it to be 1.8 or newer. Do you have the time/interest to make this part of that work? Was this one close to working, and would finishing be simplified if we remove the requirement to support 1.7 and 1.8?

Regarding "newer", definitely pin <1.9 to be safe.

sevyharris:
Oof, my bad for losing track of this-- it was supposed to be merged months ago. It's all ready to go. Works for cclib versions older and newer than 1.8

JacksonBurns:
Thanks for the heads up @berquist!

@sevyharris no oof required, that's the best news I've heard in days. I will add the suggested version limit and we can see how the CI does.

JacksonBurns (approved):
LGTM! And passing CI. @sevyharris could you re-review and approve?

JacksonBurns:
ping @sevyharris could you take one last look and review?

sevyharris (approved):
Sorry for the delay on this! I looked over it again and reran my tests. Looks good to me!

JacksonBurns:
No problem, thanks for double checking!

PR #2641, opened 2024-03-22
Fixing conda binary build via optional RMS install (dependent on #2631)

JacksonBurns:
#2539 managed to fix the conda binary build but grew dramatically out of proportion.

#2631 picks up the earliest of those changes (making RMS optional and avoiding superfluous Julia operations) but there is some disagreement about moving forward with the former. This resulted in #2636, which seeks to fix the conda recipe without merging the changes in #2631 (or, at least, without making RMS optional).

This PR takes the opposite approach and instead picks up the remaining dependent changes from #2539 and will attempt to fix the conda recipe assuming that RMS is optional. This PR is blocked by #2631 but will (hopefully) demonstrate the value in merging it.

Note that there is also an ongoing effort in #2640 (and the corresponding PR in RMS) that would (hopefully) simplify merging #2636.

We will build these binaries one way or the other (hopefully)!


JacksonBurns:
The `libcrypt` header file reported missing in previous attempted builds is not related to the system-level package `libcrypt1`, which is already installed in the runners by default.

JacksonBurns:
This build worked locally (Ubuntu) after removing `nomkl` from the dependencies, which seems like a platform-specific thing (since it works on the runners). They may both pass here!

JacksonBurns:
https://github.com/ReactionMechanismGenerator/RMG-database/pull/645 should be merged before this.

JacksonBurns:
@mjohnson541 @rwest @hwpang this PR has now fixed the `conda` binaries by allowing for an optional RMS installation.

Users wishing to install RMG+RMS from `conda` can now do so by first running `conda install rmg -c rmg` and then running the typical `RMS` build instructions (i.e. add Julia, pyrms, install pycall, build RMS, etc) that are currently being done in our CI (see the updated doc page for detailed instructions).

Please let me know what you think. **Here's is a link to the diff for just the commits that are part of this PR**, since it also includes some from 2631: https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2641/files/24284bcaf5161d23f14201ed053a4f44f883dba6..f6be0ad41ef8ab52855c333b7660c206c9c0b90d

rwest:
Just wondering: What happened (or is happening) with #2636? Still feasible?
Conda install RMG then  add Julia, pyrms, install pycall, build RMS, etc etc is not an ideal user experience, if compared to just a Conda install. 

Still, it's good to make progress, so well done figuring stuff out here. I've not had a chance to look properly yet

JacksonBurns:
> Just wondering: What happened (or is happening) with #2636? Still feasible? Conda install RMG then add Julia, pyrms, install pycall, build RMS, etc etc is not an ideal user experience, if compared to just a Conda install.

Bogged down bc of build failures which were either in the conda recipe itself or in the way we extended it to use Julia. Now that the conda recipe works _here_, I will pull the relevant commits into _there_ so we can be sure the only issues remaining are because of the Julia extensions.

JacksonBurns:
TODO: add examples of installing a specific version of `rmgdatabase`, i.e. `conda install rmg rmgdatabase=3.1.1 --channel rmg`

JacksonBurns:
The blocking PR is set to automerge, so I have rebased this one onto #2631 to check if the build still works as expected

JacksonBurns:
ongoing

JacksonBurns:
ongoing

JacksonBurns:
*clarifying, many of the edits required to fix the conda recipe have been ported piecemeal into #2741 but the actual fixes to the conda recipe itself that are present in this PR will need to be re-done over there (since the codebase has changed a lot since this was last working).

PR #2598, opened 2024-02-01
Electrochemistry! (finally) a.k.a. RMG-electrocat

JacksonBurns:
# Electrochemistry!

This PR finally brings electrochemistry to RMG.
It replaces or combines #2000 and  #2316.
Most of the original work was done by @davidfarinajr and @mjohnson541.
Help and effort rebasing from @JacksonBurns and @ssun30.
With some assistance along the way by @rwest.

Over the years it been checked many times. All the tests now pass.
It goes with RMG-database pull request https://github.com/ReactionMechanismGenerator/RMG-database/pull/667 
and it uses the [marcus_development](https://github.com/ReactionMechanismGenerator/ReactionMechanismSimulator.jl/tree/marcus_development) branch of RMS (which it checks out during the CI).
_The manual installation instructions need tweaking to help people install that branch, or we need to merge that branch to main in RMS._

PR #2643, opened 2024-03-28
Make chemkin read of transport and thermo files more robust

sevyharris:
Make chemkin read of transport and thermo files more robust

### Motivation or Problem
I ran into these two issues while importing the Aramco 3.0 mechanism:
1. It couldn't read the separate thermo file because of the relative seek command, which apparently requires reading in binary format and gave this error message: `UnsupportedOperation: can't do nonzero cur-relative seeks`
2. It couldn't read the transport file because the shape_index column had been formatted as a float instead of an int.

### Description of Changes
1. Instead of using a relative f.seek() command, keep track of the previous line and just seek to that. This is how the file is parsed in the rest of chemkin.pyx.
2. Read in the shape_index as a float and then convert it to an int. This works whether the data is formatted as 0 or 0.000
--------------------------------------------------
#### Part 1 - separate thermo file
1. For the separate thermo read: here is a mechanism split into [mech.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/14804947/mech.txt), [species_dictionary.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/14804949/species_dictionary.txt), [tran.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/14804959/tran.txt), and [thermo.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/14804960/thermo.txt) files.
Try reading it with the following code:
```
import rmgpy.chemkin

mech_file = 'mech.txt'
thermo = 'thermo.txt'
transport = 'tran.txt'
dictionary_path = 'species_dictionary.txt'

species_list, reaction_list = rmgpy.chemkin.load_chemkin_file(mech_file, dictionary_path=dictionary_path, transport_path=transport, thermo_path=thermo)
```

You should get the following error when trying this on main: `UnsupportedOperation: can't do nonzero cur-relative seeks` that is fixed on `readchemkin_robust_transthermo`

--------------------------------------------------
#### Part 2 - float issue
2. For the shape_index issue, you can test out this [full_mech.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/14805007/full_mech.txt) [species_dictionary.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/14804990/species_dictionary.txt), and
[tran_float.txt](https://github.com/ReactionMechanismGenerator/RMG-Py/files/14804992/tran_float.txt).

```
import rmgpy.chemkin
mech_file = 'full_mech.txt'
[...]

PR #2644, opened 2024-03-29
Remove duplicated test data directory

rwest:
### Motivation or Problem
Jackson tried removing this test data directory as part of another pull request (#2598).
We're now having problems in a related area, caused by this, so I'm doing it as a standalone pull request.
It took a few fixes to get it working.
In any case, it would be nice to remove duplication (as noted in the commit message of 3177452d8a61987e6205f3cd6a64669bcffbaa31.)

### Description of Changes
We already have a `test/rmg_py/test_data/testing_database` which contains a mock database
for running unit tests.
This PR removes `rmgpy/data/test_data/` which served a similar purpose.
Because they were slightly different, the tests needed to be fixed.
Also moved test_remove_group (one of the broken tests) out of TestCyclicThermo into TestThermoDatabase because it had nothing to do with cyclic thermo.

 
The test test_remove_group tests a function (remove_group) that is not currently used in RMG, but has been used in the past for manipulating the database offline. Because it could be helpful to people in the future, and is not currently (after this PR) broken, we might as well leave it. But one could argue for deleting it.

### Testing
Unit tests should pass.


rwest:
We could drop the reverted commits when rebasing before merging but I'm leaving them for review so you can follow along.

JacksonBurns (commented):
Thanks for pulling in these changes... worth getting into `main` for sure.

The changes to the unit test look reasonable to me. I noticed in your commit message on 7da3470 that said unit test is testing a disused function.

Thoughts on simply removing that unit test instead, making this PR something like "remove disused test and duplicated test data"? They aren't explicitly related but are both small. Happy to merge this as-is too, just LMK

rwest:
> Thoughts on simply removing that unit test instead, making this PR something like "remove disused test and duplicated test data"? 

I advise against removing the test without removing the function that it's testing.  One could argue for that if trying to throw away all unused code. But as it's currently working, and has been helpful to people in the past, I feel we might as well leave it. (Although the chances that someone who could use it it will notice it's there is arguably small.)  Anyway, I think the current way (in this PR) is fine.


I rebased, removing the commits that had no net effect, and added one more that moves the test into the class where it belongs. When the tests pass, I suggest this is merged.

JacksonBurns:
@rwest sounds like a plan! I will let #2643 merge, rebase this, and then enble auto-merge. (Previous docs build failure was due to network drop)

JacksonBurns (approved):

PR #2646, opened 2024-04-01
Cov dep therm

12Chao:
### Motivation or Problem
This PR adds in coverage dependent thermodynamic models for heterogeneous catalysis modeling. An adsorbate's enthalpy and entropy should be affected by other adsorbates around it. The PR enables RMG to use a polynomial model to estimate the change of an adsorbate's enthalpy or entropy based on the coverage of other adsorbates on the catalyst surface. By doing this, a Cantera yaml file with thermodynamic coverage dependent data can be generated at the end of a RMG simulation, and it can be used by Cantera (>=3.0) to run PFR simulation. RMG looks into the database to read the thermo coverage dependent models.

### Description of Changes
surface.pyx file is modified to change the species enthalpy and entropy based on the coverage dependent (polynomial) models in the database. 
nasa.pyx, wilhoit.pyx, and thermodata.pyx are all modified to be able to save the thermo coverage dependent model for each species.
main.py is modified to be able to incorporate thermo coverage dependence modeling as an option and be able to write Cantera yaml file with thermo coverage models in.
### Testing
Unit tests are added in thermo and solver folders to test if a species with thermo coverage dependent model can be created and whether the simulation can be ran.
To further test the effect to the model generation, a thermo database is made up with the thermo cov dependent models, and the data base was used to make a CPOX model on Pt. In the CPOX on Pt model, OX self interaction and COX self interaction are included. The model was then ran through a PFR simulation with Cantera 3.0, a significant drop of coverage of COX on the surface was observed, and therefore validate the code modification. Test files are attached here.
[test.zip](https://github.com/ReactionMechanismGenerator/RMG-Py/files/14828446/test.zip)
### Reviewer Tips
Please try to pull the changes from this branch and used the database provided in the tes.zip to generate a model of your own choice see if it makes any difference to your model compared to without thermo cov dependent model. You can also generate a model with the RMG input file provided in the tes.zip, then in the input file turn `thermoCoverageDependence` to False to generate another model without coverage dependence. Then use simultion.py in test.zip to run simulations for the two RMG models generated to test if there is any difference in the simulation results.


rmgpy/rmg/main.py
```diff
@@ -1179,6 +1182,44 @@ def execute(self, initialize=True, **kwargs):
...
                 )
+                if self.thermo_coverage_dependence:
+                    # add thermo coverage dependence to Cantera files
+                    chem_yaml_path = os.path.join(self.output_directory, "cantera", "chem.yaml")
+                    gas = ct.Solution(chem_yaml_path, "gas")
+                    surf = ct.Interface(chem_yaml_path, "surface1", [gas])
+                    with open(chem_yaml_path, 'r') as f:
+                        content = yaml.load(f, Loader=yaml.FullLoader)
+                    
+                    content['phases'][1]['reference-state-coverage'] = 0.11
```
rwest: is this an arbitrary ~1/9 because that's where most of our DFT was done? (and we assume is the default everywhere)

rmgpy/rmg/main.py
```diff
@@ -1179,6 +1182,44 @@ def execute(self, initialize=True, **kwargs):
                     os.path.join(self.output_directory, "chemkin", "chem_annotated-gas.inp"),
                     surface_file=(os.path.join(self.output_directory, "chemkin", "chem_annotated-surface.inp")),
                 )
+                if self.thermo_coverage_dependence:
+                    # add thermo coverage dependence to Cantera files
```
rwest: I assume this might all change when we refactor the yaml-writing?

rmgpy/rmg/main.py
```diff
@@ -1232,6 +1235,44 @@ def execute(self, initialize=True, **kwargs):
...
+                    chem_yaml_path = os.path.join(self.output_directory, "cantera", "chem.yaml")
+                    gas = ct.Solution(chem_yaml_path, "gas")
+                    surf = ct.Interface(chem_yaml_path, "surface1", [gas])
```
sevyharris: Instead of loading the cantera gas and surf objects, we can get the order of species names in the cantera file with something like this:
`cantera_names = [content["species"][i]['name'] for i in range(len(content["species"]))]`

rmgpy/rmg/main.py
```diff
@@ -50,6 +50,7 @@
 import yaml
 from cantera import ck2yaml
 from scipy.optimize import brute
+import cantera as ct
```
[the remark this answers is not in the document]
bjkreitz: Removed cantera import

rmgpy/rmg/main.py
bjkreitz: changed the code as suggested

rmgpy/rmg/main.py
```diff
@@ -1232,6 +1235,44 @@ def execute(self, initialize=True, **kwargs):
...
+                        if s.contains_surface_site() and s.thermo.thermo_coverage_dependence:
+                            for dep_sp, parameters in s.thermo.thermo_coverage_dependence.items():
+                                mol = Molecule().from_adjacency_list(dep_sp)
+                                for sp in self.reaction_model.core.species:
+                                    if sp.is_isomorphic(mol, strict=False):
+                                        parameters['units'] = {'energy':'J', 'quantity':'mol'}
+                                        parameters['enthalpy-coefficients'] = [value.value_si for value in parameters['enthalpy-coefficients']]
+                                        parameters['entropy-coefficients'] = [value.value_si for value in parameters['entropy-coefficients']]
+                                        try:
+                                            content["species"][gas.n_species+surf.species_index(sp.to_chemkin())]['coverage-dependencies'][sp.to_chemkin()] = parameters
```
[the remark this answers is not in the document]
bjkreitz: changed the code as suggested

rmgpy/rmg/main.py
```diff
@@ -1232,6 +1235,44 @@ def execute(self, initialize=True, **kwargs):
...
+                                        except KeyError:
+                                            content["species"][gas.n_species+surf.species_index(sp.to_chemkin())]['coverage-dependencies'] = {sp.to_chemkin(): parameters}
```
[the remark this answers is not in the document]
bjkreitz: I agree with your comment and I changed the code accordingly.

bjkreitz:
> Just a few more things left to work through:
> 
> 1. Is it possible to remove the np_list definition in util.py? This feels like an unnecessary layer of abstraction that can perhaps be avoided by adding `.value_si` to some of the quantity variables

It was actually possible to remove the `np_list` definition. Turns out that I only had to change the code to 
`'enthalpy-coefficients': [quantity.Enthalpy(p) for p in parameters['enthalpy-coefficients']]`. I'm not entirely sure anymore why this was needed in the first place. 


> 2. ~I need to delve a little deeper into the solver.pyx code before I sign off on it~ I have poked around the matrix multiplications in the surface.pyx solver, checked some example matrices in a jupyter notebook, and convinced myself that this new addition is implemented correctly (I've attached the little explainer pdf I made for myself). ✔️
> 3. ~I want to see if it's possible to avoid adding the `import cantera` call in main.py. It's a big dependency to be adding just to get the index of the coverage-dependent species~ ✔️ yes this is possible. See suggestions below

Yes, resolved this comment

> 4. ~I was having trouble exercising the new code in main.py that writes the cantera file, so I want to confirm whether that's just a me problem or a sign of bigger trouble~ I was able to reach the code in the debugger, but there is one potential problem: thermo dependence is only written at the very end of model generation. So if the model generation gets cut short or never converges, then you end up with just chemkin files that have no coverage dependence. You'd have to run ck2yaml and then manually copy your thermo dependence into the cantera file. I think fixing this problem is outside the scope of this PR, but it's still something we should be aware of. ✔️

I would say that this is outside the scope of this PR, too. 

> 5. ~I need to test whether writing coverage-dependent thermo works for cross species interactions.~ ❌ this does not work. may be worth adding a unit test in addition to suggestions below

I still need to work on this part.

sevyharris:
✅ I've confirmed this now writes the cross-species interactions. I take back what I said earlier, and a unit test is probably not worth the effort here because the code is buried inside RMG's main loop and not easily accessible. 

❌ It seems the _as_dict tests for nasa and wilhoit don't match the new structure of coverage dependence in dictionary form. The roundtrip to and from dictionary works fine so I think it's just the test code that needs adjustment. 

(Here's how I'd rewrite it for wilhoit)
```

    def test_wilhoit_as_dict(self):
        """
        Test that a Wilhoit object can be converted to a dictionary representation properly
        """
        wilhoit_dict = self.wilhoit.as_dict()

        assert wilhoit_dict["thermo_coverage_dependence"].keys() == self.thermo_coverage_dependence.keys()
        sp_name = list(self.thermo_coverage_dependence.keys())[0]
        assert wilhoit_dict['thermo_coverage_dependence'][sp_name]['model'] == self.thermo_coverage_dependence[sp_name]['model']
        enthalpy_list = wilhoit_dict['thermo_coverage_dependence'][sp_name]['enthalpy-coefficients']
        for i in range(len(enthalpy_list)):
            assert enthalpy_list[i]['units'] == self.thermo_coverage_dependence[sp_name]['enthalpy-coefficients'][i][1]
            assert enthalpy_list[i]['value'] == self.thermo_coverage_dependence[sp_name]['enthalpy-coefficients'][i][0]
        entropy_list = wilhoit_dict['thermo_coverage_dependence'][sp_name]['entropy-coefficients']
        for i in range(len(entropy_list)):
            assert entropy_list[i]['units'] == self.thermo_coverage_dependence[sp_name]['entropy-coefficients'][i][1]
            assert entropy_list[i]['value'] == self.thermo_coverage_dependence[sp_name]['entropy-coefficients'][i][0]

        assert wilhoit_dict == {
            "comment": "C2H6",
            "B": {"units": "K", "class": "ScalarQuantity", "value": 1068.68},
            "Tmin": {"units": "K", "class": "ScalarQuantity", "value": 300.0},
            "H0": {
                "units": "kJ/mol",
                "class": "ScalarQuantity",
                "value": -782.292041536,
            },
            "Tmax": {"units": "K", "class": "ScalarQuantity", "value": 3000.0},
            "S0": {
                "units": "J/(mol*K)",
                "class": "ScalarQuantity",
                "value": -984.93235312,
            },
            "a1": -16.3067,
            "a0": 0.0977518,
            "a3": -12.6785,
            "a2": 26.2524,
            "Cp0": {
                "units": "J/(mol*K)",
                "class": "ScalarQuantity",
                "value": 33.257888,
            },
            "CpInf": {
                "units": "J/(mol*K)",
                "class": "ScalarQuantity",
                "value": 178.76114800000002,
            },
            "class": "Wilhoit",
            "thermo_coverage_dependence": {
                '1 O u0 p2 c0 {2,D} \n 2 X u0 p0 c0 {1,D}': {
                    'model': 'polynomial',
                    'enthalpy-coefficients': [{'class': 'ScalarQuantity',
                        'value': 1.0,
                        'units': 'J/mol'},
                        {'class': 'ScalarQuantity', 'value': 2.0, 'units': 'J/mol'},
                        {'class': 'ScalarQuantity', 'value': 3.0, 'units': 'J/mol'}],
                    'entropy-coefficients': [{'class': 'ScalarQuantity',
                        'value': 1.0,
                        'units': 'J/(mol*K)'},
                        {'class': 'ScalarQuantity', 'value': 2.0, 'units': 'J/(mol*K)'},
                        {'class': 'ScalarQuantity', 'value': 3.0, 'units': 'J/(mol*K)'}]
                }
            }
        }
```


and for NASA:
```
    def test_nasa_as_dict_full(self):
        """
        Test that NASA.as_dict functions properly with all attributes
        """
        nasa_dict = self.nasa.as_dict()
        assert nasa_dict["E0"]["value"] == self.E0
        assert nasa_dict["Tmin"]["value"] == self.Tmin
        assert nasa_dict["Tmax"]["value"] == self.Tmax
        assert nasa_dict["thermo_coverage_dependence"].keys() == self.thermo_coverage_dependence.keys()
        sp_name = list(self.thermo_coverage_dependence.keys())[0]
        assert nasa_dict['thermo_coverage_dependence'][sp_name]['model'] == self.thermo_coverage_dependence[sp_name]['model']
        enthalpy_list = nasa_dict['thermo_coverage_dependence'][sp_name]['enthalpy-coefficients']
        for i in range(len(enthalpy_list)):
            assert enthalpy_list[i]['units'] == self.thermo_coverage_dependence[sp_name]['enthalpy-coefficients'][i][1]
            assert enthalpy_list[i]['value'] == self.thermo_coverage_dependence[sp_name]['enthalpy-coefficients'][i][0]
        entropy_list = nasa_dict['thermo_coverage_dependence'][sp_name]['entropy-coefficients']
        for i in range(len(entropy_list)):
            assert entropy_list[i]['units'] == self.thermo_coverage_dependence[sp_name]['entropy-coefficients'][i][1]
            assert entropy_list[i]['value'] == self.thermo_coverage_dependence[sp_name]['entropy-coefficients'][i][0]
        assert nasa_dict["comment"] == self.comment
        assert tuple(nasa_dict["polynomials"]["polynomial1"]["coeffs"]["object"]) == tuple(self.coeffs_low)
        assert tuple(nasa_dict["polynomials"]["polynomial2"]["coeffs"]["object"]) == tuple(self.coeffs_high)
        assert nasa_dict["polynomials"]["polynomial1"]["Tmin"]["value"] == self.Tmin
        assert nasa_dict["polynomials"]["polynomial1"]["Tmax"]["value"] == self.Tint
        assert nasa_dict["polynomials"]["polynomial2"]["Tmin"]["value"] == self.Tint
        assert nasa_dict["polynomials"]["polynomial2"]["Tmax"]["value"] == self.Tmax
```

Anyways, I think it should be ready for final approval after those tests are passing again.

bjkreitz:
Thanks for the feedback, Sevy! I think there is still something broken with formatting of the coverage dependence parameters. The output yaml files are screwed up and I have to investigate what causes this issue.

rwest:
Hmm - this is going to cause work for https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2800.
Not saying to delay merging this, just cross-linking and highlighting for that effort....

rmgpy/thermo/nasa.pyx
```diff
@@ -243,7 +247,7 @@ cdef class NASA(HeatCapacityModel):
         """
         A helper function used when pickling an object.
         """
-        return (NASA, (self.polynomials, self.Tmin, self.Tmax, self.E0, self.Cp0, self.CpInf, self.label, self.comment))
+        return (NASA, (self.polynomials, self.Tmin, self.Tmax, self.E0, self.Cp0, self.CpInf, self.label, self.thermo_coverage_dependence, self.comment))
 
     cpdef dict as_dict(self):
         """
```
rwest, line 253: Does as_dict include the thermo_coverage_dependence?
Because it's stored internally as _thermo_coverage_dependence prefixed with an underscore, it might not be. I can't guess from just inspecting the code.

rmgpy/thermo/nasa.pyx
```diff
@@ -399,6 +419,7 @@ cdef class NASA(HeatCapacityModel):
             poly.change_base_entropy(deltaS)
         return self
 
+    # need to modify this to include the thermo coverage dependence
```
rwest, line 424: What's the plan for adding to the to_cantera method?

rmgpy/thermo/nasa.pyx
rwest: Claude Code tells me that it will be included. In fact, it will always be inculuded, even when it should be None, it'll instead be an empty dictionary.
This, apparently, is also true for `__repr__` calls. I'll make a commit that tweaks this.

rmgpy/util.py
```diff
@@ -33,6 +33,7 @@
 import shutil
 import time
 from functools import wraps
+import numpy as np
```
sevyharris:
```suggestion
```

Sorry, this is very nitpicky but can we delete this line now that the np requirement has been eliminated?

rmgpy/rmg/main.py
bjkreitz: All our DFT was done at 1/9 and this is currently also the standard in the literature for metal surfaces.

bjkreitz:
There is still a bug with the yaml writer. The format of the yaml files is different from the standard RMG output.  
Screenshot 2026-04-09 at 9 37 17 AM

rwest:
I just pushed something that I hope addresses the YAML layout issue....?
I also squashed the two minor fixup commits on util.py

rmgpy/thermo/nasa.pyx
rwest: I've addressed this in #2800 (on a commit I'll push soon)

rmgpy/rmg/main.py
rwest: (will be) addressed in #2800

rwest:
This yaml was made by Sevy's notebook from #2918, running on my latest commit on this branch, to show the new yaml syntax/formatting:
Screenshot 2026-04-09 at 2 37 47 PM

Screenshot 2026-04-09 at 2 37 24 PM


At first glance it is better (than it was).

But it's still not ideal.
This reaction from the non-cov-dep:
Screenshot 2026-04-09 at 2 44 17 PM
became this in the cov-dep:
Screenshot 2026-04-09 at 2 44 40 PM

rwest:
I changed the way we edit the cantera yaml file. Now it makes minimal changes - most of the file remains intact.

Here's the new diff between the non-cov-dep and the cov-dep:

```diff
 diff temp/base_rmg_run/cantera/chem_annotated.yaml temp/cov_dep_rmg_run/cantera/chem_annotated.yaml                           
5c5
< date: Thu, 09 Apr 2026 18:25:54 -0400
---
> date: Thu, 09 Apr 2026 18:32:50 -0400
24c24,25
<   thermo: ideal-surface
---
>   thermo: coverage-dependent-surface
>   reference-state-coverage: 0.11
526a528,533
>   coverage-dependencies:
>     COX(23):
>       model: polynomial
>       enthalpy-coefficients: [30103.426817162992, -31164.765583152708, 85871.95470280468]
>       entropy-coefficients: [10.70987300225991, -6.252250185103083, -15.72711080512041]
>       units: {energy: J, quantity: mol}
```

Sevy's notebook in #2918 seems to show it still works. 
Are we good to merge?

Issue #2642, opened 2024-03-25
RMG cannot generate resonance structures for the fragment "C[C]1C=CC=CC1=CC=CR"


donerancl:
### Bug Description
Resonance structures cannot be generated from specific fragment "C[C]1C=CC=CC1=CC=CR" , despite success for the same smiles without the "R" and for other fragments with resonance.

### How To Reproduce
```
from rmgpy.molecule.fragment import Fragment
f = Fragment().from_smiles_like_string("C[C]1C=CC=CC1=CC=CR")
f.generate_resonance_structures()
```
result
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/tmp/ipykernel_1087842/3088769613.py in <module>
      1 f = Fragment().from_smiles_like_string("C[C]1C=CC=CC1=CC=CR")
----> 2 f.generate_resonance_structures()

~/RMG-Py/rmgpy/molecule/molecule.cpython-37m-x86_64-linux-gnu.so in rmgpy.molecule.molecule.Molecule.generate_resonance_structures()

~/RMG-Py/rmgpy/molecule/molecule.cpython-37m-x86_64-linux-gnu.so in rmgpy.molecule.molecule.Molecule.generate_resonance_structures()

~/RMG-Py/rmgpy/molecule/resonance.cpython-37m-x86_64-linux-gnu.so in rmgpy.molecule.resonance.generate_resonance_structures()

~/RMG-Py/rmgpy/molecule/resonance.cpython-37m-x86_64-linux-gnu.so in rmgpy.molecule.resonance.generate_resonance_structures()

~/RMG-Py/rmgpy/molecule/resonance.cpython-37m-x86_64-linux-gnu.so in rmgpy.molecule.resonance._generate_resonance_structures()

~/RMG-Py/rmgpy/molecule/resonance.cpython-37m-x86_64-linux-gnu.so in rmgpy.molecule.resonance._generate_resonance_structures()

~/RMG-Py/rmgpy/molecule/resonance.cpython-37m-x86_64-linux-gnu.so in rmgpy.molecule.resonance.generate_kekule_structure()

~/RMG-Py/rmgpy/molecule/resonance.cpython-37m-x86_64-linux-gnu.so in rmgpy.molecule.resonance.generate_kekule_structure()

AttributeError: 'NoneType' object has no attribute 'label'
```

### Expected Behavior
successful molecule example:
```
[...]

PR #2647, opened 2024-04-02
Fix Fragment Resonance Structure Bug

donerancl:
Begins to address [Issue #2642](https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2642)
### Motivation or Problem

### Description of Changes

### Testing

### Reviewer Tips

### DIFF
--- a/rmgpy/molecule/resonance.py
+++ b/rmgpy/molecule/resonance.py
@@ -62,6 +62,7 @@ from rmgpy.molecule.adjlist import Saturator
 from rmgpy.molecule.graph import Vertex
 from rmgpy.molecule.kekulize import kekulize
 from rmgpy.molecule.molecule import Atom, Bond, Molecule
+from rmgpy.molecule.fragment import CuttingLabel
 
 
 def populate_resonance_algorithms(features=None):
@@ -827,7 +828,7 @@ def generate_kekule_structure(mol):
     cython.declare(atom=Vertex, molecule=Graph)
 
     for atom in mol.atoms:
-        if not isinstance(atom, Atom):
+        if isinstance(atom,CuttingLabel):
             continue
         if atom.atomtype.label == 'Cb' or atom.atomtype.label == 'Cbf':
             break


JacksonBurns:
Hmm so at face value this looks reasonable, and all of the tests pass...

`kekulize` is in two places in RMG, the `molecule.kekulize` method and the `kekulize` function in `kekulize.pyx` - the former simply calls the latter on itself.

I looked around the codebase for other uses of `kekulize` and found this: https://github.com/ReactionMechanismGenerator/RMG-Py/blob/a458b1deb3c26f69cff627396e3a635538c1cff6/rmgpy/data/transport.py#L404

which makes it seem like, in other cases, if kekulization fails we expect it to just throw the error and stop execution.

This `generate_kekule_structure` function is the only place where we actually catch and deal with the error.

Overall I am somewhat unsure if this is an acceptable change.. @hwpang could you take a look at this?

donerancl:
I think I found the deeper issue. `generate_kekule_structures` loops through atoms and checks for benzene/fused benzene atoms. If it finds one, it immediately tries kekulization. The error arises when `generate_kekule_structures` checks the `atom.atomtype.label` and finds `None`.

donerancl:
Ok, I fixed the check for `CuttingLabel`s in `generate_kekule_structures`, but any input on whether it is good practice is welcome @JacksonBurns

JacksonBurns (changes requested):
RMG subgroup discussion - better to directly test if `isinstance(atom, CuttingLabel`. Makes it more obvious what this does.

rmgpy/molecule/resonance.py
```diff
@@ -827,7 +828,7 @@ def generate_kekule_structure(mol):
     cython.declare(atom=Vertex, molecule=Graph)
 
     for atom in mol.atoms:
-        if type(atom) != type(Atom()):
+        if not isinstance(atom,CuttingLabel):
```
JacksonBurns, line 831: I think it should be just plain `isinstance` instead of `not isinstance`? i.e. we should skip the atom if it is a cutting label

JacksonBurns:
This looks good! Could you squash these commits into one (since some of them are just undoing the previous ones) and then re-request a review? Thanks!

JacksonBurns (approved):
LGTM! Thanks for the patch

PR #2648, opened 2024-04-03
Run Complete CI on MacOS and Ubuntu

JacksonBurns:
Right now we run the unit tests and Julia install on only Ubuntu, since historically the MacOS runners weren't capable for hardware reasons. This is checking if that is still the case.


JacksonBurns:
@hwpang I've set up the MacOS runners to actually install Julia and execute the tests, but frustratingly a unit test for the arkane example fails on Mac. Could you take a look at: https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/8556581181/job/23446739913?pr=2648#step:10:15553

JacksonBurns:
Closing for the time being - significant changes coming in #2687 will require re-doing this anyway.

Issue #2649, opened 2024-04-10
ValueError: 'site' is not in list when using Deutschmann_Ni as seed mechanism


ccaitlien:
### Bug Description
In my RMG input file, I'm trying to use the Surface/Methane/Deutschmann_Ni reaction library as a seed mechanism for a simulation involving a SurfaceReactor but I'm getting a ValueError saying that 'site' is not in list. This is what the full traceback looks like:
```  
  File "/home/RMG-Py/rmg.py", line 118, in <module>
    main()
  File "/home/RMG-Py/rmg.py", line 112, in main
    rmg.execute(**kwargs)
  File "/home/RMG-Py/rmgpy/rmg/main.py", line 760, in execute
    self.initialize(**kwargs)
  File "/home/RMG-Py/rmgpy/rmg/main.py", line 624, in initialize
    self.reaction_model.add_seed_mechanism_to_core(seed_mechanism, react=False)
  File "/home/RMG-Py/rmgpy/rmg/model.py", line 1692, in add_seed_mechanism_to_core
    self.add_reaction_to_core(rxn)
  File "/home/RMG-Py/rmgpy/rmg/model.py", line 1502, in add_reaction_to_core
    self.core.phase_system.interfaces[frozenset({"Default", "Surface"})].add_reaction(rxn, species_names, rms_species_list)
  File "/home/RMG-Py/rmgpy/rmg/reactors.py", line 340, in add_reaction
    self.reactions.append(to_rms(rxn, species_names=species_names, rms_species_list=rms_species_list))
  File "/home/RMG-Py/rmgpy/rmg/reactors.py", line 754, in to_rms
    reactantinds = [species_names.index(spc.label) for spc in obj.reactants]
  File "/home/RMG-Py/rmgpy/rmg/reactors.py", line 754, in <listcomp>
    reactantinds = [species_names.index(spc.label) for spc in obj.reactants]
ValueError: 'site' is not in list
```

### How To Reproduce
I'm using the methane steam reforming example in the RMG docs [here](https://reactionmechanismgenerator.github.io/RMG-Py/users/rmg/surfaces.html#example-input-file-methane-steam-reforming), except I'm using Surface/Methane/Deutschmann_Ni as a seed mechanism instead of a reaction library:

```
database(
    thermoLibraries=['surfaceThermoNi111', 'surfaceThermoPt111', 'primaryThermoLibrary', 'thermo_DFT_CCSDTF12_BAC'], 
[...]

PR #2653, opened 2024-04-20
FIx Handling of Initial Species with Seed Mechanisms

mjohnson541:
In #2649 it seems like we are having issues with adding seed mechanism containing a matching initial species while the initial species are in the edge. 

Because of how we process creating the initial core-edge and running estimation we add the initial species to the edge first later add the seed mechanisms and reaction libraries and lastly add the initial species to core and react them. 

In #2649 we run into a problem because the function for adding seed mechanisms searches for "site" and finds it (it is in the edge), since it finds it, it seems to assume it doesn't need to do anything which becomes a problem when it tries to add the associated reaction directly to core since "site" isn't in the core. 

I have modified add_seed_mechanism_to_core to track species that are in the edge and add them to core before adding the reactions. I have modified enlarge to be okay with reacting a species that has already been added to core.

### DIFF
--- a/rmgpy/rmg/model.py
+++ b/rmgpy/rmg/model.py
@@ -635,7 +635,10 @@ class CoreEdgeReactionModel:
                     display(new_species)  # if running in IPython --pylab mode, draws the picture!
 
                 # Add new species
-                reactions_moved_from_edge = self.add_species_to_core(new_species)
+                if new_species not in self.core.species:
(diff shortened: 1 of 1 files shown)


mjohnson541:
I tested this on the example in #2649 and it doesn't seem to be doing anything notable in the regression tests. However, this is a rather sensitive and opaque part of RMG.

JacksonBurns (approved):
Thanks for working on this and taking a look over the regression tests! I'll merge now!

PR #2655, opened 2024-05-03
Test Regression Example added in 2514

JacksonBurns:
Double checking that https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2514 was successful.

This has to be done after the fact because we had to wait for `main` to run the CI to have a reference result.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,4 +1,4 @@
-# RMG Code of Conduct
+# RMG Code of Conduct :+1:
 
 ## Our Pledge
 


JacksonBurns:
@sevyharris looks like there is a small error with the new regression test, could you take a look?

sevyharris:
> @sevyharris looks like there is a small error with the new regression test, could you take a look?

Here's my patch for the problem:
https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2656

We'll see in about an hour if it actually fixes the minimal_surface regression test.

JacksonBurns:
Closing in favor of #2656

PR #2656, opened 2024-05-03
Patch for surface regression diffmodels

sevyharris:
Patch for surface regression diffmodels

### Motivation or Problem
The .github actions script calling scripts/checkModels.py does not use the correct arguments or filenames for checking a surface model. So this adds a check for whether the existence of surface files to rmgpy/tools/diffmodels.py

### Description of Changes
Rather than changing this part of https://github.com/ReactionMechanismGenerator/RMG-Py/blob/main/.github/workflows/CI.yml ...
```
# Compare the edge and core
            if python-jl scripts/checkModels.py \
                "$regr_test-core" \
                $REFERENCE/"$regr_test"/chemkin/chem_annotated.inp \
                $REFERENCE/"$regr_test"/chemkin/species_dictionary.txt \
                test/regression/"$regr_test"/chemkin/chem_annotated.inp \
                test/regression/"$regr_test"/chemkin/species_dictionary.txt
            then
```
... I think it would be easier to add a check in Python to see if the surface files exist and chem_annotated.inp does not, and then make appropriate changes from there.

### Testing
I ran the checkModels.py script locally and confirmed these changes fix the problem for the minimal_surface example without breaking the minimal example.

**Even if this fix works, the minimal_surface regression test for this PR is not expected to pass because there are no stable regression edge files to compare against.**

### Reviewer Tips
Let me know if you have thoughts about the level at which this bug should be fixed. In theory, we could rewrite the CI.yml to call the surface tests separately from the others, but that sounds like a pain. Alternatively, we could move the check for surface files inside the checkModels.py script instead of adding it to rmgpy/tools/diffmodels.py

### DIFF
--- a/rmgpy/tools/diffmodels.py
+++ b/rmgpy/tools/diffmodels.py
@@ -344,6 +344,23 @@ def execute(chemkin1, species_dict1, thermo1, chemkin2, species_dict2, thermo2, 
     model1 = ReactionModel()
     model2 = ReactionModel()
 
+    # check if this is supposed to be comparing surface mechanisms
+    chemkin_gas1 = chemkin1.replace('.inp', '-gas.inp')
+    chemkin_surface1 = chemkin1.replace('.inp', '-surface.inp')
+    if not os.path.exists(chemkin1) and \
+            os.path.exists(chemkin_surface1) and \
+            os.path.exists(chemkin_gas1):
+        chemkin1 = chemkin_gas1
+        kwargs['surface_path1'] = chemkin_surface1
+
+    chemkin_gas2 = chemkin2.replace('.inp', '-gas.inp')
+    chemkin_surface2 = chemkin2.replace('.inp', '-surface.inp')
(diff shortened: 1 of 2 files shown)


sevyharris:
Although this is still failing with:
```
FileNotFoundError: [Errno 2] No such file or directory: 'stable_regression_results/minimal_surface/chemkin/species_edge_dictionary.txt'
```

This failure is expected because the last run of the regression tests on main didn't produce the edge files. It's the same expected error as when we first pushed the minimal_surface regression test, so I think this is ready for review.

sevyharris:
> If we merge this PR would that enable #2657 to pass normally (assuming the bug can be fixed)?

Yes. And then https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2658 is my fix for that bug.

JacksonBurns (approved):

JacksonBurns:
Great - will send this in now.

PR #2662, opened 2024-05-16
Send P in SI units to Chebyshev.fit_to_data() when reversing a rate

alongd:
The Chebyshev.fit_to_data() method treats the Pmin and Pmax args as SI (Pa units), see here for example: `self.Pmin = (Pmin * 1e-5, "bar")`. Make sure we transmit the arguments in SI units and not in the arbitrary units they were defined in.


JacksonBurns:
I think the CI failed because the network dropped (https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/9134588671) I will restart it.

PR #2667, opened 2024-06-04
Garbage Commit to Trigger CI (see #2663)

JacksonBurns:
Title

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -220,7 +220,7 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          run_id=$(gh run list -R ReactionMechanismGenerator/RMG-Py --workflow="Continuous Integration" --branch main --event schedule --limit 1 --json databaseId --jq '.[0].databaseId')
+          run_id=$(gh run list -R ReactionMechanismGenerator/RMG-Py --workflow="Continuous Integration" --branch main --limit 1 --json databaseId --jq '.[0].databaseId')
           echo "CI_RUN_ID=$run_id" >> $GITHUB_ENV
       - name: Retrieve Stable Regression Results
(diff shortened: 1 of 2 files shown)


alongd:
This one fails after >199 min (one of them at 212, but maybe it's a glitch of the 200 min rule)
In the [previous PR](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2663) we discussed, we know that the regression tests did run originally, and only lately we get this strange behavior.
I'm also at lost, can suspect that one of the dependencies has changes versions, but have no lead.

rwest:
Looks like you're making progress! If you need to commit to main to get the reference artifact uploaded, feel free IMHO.
I noted there's also a warning to update `conda-incubator/setup-miniconda@v2` - could be worth doing while you're at it.

JacksonBurns:
@rwest thanks for the note, have done that here. Interestingly, changing `conda-incubator` to v3 allowed the `macos-latest` runner to resolve to macos-14, which runs on the new apple hardware instead of the intel chips. This then caused the CI to fail because none of the RMG channel conda packages are available for the apple hardware.

As a fix now I have set the runner to macos-13, which runs on intel. In the future will would need conda binaries on our channel that are built for apple hardware.

Rather than merge this PR I will just get it to fail the way I want it to (it should discover the correct run of the CI), and then I will move these changes to a branch on my fork, open a new PR, and merge that if it 'fails right'. Just want to make sure this works from forks.

JacksonBurns:
Yes for sure. #2641 would be a big part of that, since it should allow us to rebuild the RMG binaries for the first time in a while.

JacksonBurns:
Closing in favor of #2669

JacksonBurns:
> Great.
> 
> Yes, let's aim for Intel Mac and Mac Silicon binaries for all conda dependencies - maybe this summer? (Not this PR)

I have started this here: https://github.com/ReactionMechanismGenerator/conda-recipes/pull/18

PR #2669, opened 2024-06-07
github actions CI maintainence

JacksonBurns:
- increment miniconda setup to v3 from v2
 - set macos build runner to use intel macs rather than m1, since we don't have binaries for m1
 - switch download artifact step to use the official github one, which now allows grabbing artifacts from other runs

Resolves https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2611

See https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2667 for more history

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -51,7 +51,7 @@ env:
 
 jobs:
   build-osx:
-    runs-on: macos-latest
+    runs-on: macos-13
     # skip scheduled runs from forks
     if: ${{ !( github.repository != 'ReactionMechanismGenerator/RMG-Py' && github.event_name == 'schedule' ) }}
     defaults:
@@ -63,7 +63,7 @@ jobs:
 
       # configures the mamba environment manager and builds the environment
       - name: Setup Mambaforge Python 3.7
-        uses: conda-incubator/setup-miniconda@v2
+        uses: conda-incubator/setup-miniconda@v3
         with:
           environment-file: environment.yml
           miniforge-variant: Mambaforge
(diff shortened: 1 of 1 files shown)


JacksonBurns:
This should fail at the 'download stable regression results' step, complaining that it could not find the artifact, possibly specifying that the reason it couldn't find it is because it was upload with a previous version of the upload artifact action. That would be resolved by ignoring this failure and merging anyway, then letting the CI run on main.

.github/workflows/CI.yml
```diff
@@ -215,23 +215,24 @@ jobs:
         run: mkdir stable_regression_results
 
       # Retrieve Stable Results for reference
-      # Will need to use this -> https://github.com/dawidd6/action-download-artifact
+      - name : ChatGPT said this would work - Get most recent CI run ID
```
rwest, line 218: I suggest remove the chatGPT bit

rwest (commented):
Looks good - I'd it works as expected. 
One suggested change. 
Do we need to do anything to RMG-database also?

JacksonBurns:
> Do we need to do anything to RMG-database also?
yes! thanks for the reminder

.github/workflows/CI.yml
JacksonBurns: Haha forgot i had that in there, removing now

JacksonBurns:
@rwest could you take a look at https://github.com/ReactionMechanismGenerator/RMG-database/pull/652? It's the database half of this PR. Both should 'fail' in the same way.

PR #2635, opened 2024-03-15
Upgrading to Python 3.11

JacksonBurns:
This PR will (eventually) move RMG from Python 3.7 to 3.11/2. Python 3.7 is deprecated, which reason enough on its own to move, but doing so will also unblock a number of other improvements.

Opening this now to track the progress of the change and so that I can use the CI to test what packages need to be rebuilt, etc. Will edit this description and clean up the git history when things actually stand a chance of working.

PR #2671, opened 2024-06-07
Switch `muq`, `lpsolve55`, and `RingDecomposerLib` from `rmg` to `conda-forge`

jonwzheng:
This PR changes a number of `rmg` channel packages to their official `conda-forge` variants maintained by other people:
 - `rmg::muq2` -> `conda-forge::muq` (different name, same version)
 - `rmg::lpsolve55` -> `conda-forge::lpsolve55`
 - `rmg::py_rdl` -> `conda-forge::ringdecomposerlib-python` (different name, same package and version)

The first change also required adding a new line to the `Makefile` to ensure that correct compiler is used when installing RMG.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -81,13 +81,13 @@ dependencies:
   - conda-forge::gprof2dot
   - conda-forge::numdifftools
   - conda-forge::quantities
+  - conda-forge::muq
+  - conda-forge::lpsolve55
+  - conda-forge::ringdecomposerlib-python
 
 # packages we maintain
-  - rmg::lpsolve55
-  - rmg::muq2
   - rmg::pydas >=1.0.3
   - rmg::pydqed >=1.0.3
-  - rmg::pyrdl
   - rmg::pyrms
   - rmg::symmetry
 
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,9 @@
 #
 ################################################################################
 
+CC=gcc
+CXX=g++
+
 .PHONY : all minimal main solver check pycheck arkane clean install decython documentation test q2dtor
 
 all: pycheck main solver check


JacksonBurns:
So this simple change actually dramatically changed the conda environment, see the diff here: https://www.diffchecker.com/RJ4dcsLN/

JacksonBurns:
@jonwzheng I suspect that the issue is `clang` - the new version of `muq2` results in `clang` being installed and used for compiling RMG, whereas we advise using `gcc`. Going to tack on a commit to try and deal with this.

JacksonBurns:
My trick worked, so the Mac build works again, if we do want to stick with this, we should add the specification of the environment variables to the docs.

I'm getting ahead of myself here, but `muq` only has conda binaries supporting up to Python 3.10 - when we want to make the transition to 3.11+ (#2635) this would be a problem.
We use `muq` in one place (the global uncertainty module) and we list it as 'optional' in the docs. I'm actually tempted to say that we should actually _remove_ it from the environment file and instead add installation instructions to the docs for those who want it.

Long term we should really think about the future of `muq` - the `conda-forge` team considers the conda recipe "comatose": https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/5231#issue-2022972006

JacksonBurns:
Summarizing some offline discussion:
 - we are going to poke the developers of `muq` to see if they are willing to take up the conda binary building again and get versions out that supporting python 3.11+
 - we can make this change for now anyway since all of the supported python versions are present
 - when we upgrade python versions, we can either add retroactive support for Python 3.10 and/or just require users to install `muq` themselves

Regardless, @jonwzheng can you add a line in the docs about specifying `CC` and `CXX` to avoid Mac using a compiler other than `gcc`?

jonwzheng:
Just sent an email poking the `muq2` team. Will work on updating the docs shortly.

jonwzheng:
Instead of requiring users to specify compiler, do you know if we could just modify the Makefile to use those compilers by default when `make` is called?

JacksonBurns:
Good point, I think we could just put `CC=gcc` in the top of the `Makefile` actually

JacksonBurns:
Nice, look like this worked!

I think the only outstanding thing to do for this PR is to get another review on it. Could you request someone @jonwzheng?

When we do the Python upgrade we will also target Python 3.10 so that the `muq` binaries can be installed, or else update the build instructions to tell people to install `muq` some other way.

jonwzheng:
sounds good, thanks! NGL, I wasn't confident that change would work on the first pass through...

JacksonBurns:
After all of the changes we've made that _should've_ worked on the first try, I am willing to accept this nice little treat from the universe.

JacksonBurns:
@jonwzheng this PR as well as #2680 and #2681 all basically do the same thing - do you want me to just merge those two into this one, and then this one into `main` to avoid a bunch of extra rebasing. All three of them seem to be working, BTW

jonwzheng:
sure, that sounds good

JacksonBurns:
@rwest could you take a look at this? We just changed some packages to the `conda-forge` version, and one minor change to the Makefile to ensure the correct C compiler is used. More thorough summary at the top of the PR, if needed. Thanks!

environment.yml
```diff
@@ -81,10 +81,11 @@ dependencies:
...
+  - conda-forge::lpsolve55
+  - conda-forge::ringdecomposerlib-python
+  - conda-forge::muq
 
 # packages we maintain
-  - rmg::lpsolve55
-  - rmg::muq2
   - rmg::pydas >=1.0.3
   - rmg::pydqed >=1.0.3
   - rmg::pyrdl
```
rwest, line 91: Was the intention to remove `rmg::py_rdl` ?

rwest (commented):
Was the intention to remove `rmg::py_rdl`? I don't see this change in the diff - only the addition of `conda-forge::ringdecomposerlib-python`.

environment.yml
JacksonBurns: Oops thanks for noticing, resolved a previous merge conflict incorrectly

JacksonBurns:
@rwest thanks for the quick review and bug catch! I've cleaned up the commit history and fixed the line.

rwest (approved):
I made a fresh conda environment with the latest environment.yml, followed the instructions for installing julia and RMS etc.,  and  ran a few examples. Seems to be working, on my  Intel Core i7 MacBook Pro running MacOS Sonoma 14.5. 
If this works on the CI too, then I assume it's good to go.

PR #2321, opened 2022-07-13
Create YAML file outputs in RMG for use in Cantera

Nora-Khalil:
### Motivation or Problem
Previously, RMG created CHEMKIN files as outputs when generating a mechanism. The final CHEMKIN file for a mechanism is then converted to a CTI file, which can be used in Cantera simulations. When dealing with more complex mechanisms, RMG has a problem converting the final CHEMKIN file to a CTI file (because of numerous unmarked duplicate reactions in CHEMKIN file). To fix this issue, the CHEMKIN file has to be manually edited before file conversion can take place. This is time-consuming and tedious. Additionally, CTI input files are depreciated in Cantera 2.5 and will be removed in Cantera 3.0. For those using Cantera, it would be beneficial if RMG could directly output Cantera-formatted YAML files.

**With the improvements proposed in this pull request, RMG can now directly output Cantera-formatted YAML files.**

### Description of Changes
These changes only work with Cantera 2.6 or greater. 

- new file: `RMG-Py/rmgpy/cantera.py` which creates a new Cantera-formatted YAML file each time another species is added to the mechanism, much like the CHEMKIN output files. Formats the YAML file depending on type of system (gas-phase v. catalysis). 
- changes to `RMG-Py/rmgpy/reaction.py`:
              - `to_cantera()` method updated so that Cantera objects are correctly created for Troe, SurfaceArrhenius, 
                  StickingCoefficient, and Lindemann reactions. 
              - `fix_reaction()` method added so the equations of StickingCoefficient reactions don't include the SMILES of the 
                  surface  species. 
- changes to `rmgpy/kinetics/falloff.pyx`: 
              - edited/added`set_cantera_kinetics()` and `to_cantera_kinetics()` in Lindemann and Troe classes
- changes to `rmgpy/kinetics/surface.pyx`: 
              - edited/added `set_cantera_kinetics()` and `to_cantera_kinetics()` in StickingCoefficient and SurfaceArrhenius 
                 classes
- changes to `rmgpy/species.py`:
              - edited `to_cantera()` to account for surface sites in surface reactions. 
 changes to `rmgpy/rmg/main.py`: 
              - edited `register_listeners()`, attached CanteraWriter

              
### Testing
Tested yaml_writer_addition branch on the following examples:
     - `examples/rmg/catalysis/methane_steam`
     - `examples/rmg/catalysis/ch4_o2`
     - `examples/rmg/minimal`
     - `examples/rmg/minimal_surface`

### Reviewer Tips
[...]

PR #2673, opened 2024-06-08
fix species_to_dict function to use correct parameter as reaction species

ntietje1:
### Motivation or Problem
The yaml_writer_addition branch no longer correctly adds reaction species to the Cantera yaml output file. Species block is always a list of nulls. This issue originated during a rebase of the branch.

### Description of Changes

* changes to `rmgpy/cantera_.py`:
    - `species_to_dict()` method updated to correctly use the parameter `obj` as the RMG species object.

### Testing
Tested on minimal example and minimal_surface example (minimal_surface errors unless commit [feb70e2](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2321/commits/feb70e275fe16e53a14788c83bc7b9404dc7510f) is reverted/fixed)

### Reviewer Tips
Comment out the temporary "raise exception" calls in `to_cantera_kinetics()` to test on any example that requires this method.

### DIFF
--- a/rmgpy/cantera_.py
+++ b/rmgpy/cantera_.py
@@ -302,18 +302,18 @@ def reaction_to_dicts(obj, spcs):
 
 def species_to_dict(obj, spc, names=None, label="solvent"):
     """
-    Takes an RMG species object (spc), returns a list of dictionaries
+    Takes an RMG species object (obj), returns a list of dictionaries
     for YAML properties. Also adds in the number of surface sites
     ('sites') to dictionary.
     """
 
     result_dict = dict()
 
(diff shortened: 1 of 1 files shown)


rwest (commented):
Yay. Thanks. I think we should rename some of these variables to reduce errors of that kind. Like obj becomes species and spc becomes list_of_all_species. 
Actually, do we need that variable at all? Maybe it should just be removed.

ntietje1:
Yes, I agree the variable names are confusing. All of the function parameters are unused besides `obj`, so I would think they could be removed and `obj` renamed to `species`?

PR #2595, opened 2024-01-23
Change method for loading energies from Gaussian log files to regex search to avoid errors 

donerancl:
### Motivation or Problem
Sometimes, in  gaussian log files, the line that contains the energy has extra characters (more than one line printed in the same line) that prevent the load_energy function in arkane/ess/gaussian.py from parsing the energy. (e.g., it will try to use an "=" as the energy)

### Description of Changes
Before: find the line that has "SCF Done:", split it, take the 4th index as the energy

After: find the line that has "SCF Done:", find the substring that matches "SCF Done: E(...) = {energyfloat}, split the substring, and take the last item

the same can be done for E2, MP2, E(CORR), CCSD(T), CBS-QB3 (0 K), E(CBS-QB3)=, CBS-4 (0 K)=, G3(0 K), G3 Energy=,G4(0 K), G4 Energy=, G4MP2(0 K), G4MP2 Energy= in the same function

### Testing
I plan to try the load_energy function on many gaussian log files, some which have extra characters in the line containing the energy

### Reviewer Tips
Suggestions for verifying that this PR works or other notes for the reviewer.

PR #2674, opened 2024-06-09
Patch `minimal_surface` Regression Test

JacksonBurns:
Edit 9/15/24: the runner is crashing due to OOM errors. There is a solution to this that involves reducing the memory usage in the regression tests, but implementing this solution is hard. We have opted for an easier solution: severely limit the size of the mechanism. This does not cover as much 'science' as we would like the regression test to do, but does at least offer some basic correctness checks. At this point this is preferable to not testing at all, as we are currently doing since the 'better' solution is quite challenging.

Original PR content follows:

### Problem
I discovered in #2595 and #2669 that the `minimal_surface` regression test comparison causes the runners to crash with exit code 143. To allow things to still be merged in the meantime, I have removed that test from the CI and intend to fix it in this PR.

### TODO:
 - diagnose the issue(s)
 - implement a patch
 - drop 862adb831fdb2e43f36f13c94704932b506dfea0
 - merge, hopefully passing all tests (unit and regression)

Calling in @sevyharris for some help on this one!

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -167,7 +167,7 @@ jobs:
         id: regression-execution
         timeout-minutes: 60
         run: |
-          for regr_test in aromatics liquid_oxidation nitrogen oxidation sulfur superminimal RMS_constantVIdealGasReactor_superminimal RMS_CSTR_liquid_oxidation RMS_liquidSurface_ch4o2cat fragment RMS_constantVIdealGasReactor_fragment;
+          for regr_test in aromatics liquid_oxidation nitrogen oxidation sulfur superminimal RMS_constantVIdealGasReactor_superminimal RMS_CSTR_liquid_oxidation RMS_liquidSurface_ch4o2cat fragment RMS_constantVIdealGasReactor_fragment minimal_surface;
           do
             if python-jl rmg.py test/regression/"$regr_test"/input.py; then
               echo "$regr_test" "Executed Successfully"
@@ -243,7 +243,7 @@ jobs:
         run: |
           exec 2> >(tee -a regression.stderr >&2) 1> >(tee -a regression.stdout)
           mkdir -p "test/regression-diff"
-          for regr_test in aromatics liquid_oxidation nitrogen oxidation sulfur superminimal RMS_constantVIdealGasReactor_superminimal RMS_CSTR_liquid_oxidation fragment RMS_constantVIdealGasReactor_fragment;
+          for regr_test in aromatics liquid_oxidation nitrogen oxidation sulfur superminimal RMS_constantVIdealGasReactor_superminimal RMS_CSTR_liquid_oxidation fragment RMS_constantVIdealGasReactor_fragment minimal_surface;
           do
             echo ""
             echo "### Regression test $regr_test:"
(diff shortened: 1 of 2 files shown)


JacksonBurns:
It's actually not that, but instead:
```bash
              if python-jl rmgpy/tools/regression.py \
                test/regression/"$regr_test"/regression_input.py \
                $REFERENCE/"$regr_test"/chemkin \
                test/regression/"$regr_test"/chemkin
```

sevyharris:
Not particularly helpful, but I pulled the latest from main and checked that the minimal_surface example runs without issue. I can call regression.py on the result (using the same chemkin folder for old and new) and it passes, so no there are no obvious problems with the surface example itself.

I'm going to create a second PR like this one so I can play around with the CI.yml and hopefully get some additional diagnostic info.

JacksonBurns:
Beat me to it! Thanks for running that. Given that it works locally, that leads me to believe that we are probably just running out of memory on the GitHub actions runners...

sevyharris:
It looks like you're right about the memory/model size being an issue since it passed with the size limits. ^

Richard has suggested that since it looks like it's failing after it's already generated the model, the problem could be that the regression is trying to compare too many time points. I will see if that's an obvious culprit and if so if it can be reduced to a more reasonable number.

JacksonBurns:
Sounds like a plan! Feel free to edit on this PR.

I will defer to your and Richard's judgement about how to reduce the computational power needed for this test while still keeping it scientifically meaningful. I've cut the number of species by about a factor of 10, which I am assuming is too much...

rwest:
> Richard has suggested that since it looks like it's failing after it's already generated the model, the problem could be that the regression is trying to compare too many time points. I will see if that's an obvious culprit and if so if it can be reduced to a more reasonable number.

Looking a little bit more at this (I was not recalling how the regression comparisons are done) it is also running Cantera simulations at this step,  via /rmgpy/tools/observablesregression.py which calls into rmgpy/tools/canteramodel.py. So it could be any of these steps that are taking the resources - running cantera or comparing the cantera results. Should probably do a bit more instrumentation before guessing the cause.

JacksonBurns:
@sevyharris I just realized we never added the `minimal_surface` regression test to RMG-database, so we should do that once we have something acceptable here.

JacksonBurns:
hi @sevyharris i realized this PR and https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2675 fell off my radar. Do we want to merge an 'easy' fix like the one that I have here? Or still try and figure out the root cause and get a 'better' fix?

sevyharris:
> hi @sevyharris i realized this PR and #2675 fell off my radar. Do we want to merge an 'easy' fix like the one that I have here? Or still try and figure out the root cause and get a 'better' fix?

Hey @JacksonBurns, I'm in favor of merging this version now. (I totally forgot about this too!) I'll try to come up with a fix to bring down the memory usage for testing without limiting the number edge species so severely, but it'd be great to have at least something in place. Thanks!

JacksonBurns:
@sevyharris and/or @rwest please review. This should fail CI since there are no "stable" results for this test to compare against, so I will bypass and merge the failing CI with an approval here. Thanks!

JacksonBurns:
I confirm that the CI is failing where expected.

sevyharris (approved):
Looks good to me!

PR #2679, opened 2024-06-13
fake PR to run coverage & tests

ntietje1:
Using this PR to generate coverage reports & run tests. If there is a better way to generate the coverage report please let me know.

### DIFF
--- a/rmgpy/kinetics/surface.pyx
+++ b/rmgpy/kinetics/surface.pyx
@@ -285,6 +285,34 @@ cdef class StickingCoefficient(KineticsModel):
 
         return string
 
+    def to_cantera_kinetics(self):
+        """
+        Converts the Arrhenius object to a cantera Arrhenius object
+
+        Arrhenius(A,b,E) where A is in units of m^3/kmol/s, b is dimensionless, and E is in J/kmol
+        """
+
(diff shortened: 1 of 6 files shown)


rwest (commented):
I think the dimensionality in this dict should basically be the power that is used for molecules (or moles). So cm5/mol2/s should be 2  ?

rmgpy/kinetics/surface.pyx
```diff
@@ -570,6 +601,55 @@ cdef class SurfaceArrhenius(Arrhenius):
...
+                                     's^-1': 0,
+                                     'm^2/(mol*s)': 1,
+                                     'm^4/(mol^2*s)': 2,
+                                     'cm^2/(mol*s)': 1,
+                                     'cm^4/(mol^2*s)': 2,
+                                     'm^2/(molecule*s)': 1,
+                                     'm^4/(molecule^2*s)': 2,
+                                     'cm^2/(molecule*s)': 1,
+                                     'cm^4/(molecule^2*s)': 2,
+                                     'cm^5/(mol^2*s)': 3,
```
rwest, line 624:
```suggestion
                                     'cm^5/(mol^2*s)': 2,
```

PR #2623, opened 2024-03-05
Reimplement Clar Optimization with Scipy MILP

xiaoruiDong:
### Motivation or Problem
Clar optimization was originally implemented with lpsolve55, potentially introducing difficulties in maintenance. This PR replaces lpsolve55 with `scipy.optimize.milp` with a hope to alleviate the burden in maintaining this external package. 

### Description of Changes
1. Replace lpsolve APIs with the scipy milp APIs. The new implementation potentially has a slightly better performance (due to vectorization, less data transfer, comparable solver performance, etc.) and improved readability.
2. Decouple the MILP solving step (as _solve_clar_milp ) from the MILP formulation step. The motivation is to avoid unnecessary computation. The original approach includes molecule analysis (specifically `get_aromatic_rings`) into the recursive calls. However, this is not necessary, as molecules are just copied over and not modified at all. Therefore, analyzing once is enough.

### Testing
No new tests are added. The replacement expects no failure in previous tests.

### Reviewer Tips
A short meeting with me @xiaoruiDong before the reviewing is preferred, which should help the reviewer understand changes.

PR #2680, opened 2024-06-13
switch `rmg::lpsolve55` to `conda-forge::lpsolve55` channel

JacksonBurns:
In the same vein as #2671, I just discovered that someone has been thanklessly maintaining an official `conda-forge` binary of `lpsolve55`: https://github.com/conda-forge/lpsolve55-feedstock

This PR will see if it works for us.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -81,9 +81,9 @@ dependencies:
   - conda-forge::gprof2dot
   - conda-forge::numdifftools
   - conda-forge::quantities
+  - conda-forge::lpsolve55
 
 # packages we maintain
-  - rmg::lpsolve55
   - rmg::muq2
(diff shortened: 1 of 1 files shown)


jonwzheng:
Thanks for working on this. Just noting here that efforts in #2623 are attempting to replace `lpsolve` with `scipy`. Do we still want to replace it given that it can be found on conda-forge?

JacksonBurns:
Following up on some offline conversation:
 - #2596 is also working on implementing linear solving with scipy, which combined with #2623 would remove our need of lpsolve55
 - the former is blocked by upgrading the Python version (#2635) which will be made easier by merging _this_ pr
 - thus, we merge this PR as a stopgap to (possibly) eventually removing lpsolve completely

PR #2682, opened 2024-06-14
Fix Failing Joback Method Unit Test

JacksonBurns:
#2680 and #2681 both fail this unit test despite not touching the associate code in any way:
```
            if comment:
>               assert transport_data.comment == comment
E               AssertionError: assert 'Epsilon & sigma estimated with Tc=500.53 K, Pc=48.02 bar (from Joback method)' == 'Epsilon & sigma estimated with Tc=500.53 K, Pc=47.11 bar (from Joback method)'
E                 - Epsilon & sigma estimated with Tc=500.53 K, Pc=47.11 bar (from Joback method)
E                 ?                                                 ^ ^^
E                 + Epsilon & sigma estimated with Tc=500.53 K, Pc=48.02 bar (from Joback method)
E                 ?                                                 ^ ^^

test/rmgpy/data/transportTest.py:181: AssertionError
```

...I _think_ because we updated the Joback parameters in RMG-database here: https://github.com/ReactionMechanismGenerator/RMG-database/pull/636

I've added a dummy commit just to trigger the CI and check that this fails again, and then I intend to update the expected values to what we are getting now.

cc @jonwzheng TL;DR: the Joback coefficient changes we made in RMG-database are failing a unit test on RMG-Py, this PR will fix it

### DIFF
--- a/test/rmgpy/data/transportTest.py
+++ b/test/rmgpy/data/transportTest.py
@@ -151,7 +151,7 @@ class TestTransportDatabase:
             [
                 "acetone",
                 "CC(=O)C",
-                Length(5.36421, "angstroms"),
+                Length(5.32979, "angstroms"),
                 Energy(3.20446, "kJ/mol"),
                 "Epsilon & sigma estimated with Tc=500.53 K, Pc=48.02 bar (from Joback method)",
             ],
--- a/test/rmgpy/data/transportTest.py
+++ b/test/rmgpy/data/transportTest.py
@@ -153,7 +153,7 @@ class TestTransportDatabase:
                 "CC(=O)C",
                 Length(5.36421, "angstroms"),
                 Energy(3.20446, "kJ/mol"),
-                "Epsilon & sigma estimated with Tc=500.53 K, Pc=47.11 bar (from Joback method)",
+                "Epsilon & sigma estimated with Tc=500.53 K, Pc=48.02 bar (from Joback method)",
             ],
             [
                 "cyclopenta-1,2-diene",


jonwzheng:
whoops! looks like we did need to consider the unit tests a bit more ...

from reading the code looks like we expect it to be 47.11 but it should be 48.02 (with our new code)

I will also double check against JRgui to see what that implementation of the joback parameters returns.

jonwzheng:
JRgui agrees that Pc of acetone should be 48.02 bar. We should just amend the unit test comment to be 48.02, i.e.:

```
        self.testCases = [
            [
                "acetone",
                "CC(=O)C",
                Length(5.36421, "angstroms"),
                Energy(3.20446, "kJ/mol"),
                "Epsilon & sigma estimated with Tc=500.53 K, Pc=48.02 bar (from Joback method)",
            ],
```

jonwzheng:
Looks like the transport properties are good to go now, but any idea about the CI failure @JacksonBurns? Looks like there's an issue with the GH actions artifact download, which is odd to me as it wasn't showing up before.

JacksonBurns:
Ah this is vexing, the CI is trying to download the `stable_regression_results` from the last run of the CI on the `main` branch, which was the scheduled run last night. Unforunately that run failed because of the unit tests that are being fixed in this PR...

@jonwzheng I think we just need to bypass merge this PR. Even if it does introduce regression test changes, we know that they are 'correct' - the equation that we fixed was unambiguously wrong.

jonwzheng:
Okay sounds good. Pleas remove the dummy commit ("DROPME") and I'll go ahead and approve.

JacksonBurns:
Done! Swag removed

jonwzheng (approved):
LGTM (noting that CI fail in this is okay. Unit tests have been fixed.)

PR #2691, opened 2024-07-12
Allow scratch dir as input for arkane symmetry calculation

hwpang:
### Motivation or Problem
Currently, the scratch directory in Arkane symmetry calculation is hardcoded to be "./scratch". It makes parallelization difficult as all the jobs will use the same scratch directory name.

### Description of Changes
I allow the scratch directory to be a user input with default being "./scratch".

### Testing
After this changes, my script was able to run in parallel without crashing due to scratch directory related errors.

### DIFF
--- a/arkane/ess/adapter.py
+++ b/arkane/ess/adapter.py
@@ -45,10 +45,11 @@ class ESSAdapter(ABC):
     An abstract ESS Adapter class
     """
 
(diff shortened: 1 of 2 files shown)


sevyharris (approved):
Looks good to me! I also ran the code as a sanity check and encountered no problems there.

PR #2693, opened 2024-07-18
Draw reversible arrow if reaction is reversible

sevyharris:
### Motivation or Problem
RMG reactions are typically reversible, but they are always drawn with a forward arrow → instead of the double harpoons. ⇌

### Description of Changes
This PR makes it the default to draw the arrow as described by the `reaction.reversible` property (→ for `False` and ⇌ for `True`). If someone wants to revert to the way things were and always draw the arrow as →, they can set the ReactionDrawer option 'drawReversibleArrow' to `False`.

### Testing
Here's some code to test out the results in a Jupyter notebook (note the last example writes to a file)

```
import rmgpy.reaction

# make an example reaction
my_reaction = rmgpy.reaction.Reaction()
my_reaction.reactants = [rmgpy.species.Species(smiles='CCCC'), rmgpy.species.Species(smiles='[OH]')]
my_reaction.products = [rmgpy.species.Species(smiles='O'), rmgpy.species.Species(smiles='[CH2]CCC')]

# show it with reversible arrows
display(my_reaction)

# show it with forward arrow
my_reaction.reversible = False
display(my_reaction)

# save an image of the reaction using the old method of always drawing just a forward arrow
my_reaction.reversible = True
rmgpy.molecule.draw.ReactionDrawer(options={'drawReversibleArrow': False}).draw(my_reaction, 'png', 'temp_reaction.png', )
```

### Reviewer Tips
Any other thoughts on changing the default from →? Possible consequences downstream?

### DIFF
--- a/rmgpy/molecule/draw.py
+++ b/rmgpy/molecule/draw.py
@@ -1644,6 +1644,7 @@ class ReactionDrawer(object):
         self.options = MoleculeDrawer().options.copy()
         self.options.update({
             'arrowLength': 36,
+            'drawReversibleArrow': True
         })
         if options:
             self.options.update(options)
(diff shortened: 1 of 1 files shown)


sevyharris:
(top arrow was pointing backward and bottom pointing forward, so I fixed that and force pushed)

sevyharris:
Hmm... for tall reactions the arrows are too far apart. 
image

Maybe I'll set a maximum space between the harpoons.

sevyharris:
Made the spacing a fixed number of pixels. Seems to have fixed it.

image

rwest (approved):
Looks good to me

rwest:
> but this example isn't too bad. It still appears to be in the approximate center of mass or so.

I agree that looks fine to me - near the center of mass. Unless you have examples of it looking much weirder (perhaps one big one small?)  _and_ being the fault of this PR,  I would call this good.

sevyharris:
Yeah, here are some larger images that look fine. Probably good to merge.

Untitled

Untitled

Untitled

PR #2698, opened 2024-07-30
Better drawing of poly-dentate adsorbates (with more than one surface site)

rwest:
### Motivation or Problem
Adsorbates with more than one surface site were drawn weirdly, with the second or third surface site often floating up in the air. Examples:

Screenshot 2024-08-07 at 5 12 52 PM

### Description of Changes
With this change, it temporarily forms a bond between the surface sites, so the layout algorithm sees them as part of a ring, then it rotates the molecule so they are at the bottom, and then it removes the bond once the coordinates are determined. 
**Update** after many commits, the latest (2024-08-07) version looks like this:
Screenshot 2024-08-07 at 5 13 14 PM

You'll see that poly-dentate adsorbates with fused rings are still not perfect but
1) these are hard to do, even by hand
2) these are not worse than they used to be
3) our chemistry is so bad for these that for anyone with these in their model, pretty pictures are the least of their worries

...so I suggest fixing those is left for another day.

### Testing
I made a unit test that draws a ton of polydentate adsorbates, then looked at them all (see above). 
This test set is mostly weirder than molecules that are likely to actually end up in a model core. 

_Keywords: multidentate bidentate polydentate_

### DIFF
--- a/test/rmgpy/molecule/drawTest.py
+++ b/test/rmgpy/molecule/drawTest.py
@@ -287,6 +287,8 @@ class TestMoleculeDrawer:
 25 X u0 p0 c0 {8,S}
 26 X u0 p0 c0 {3,S}
 """))
+        test_molecules.append(Molecule(smiles="*CC(*)(C*)OCC#*"))
+        test_molecules.append(Molecule(smiles="*CC(*)(C*)C*"))
         for number, molecule in enumerate(test_molecules, 1):
             path = f"test_polydentate_{number}.png"
             if os.path.exists(path):


bjkreitz:
Cool, thanks for making these changes, Richard! I'm going to review this soon.

rwest:
I think this is a bit better.
Screenshot 2024-08-01 at 12 27 55 PM

rwest:
I tried it on the `examples/rmg/catalysis/ch4_o2/input.py` example.
The edge species got a bit out of hand....

[new-species.zip](https://github.com/user-attachments/files/16459841/new-species.zip)
Screenshot 2024-08-01 at 1 01 51 PM


But was it much better before?
[old-species.zip](https://github.com/user-attachments/files/16460110/old-species.zip)
Screenshot 2024-08-01 at 1 13 41 PM

I think the "before" pictures are probably clearer, for n>4 ?

Suggestions (comments welcome):
- [ ] (in this PR) disable the new drawing method for n>4 sites ?  maybe even n>3?  
- [ ] (different issue) add a "maximum surface sites" constraint on the `examples/rmg/catalysis/ch4_o2/input.py` example!
- [ ] (different issue) get the `examples/rmg/catalysis/methane_steam/input.py` working again.

rwest:
I think this (skipping the new drawing method for n>4 surface sites) is a decent compromise?

Screenshot 2024-08-01 at 2 49 06 PM

bjkreitz (commented):
This is a nice improvement of RMG and the structures of the polydentate adsorbates look much better (with some exceptions for cyclic species). I think it is a good compromise to truncate this new drawing style to adsorbates with less than 4 surface sites (We shouldn't be looking at those anyway at the moment since all our estimates are nonsense).  

I generated a mechanism and everything works fine. I don't have suggestions for the code. There are a couple of structures that we could display even cleaner, e.g. like 
Pt C(C= Pt )CC# Pt (135)
 
Would it be possible to adjust the length of all bonds to the surface, so that they are all the same length? So that it looks more like 
Pt C(C= Pt )C= Pt (132)
I'm not sure if such a change would make the structure of other adsorbates worse, though.

rwest:
I see what you mean. But I can't currently think of a way to tweak the current algorithm, without a fairly specific "look for this exact case" change. I'll keep thinking though.

rwest:
I found a few more issues. Think I've now fixed them.
| Before this PR |  before the most recent fix | latest version |
|----|-----|----|
test_49| test_49_A| test_49_B
test_51| test_51_A| test_51_B
test_82| test_82_A |  test_82_B
test_40| test_40_A | test_40_B
test_39 | test_39_A| test_39_B


The last one is still ugly, but at least none of them are misleading any more.

rwest:
I updated the pictures at the top of the pull request with the latest version. I think we're good to go?

bjkreitz:
I tested this feature and generated mechanisms with RMG for ethane oxidation on Pt(111). The new drawing method slows down the mechanism generation slightly (04:43 compared to 04:30), but otherwise it works. 
While most structures look pretty good (it really looks better for most of the structures), there are a few exceptions where the drawing is wrong. Here are a few of the structures that are wrong:

bidentate_drawing

For some of the multidentates with 4 binding sites, there is an overlap of the X atoms. Is it possible to enforce a minimum distance between the X atoms? Alternatively, could we default to the old drawing style if we detect an overlap? I glimpsed at the code, but I am not sure if this is easily doable. Also, this goes away if we restrict the new drawing method to tridentates. It will be a while before we can handle anything with more than 3 binding sites accurately.  

I think that it will be challenging for the code to draw adsorbates with more than 6 heavy atoms (C, O, N) accurately with this new method. What do you think of adding this criterion as an easy fix? The old method is not better or easier to read, just because the 2D representation isn't great, but at least it is cleaner.

bjkreitz (commented):
I tested this feature and generated mechanisms with RMG for ethane oxidation on Pt(111). The new drawing method makes the structures look pretty good and it doesn't affect the mechanism generation time. 

Apologies for the earlier comment, I checked out this branch at a previous commit, which led to the wrong structures. Everything is cool! 

I found a duplicate code snippet that I removed. I don't think that it was there on purpose, but @rwest please review and confirm. Then I'll leave an approving review.

rwest:
I merged this in commit b30535b45e7033dd9f8ec1990ae8c5b2d4e1762d
Unfortunately I messed up the interactions between github and the git command line
and so github now thinks this pull request has zero commits and was closed,
rather than containing many commits and being merged.
Sorry. At least it's all there.

PR #2701, opened 2024-08-02
Improve SMILES translation for surface adsorbates

rwest:
### Motivation or Problem
I recently discovered (a happy accident) that you can put `*` in a SMILES string and RDKit will use it as a dummy atom, with an atomic number of 0, which RMG happens to interpret as a surface site, so you can easily enter adsorbates as SMILES strings this way. (Was I the last to realize this? seems super convenient!). 

Anyway, the round trip didn't work,  because it would output the atom as `[Pt]` when you convert things _into_ a SMILES string.

### Description of Changes
This PR changes it so you can do round-trip read and write SMILES with `*` for surface sites.

The default atom for a surface site being turned into an RDKit molecule is now the wildcard 0. When you're dealing with InChI conversion it instead uses atom 78 (platinum), because otherwise the inchi stuff crashes. 
Molecules with an N atom in are by default converted to SMILES by OpenBabel, which used the `[Pt]` syntax, which made things look inconsistent and prevented a round-trip conversion for those adsorbates only. So then I made the OpenBabel converter replace `[Pt]` with `*` after it's made a SMILES.

Now you can do round-trip conversion to and from SMILES for at least these adsorbates:
`*C`, `*CC`, `*=C`, `*[H]`, `*=O`, `*COC*`, `*CNC`, `*C1CCCC1`, `*N`, `*N(Cl)`

### Testing
I wrote unit tests, (that was the lion's share of the work, as usual).
Locally, I'm getting some weird augmented inchi error when I debug in VSCode (`'CH2O2/c2-1-3/h1-2H/u1,3' == 'CH2O2/c2-1-3/h1H,(H,2,3)/u1,2'`) but I can't see why that would have changed.  When I run in my console with `make test` I instead get a TestSoluteDatabase::test_saturation_density error (`1.93 == 0`). 
I'm hoping neither of these are actual problems and the GitHub CI runs smoothly. 🤞

### Reviewer Tips
The output, log files, etc. will look different.
Species names are now more likely to have `*` in them (than `[Pt]`).
Hopefully filenames with `*` in aren't a problem.
Maybe there are other unintended consequences we'll have to deal with?
Will this mess up anyone's workflows (in ways that they shouldn't just improve their workflows?).
Try it out and report back.

### DIFF
--- a/rmgpy/molecule/converter.pxd
+++ b/rmgpy/molecule/converter.pxd
@@ -29,7 +29,7 @@ cimport rmgpy.molecule.molecule as mm
 cimport rmgpy.molecule.element as elements
 
 
-cpdef to_rdkit_mol(mm.Molecule mol, bint remove_h=*, bint return_mapping=*, bint sanitize=*, bint save_order=?)
+cpdef to_rdkit_mol(mm.Molecule mol, bint remove_h=*, bint return_mapping=*, bint sanitize=*, bint save_order=?, int X=?)
 
 cpdef mm.Molecule from_rdkit_mol(mm.Molecule mol, object rdkitmol, bint raise_atomtype_exception=?)
 
--- a/rmgpy/molecule/converter.py
+++ b/rmgpy/molecule/converter.py
@@ -49,11 +49,13 @@ import rmgpy.molecule.molecule as mm
 from rmgpy.exceptions import DependencyError
 
 
-def to_rdkit_mol(mol, remove_h=True, return_mapping=False, sanitize=True, save_order=False):
+def to_rdkit_mol(mol, remove_h=True, return_mapping=False, sanitize=True, save_order=False, X=0):
     """
     Convert a molecular structure to a RDKit rdmol object. Uses
     `RDKit <http://rdkit.org/>`_ to perform the conversion.
     Perceives aromaticity and, unless remove_h==False, removes Hydrogen atoms.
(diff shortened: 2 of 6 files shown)


rwest:
@mjohnson541 we were just discussing how RMS uses SMILES for some things and uses either RDKit or RMG. Will this have impacts on RDKit that need coordination?

mjohnson541:
I don't think so, right now RMS can't do anything with surface smiles, this should allow RMS to accept smiles for surface species.

rwest:
Chatting with Bjarne, we thought that having `*` in filenames could be a recipe for disaster (imagine people trying to copy, move, rename, or delete a file....).
So probably this needs more thought.

Nice to have?
1. round trip so a SMILES string written by RMG can be read by RMG
2. compatibility with other tools, so a SMILES string written by RMG can be read by other things. (by what? RDKit? what else is used by our users?)
3. short names. `X` and `*` are each 1 character whereas `[Pt]` is 4
4. human-friendly; looking at it makes sense and is easy to understand.
5. file-system friendly. avoid filenames with `*` in.
6. not confusing when you change metal. i.e. using [Pt] is hard to explain to people when you're simulating copper, nickel, ruthenium, an arbitrary hypothetical metal, or an alloy.

extend the list?
cross any off as unimportant?
identify if any of these are in direct conflict?
think of some solutions that match as much as possible?

PR #2702, opened 2024-08-04
Restrict pyrms to <2

mjohnson541:
I'm planning to release pyrms v2.0.0 which will mark a switch to pyjuliacall. This PR prevents current RMG-Py from grabbing the new version it isn't compatible with.


JacksonBurns (approved):
Ty!

PR #2711, opened 2024-09-08
Parse RS2C (CASPT2) energy in Molpro

alongd:
### Motivation or Problem
Our Arkane Molpro adapter does not parse the electronic energy of any level of theory, just the comon ones we use.

### Description of Changes
Here a small modification was added to allow Molpro to recognize and parse `RS2C` energies.
For the long run, we should carefully examine whether it is enough to just recognize the `energy= ` token, so our Molpro adapter will be more generalized instead of hard coding it for every level we use.

### Testing
A test was added


JacksonBurns:
@donerancl have you used MolPro and could review this PR?

PR #2715, opened 2024-09-16
Draw Radicals on Groups

mjohnson541:
Right now you can't visually tell where radicals are on drawn groups structures, this draws the radical on GroupAtoms where the attribute radical_electrons == [1].


rwest:
Looks like `test/rmgpy/molecule/groupTest.py::TestGroup::test_draw_group` needs updating.

PR #2718, opened 2024-10-01
Change CI to use Miniforge3 instead of Mambaforge

ssun30:
### Motivation or Problem
Mambaforge will be [sunsetted](https://conda-forge.org/news/2024/07/29/sunsetting-mambaforge/) starting from Oct 1, 2024 and only Miniforge will be maintained going forward. This will cause our CI to fail at the environment creation step. 

### Description of Changes
The `miniforge-variant` parameter is changed to `Miniforge3`.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -219,16 +219,16 @@ jobs:
       - name : Find ID of Reference Results
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        # this will search for the last successful execution of CI on main
         run: |
-          run_id=$(gh run list -R ReactionMechanismGenerator/RMG-Py --workflow="Continuous Integration" --branch main --limit 1 --json databaseId --jq '.[0].databaseId')
+          run_id=$(gh run list -R ReactionMechanismGenerator/RMG-Py --workflow="Continuous Integration" --branch main --limit 15 --json databaseId,conclusion --jq 'map(select(.conclusion == "success")) | .[0].databaseId')
           echo "CI_RUN_ID=$run_id" >> $GITHUB_ENV
 
       - name: Retrieve Stable Regression Results
         if: ${{ env.REFERENCE_JOB == 'false' }}
         uses: actions/download-artifact@v4
         with:
-        # this will search for the last successful execution of CI on main and download
-        # the stable regression results
+        # download stable regression results
           run-id: ${{ env.CI_RUN_ID }}
           repository: ReactionMechanismGenerator/RMG-Py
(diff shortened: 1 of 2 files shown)


rwest (approved):
Looking at the diff and the [conda-incubator/setup-miniconda](https://github.com/conda-incubator/setup-miniconda)  readme, I approve. But suggest waiting until the CI passes.

ssun30:
> Looking at the diff and the [conda-incubator/setup-miniconda](https://github.com/conda-incubator/setup-miniconda) readme, I approve. But suggest waiting until the CI passes.

I need to add the change logs to the workflow file and also change the miniforge version in doc build (that has failed). Will have another go in an hour or two.

ssun30:
The CI failed at the step where it's trying to retrieve stable regression results:
```
Run actions/download-artifact@v4
  with:
    run-id: 11121550302
    repository: ReactionMechanismGenerator/RMG-Py
    github-token: ***
    name: stable_regression_results
    path: stable_regression_results
    merge-multiple: false
  env:
    RMG_DATABASE_BRANCH: main
    REFERENCE_JOB: false
    INPUT_RUN_POST: true
    CONDA: /home/runner/miniconda3
    CONDA_PKGS_DIR: /home/runner/conda_pkgs_dir
    PYTHONPATH: /home/runner/work/RMG-Py/RMG-Py/RMG-Py:
    CI_RUN_ID: 11121550302
Downloading single artifact
Error: Unable to download artifact(s): Artifact not found for name: stable_regression_results
        Please ensure that your artifact is not expired and the artifact was uploaded using a compatible version of toolkit/upload-artifact.
```
@rwest @mjohnson541 @JacksonBurns
It's trying to retrieve the stable regression results from the [most recent reference CI job](https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/11121550302) that failed yesterday and didn't upload any regression results, so no artifact was found. It appears to me the behavior is that non-reference CI jobs are downloading the stable regression results from the most recent reference CI job instead of the [latest successful reference CI job](https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/11102339269).

This might be a weakness going forward as any failure in our scheduled reference CI job (for example, a package or Github action gets deprecated) will cause all other jobs, including PRs that intend to fix that failure, to fail as well.

rwest:
Looks like a good patch  - if it works. Thanks Su

ssun30:
Problem with retrieving regression results solved. Squashed a couple commits. Should be ready to deploy.

JacksonBurns (approved):
Looks good thanks for the thorough patch. Will auto-merge for when CI passes

PR #2712, opened 2024-09-09
Remove `default` Channel from `environment.yml`

mjohnson541:
Update Before Merging from @JacksonBurns:

This PR tested and showed that we can remove the `default` channel from our `environment.yml` file. This is important since Anaconda has changed their Terms of Service, and this change brings many of our users back in line with said ToS.

PR #2719, opened 2024-10-03
update user and developer installation instructions to point towards miniforge

JacksonBurns:
Installing `conda` via other providers can result in non-free channels being used, violating the `conda` terms of service for many of our users. Pointing users towards this new installation method for `conda` guarantees that they will not access non-free channels.

Related to #2712


JacksonBurns:
Note that as of #2718 we also use `Miniforge` in the CI

rwest (approved):
Looks fine to me.

Not meaning to delay this PR, but have any Apple Sicilon users tested our instructions recently?

JacksonBurns:
Just tested on an M2 Mac mini and it worked.

Issue #2722, opened 2024-10-10
Could not solve environment on Linux due to the removal of the `defaults` channel from `environment.yml`


ssun30:
### Bug Description
**_Specific to Linux:_** 

Users may encounter a problem when installing RMG following our updated instructions introduced in PR #2719 . The following error message occurs when the user attempts to create the conda environment:

```
Solving environment: ...working... failed
  Warning: 
  LibMambaUnsatisfiableError: Encountered problems while solving:
    - package cantera-2.6.0-py310h6cd0baa_0 requires mkl, but none of the providers can be installed
  
  Could not solve for environment specs
  The following packages are incompatible
  ├─ cantera 2.6*  is installable and it requires
  │  └─ mkl, which can be installed;
  └─ nomkl is not installable because it requires
     └─ mkl <0.a0 , which conflicts with any installable versions previously reported.
```

### How To Reproduce
By following instructions "Installation by Source Using Anaconda Environment for Unix-based Systems: Linux and Mac OSX" steps 1 through 7 (step 6 skipped). 

Users who encounter this problem either have a blank `.condarc` file or not have the `defaults` channel in `.condarc`. 

### Installation Information
Describe your installation method and system information.
 - Ubuntu 22.04 LTS, CentOS Linux 7 (on Discovery)
 - x86_64 hardware architecture
 - Installation from source with Miniforge3
 - RMG version information:
   
   - RMG-Py: 3.2.0-357-g74ef5e699
   - RMG-database: 3.2.0-111-gcc0ecdb86

### Additional Context
We recently removed the `defaults` channel from `environment.yml` due to changes to Anaconda's Terms of Service (PR #2712). Now we aren't installing any package from the official channel. But this introduced the error shown above. 

The problem goes away with either one of these workarounds:

- Adding the `defaults` channel in user's `.condarc` file. 
- Removing the `-nomkl` mutex metapackage from `environment.yml`. 

From my understanding, it seems that [Cantera 2.6](https://anaconda.org/Cantera/cantera/files?version=2.6.0) on Linux uses `mkl` instead of `openblas`, so the `-nomkl` toggle may result in an incompatibility when solving the environment. But for reasons beyond my understanding, adding the `defaults` channel bypasses this problem. 

This problem escaped our CI because the `conda-incubator/setup-miniconda` Github Action automatically creates a `.condarc` file in the runner's home directory that includes the `defaults` channel. 

[...]

PR #2723, opened 2024-10-11
Deprecate the nomkl mutex metapackage

ssun30:
### Motivation or Problem
Resolves issue #2722 .

### Description of Changes

- Added a step in the CI workflow to create a custom conda config files that excludes the default channel. 
- Replaced the nomkl mutex metapackage that controls whether packages should use MKL or OpenBLAS.


JacksonBurns (approved):
LGTM - thanks for the quick fix

PR #2724, opened 2024-10-26
Changed Dockerfile to use Miniforge3

ssun30:
This is a follow-up to #2718 . We missed the Docker build file when we updated our CI to use Miniforge3. 

I also added some arguments in the Dockerfile so the CI or an end user can configure it to use different branches for RMG-Py, RMG-database, and RMS. They are currently only used by the `echem-rebase` branch to push an `electrocat` version of the Docker image, but could also provides some flexibility for future projects. 

### Description of Changes

- Changed the Docker build file to use Miniforge3 instead of Miniconda and Mamba. 

- Added arguments in the Docker build file to allow different branches to be pulled instead of main. Default to main. 

- Updated the `docker-build-and-push` action to use the latest version.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -401,7 +401,7 @@ jobs:
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
       - name: Build and Push
-        uses: docker/build-push-action@v4
+        uses: docker/build-push-action@v6
         with:
           push: true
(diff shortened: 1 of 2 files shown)


JacksonBurns (approved):
Ah, thanks for catching this, can't believe I forgot earlier. Since the Dockerfile tests itself during running, I think this is good to merge.

Issue #2726, opened 2024-10-31
Barrier heights not fixed for linear scaling of libraries


sevyharris:
### Bug Description
I tried regenerating the cpox mechanism on Rh and found that the O2 dissociation reaction has crazy fast kinetics compared to [Emily's original model](https://github.com/comocheng/uncertainty_estimator/blob/main/cpox_pt/cpox_rh_emily/chem_annotated-surface.inp). 
![Untitled](https://github.com/user-attachments/assets/c9e71eaa-9d3b-4135-ae7d-6f2d15c0a6b1)
![Untitled](https://github.com/user-attachments/assets/b0dbb841-c58c-4d11-9115-20cfbc0266e1)

I think the barrier heights are the issue here. 
The reaction is defined in the library in the desorption direction: 
![Untitled](https://github.com/user-attachments/assets/c34e9752-2be0-436b-8360-cf7f2e86c749)

And the barrier height is not raised to the endothermicity of the reaction in the model I generated.
The three points plotted are 1. enthalpy of reactants (set to 0 for reference), 2. Ea for the reaction, 3. enthalpy of reaction/enthalpy of products relative to reactants
![Untitled-1](https://github.com/user-attachments/assets/3677fa24-e67f-4744-9772-ebfd38685607)

If you raise the barrier, you get more reasonable kinetics for O2 adsorption. So, I think this reaction's barrier was supposed to be fixed due to the scaling, but fix_barrier_height is intentionally not called for libraries in the rmgpy.rmg.model.CoreEdgeReactionModel class.

### How To Reproduce
Run the latest RMG on the main branch using this input file (basically the same as Emily's cpox input file but now using Deutschmann2006_adjusted instead of Deutschmann2006: 
https://github.com/comocheng/uncertainty_estimator/blob/main/cpox_pt/cpox_rh_20241028/input.py

You can see the mechanism it generated here: https://github.com/comocheng/uncertainty_estimator/blob/main/cpox_pt/cpox_rh_20241028/chem_annotated-surface.inp

You can also just look at the values in Surface/CPOX_Pt/Deutschmann2006_adjusted using this code:

```
import numpy as np
import os
import copy

import rmgpy.chemkin

import matplotlib.pyplot as plt
%matplotlib inline


database = rmgpy.data.rmg.RMGDatabase()

[...]

PR #2729, opened 2024-11-12
Fix library reaction barrier heights for linear scaling

sevyharris:
### Motivation or Problem
If you include a library for a surface mechanism that uses linear scaling, this can cause some reaction barriers to be below the endothermicity of the reaction. This PR fixes the barrier heights for library reactions that have been scaled across metals.

See https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2726 for more details of the issue.

### Description of Changes
This PR adds a call to fix_barrier_height() if the library reaction is a surface reaction with any species thermo that has been scaled across metals.

### Testing
I have tested with this CPOX on Rh RMG [input.txt](https://github.com/user-attachments/files/17718558/input.txt) and confirmed that it raises the barrier for the O2 dissociative adsorption reaction taken from the Surface/CPOX_Pt/Deutschmann2006_adjusted library:

```
! Reaction index: Chemkin #4; RMG #6
! Library reaction: Surface/CPOX_Pt/Deutschmann2006_adjusted
! Flux pairs: OX(25), O2(3); OX(25), X(1); OX(25), X(1);
!
! Ea raised from 278.7 to 418.8 kJ/mol to match endothermicity of reaction.
OX(25)+OX(25)<=>X(1)+X(1)+O2(3)                     3.700000e+21 0.000     100.098
```

vs, the original:

```
! Reaction index: Chemkin #4; RMG #6
! Library reaction: Surface/CPOX_Pt/Deutschmann2006_adjusted
! Flux pairs: OX(25), O2(3); OX(25), X(1); OX(25), X(1);
OX(25)+OX(25)<=>X(1)+X(1)+O2(3)                     3.700000e+21 0.000     66.611
```

<s>I have not yet tested that it _doesn't_ scale the same input file if you try to generate a Pt mechanism, but I plan to.</s>
have now confirmed it doesn't bother fixing barriers for regular Pt mechanism.

### DIFF
--- a/rmgpy/rmg/model.py
+++ b/rmgpy/rmg/model.py
@@ -548,6 +548,10 @@ class CoreEdgeReactionModel:
             #  correct barrier heights of estimated kinetics
             if isinstance(forward, (TemplateReaction, DepositoryReaction)):  # i.e. not LibraryReaction
                 forward.fix_barrier_height()  # also converts ArrheniusEP to Arrhenius.
+            elif isinstance(forward, LibraryReaction) and forward.is_surface_reaction():
+                # do fix the library reaction barrier if this is scaled from another metal
+                if any(['Binding energy corrected by LSR' in x.thermo.comment for x in forward.reactants + forward.products]):
+                    forward.fix_barrier_height()            
 
             if self.pressure_dependence and forward.is_unimolecular():
                 # If this is going to be run through pressure dependence code,


bjkreitz:
This PR looks good to me. The bigger question is if it makes sense to use a library value when you scale across the periodic table since your estimate of the barrier would not be great anyway. At least this adjustment makes sure that the barrier is somewhat reasonable. Let me know when you have tested this on a Pt mechanism and I'll leave an approving review.

sevyharris:
Thank you @bjkreitz for reviewing this! I have now done the test for Pt and it looks good-- doesn't change any barrier heights.

I will keep that in mind about scaling from a library.

PR #2734, opened 2024-11-20
Draw Hydrogen bonds (again)

rwest:
### Motivation or Problem 
Pull request  https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1961  introduced the ability to draw hydrogen bonds.

To do so, Hydrogen atoms that have H bonds shouldn't be removed from the drawings.
This check was  removed without explanation in ec9515351ea9964a5a8499c1f4066378544eca41
which was designed to make drawing work for fragments.
That broke drawing of hydrogen bonds.
The unit test for drawing hydrogen bonds exists, but just checks that a PDF is made, not what it looks like. 
So the regression was not detected.

### Description of Changes
This restores the check that H atoms in H-bonds are not removed from drawings.
It also removes code duplication that was introduced in #1961 through erroneous rebasing, that @bjkreitz noticed and removed in a different PR.

### Testing
It was while testing the second commit (Bjarne's commit that removes the code duplication)
that I found the title problem (the H bonds being broken). Now I've tested both, by running the original tests from #1961 
```python
example =  Molecule().from_adjacency_list("""
example
1  O u0 p3 c-1 {2,S} {10,H}
2  N u0 p0 c+1 {1,S} {3,D} {4,S}
3  O u0 p2 c0 {2,D}
4  O u0 p2 c0 {2,S} {7,S}
5  N u0 p1 c0 {6,S} {8,S} {9,S} {7,H}
6  O u0 p2 c0 {5,S} {10,S}
7  H u0 p0 c0 {4,S} {5,H}
8  H u0 p0 c0 {5,S}
9  H u0 p0 c0 {5,S}
10 H u0 p0 c0 {6,S} {1,H}
""")
example
display(example)
for m in example.generate_h_bonded_structures():
    display(m)
```

Restoring the check makes the H bonds work. This particular molecule still looks ugly, as it always did. Removing the duplicate code doesn't make them worse. 

### Reviewer Tips
Not a big change, and not worth a lot of time.

### DIFF
--- a/rmgpy/molecule/draw.py
+++ b/rmgpy/molecule/draw.py
@@ -184,7 +184,8 @@ class MoleculeDrawer(object):
         surface_sites = []
         for atom in self.molecule.atoms:
             if isinstance(atom, Atom) and atom.is_hydrogen() and atom.label == '':
-                atoms_to_remove.append(atom)
+                if not any(bond.is_hydrogen_bond() for bond in atom.bonds.values()):
+                    atoms_to_remove.append(atom)
             elif atom.is_surface_site():
                 surface_sites.append(atom)
         if len(atoms_to_remove) < len(self.molecule.atoms) - len(surface_sites):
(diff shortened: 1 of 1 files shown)


bjkreitz (approved):
This change looks good to me. The structures of the example molecule are indeed pretty ugly, but reintroducing this check allows to draw the hydrogen bonds again. RMG also works normally.

PR #2735, opened 2024-11-20
Port Optional RMS Changes into Ongoing SciPy CLAR Implementation

JacksonBurns:
Can't debug the CLAR implementation right now because the new `juliaup` based installation isn't working. Going to allow running without Julia on this PR, which we were planning on doing in the future, and then add support back for it later. Can't fix two things at once.


JacksonBurns:
Manually resolved locally and merged.

PR #2694, opened 2024-07-24
[Python 3.9] Clar SciPy Re-implementation, `pythoncall` Debugging, Optional RMS

JacksonBurns:
This PR replaces #2623 as discussed in this comment: https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2687#issuecomment-2248623916

PR #2737, opened 2024-11-22
Fix Conda build workflow

ssun30:
This is the mirror PR to [this](https://github.com/ReactionMechanismGenerator/RMG-database/pull/671).

### Motivation or Problem
The old Conda build workflow to set up Miniconda no longer works due to deprecation of the `default` channel. Now only the `conda-forge` channel will be used. Furthermore, the new workflow directly uses Miniforge3 which has built-in mamba solver. 

### Testing
Currently, this workflow only triggers upon releasing a stable version of RMG-Py. So I've set this PR to WIP until I've tested it to prevent overwriting the current stable RMG-Py package on the rmg channel.

### DIFF
--- a/.github/workflows/conda_build.yml
+++ b/.github/workflows/conda_build.yml
@@ -11,48 +11,88 @@ jobs:
       run:
         shell: bash -l {0}
     steps:
-      - uses: actions/checkout@v2
-      - uses: conda-incubator/setup-miniconda@v2
+      - uses: actions/checkout@v4
+      - name: Create custom conda config file
+        run: |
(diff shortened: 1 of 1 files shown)


JacksonBurns:
Just FYI that the conda-build workflow doesn't work on RMG-Py, see #2641 which tried to fix it by making RMS optional (now in #2694) and  #2636 which attempted to fix it as-is

You can probably just port whatever changes you make in RMG-database into this workflow but don't actually run it or anything, just bring the changes over so we don't forget

PR #2739, opened 2024-12-01
Orca parser fixup

kfir4444:
### Motivation or Problem
The `OrcaLog.load_conformers` required some adaptations to function with the orca version 5.0.4.

### Description of Changes
1) Made sure the frequencies are read correctly with the new format
2) Verified the LinearRotor object is created with no issues (the past implementation got an error:`TypeError: Cannot convert rmgpy.quantity.ArrayQuantity to rmgpy.quantity.ScalarQuantity` due to parsing an array instead of a single value)
3) Added a new frequencies log to ess/data with the 5.0.4 version

### Testing
Added a test with the new Orca log.

### DIFF
--- a/arkane/data/orca/freq_orca_5.0.4.log
+++ b/arkane/data/orca/freq_orca_5.0.4.log
@@ -1,0 +1,1709 @@
+
+                                 *****************
+                                 * O   R   C   A *
+                                 *****************
+
+                                            #,                                       
+                                            ###                                      
+                                            ####                                     
+                                            #####                                    
+                                            ######                                   
+                                           ########,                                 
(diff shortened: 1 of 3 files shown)


kfir4444:
> My only suggestion is that you call the new file "freq_orca_5.0.4.log" or something like that to indicate the version, rather than just "new". When ORCA releases more new versions in the future, this will get confusing otherwise.

Thanks for the comment, I changed the name of the file. Could we proceed with this PR?

JacksonBurns (approved):
Looks good, thanks! I'll set to auto-merge, assuming tests pass.

PR #2741, opened 2024-12-03
Python 3.9

JacksonBurns:
## Upgrading RMG-Py to Python 3.9

The current version of Python used by RMG (3.7) is deprecated, so we are missing out on the latest updates to our various dependencies. This PR upgrade RMG to use Python 3.9 by doing the following:
 - use of `lpsolve` is removed and replaced with `scipy`'s Linear Programming Solver (credit @xiaoruiDong and @JacksonBurns) because `lpsolve` doesn't support 3.9+ (no builds exist, and we can't get our build to work). Ported from #2623
 - various improvements by @hwpang related to Arkane tests, Isodesmic and element balance constraints in #2596 and #2640
 - makes RMS optional to enable running RMG in Pure Python only mode, new CI added for that setup, ported from #2631 (this will also unblock fixing the conda recipe, see https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2641#issuecomment-2590834212
 - Yet another `julia` overhaul, see below.

## Getting this merged
One feature that remains to be implemented and will have to be figured out in a separate PR _into this one_ is the new `juliaup` + `juliacall` setup implemented by @hwpang. It will be much better, but there are currently some bugs preventing it from working (at all), primarily related to the installed `julia` executable being able to 'see' the Python environment and its installed dependencies.

After that, this should be good to merge. RMG-Py will be appreciably faster, easier to extend, and less buggy.


JacksonBurns:
@jonwzheng @donerancl the current unit test failure seems to indicate that benzene is not being perceived as aromatic, preventing it from being matched in the tree (maybe?)

JacksonBurns:
0a13902 fixed the single remaining failing unit test. Rebased on the latest version on `main`, will now let the tests run here.

JacksonBurns:
@jonwzheng I've rebased this onto RMG/main and the branch of RMS we are using onto RMS/for_rmg so let's see what happens

JacksonBurns:
TODO: update the docs to reference the new install_rms.sh script

JacksonBurns:
Julia hanging during the RMS install seems like a known limitation: https://discourse.julialang.org/t/running-ci-for-julia-1-6-on-github-macos-runners-seems-to-fail/116577/4

So if it doesn't pass this time, we can probably just remove it.

jonwzheng:
thanks for working on this Jackson. 
do you intend on updating the docs in a future PR or on this one?

JacksonBurns:
> thanks for working on this Jackson. do you intend on updating the docs in a future PR or on this one?

Oops I had the changes locally but forgot to push - they are up now in ec57780

install_rms.sh
```diff
@@ -0,0 +1,31 @@
+# install_rms.sh
+#
+# Convenience script to install ReactionMechanismSimulator inside your rmg_env conda environment
+#
+# Note that you will have to manually install juliaup before running this script:
```
jonwzheng, line 5:
```suggestion
# Note that you will have to manually install juliaup before running this script by uncommenting and executing the following lines:
```

pytest.ini
```diff
@@ -17,7 +17,7 @@ filterwarnings =
 # -n auto # will use all available cores, if you first pip install pytest-xdist, but it doesn't work well with julia (so add --no-julia? or give up?)
```
jonwzheng, line 17: could provide more helpful advice than `or give up` :stuck_out_tongue:

rmgpy/kinetics/kineticsdata.pyx
```diff
@@ -249,9 +249,8 @@ cdef class PDepKineticsData(PDepKineticsModel):
                         Plow = Pdata[j]
                         Phigh = Pdata[j + 1]
                         if Plow <= P and P <= Phigh:
-                            klow = kdata[i, j] * (kdata[i + 1, j] / kdata[i, j]) ** ((T - Tlow) / (Thigh - Tlow))
-                            khigh = kdata[i, j + 1] * (kdata[i + 1, j + 1] / kdata[i, j + 1]) ** (
-                                        (T - Tlow) / (Thigh - Tlow))
+                            klow = kdata[i, j] * (kdata[i + 1, j] / kdata[i, j]) ** ((T - Tlow) / (Thigh - Tlow)).real
+                            khigh = kdata[i, j + 1] * (kdata[i + 1, j + 1] / kdata[i, j + 1]) ** ((T - Tlow) / (Thigh - Tlow)).real
```
jonwzheng, line 253: if we want only the real components should `.real` apply to the entire term here and not just the T term? I saw from eb173d323498a7d83312baee5fadd14e41ca433e (great song choices btw) that this was required to avoid cython issues; i'm curious why there are imaginary components to the temperatures in the first place. Makes me wonder if there's an issue upstream w/ the T calculations or if it's just a numerical artifact.

rmgpy/molecule/resonance.py
```diff
@@ -927,19 +929,18 @@ def generate_clar_structures(mol, save_order=False):
         mol.assign_atom_ids()
 
     try:
-        output = _clar_optimization(mol, save_order=save_order)
-    except ILPSolutionError:
+        solutions = _clar_optimization(mol, save_order=save_order)
+    except (RuntimeError, ValueError):  # either a crash during optimization or the result was an empty tuple
```
jonwzheng, line 933: I'm not too familiar w/ this algorithm, but is this something that warrants a warning/logging message or no?

rmgpy/molecule/group.py
```diff
@@ -46,8 +46,91 @@
 from rmgpy.molecule.graph import Vertex, Edge, Graph
 from rmgpy.molecule.fragment import CuttingLabel
 
+# helper functions
+# these were originall nested inside the indicated parent function, but when we upgraded to
```
jonwzheng, line 50:
```suggestion
# these were originally nested inside the indicated parent function, but when we upgraded to
```

rmgpy/rmg/reactionmechanismsimulator_reactors.py
jonwzheng, line 1: `reactionmechanismsimulator_reactors` is quite a mouthful. Any thoughts on just naming this `rms_reactors` or do you prefer the clarity of the full written name? (Throughout the rest of the code base, for example in `requires_rms` we use the `rms` acronym, so I don't think it's losing too much clarity by just using that.) Or is there some other consideration I haven't thought about?

utilities.py
```diff
@@ -271,7 +228,7 @@ def clean(subdirectory=''):
     if platform.system() == 'Windows':
         extensions = ['.pyd', '.pyc']
```
jonwzheng, line 229: does this need `.c`?

jonwzheng (commented):
first of all thanks for working on this huge upgrade! This is really great work. 
In addition to the comments here, I wonder about our long-term strategy for getting RMS tests for M1 macOS. Our current workflow is blocked by Julia hanging w/ RMS install using the Github runners; maybe that's something we need RMS maintainers to prioritize if possible?

documentation/source/users/rmg/installation/anacondaDeveloper.rst
```diff
@@ -91,40 +81,33 @@ Installation by Source Using Anaconda Environment for Unix-based Systems: Linux
...
-   Also, add your RMG-Py folder to PATH to launch ``rmg.py`` from any folder.
+#. **Optional**: add your RMG-Py folder to ``PATH`` to launch ``rmg.py`` from any folder.
 
-   In general, these commands should be placed in the appropriate shell initialization file.
+   In general, this commands should be placed in the appropriate shell initialization file.
    For Linux users using bash (the default on distributions mentioned here), these should be placed in ``~/.bashrc``.
    For MacOS users using bash (default before MacOS Catalina), these should be placed in ``~/.bash_profile``, which you should create if it doesn't exist.
    For MacOS users using zsh (default beginning in MacOS Catalina), these should be placed in ``~/.zshrc``. ::
 
-    export PYTHONPATH=YourFolder/RMG-Py/:$PYTHONPATH
```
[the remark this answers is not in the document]
JacksonBurns: The installation process for RMG has been updated in this PR. Now, we actually run `pip install` which results in RMG being added to the `site-packages` directory, making it accessible from anywhere (as setting this environment variable used to do prior to this PR when we were building in-place). This change was made because setting `PYTHONPATH` breaks ARM Mac when Julia is installed.

environment.yml
```diff
@@ -42,69 +46,54 @@ dependencies:
...
-  # It is only on pip, so we will need to do something like:
-  # https://stackoverflow.com/a/35245610
-  # Note that _some other_ dep. in this list requires diffeqpy in its recipe
-  # which will cause it to be downloaded from the rmg conda channel
-
 # configure packages to use OpenBLAS instead of Intel MKL
   - blas=*=openblas
 
+# optional dependencies for using ReactionMechanismSimulator
+# remove the leading '#' to install the required dependencies
```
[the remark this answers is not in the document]
JacksonBurns: Good idea, added

examples/arkane/networks/acetyl+O2_slse/input.py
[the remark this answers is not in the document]
JacksonBurns: We execute both of these separately during the tests to cover the different methods, so for ease of testing there I am not thinking we should separate these.

I agree that the current setup is confusing, though, so I have added mention to the top of each about what changes.

pytest.ini
JacksonBurns: Haha this is an artifact of @rwest

i will actually update it though since now we can run the tests in parallel in pure python mode (:tada:)

rmgpy/kinetics/kineticsdata.pyx
JacksonBurns: Yeah NGL this one doesn't make sense to me. This quantity should never end up imaginary. My _assumption_ (read _guess_) was that, because the result of this calculation is packed into a numpy array that only support real-valued floats, the C-typed `T`-family of variables need to be modified to reflect that.

rmgpy/molecule/resonance.py
JacksonBurns: No logging is done in this location because it is handled lower down in `_clar_optimization`. There, we log if generation fails after at least one structure is enumerated. This is because, except in rare cases, failure on the first iteration means that there is no Clar formula to be found. In other cases, the user will see why we stopped.

rmgpy/rmg/reactionmechanismsimulator_reactors.py
JacksonBurns: Thinking ahead for when we might add `cantera_reactors` or the like - trying to keep an option open for consistent (though verbose) naming.

utilities.py
JacksonBurns: Yeah, it's possible to generate these files while debugging the Cython code.

JacksonBurns:
@jonwzheng this is just about ready. All tests pass on Ubuntu with and without Julia. Without Julia Mac OS and Mac OS Latest pass. [fd4b979](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2741/commits/fd4b9799199bd89bdd3f6155f6bd88cd3252b892) is my last attempt to get Julia to work on Mac OS Latest. If that doesn't work we can just skip that test (as we do on `main` right now) and run a manual check to confirm we can get Julia + RMG working on a Mac.

JacksonBurns:
One last (hopefully) hiccup - the regression test observables comparison seg faults with Julia installed but not otherwise. Going to diagnose locally.

environment.yml
```diff
@@ -18,104 +18,84 @@
 #   moved diffeqpy to pip and (temporarily) removed chemprop
 # - April 17, 2024 Limit versions of cclib at advice of maintainers.
 # - August 4, 2024 Restricted pyrms to <2
+# - May 14, 2024 Removed diffeqpy by switching to call SciMLBase and Sundials using JuliaCall
+# - March 15, 2024 - started migration to Python 3.9
+# - Mar 11, 2024 Removed Julia dependencies, now considered optional
```
jonwzheng, line 23: a tiny thing, but how about
```suggestion
# - April 14, 2025 Removed Julia dependencies, now considered optional
```
? I think the year is off and Mar is shorthand

examples/arkane/networks/acetyl+O2_sls/input.py
```diff
@@ -0,0 +1,286 @@
+################################################################################
+#
+#   Arkane input file for acetyl + O2 pressure-dependent reaction network
+#
+#  uses method = 'simulation least squares eigen' for PressureDependence
```
jonwzheng, line 5:
```suggestion
#  uses method = 'simulation least squares' for PressureDependence
```

jonwzheng (commented):
thanks for this work Jackson. Just a couple of very minor things (and also will await the seg fault hotfix before approving)

rmgpy/molecule/group.py
jonwzheng: i think this may have been prematurely resolved

JacksonBurns:
Dug this up with GDB for when the regression testing script for observables is called:
```bash
Thread 1 "python" received signal SIGSEGV, Segmentation fault.
0x00007ffecbbf8b25 in Cantera::CVodesIntegrator::setTolerances(double, unsigned long, double*) ()
   from /home/jackson/miniconda3/envs/rmg_env/lib/python3.9/site-packages/cantera/../../../libcantera.so.2
```

probably correlated with this error also in the job logs, which may be accounting for the reduced number of species found by RMS:
```bash
2025-04-14T01:20:18.6394596Z Making seed mechanism...
2025-04-14T01:20:18.6422368Z Conducting simulation of reaction system 1...
2025-04-14T01:21:21.6367141Z 
2025-04-14T01:21:21.6367540Z [CVSLS ERROR]  CVodeSetLinearSolver
2025-04-14T01:21:21.6367994Z   LS must be non-NULL
2025-04-14T01:21:21.6368199Z 
2025-04-14T01:21:21.6368221Z 
2025-04-14T01:21:21.6374300Z [CVSLS ERROR]  CVodeSetJacFn
2025-04-14T01:21:21.6374652Z   Linear solver memory is NULL.
2025-04-14T01:21:21.6374868Z 
2025-04-14T01:21:27.8278294Z 
2025-04-14T01:21:27.8278826Z [CVODE ERROR]  cvNlsInit
2025-04-14T01:21:27.8279349Z   Setting linear solver solve function failed
2025-04-14T01:21:27.8279715Z 
2025-04-14T01:21:27.8279721Z 
2025-04-14T01:21:27.8279887Z [CVODES ERROR]  cvInitialSetup
2025-04-14T01:21:27.8280349Z   The nonlinear solver's init routine failed.
```

JacksonBurns:
Looks like Julia might be changing the installed/used version of Sundials/CVODE and causing Cantera to segfault?

JacksonBurns:
ChatGPT agrees with the above and points out that it is also possible that Julia is setting `LD_LIBRARY_PATH` and in doing so pointing Python towards the wrong libraries for CVODE

JacksonBurns:
Well @jonwzheng this seg fault error is hard to pin down. Going to just try and bump the Julia version... would appreciate any input you have on this as well!

rwest:
>  Doesn't work on Intel mac but I don't think we need to worry about it.

Guess it's time for me to upgrade my laptop? 👨‍💻



I think the segfault is indeed quite likely different CVODEs libraries conflicting. Wondering if @ssun30 has any tricks up his sleeve, having worked with both Cantera and RMS.

JacksonBurns:
@rwest @ssun30 from the installation instruction on Sundials it looks like we can tell it **not** to re-download CVODES by doing something like this: https://github.com/tshort/Sundials.jl/blob/master/README.md#installation

JacksonBurns:
@rwest I'll clarify that it doesn't work on intel mac _runners_ - I believe it should still work on intel mac locally, it's just that the runners aren't up to it.

@ssun30 I've taken an initial pass at trying to 'protect' the pre-installed libs from Julia. Now, in the `install_rms.sh` script I follow the Sundials.jl suggestion for using pre-installed libs, but I'm really not sure about the correct name for the lib to be protecting in the first place (or, if this is even the right approach...)

documentation/source/users/rmg/installation/anacondaDeveloper.rst
```diff
@@ -91,40 +81,33 @@ Installation by Source Using Anaconda Environment for Unix-based Systems: Linux
...
     export PATH=YourFolder/RMG-Py/:$PATH
 
    NOTE: Make sure to change ``YourFolder`` to the path leading to the ``RMG-Py`` code. Not doing so will lead to an error stating that python cannot find the module ``rmgpy``.
 
    Be sure to either close and reopen your terminal to refresh your environment variables (``source ~/.bashrc`` or ``source ~/.zshrc``).
 
-#. Install and Link Julia dependencies: ::
-
-     julia -e 'using Pkg; Pkg.add("PyCall");Pkg.build("PyCall");Pkg.add(PackageSpec(name="ReactionMechanismSimulator",rev="for_rmg")); using ReactionMechanismSimulator;'
+#. **Optional (Recommended)**: Install and Link Julia dependencies using ``install_rms.sh``. Ensure that you have modified your environment variables as described above, and then run the following: ::
```
rwest, line 101: As written I think it's important to do the step of adding RMG-Py to the $PATH, which was marked as **Optional** above, with the note "to launch ``rmg.py`` from any folder.". That's a bit confusing. I imagine many people would skip the apparently-optional step, thinking they don't want to run rmg.py from anywhere.

JacksonBurns:
@jonwzheng  e675447 implements the attempt at setting the sundials .so path correctly that we discussed in person

JacksonBurns:
An update:

I confirmed locally that the sundials binaries are definitely at the indicated location, so `Sundials.jl` is just choosing to ignore them during install (and apparently installing its own version of Sundials via `Sundials_jll.jl`, the Julia `BinaryBuilder.jl`-built binaries). This continues to cause a segfault during Cantera execution during regression testing comparison to baseline observables.

A new error has also come up - for some reason, Julia re-precompiles RMS every time it is called. This drops a few thousand lines of output into the [regression test output](https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/14720665855#summary-41316301867) and breaks the bot that comments regression test results.

Current plan is to make a separate conda environment that has RMS in it and use that just for executing the regression tests.

JacksonBurns:
:tada: this is almost done! The issue with RMS simulations always crashing has been fixed by specifying a `linear_solver` for CVODE_BDF explicitly. Here's the TODO-list to push this PR across the finish line:
 - [ ] clean up the latest commit history (primarily all the `[debug]` commits)
 - [ ] add a line to the installation docs about how RMS and Cantera's simulations cannot be run in the same environment
 - [ ] open an issue to track the compatibility of RMS with Cantera simulations that results in a segfault w/n Cantera when calling out to Sundials (namely that we have tried getting RMS to use Canter's Sundials via DL_LOAD_PATH and getting Cantera to use its own Sundials via DL_LOAD_PATH)
 - [ ] clean up the additional output from the `logger` that is crowding up the bot PR comment
 - [ ] review and merge

documentation/source/users/rmg/installation/anacondaDeveloper.rst
JacksonBurns: Circling back to this now that the rest of the PR seems nearly fixed.

I marked this as optional because, in the docker container, the user is dropped into the same directory as `rmg.py` so they can call `python rmg.py ...`. I think when doing a developer install, we should expect that the users would run `python rmg.py /path/to/file/input.py` from the root dir in general, and if they really want Python to be able to find `rmg.py` from wherever they are they can take this step.

rwest: I think that previously it's been crucial to set the environment paths _before_ doing some julia step. Is that still the case? It sounds like it from "Ensure that you have modified your environment variables as described above". 

That's what I meant about it not being optional.

JacksonBurns: AH I see, no this step is not required in order to install RMS

JacksonBurns:
@jonwzheng this is working now, can you review this PR and also the corresponding PR on RMS: https://github.com/ReactionMechanismGenerator/ReactionMechanismSimulator.jl/pull/268

Many thanks!

jonwzheng:
LGTM. To summarize, the latest changes run tests in two separate environments: one with RMS and one without. This way the conflicting Cantera code never runs into the RMS installation of Sundials.

The one thing I would suggest is a note somewhere saying that usage of Cantera with the Julia enviroment is not fully supported and users may encounter crashes.

jonwzheng (approved):

documentation/source/users/rmg/installation/anacondaDeveloper.rst
```diff
@@ -6,6 +6,11 @@ Installation by Source Using Anaconda Environment for Unix-based Systems: Linux
 
 #. Install the `conda` package manager via `miniforge`, if you do not already have it (or Anaconda), by following the `Miniforge installation instructions <https://github.com/conda-forge/miniforge?tab=readme-ov-file#install>`_.
 
+#. If your `conda` version is older than 23.10.0, manually switch the solver backend to `libmamba` ::
```
rwest, line 9:
```suggestion
#. If your `conda` version is older than 23.10.0, manually switch the solver backend to `libmamba` (or update your conda) ::
```

documentation/source/users/rmg/installation/anacondaDeveloper.rst
```diff
@@ -91,40 +81,33 @@ Installation by Source Using Anaconda Environment for Unix-based Systems: Linux
...
 
    Be sure to either close and reopen your terminal to refresh your environment variables (``source ~/.bashrc`` or ``source ~/.zshrc``).
 
-#. Install and Link Julia dependencies: ::
-
-     julia -e 'using Pkg; Pkg.add("PyCall");Pkg.build("PyCall");Pkg.add(PackageSpec(name="ReactionMechanismSimulator",rev="for_rmg")); using ReactionMechanismSimulator;'
+#. **Optional (Recommended)**: Install and Link Julia dependencies using ``install_rms.sh``. Ensure that you have modified your environment variables as described above, and then run the following: ::
 
-     python -c "import julia; julia.install(); import diffeqpy; diffeqpy.install()"
+     . install_rms.sh
```
rwest, line 103:
```suggestion
     source install_rms.sh
```

Doing `. install_rms.sh` on my Mac fails with `.: no such file or directory: install_rms.sh`.

Doing `source install_rms.sh` it at least tries to run the script, and does install pyjuliacall.

rwest (changes requested):
Following the current installation instructions, with a recently updated version of conda, and making a fresh conda environment, I got as far as the `install_rms.sh` script which failed with:
```
install_rms.sh:19: command not found: julia
RMS install error - continuing anyway ¯\_(ツ)_/¯
```

rwest:
The instructions don't seem to mention it, but is the advised way to install a system-wide official julia (`curl -fsSL https://install.julialang.org | sh` ) rather than put one in your conda environment (eg. `conda install -c conda-forge julia`) ? 

Does the latter fail? it's tempting to keep things isolated/sandboxed if possible...

JacksonBurns:
> Following the current installation instructions, with a recently updated version of conda, and making a fresh conda environment, I got as far as the `install_rms.sh` script which failed with:
> 
> ```
> install_rms.sh:19: command not found: julia
> RMS install error - continuing anyway ¯\_(ツ)_/¯
> ```

Oops haha I forgot about the step where you need install juliaup - I put it in the top of the `install_rms.sh` script for later, now is later. Let me revise that.

install_rms.sh
```diff
@@ -0,0 +1,32 @@
...
+# juliaup remove release
+
+#!/bin/bash
+
+conda install -y conda-forge::pyjuliacall
+
+export JULIA_CONDAPKG_BACKEND=Null
+export JULIA_PYTHONCALL_EXE=$CONDA_PREFIX/bin/python
+
+julia -e 'using Pkg; Pkg.add(Pkg.PackageSpec(name="ReactionMechanismSimulator", url="https://github.com/hwpang/ReactionMechanismSimulator.jl.git", rev="fix_installation")); using ReactionMechanismSimulator; Pkg.instantiate()' || echo "RMS install error - continuing anyway ¯\_(ツ)_/¯"
```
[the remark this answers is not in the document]
JacksonBurns: We are already using `for_rmg` on the RMS repo.

We want to go back to this, which will require merging: https://github.com/ReactionMechanismGenerator/ReactionMechanismSimulator.jl/pull/268

Once that is merged, we can update this PR to point to `for_rmg` again

JacksonBurns:
> The instructions don't seem to mention it, but is the advised way to install a system-wide official julia (`curl -fsSL https://install.julialang.org | sh` ) rather than put one in your conda environment (eg. `conda install -c conda-forge julia`) ?
> 
> Does the latter fail? it's tempting to keep things isolated/sandboxed if possible...

Yeah the intended way is to install with `juliaup`, the current recommended Julia installer. This is at the top of `install_rms.sh` but I meant (and forgot) to make it more obvious.

The latter may or may not work with the new setup. Haven't tested, because it's not advised by Julia.

rwest:
> This is nice I'll give it a try. I think it actually needs to apply to the `rmg_env_with_rms` environment, though.

I was dealing with the user instructions, rather than the CI scripts.

I think it works there. 
I think the correct order of operations is
* make conda environment
* install julia and juliaup
* restart your terminal (so `which julia` works) and reactivate your conda environment
* run these commands to modify the variables in your conda environment
* deactivate and reactivate your enivronment (so the variables are set)
* build RMG (`make`) 
* install RMS (could run install_rms.sh, but by now there are only 2 useful lines left)
* test RMG

FWIW, running `make eg0` now runs the whole pip installing stuff, because it's part of the `all` target.
When is this step now necessary?

rwest:
one more thought. is `$HOME/.julia/packages` robust? could we make it more general? (what if someone puts their julia packages elsewhere?)

JacksonBurns:
> I was dealing with the user instructions, rather than the CI scripts.
> 
> I think it works there. I think the correct order of operations is
> 
>     * make conda environment
> 
>     * install julia and juliaup
> 
>     * restart your terminal (so `which julia` works) and reactivate your conda environment
> 
>     * run these commands to modify the variables in your conda environment
> 
>     * deactivate and reactivate your enivronment (so the variables are set)
> 
>     * build RMG (`make`)
> 
>     * install RMS (could run install_rms.sh, but by now there are only 2 useful lines left)
> 
>     * test RMG

I'm not sure I understand what you mean here. I've implemented what I think is a reasonably straightforward install process here: e254b5e37a86dd672d3d1bcab92eed28b7a49951

> FWIW, running `make eg0` now runs the whole pip installing stuff, because it's part of the `all` target. When is this step now necessary?

This happens because the build now uses this command: `python -m pip install -vv --no-cache-dir -e .` which doesn't detect when some files have already been built. This was a cheating shortcut for me so I didn't have to run `make clean && make` everytime. I've updated the makefile to undo this.

> one more thought. is $HOME/.julia/packages robust? could we make it more general? (what if someone puts their julia packages elsewhere?)

I have no idea if this is robust. Users could presumably set this themselves by configuring a Julia package-related environment variable.

JacksonBurns:
@rwest the revised install process with `conda config` seems to be working, I think I addressed all your comments [here](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2741#issuecomment-2908060705) please let me know how things look, then we can (1) merge the RMS PR (2) modify this PR (3) merge this PR

JacksonBurns:
@rwest those look good - let me know about the RMS PR

I think the theory behind using the global julia packages dir is that one can then access RMS without being inside the RMG conda envionrment

setup.py
```diff
@@ -215,41 +133,27 @@
...
-                module = 'arkane' + root.partition('arkane')[-1].replace('/', '.') + '.' + f.partition('.py')[0]
-                modules.append(module)
-
 # Read the version number
 exec(open('rmgpy/version.py').read())
 
 # Initiate the build and/or installation
 setup(
-    name='RMG-Py',
+    name='reactionmechanismgenerator',
```
rwest, line 143: Is this renaming a significant (possibly breaking) change?
what are the consequences and what did it enable?

JacksonBurns (commented):
Latest commits form @rwest are good to me

PR #2749, opened 2025-01-14
Fix Python 3.9 `julia` Overhaul

JacksonBurns:
The Python 3.9 Upgrade is complete, but the new `juliacall` setup does not work. Current theory is that `julia` cannot 'see' the Python environment and/or its dependencies. Will address here.


JacksonBurns:
The pure-python tests fail on MacOS because between Python 3.7 and 3.9 the way multiprocessing works was changed (see https://stackoverflow.com/a/72830812) - we violate the requirements for using this new method (see https://docs.python.org/3/library/multiprocessing.html#the-spawn-and-forkserver-start-methods) so we may just need to set the method back to the old way.

JacksonBurns:
Using PythonCall with an existing conda environment: https://discourse.julialang.org/t/how-to-use-pythoncall-with-a-previous-conda-environment/87419

JacksonBurns:
Setting `JULIA_CONDAPKG_BACKEND` to `System` caused RMS to install new conda packages during its build, since it could not 'see' the `rmg_env` conda environment. This resulted in the newly installed Python not being on the path, and then failing to find its `encodings` module during import (and thus killing the CI run).

Now attempting to set `JULIA_CONDAPKG_BACKEND` to `Current`, which should make RMS 'see' the `rmg_env`.

JacksonBurns:
Julia sees the correct environment (I think), so next is solving the error that Julia can't see the libraries. I think this may fix it: https://stackoverflow.com/a/77206459

JacksonBurns:
`PYTHONPATH` is likely set incorrectly, see: https://github.com/pythonnet/pythonnet/discussions/2147

JacksonBurns:
Same error; no solution: https://discourse.julialang.org/t/fatal-python-error-when-running-julia-on-mac-m1-vs-code/109540/4

JacksonBurns:
More related: https://github.com/JuliaPy/PyCall.jl/issues/410#issuecomment-406794511

JacksonBurns:
Next debugging step is to set JULIA_PYTHONCALL_EXE to the python executable rather than setting PYTHON: https://juliapy.github.io/PythonCall.jl/stable/pythoncall/#If-you-already-have-Python-and-required-Python-packages-installed

JacksonBurns:
Seems that for Ubuntu the install works, but then during tests Python can't find it's `libs` - likely candidate is Python home being set incorrectly (i.e. at all)

JacksonBurns:
The RMG modules are built in place, which is why we need to set PYTHONPATH: https://stackoverflow.com/a/7851585

Maybe we also need to set the system path to point toward the libs in the conda environment?

JacksonBurns:
Intel Mac now fails only the tests that are due to the change in `fork` - why can't ubuntu see the `libs`? Going to set the library path for ld

JacksonBurns:
Current theory is that we have to set PYTHONPATH because we aren't actually installing rmgpy or arkane, just building them, so that Julia can see arkane and rmgpy during its installation process. But this then messes up the actual execution, since setting PYTHONPATH seems to mess up the normal path, making python unable to find its libs

Potential fix to this potential problem is to make the code actually installable

JacksonBurns:
@jonwzheng a familiar name has popped up on this PR:
``` 
E   ImportError: /home/runner/miniconda3/envs/rmg_env/lib/python3.9/site-packages/scipy/sparse/../../../../libstdc++.so.6: version `GLIBCXX_3.4.31' not found (required by /home/runner/miniconda3/envs/rmg_env/lib/python3.9/site-packages/rdkit/Chem/../../../../libboost_serialization.so.1.86.0)
```

JacksonBurns:
I have now confirmed that _without_ installing Julia, the tests work. Something about installing `juliaup` and then RMS is causing python to be unable to find its libraries. Going to just printenv and see if juliaup is severely messing things up

JacksonBurns:
Comparison of the environment variables and the conda environment before and after the julia install: https://www.diffchecker.com/5ETXidNB/

of note is that the rdkit version is changed, as well as `libstdcxx-ng`

JacksonBurns:
Current theory:
 - we build rmg, then install julia, in the process of installing Julia we change some versions of some deps, this causes the built version of RMG to no longer work

Possible solutions (?):
 - rebuild RMG after installing Julia (will this cause a chicken-and-egg problem where Julia won't be compatible anymore?)
 - build RMG after installing Julia (will Julia still correctly be able to import and use RMG if it is not built? I think no)
 - set dependency versions to be more compatible with Julia (probably easier this way)

Note also that the PATH was changed when Juliaup was installed, but I don't think this is causing the problem

JacksonBurns:
Update: the MacOS actually _worked_ - it appears that during Julia installation on ubuntu the change to libstdcxx causes issues, but on Mac this isn't touched

JacksonBurns:
Candidate for why ARM mac is broken: https://stackoverflow.com/a/71554275

JacksonBurns:
well, at least the CI is passing on all platforms in pure-python mode now...

JacksonBurns:
@jonwzheng where does this stand now?

jonwzheng:
I've tried updating LD_LIBRARY_PATH in the environment but nothing seemed to change. May need to check/confirm that the path was the correct one

JacksonBurns:
Ok, thanks for giving that a shot. I''ve got a new idea - rather than trying to get Julia to 'see' the packages we want and to not change the packages we already have, we can just disallow it from doing the latter entirely: https://juliapy.github.io/PythonCall.jl/stable/pythoncall/#If-you-already-have-Python-and-required-Python-packages-installed

In theory, if we install all of the packages that RMS would need manually, then Julia won't attempt to change them, and thus hopefully won't break the libstdc++ lib

JacksonBurns:
Possibly need to change RMS, getting this error: `ERROR: LoadError: Can not get envdir when backend is Null.`

JacksonBurns:
a84641c introduces changes to RMS that will stop the current issue, at least for now. If RMS attempts to install things still, it might go awry.

Cancelling and re-starting CI now

JacksonBurns:
Miracle of miracles, that actually worked...

JacksonBurns:
Now addressing the still-failing Apple Silicon in this PR: https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2751

JacksonBurns:
All of the RMS-calling regression tests are failing with this error:
```bash
2025-01-22T18:46:35.2662765Z AttributeError: Julia: UndefVarError: `TerminationConversion` not defined in `Main`
2025-01-22T18:46:38.4740641Z RMS_constantVIdealGasReactor_fragment Failed to Execute
```

JacksonBurns:
Ah, right at the start of the tests this happens:
```
2025-01-22T19:32:05.1095495Z collecting ... [juliapkg] Found dependencies: /home/runner/miniconda3/envs/rmg_env/lib/python3.9/site-packages/juliacall/juliapkg.json
2025-01-22T19:32:05.1096946Z [juliapkg] Found dependencies: /home/runner/miniconda3/envs/rmg_env/lib/python3.9/site-packages/juliapkg/juliapkg.json
2025-01-22T19:32:05.1097902Z [juliapkg] Locating Julia ~1.6.1, ~1.7, ~1.8, ~1.9, =1.10.0, ^1.10.3
2025-01-22T19:32:05.1098824Z [juliapkg] Installing Julia 1.11.3 using JuliaUp
2025-01-22T19:32:05.1099756Z [juliapkg] Using Julia 1.11.3 at /home/runner/.julia/juliaup/julia-1.11.3+0.x64.linux.gnu/bin/julia
2025-01-22T19:32:05.1100538Z [juliapkg] Using Julia project at /home/runner/miniconda3/envs/rmg_env/julia_env
2025-01-22T19:32:05.1101074Z [juliapkg] Installing packages:
2025-01-22T19:32:05.1101417Z            julia> import Pkg
2025-01-22T19:32:05.1101749Z            julia> Pkg.Registry.update()
2025-01-22T19:32:05.1102577Z            julia> Pkg.add([Pkg.PackageSpec(name="PythonCall", uuid="6099a3de-0909-46bc-b1f4-468b9a2dfc0d")])
2025-01-22T19:32:05.1103188Z            julia> Pkg.resolve()
2025-01-22T19:32:05.1103512Z            julia> Pkg.precompile()
2025-01-22T19:32:05.1104234Z Detected IPython. Loading juliacall extension. See https://juliapy.github.io/PythonCall.jl/stable/compat/#IPython
```
`juliacall` can't see the version of Julia we went to all of the trouble of preparing

JacksonBurns:
How to configure the above to hopefully not happen: https://github.com/JuliaPy/PyJuliaPkg?tab=readme-ov-file#details

JacksonBurns:
When attempting to run this locally I have been running into segfaults during the RMS install. Turns out they were actually just on the _exit_ of the install command. Some reading suggests that this is expected or at least OK: https://discourse.julialang.org/t/my-julia-code-has-segfault-only-on-exit-is-there-a-way-to-simply-stop-julia-without-calling-into-this-finalization-stuff/106538

Adding a step to the install script that will continue the install even if the command actually raises an error...

JacksonBurns:
Regression tests now run locally - fail with some errors related to the changes made to make RMS optional (phases is None would it should be an RMS object). Just need to debug that now, going to let CI list out all the errors.

JacksonBurns:
> How to configure the above to hopefully not happen: https://github.com/JuliaPy/PyJuliaPkg?tab=readme-ov-file#details

May need to also set PYTHON_JULIAPKG_PROJECT to ensure we see the installed RMS package

JacksonBurns:
... or just let juliacall install everything?

JacksonBurns:
> May need to also set PYTHON_JULIAPKG_PROJECT to ensure we see the installed RMS package

cc2c9ba tries this

Note that we will want to, at some point, run the regression tests on only one platform (right now, if this code were all working, we would be running them on all three platforms).

JacksonBurns:
Fixed the last couple issues I believe! I am moving this work to another PR that is cleaner.

PR #2753, opened 2025-01-24
Fix database issues with loading new kinetic libraries

jonwzheng:
See https://github.com/ReactionMechanismGenerator/RMG-database/issues/678. 

Some of the new libraries are causing issues with loading the relevant kinetics data into the RMG database.

(1) `LithiumPrimaryChargedKinetics`. The cause of this error was `ArrheniusChargeTransfer` not being included in the local context. **This has been fixed in this PR.**
(2) `LithiumAnalogyKinetics`. The cause of this error is not clear to me yet; still investigating.  As this gets fixed (and if any other errors pop up), they will be addressed in this PR.

To reproduce, try:

```
from rmgpy.data.rmg import RMGDatabase
from rmgpy import settings

database = RMGDatabase()

database.load(settings['database.directory'], thermo_libraries = []', 
              kinetics_families=[], kinetics_depositories=[], reaction_libraries=None) # None loads all libraries
```

### DIFF
--- a/rmgpy/data/kinetics/database.py
+++ b/rmgpy/data/kinetics/database.py
@@ -268,7 +268,16 @@ class KineticsDatabase(object):
                 for f in files:
                     if f.lower() == 'reactions.py':
                         library_file = os.path.join(root, f)
-                        label = os.path.dirname(library_file)[len(path) + 1:]
+                        dirname = os.path.dirname(library_file)
+                        if dirname == path:
+                            label = os.path.basename(dirname)
+                        else:
+                            label = os.path.relpath(dirname, path)
+
+                        if not label:
+                            logging.warning(f"Empty label for {library_file}. Using 'default'.")
+                            label = "default"
+                        
                         logging.info(f'Loading kinetics library {label} from {library_file}...')
                         library = KineticsLibrary(label=label)
                         try:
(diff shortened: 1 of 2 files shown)


ssun30:
Both issues are solved, plus some extras for dealing with library labeling and debugging help for future developers. This PR should be ready to deploy.

jonwzheng:
LGTM, thanks for the fixes

rmgpy/data/kinetics/database.py
```diff
@@ -84,11 +87,13 @@ def __init__(self):
             'SurfaceChargeTransfer': SurfaceChargeTransfer,
             'R': constants.R,
             'ArrheniusBM': ArrheniusBM,
+            'ArrheniuisChargeTransferBM': ArrheniusChargeTransferBM,
```
jonwzheng, line 90: actually just saw this: should be `ArrheniusChargeTransfer`

jonwzheng (commented):
just saw this one last thing that needs to be fixed ^

JacksonBurns (approved):
Lgtm thanks

rwest:
Sorry, I could answer this myself if I had time to look properly, but did we (or can we) add any unit tests or database tests that would've helped us catch this earlier?

jonwzheng:
Su added debug messages that greatly helps figure out the source of these issues, but no we didn't add any tests.
 
We already have one for loading the databases the "canonical" way, so we should add some tests for loading various databases, something like:

```
database = RMGDatabase()
database.kinetics = KineticsDatabase()
database.kinetics.load_libraries(os.path.join(settings["database.directory"], "kinetics", "libraries"),)
```
and maybe also for thermo and transport, and ensure that they are identical to the database loaded with `RMGDatabase().load()`

PR #2756, opened 2025-02-03
Fxi UnboundLocalError: local variable 'e_electronic_with_corrections' referenced before assignment

alongd:
### Motivation or Problem
Arkane's `statmech.py` has a variable defined within an if statement, but used outside of it. When the statement evaluates to `False`, Arkane crushes.

### Description of Changes
Defined `e_electronic_with_corrections` and `zpe` formally outside the if statement, and properly deal with `e0` assignment after the condition.

### Testing
An Arkane example for feeding in E0 directly for a species was added
(Arkane's examples are automatically executed as functional tests)

### DIFF
--- a/examples/arkane/species/OCHO/OCHO.py
+++ b/examples/arkane/species/OCHO/OCHO.py
@@ -1,0 +1,14 @@
+#!/usr/bin/env python3
+# encoding: utf-8
+
+externalSymmetry = 2
+
+spinMultiplicity = 2
+
+opticalIsomers = 1
+
+energy =  (-342.84,'kJ/mol')
+
+geometry = Log('freq.out')
+
+frequencies = Log('freq.out')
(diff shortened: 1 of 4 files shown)


examples/arkane/species/OCHO/OCHO.py
```diff
@@ -0,0 +1,14 @@
...
+
+externalSymmetry = 2
+
+spinMultiplicity = 2
+
+opticalIsomers = 1
+
+energy =  (-342.84,'kJ/mol')
+
+geometry = Log('OCHO/freq.out')
```
sevyharris, line 12:
```suggestion
geometry = Log('freq.out')
```

examples/arkane/species/OCHO/OCHO.py
```diff
@@ -0,0 +1,14 @@
...
+
+frequencies = Log('OCHO/freq.out')
```
sevyharris, line 14:
```suggestion
frequencies = Log('freq.out')
```

alongd:
Thanks for the catch, @sevyharris! I updated the path now

sevyharris (approved):

Issue #2757, opened 2025-02-05
Projecting out rotors in Arkane leads to zero frequency


sevyharris:
### Bug Description
I am trying to compute the thermodynamics for a species (smiles='CCCCOO') by running an Arkane statmech + thermo job, but there is a very large difference between my entropy and the reference data I'm using from the AramcoMech3.0, leading to a 27 kcal/mol error in Gibbs free energy at 1000K.
![Image](https://github.com/user-attachments/assets/97e864b0-5fbc-432b-abad-f2d45e09598d)

This led me to look at the rotors. I ran 1-D hindered rotor scans (at 20 degree increments) for the molecule and they looked reasonable in Gaussview, but @rwest noticed that after Arkane projects out the rotors, one of the vibrational frequencies is approximately zero (1e-5). 

```
HarmonicOscillator(
        frequencies = ([4.79104e-05, 424.798, 521.468, 584.421, 778.857, 820.924, 923.676, 952.077, 1015.85, 1044.41, 1084.27, 1107.11, 1156.83, 1241.45, 1259.22, 1294.35, 1298.53, 1335.14, 1346.66, 1375.05, 1392.76, 1419.19, 1430.83, 1464.16, 1477.48, 1501.56, 1553.74, 2611.14, 2672.06, 2713.94, 2729.57, 2761.52, 2804.99, 2828.13, 2855.04, 2875.21, 2981.48], 'cm^-1'),
)
```
You can see all the statmech modes in the [full output.py](https://github.com/user-attachments/files/18678480/output.txt)

I'm trying to figure out whether 1. I've messed up my input to Arkane, 2. Arkane is projecting the rotors out incorrectly, or 3. the rotation and vibrational modes are too tightly coupled for this kind of calculation to work. [This is known to be a problem for OOQOOH radicals](https://doi.org/10.1016/j.proci.2024.105618) but my species is a little simpler. I have successfully calculated dozens of species thermodynamics with my current workflow, but I also see this rotors/zero-frequency issue with a handful of other species: (smiles='CC(CC=O)OO', 'CC(CCOO)O[O]', 'CC(CCO[O])OO', and 'O=CCCCOO').

I would appreciate any and all hints for how to tackle debugging this issue.

### How To Reproduce
These are the files I'm using (.py extensions changed to .txt). For simplicity, I have excluded my single point energy calculation (used to generate the plots above), and energy, frequencies and geometry are all taken from the conformer_0002.log file.

[...]

PR #2758, opened 2025-02-07
Arkane: add check for iop(2/9=2000) in Gaussian load geometry

sevyharris:
### Motivation or Problem
By default, Gaussian only prints Cartesians in the input orientation for small cases. Add a check for the IOP(2/9=2000) keyword so users don't get tripped up like I did here: https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2757 with coordinates that don't match the Hessian.

### Description of Changes
I added a check for the IOP(2/9=2000 keyword) when loading the geometry. If the molecule has 14+ atoms (this seems to be the threshold according to my experiments), this will raise a LogError if the keyword hasn't been specified. 

### Testing
I tested this using one Gaussian logfile that has the IOP and one that doesn't.
[conformer_0002_iop.log](https://github.com/user-attachments/files/18709550/conformer_0002_iop.log)
[conformer_0002_no_iop.log](https://github.com/user-attachments/files/18709559/conformer_0002_no_iop.log)

### DIFF
--- a/arkane/ess/gaussian.py
+++ b/arkane/ess/gaussian.py
@@ -132,8 +132,13 @@ class GaussianLog(ESSAdapter):
         only the last is returned. The units of the returned force constants
         are J/m^2. If no force constant matrix can be found in the log file,
         ``None`` is returned.
+        Also checks that the force constant matrix was computed using the correct
+        (input orientation Cartesian) coordinates.
+        IOP(2/9=2000) must be specified for large cases (14+ atoms)
         """
         force = None
+
+        iop2_9_equals_2000 = False
 
         n_atoms = self.get_number_of_atoms()
         n_rows = n_atoms * 3
@@ -141,6 +146,9 @@ class GaussianLog(ESSAdapter):
         with open(self.path, 'r') as f:
             line = f.readline()
             while line != '':
+                if '2/9=2000' in line:
+                    iop2_9_equals_2000 = True
+
                 # Read force constant matrix
                 if 'Force constants in Cartesian coordinates:' in line:
                     force = np.zeros((n_rows, n_rows), float)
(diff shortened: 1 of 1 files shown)


arkane/ess/gaussian.py
```diff
@@ -164,12 +164,18 @@ def load_geometry(self):
...
+        IOP(2/9=2000) must be specified for large cases (14+ atoms)
         """
         number, coord, mass = [], [], []
 
+        iop2_9_equals_2000 = False
         with open(self.path, 'r') as f:
             line = f.readline()
             while line != '':
+                if '2/9=2000' in line:
+                    iop2_9_equals_2000 = True
```
sevyharris, line 177: This is the wrong place in the code for the check. It isn't about the Cartesian coordinates-- it's the fact that the Hessian isn't consistent with those Cartesian coordinates, so move this to the load_force_matrix function

sevyharris:
This is currently failing the unit tests because the toluene and benzyl Arkane examples have Gaussian logfiles with more than 14 atoms. I'm rerunning these calculations to see how much it changes the results. If it's a lot, I'll replace the logfiles. If it's not much, that means I've been overstating the error due to this bug and I may not fully understand the problem I thought I was diagnosing.

rwest:
It fails the unit test, because a log file for Toluene.

> FAILED test/rmgpy/statmech/ndTorsionsTest.py::TestHinderedRotorClassicalND::test_hindered_rotor_nd - arkane.exceptions.LogError: Gaussian output file /Users/runner/work/RMG-Py/RMG-Py/arkane/data/TolueneFreq.log contains more than 13 atoms. Please add the `iop(2/9=2000)` keyword to your input file so Gaussian will compute force matrix using the input orientation Cartesians.

Is the test too restrictive? or should the log file be replaced? or...?

sevyharris:
> It fails the unit test, because a log file for Toluene.
> 
> > FAILED test/rmgpy/statmech/ndTorsionsTest.py::TestHinderedRotorClassicalND::test_hindered_rotor_nd - arkane.exceptions.LogError: Gaussian output file /Users/runner/work/RMG-Py/RMG-Py/arkane/data/TolueneFreq.log contains more than 13 atoms. Please add the `iop(2/9=2000)` keyword to your input file so Gaussian will compute force matrix using the input orientation Cartesians.
> 
> Is the test too restrictive? or should the log file be replaced? or...?

So, as we just saw with David's workflow, I think the problem is only when you combine optimization and frequency. Here, frequency is its own separate calculation, so it's fine. I should add a check to see if the Gaussian job did any optimization in the same job as the frequency calculations.

sevyharris:
I changed it so it only raises this error if it's grabbing a Hessian from a calculation that also included an optimization (which might change the orientation).

So calculations like the Arkane examples that just do the frequency are fine.

It should now pass the test it was failing earlier.

PR #2766, opened 2025-02-26
require a more recent version of mkl for Cantera

sevyharris:
### Motivation or Problem
The RMG-database continuous integration is failing because it can't find the right mkl library file: https://github.com/ReactionMechanismGenerator/RMG-database/issues/683

Rather than slap a bandaid over the problem (this PR: https://github.com/ReactionMechanismGenerator/RMG-database/pull/684), this PR tries to fix the versions conda requires.

From what I've seen, I believe it's Cantera that has the MKL dependency. It doesn't specify a version and then errors out when conda grabs something so old it can't find the right file name.

### Description of Changes
I added a version requirement for MKL to the RMG-Py environment file. I chose the version 2023.1.0 because it's what I have on my local Window's subsystem linux and I was worried that choosing the latest 2025 version would be too strict and cause the conda solver to fail.

### Testing
I'm not really sure how to test this besides letting the usual Github actions run. Then we merge it and see if that fixes the database tests? Let me know if you have any ideas.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -110,6 +110,7 @@ dependencies:
 # depends on OpenBLAS instead
   - conda-forge::mkl >=2023.1.0
   - conda-forge::libcurl <= 8.9  # this is because of a Julia bug associated with libcurl 8.10+
+  - conda-forge::pyopenssl >20
 
 # additional packages that are required, but not specified here (and why)
   # pydqed, pydas, mopac, and likely others require a fortran compiler (specifically gfortran)
--- a/environment.yml
+++ b/environment.yml
@@ -109,6 +109,7 @@ dependencies:
 # but if we can switch over to the conda-forge version, we should be able to remove this line because it
 # depends on OpenBLAS instead
   - conda-forge::mkl >=2023.1.0
+  - conda-forge::libcurl <= 8.9  # this is because of a Julia bug associated with libcurl 8.10+
 
 # additional packages that are required, but not specified here (and why)
   # pydqed, pydas, mopac, and likely others require a fortran compiler (specifically gfortran)
--- a/environment.yml
+++ b/environment.yml
@@ -105,12 +105,17 @@ dependencies:
 # configure packages to use OpenBLAS instead of Intel MKL
   - blas=*=openblas
 
-# except that Cantera is configured to use Intel MKL, so we have to make sure its version is recent
-# but if we can switch over to the conda-forge version, we should be able to remove this line because it
-# depends on OpenBLAS instead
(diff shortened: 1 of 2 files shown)


JacksonBurns:
@sevyharris sorry I didn't think of this earlier, but we could also consider changing the channel for `cantera` from `cantera` to `conda-forge` - this is something we did over in #2741

sevyharris:
> @sevyharris sorry I didn't think of this earlier, but we could also consider changing the channel for `cantera` from `cantera` to `conda-forge` - this is something we did over in #2741

Ooh, I like that idea! I'll try that as soon as this round of tests finishes. I'm curious to see if this set of changes passes.

sevyharris:
Mamba can't solve the environment for OSX using Python3.7 and conda-forge Cantera, so I'm gonna roll this back to requiring conda-forge::mkl >=2023.1.0

sevyharris:
So it seems this MKL requirement has a cascading effect that breaks the PyCall installation for Julia in the Build Documentation test. The curl versions changed and I guess PyCall can't handle the new one? 

https://www.diffchecker.com/MPHzDj0H/ (left passed and right failed)

I'll play around with the versions and see if I can get the Build Documentation test working with an environment that enforces a more recent MKL version.

And thanks @JacksonBurns for showing me the diffchecker site. It's very convenient!

sevyharris:
This thread says the error happens for libcurl>=8.10 https://discourse.julialang.org/t/curl-multi-assign-and-segmentation-fault-when-installing-package/120901/3

sevyharris:
The indignity of being failed by a unit test with a typo in the name 😅

```
2025-02-26T20:09:20.9561378Z =========================== short test summary info ============================
2025-02-26T20:09:20.9562158Z FAILED test/rmgpy/rmg/inputTest.py::TestInputThemoCentralDatabase::test_themo_central_database - AttributeError: module 'lib' has no attribute 'X509_V_FLAG_CB_ISSUER_CHECK'
2025-02-26T20:09:20.9562976Z ===== 1 failed, 1968 passed, 38 skipped, 39 warnings in 2206.14s (0:36:46) =====
```

Looks like this is actually another version error but this time with openssl, so I'll try updating that.
https://stackoverflow.com/questions/73830524/attributeerror-module-lib-has-no-attribute-x509-v-flag-cb-issuer-check

sevyharris:
Okay, it should be ready to go, but let me know if you have any more suggestions.

JacksonBurns (approved):
thanks for the thorough fix! the good news is that the Python 3.9 upgrade will remove our use of PyCall

sevyharris:
> thanks for the thorough fix! the good news is that the Python 3.9 upgrade will remove our use of PyCall

🙏I can't wait!

PR #2767, opened 2025-02-26
Fake PR to test build documentation test

sevyharris:
### Motivation or Problem
Did adding the mkl requirement break the build documentation test? I don't understand how that would be possible. Trying to see if I can get that test to fail without any changes to the main branch.

https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2766


sevyharris:
Dang. The documentation builds fine here. Maybe it is my code.

PR #2775, opened 2025-03-28
Fix bug in electrons initialization for depository reaction data

jonwzheng:
This PR fixes a small bug:

- `Reaction` expects `electrons` to be an integer, but the constructor for Depository reactions initializes `electrons` as `None`. This can lead to crashes (as was observed for the `RMG-website`: https://github.com/ReactionMechanismGenerator/RMG-website/issues/288)
- This PR simply changes the constructor to use a default value of `0` rather than `None` like other classes that inherit from `Reaction`.

### DIFF
--- a/rmgpy/data/kinetics/depository.py
+++ b/rmgpy/data/kinetics/depository.py
@@ -62,7 +62,7 @@ class DepositoryReaction(Reaction):
                  depository=None,
                  family=None,
                  entry=None,
-                 electrons=None,
+                 electrons=0,
                  ):
         Reaction.__init__(self,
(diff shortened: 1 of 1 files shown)


JacksonBurns:
I see that the default value in `Reaction` is already zero ([here](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/55464c54d1fa61b531e865682df598d33718597d/rmgpy/reaction.py#L115C18-L115C27)) - this change makes sense to me.

JacksonBurns (approved):

PR #2780, opened 2025-04-08
Update Disproportionation family in the test database

sevyharris:
### Motivation or Problem
The Disproportionation family was converted to an autogenerated SIDT tree years ago, but we're still using a highly outdated version for testing because there's no automatic transfer from RMG-database, and it takes effort to update since a lot of unit tests were written for that version.

The fact that this version of the database is so old should be reason enough to change it: what's the point of testing RMG-Py using entries that no longer resemble RMG-database? But I also need a decent sized autogenerated kinetics tree for some of the tests I'll be writing for uncertainty estimation, and Disproportionation has been my go-to example. So, that's why I'm updating this family in the test database.

### Description of Changes
I copied the latest Disproportionation family from RMG-database cbc8f9f5a11c179208ad078d4010afc8355698fa and put it inside RMG-Py's test database. Now it's just a matter of updating all the failing unit tests so they know to expect a newer version of this kinetics tree:

Some of the highlights so far are
- kineticsTest.py::TestKineticsCommentsParsing::test_parse_kinetics FAILED
- kineticsTest.py::TestReactionDegeneracy::test_degeneracy_same_reactant_different_resonance_structure FAILED

### Testing
This change only affects the RMG-Py tests, and so it should be enough to make sure the tests are still passing.

### DIFF
--- a/test/rmgpy/data/kinetics/kineticsTest.py
+++ b/test/rmgpy/data/kinetics/kineticsTest.py
@@ -607,8 +607,6 @@ class TestReactionDegeneracy:
 
         reaction_list = self.assert_correct_reaction_degeneracy(reactants, correct_rxn_num, correct_degeneracy, family_label, products)
 
-        assert set(reaction_list[0].template) == {"C_rad/H2/Cd", "Cmethyl_Csrad/H/Cd"}
-
     def test_degeneracy_multiple_ts_different_template(self):
         """Test that reactions from different transition states are marked as duplicates."""
         family_label = "intra_H_migration"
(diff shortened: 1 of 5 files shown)


sevyharris:
I understand why nobody ever updates this. You'd also have to regenerate a lot of the example chemkin files (see RMG-Py/test/rmgpy/test_data/parsing_data/chem_annotated.inp) and I don't know how you get RMG to run using the test database instead of the regular database. Probably people have some manual workaround.

This update is far more complicated than the change that motivated this PR, so I'm tabling this for now. I might have to come back and update the test database when it comes time to implement tests for the uncertainty module as a whole, but for now I just need tests for the extract_source_from_comment() function, and I can do that with the current version of the database.

Issue #2782, opened 2025-04-10
Dissociative adsorption generates wrong flux pairs?


sevyharris:
### Bug Description
I'm trying to make some flux diagrams for surface mechanisms, and I've noticed a few missing arrows for surface dissocative adsorption. For example, this reaction:
![Image](https://github.com/user-attachments/assets/0e07d98c-3325-4f80-822c-40058128b1c8)

Is missing the arrow from NH3 to HX:
![Image](https://github.com/user-attachments/assets/20e1f942-5f2e-428e-8ce9-fcf84afa053c)

### How To Reproduce
The flux diagram is experimental code, but you can see the annotated flux pairs RMG generates for the reaction:
```
! Reaction index: Chemkin #10; RMG #18
! Template reaction: Surface_Adsorption_Dissociative
! Flux pairs: NH3(8), H2NX(17); X(1), HX(3); X(1), HX(3); 
! Estimated using template [Adsorbate;VacantSite1;VacantSite2] for rate rule [N;VacantSite1;VacantSite2]
! Euclidian distance = 1.0
! Multiplied by reaction path degeneracy 3.0
! family: Surface_Adsorption_Dissociative
X(1)+X(1)+NH3(8)<=>HX(3)+H2NX(17)                   4.500e-02 0.000     1.195    
    STICK
```

Or generate them yourself with this code:
```
import rmgpy.species
import rmgpy.reaction

my_reaction = rmgpy.reaction.Reaction()
NH3 = rmgpy.species.Species(smiles='N')
X = rmgpy.species.Species().from_adjacency_list(
"""1 X u0 p0 c0"""
)
H2NX = rmgpy.species.Species().from_adjacency_list(
"""1 N u0 p1 c0 {2,S} {3,S} {4,S}
2 H u0 p0 c0 {1,S}
3 H u0 p0 c0 {1,S}
4 X u0 p0 c0 {1,S}"""
)
HX = rmgpy.species.Species().from_adjacency_list(
"""1 H u0 p0 c0 {2,S}
2 X u0 p0 c0 {1,S}"""
)

my_reaction.reactants = [NH3, X, X]
my_reaction.products = [H2NX, HX]

display(my_reaction)

my_reaction.generate_pairs()
for pair in my_reaction.pairs:
    print(f'{pair[0].smiles}\t->{pair[1].smiles}')
```

### Expected Behavior
[...]

PR #2783, opened 2025-04-10
Add handling for dissociative desorption reactions to flux pair generation

sevyharris:
### Motivation or Problem
RMG misses a flux pair when generating the flux pairs for surface dissocative adsorption reactions. See https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2782 describing the issue.

### Description of Changes
This code adds a case for handling dissocative adsorption reactions in the reactions.generate_pairs() function so that the gas-phase reactant gets paired with both surface products instead of just the bigger one.

### Testing
I need to add some formal unit tests for these.

### Reviewer Tips
Are there any other places in the code besides the flux diagrams that use the flux pairs? I'll need to make sure this still works for those applications.

### DIFF
--- a/test/rmgpy/reactionTest.py
+++ b/test/rmgpy/reactionTest.py
@@ -785,6 +785,68 @@ class TestReaction:
         )
         self.reaction4_pairs = [(PO3, HOPO2), (H2, H_atom)]
 
+        # Reaction definitions for testing flux pair generation
+        # Isomerization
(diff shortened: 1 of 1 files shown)


sevyharris:
It now draws the [missing arrow](https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2782) in my flux diagram for NH3 -> HX:

image

PR #2788, opened 2025-04-26
Add details to thermo comments.

rwest:
### Motivation or Problem
When debugging weird thermo, it's helpful to know how things were estimated.

### Description of Changes
Linear Scaling Relations (LSRs) for moving from one catalyst to another, and solvation corrections, can both change thermo. The comments should now tell you how much it changed by.

### Testing
I ran the unit test suite.

### DIFF
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -1538,15 +1538,16 @@ class ThermoDatabase(object):
 
         # now edit the adsorptionThermo using LSR
         comments = []
-        for element,bond in normalized_bonds.items():
+        change_in_binding_energy = 0
+        for element, bond in normalized_bonds.items():
             if bond:
                 try:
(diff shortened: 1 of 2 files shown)


rwest:
@ssun30 I'm afraid I only checked that it doesn't crash. Before approving, please check that it actually reports helpful comments (which are formatted OK).

ssun30:
> @ssun30 I'm afraid I only checked that it doesn't crash. Before approving, please check that it actually reports helpful comments (which are formatted OK).

Yes, these reports work as expected and are formatted properly.

ssun30 (approved):

rwest:
Thanks for the review @ssun30.
I fixed the `))` typo and will set to automerge.

ssun30:
> Thanks for the review @ssun30. I fixed the `))` typo and will set to automerge.

I'm not a reviewer with write access so the automerge is still blocked.

JacksonBurns (approved):

PR #2791, opened 2025-05-13
Remove incorrectly calculated and applied energy_correction from pdep.py

donerancl:
After finding many submerged barriers when running RMG in pdep mode, I discovered that the `energy_correction` is calculated incorrectly *and* only applied to the transition state. The code calculates the `energy_correction` by averaging all species E0s. However, it should average the total energy for each stationary point (some of which are bimolecular), and I believe it would be more intuitive to use the minimum energy on the surface. A simple solution would be to remove the `energy_correction` altogether (what I've done here for now). If it is desired, we need to apply `energy_correction` to all stationary points, not just transition states, so that barriers are not submerged. 

The `energy_correction` could be calculated as follows...
```
stationary_point_E0s = [sum([spec.conformer.E0.value_si for spec in stationary_point]) for stationary_point  in self.reactants + self.isomers + self.products])]
energy_correction = -np.array(stationary_point_E0s).min()
```

I would need to track down where to apply it to the wells though.

### Motivation or Problem

### Description of Changes

### Testing

### Reviewer Tips

### DIFF
--- a/rmgpy/rmg/pdep.py
+++ b/rmgpy/rmg/pdep.py
@@ -816,14 +816,6 @@ class PDepNetwork(rmgpy.pdep.network.Network):
                     spec.conformer = Conformer(E0=spec.get_thermo_data().E0)
                 E0.append(spec.conformer.E0.value_si)
 
-        # Use the average E0 as the reference energy (`energy_correction`) for the network
-        # The `energy_correction` will be added to the free energies and enthalpies for each
-        # configuration in the network.
-        energy_correction = -np.array(E0).mean()
-        for spec in self.reactants + self.products + self.isomers:
-            spec.energy_correction = energy_correction
-        self.energy_correction = energy_correction
-
         # Determine transition state energies on potential energy surface
         # In the absence of any better information, we simply set it to
         # be the reactant ground-state energy + the activation energy
@@ -851,9 +843,9 @@ class PDepNetwork(rmgpy.pdep.network.Network):
                                 'type "{2!s}".'.format(rxn, self.index, rxn.kinetics.__class__))
             rxn.fix_barrier_height(force_positive=True)
             if rxn.network_kinetics is None:
-                E0 = sum([spec.conformer.E0.value_si for spec in rxn.reactants]) + rxn.kinetics.Ea.value_si + energy_correction
+                E0 = sum([spec.conformer.E0.value_si for spec in rxn.reactants]) + rxn.kinetics.Ea.value_si 
             else:
-                E0 = sum([spec.conformer.E0.value_si for spec in rxn.reactants]) + rxn.network_kinetics.Ea.value_si + energy_correction
+                E0 = sum([spec.conformer.E0.value_si for spec in rxn.reactants]) + rxn.network_kinetics.Ea.value_si
             rxn.transition_state = rmgpy.species.TransitionState(conformer=Conformer(E0=(E0 * 0.001, "kJ/mol")))
 
         # Set collision model


donerancl:
@JacksonBurns @alongd would you be willing to review this PR? It is rather straightforward if we are willing to not apply `energy_correction` in RMG's pdep mode. This correction is only intended to change the reference E0 and only the relative E0 is important, so I think it is not necessary.

rwest:
Oh, boy, now I feel bad. I came across this problem (I think) back in December, and was half way to opening an issue and/or pull request, then got distracted.

I was debugging a PDep network and as commonly done I we would find the `network_X_Y.py` file and make Arkane draw the PDF. Getting tired of this in  in https://github.com/ReactionMechanismGenerator/RMG-Py/commit/9a6bd590bbb99a4d4f63fa0bee7d989b16b56757 I made RMG save the PDF of the network automatically as it crashes. Strangely, the two outputs were different!
Something like these two.

Screenshot 2025-05-14 at 2 24 41 PM

I got about as far as  @donerancl in realizing it's the inconsistently applied "correction".

I think I determined that during running of the RMG job, the correction _was_ correctly applied, but it was inconsistently (and incorrectly) saved in the arkane input files. The conformers would have the correction applied, but not the TS, or something like that.


I kept the PDFs and windows and tabs open for about a month hoping to get around to finishing my investigation, but eventually had to reboot the laptop. I have one more stashed commit from February when I changed branch to do something else, where around line 26 of  `rmgpy/pdep/configuration.pyx`  I added 
```
        if self.energy_correction != 0.0: string += 'energy_correction={0}, '.format(self.energy_correction)
```
I can't remember if that fixed it.


Based on when David (I think) added the correction, I seem to recall it helped numerical issues (you end up taking exponentials of this energy when evaluating a partition function?) and was not just to make the PES figures look nice.
That gives me reason to pause before approving the current PR. We should check several scenarios to see if the results change and it's still numerically stable and doesn't give overflows.
eg. the commit message ot https://github.com/ReactionMechanismGenerator/RMG-Py/commit/169a34ec00c1e4275e38a690a49fc1d86bdd8b4c.
see also https://github.com/ReactionMechanismGenerator/RMG-Py/commit/7f26746d430b591cc2f319382c88b88e843b5ff5

mjohnson541:
I think that's definitely the most readable way to handle this, but it does require identifying all locations where this might happen and some of the algorithmic fixes may be much more than couple line fixes. I suspect we have object stored vectors that may become 0/Inf/NaN without the correction which may require more involved refactoring. 

It would be good to fix that, but given the benefits/work cost perhaps its most practical now to ensure the correction is applied to all the energies correctly and add some detailed comments explaining what is going on.

donerancl:
notes for my reference on places where `energy_correction` is applied...
in `rmgpy/rmg/pdep.py`
```
    def update_configurations(self, reaction_model):
        """
        Sort the reactants and products of each of the network's path reactions
        into isomers, reactant channels, and product channels. You must pass 
        the current `reaction_model` because some decisions on sorting are made
        based on which species are in the model core. 
        """
        ...
        if self.energy_correction:
            for spec in self.reactants + self.products + self.isomers:
                spec.energy_correction = self.energy_correction
```
in `rmgpy/pdep/configuration.pyx`
```
cpdef class Configuration(object):
    ...
    def __init__(self, *species):
         ...
        self.energy_correction = 0.0
    ...
    property E0:
        """The ground-state energy of the configuration in J/mol."""
        def __get__(self):
            return sum([float(spec.conformer.E0.value_si) for spec in self.species]) + self.energy_correction 
    ...
    cpdef double get_enthalpy(self, double T) except 100000000:
        """
        Return the enthalpy in kJ/mol at the specified temperature `T` in K.
        """
        cdef double h = 0.0
        for spec in self.species:
            h += spec.get_enthalpy(T)
        h += self.energy_correction
        return h
    ...
    cpdef double get_free_energy(self, double T) except 100000000:
        """
        Return the Gibbs free energy in kJ/mol at the specified temperature
        `T` in K.
        """
        cdef double g = 0.0
        for spec in self.species:
            g += spec.get_free_energy(T)
        g += self.energy_correction
        return g        
```
in `rmgpy/pdep/configuration.pxd`
```
cdef class Configuration(object):

    cdef public list species
    cdef public np.ndarray e_list
    cdef public np.ndarray dens_states
    cdef public np.ndarray sum_states
    cdef public bint active_j_rotor
    cdef public bint active_k_rotor
    cdef public float energy_correction
```

in `arkane/explorer.py`
```
    def execute(self, output_file, plot, file_format='pdf', print_summary=True, species_list=None,
                thermo_library=None, kinetics_library=None):
        """Execute an ExplorerJob"""
        ...
        for network in self.networks:
            for rxn in network.path_reactions:
                if rxn.transition_state is None:
                    E0 = sum([spec.conformer.E0.value_si for spec in rxn.reactants])
                    if rxn.network_kinetics is None:
                        E0 += rxn.kinetics.Ea.value_si
                    else:
                        E0 += rxn.network_kinetics.Ea.value_si
                    if network.energy_correction:
                        E0 += network.energy_correction
                    rxn.transition_state = rmgpy.species.TransitionState(conformer=Conformer(E0=(E0 * 0.001, "kJ/mol")))
       
```

jonwzheng:
What's the latest status of this PR? Had we decided on a path forward, and is someone actively working on these changes?

rwest:
I think the immediate problem is fixed in #2835, which is just waiting on #2825 and a review. And this PR can be closed without merging. 

But there may be other improvements possible.

PR #2751, opened 2025-01-22
update cantera version to remove need for x86 emulation on arm mac

JacksonBurns:
WIP

would resolve #2676

PR #2796, opened 2025-05-25
Use `rdkit` for SSSR and RCs (bug fix + Python upgrade)

JacksonBurns:
Currently we use `RingDecomposerLib` for finding the Smallest Set of Smallest Rings and getting the Relevant Cycles. This package does not support Python 3.10+ and is thus blocking further upgrades to RMG.

@knathanm in particular is looking to get RMG to Python 3.11 so as to add support for ChemProp v2.

I believe we can just use RDKit to do these operations instead. The original paper mentions that the functionality was being moved upstream to RDKit. With the help of AI I've taken just a first pass at reimplementing, with the special note that:
 - I opted to use the Symmetric SSSR in place of the 'true' SSSR. This is because the latter is non-unique (see [RDKit's "The SSSR Problem"](https://www.rdkit.org/docs/GettingStartedInPython.html#the-sssr-problem)). This should actually resolve  https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2562
 - I need to read more about the "Relevant Cycles"

This PR will be a draft for now, as it is predicated on Python 3.9 already being available (which it nearly is in https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2741)

### Motivation or Problem

### Description of Changes

### Testing

### Reviewer Tips


JacksonBurns:
Cantera 2.6 isn't available for Python 3.12, so this PR will also need to upgrade the Cantera version to 3 as mostly completed in #2751

JacksonBurns:
A note for the path forward on this PR - the `get_relevant_cycles` and `get_smallest_set_of_smallest_rings` functions will need to be moved to `Molecule`. Currently they are in `Graph`, and `Graph` has no way to send itself to RDKit in order to be subject to RDKit `GetSymmSSSR` _but_ `Molecule` can be converted to RDKit via `converted.to_rdkit_mol` (also exposed as `Molecule.to_rdkit_mol`, and then RDKit can be used. This may have implications for inheritance elsewhere in the codebase, but should be OK.

jonwzheng:
I ended up moving all the functions in `Graph` that call `get_relevant_cycles` or `get_smallest_set_of_smallest_rings` to `Molecule` as well. Still have a couple of unit tests to update, but if this is the direction we want to take, hopefully this should be adequate

jonwzheng:
We are now very close to passing all of the unit tests. There are just a couple major bugs to resolve, namely: 
- how to handle the R/L cutting fragment labels in the RDKit mol object? need to replace with placeholder atom? 
- some bond orders are weird now, so need to investigate where something is going wrong. Due to sanitization maybe?

jonwzheng:
One of the last failing tests is `test_draw_hydrogen_bond_adsorbate`, which actually fails due to an assertion error in the codebase:

https://github.com/ReactionMechanismGenerator/RMG-Py/blame/d0d95b3a6bb9cd5d873c8d5b552ec112d6ba9283/rmgpy/molecule/draw.py#L541

@rwest, since you're the last remaining dev who was involved in some capacity with this part of the code, do you know why we have that assert statement? I would think it's better to just plot a "dot" than to outright crash out. 
I'm also not sure why this only popped up in this PR - maybe better to address any underlying problem in this code - but thought I'd ask, since this `assert` might cause problems in the future. There's no safeguards against passing a molecule with no backbone, unless there's something about how this is called that I'm not getting.

While we're at it, the `_generate_coordinates` also didn't use `RDKit` to draw if the charge != 0, but maybe that can be relaxed with new versions?

rwest:
> One of the last failing tests is `test_draw_hydrogen_bond_adsorbate`, which actually fails due to an assertion error in the codebase:
> 
> https://github.com/ReactionMechanismGenerator/RMG-Py/blame/d0d95b3a6bb9cd5d873c8d5b552ec112d6ba9283/rmgpy/molecule/draw.py#L541
> 
> @rwest, since you're the last remaining dev who was involved in some capacity with this part of the code, do you know why we have that assert statement? 

I don't. I can see that if there is no straight chain backbone that a function to `_find_straight_chain_backbone` shouldn't work. It should either fail, and whatever called it deals with the exception (maybe a ValueError is better than AssertError, but either way it's an error to be handled). Or it could return `None`, but then the calling code would also have to be modified to handle that without crashing.  Or the calling code shouldn't try calling it - but then it'd have to know there is no straight chain backbone (and perhaps the easiest way to find that out is to ask for it and fail.)  So I think either way the calling code needs updating.

> There's no safeguards against passing a molecule with no backbone, unless there's something about how this is called that I'm not getting.

I haven't investigated the full stack trace. But some safeguard, or exception handling, sounds appropriate.


> 
> While we're at it, the `_generate_coordinates` also didn't use `RDKit` to draw if the charge != 0, but maybe that can be relaxed with new versions?

We can try it. I expect RDKit at the time wasn't being helpful. Hopefully the commit messages offer clues? (this is why we should write helpful commit messages that explain _why_ things are being done).

rwest:
I think using RDKit for drawing as much as possible is fine, if it does a good job. Probably we ran into issues in the past that may not persist with more recent versions of RDKit, so it's fine to revisit past decisions.

jonwzheng:
Although it apparently passes the tests, I noted that the regression test comparisons are more different than usual, specifically, for systems with resonance - lots of different thermo/kinetics. The question is, are these new changes the cause of error, or were the old values mistaken (due for example to nondeterministic SSSR)? 
Also some of the regression tests didn't run because of the new `to_rdkit_mol` changes.

JacksonBurns:
@jonwzheng you're write - I was totally off there

I think I've got it fixed now in 80ce6a6d37c47b4e9cc4d9b8f119c0fc9f34f64d

JacksonBurns:
@jonwzheng could you look at the new error

jonwzheng:
ok, this error is basically because the weird way we've written the fragment code.

The `fragment` version of `to_rdkit_mol` won't work unless `return_mapping = True`, and now it forces it (even if a different kwarg is passed in).

Fragments will sometimes call `get_smallest_set_of_smallest_rings` (e.g. for drawing), which will then call the _fragment_ version of `to_rdkit_mol`, which returns a _tuple_ rather than a _mol_.

i've now attached a band-aid to the problem by double-checking the return type, 
and also warning the user in Fragment if `return_mapping = False`. Ideally there's a more elegant fix.

rwest:
For testing, I made a new conda environment, following default instructions, and was pleased to see it chose Python 3.11.13.
The RMS stuff built OK, and the unit tests ran.  Some examples seem normal.

I ran the `examples/rmg/diesel` and it is giving a lot of disconcerting errors like:
```
Warning: Bond order 1.6000001430511475 does not have string representation.
Warning: Bond order 1.2000001668930054 does not have string representation.
Warning: Bond order 1.6000001430511475 does not have string representation.
Warning: Bond order 1.2000001668930054 does not have string representation.
Warning: Bond order 1.8000001907348633 does not have string representation.
Warning: Bond order 1.2000001668930054 does not have string representation.
Warning: Bond order 1.600000262260437 does not have string representation.
Warning: Bond order 1.4000002145767212 does not have string representation.
Warning: Bond order 1.4000000953674316 does not have string representation.
Warning: Bond order 1.8000001907348633 does not have string representation.
Warning: Bond order 1.1000001430511475 does not have string representation.
Warning: Bond order 1.1000001430511475 does not have string representation.
Warning: Bond order 1.2000001668930054 does not have string representation.
Warning: Bond order 1.8000001907348633 does not have string representation.
Warning: Bond order 1.7000001668930054 does not have string representation.
Warning: Bond order 1.3000000715255737 does not have string representation.
Warning: Bond order 1.6000001430511475 does not have string representation.
Warning: Bond order 1.2000001668930054 does not have string representation.
Warning: Bond order 1.6000001430511475 does not have string representation.
Warning: Bond order 1.2000001668930054 does not have string representation.
Warning: Bond order 1.8000001907348633 does not have string representation.
Warning: Bond order 1.2000001668930054 does not have string representation.
Warning: Bond order 1.600000262260437 does not have string representation.
Warning: Bond order 1.4000002145767212 does not have string representation.
Warning: Bond order 1.4000000953674316 does not have string representation.
Warning: Bond order 1.6000001430511475 does not have string representation.
Warning: Bond order 1.2000001668930054 does not have string representation.
Warning: Bond order 1.6000001430511475 does not have string representation.
Warning: Bond order 1.2000001668930054 does not have string representation.
Warning: Bond order 1.8000001907348633 does not have string representation.
Warning: Bond order 1.2000001668930054 does not have string representation.
Warning: Bond order 1.600000262260437 does not have string representation.
Warning: Bond order 1.4000002145767212 does not have string representation.
Warning: Bond order 1.4000000953674316 does not have string representation.
Warning: Bond order 1.639999508857727 does not have string representation.
Warning: Bond order 1.6199995279312134 does not have string representation.
Warning: Bond order 1.459999442100525 does not have string representation.
Warning: Bond order 1.1799997091293335 does not have string representation.
Warning: Bond order 1.239999771118164 does not have string representation.
Warning: Bond order 1.239999771118164 does not have string representation.
Warning: Bond order 1.519999623298645 does not have string representation.
Warning: Bond order 1.4599995613098145 does not have string representation.
Warning: Bond order 1.2599997520446777 does not have string representation.
Warning: Bond order 1.3799995183944702 does not have string representation.
Warning: Bond order 1.1666666269302368 does not have string representation.
Warning: Bond order 1.833333134651184 does not have string representation.
Warning: Bond order 1.833333134651184 does not have string representation.
Warning: Bond order 1.1666666269302368 does not have string representation.
[23:38:31] Can't kekulize mol.  Unkekulized atoms: 9 11 12
Warning: Kekulization failed; sanitizing without Kekulize
Warning: Bond order 1.4000002145767212 does not have string representation.
Warning: Bond order 1.399999976158142 does not have string representation.
Warning: Bond order 1.5999999046325684 does not have string representation.
Warning: Bond order 1.2999999523162842 does not have string representation.
Warning: Bond order 1.4000002145767212 does not have string representation.
Warning: Bond order 1.600000023841858 does not have string representation.
Warning: Bond order 1.2999999523162842 does not have string representation.
Warning: Bond order 1.6666669845581055 does not have string representation.
Warning: Bond order 1.2222223281860352 does not have string representation.
Warning: Bond order 1.3333336114883423 does not have string representation.
Warning: Bond order 1.2222224473953247 does not have string representation.
Warning: Bond order 1.5555559396743774 does not have string representation.
Warning: Bond order 1.3333336114883423 does not have string representation.
Warning: Bond order 1.666667103767395 does not have string representation.
Warning: Bond order 1.6666666269302368 does not have string representation.
Warning: Bond order 1.6666665077209473 does not have string representation.
Warning: Bond order 1.3333333730697632 does not have string representation.
Warning: Bond order 1.3333333730697632 does not have string representation.
[23:38:38] Can't kekulize mol.  Unkekulized atoms: 10
Warning: Kekulization failed; sanitizing without Kekulize
Warning: Bond order 1.6666665077209473 does not have string representation.
Warning: Bond order 1.6666666269302368 does not have string representation.
Warning: Bond order 1.3333334922790527 does not have string representation.
Warning: Bond order 1.3333333730697632 does not have string representation.
```
....
```
[00:07:30] Can't kekulize mol.  Unkekulized atoms: 31
Warning: Kekulization failed; sanitizing without Kekulize
Warning: Bond order 1.200000286102295 does not have string representation.
Warning: Bond order 1.599998950958252 does not have string representation.
Warning: Bond order 1.200000524520874 does not have string representation.
Warning: Bond order 1.200000524520874 does not have string representation.
Warning: Bond order 1.5999995470046997 does not have string representation.
Warning: Bond order 1.7999988794326782 does not have string representation.
Warning: Bond order 1.400000810623169 does not have string representation.
Warning: Bond order 1.5999996662139893 does not have string representation.
Warning: Bond order 1.4000006914138794 does not have string representation.
Warning: Bond order 1.2499998807907104 does not have string representation.
Warning: Bond order 1.25 does not have string representation.
Warning: Bond order 1.624999761581421 does not have string representation.
Warning: Bond order 1.375 does not have string representation.
Warning: Bond order 1.3749998807907104 does not have string representation.
Warning: Bond order 1.624999761581421 does not have string representation.
Warning: Bond order 1.1999993324279785 does not have string representation.
Warning: Bond order 1.599998950958252 does not have string representation.
Warning: Bond order 1.1999990940093994 does not have string representation.
Warning: Bond order 1.1999990940093994 does not have string representation.
Warning: Bond order 1.5999987125396729 does not have string representation.
Warning: Bond order 1.7999987602233887 does not have string representation.
Warning: Bond order 1.3999989032745361 does not have string representation.
Warning: Bond order 1.5999987125396729 does not have string representation.
Warning: Bond order 1.3999989032745361 does not have string representation.
Warning: Bond order 1.3999991416931152 does not have string representation.
Warning: Bond order 1.3999991416931152 does not have string representation.
```

I guess this shouldn't be happening?

JacksonBurns:
@jonwzheng IMO it seems like the cause of these warnings is well understood, so it would be fair to simply catch and hide them.

rwest:
OK, interesting.  Thanks for looking deeper.

I had been thinking along the lines of: for just "what's the smallest set of smallest rings" or "find me all the rings" the type of bond doesn't matter, and if that's all that we're asking of RDKit then any type of bond will work, and we shouldn't waste time kekulizing or sanitizing the molecule. But if we're trying to ask RDKit what's the symmetry of this ring, then maybe the bonds matter?
Screenshot 2025-08-25 at 2 49 48 PM
but perhaps that's only because in this example I'm using the bonds to imply the H atoms, and if you put them all on explicitly then it wouldn't? But there's perhaps a resonance form where it might? 🤷 

Anyway, is RDKit just being used to perceive where the rings are? in which case we can probably make a lightweight RDKitModlecule without complete information.

JacksonBurns:
Here's the short summary of test failures following the latest rebase:

```
=========================== short test summary info ============================
FAILED test/rmgpy/data/thermoTest.py::TestMolecularManipulationInvolvedInThermoEstimation::test_is_bicyclic1 - AssertionError: assert False
 +  where False = is_bicyclic([<Atom 'C'>, <Atom 'C'>, <Atom 'C'>, <Atom 'C'>, <Atom 'C'>, <Atom 'C'>, <Atom 'C'>, <Atom 'C'>])
FAILED test/rmgpy/data/thermoTest.py::TestMolecularManipulationInvolvedInThermoEstimation::test_deterministic_bicyclic_decomposition - AssertionError: assert False
 +  where False = <function is_bicyclic at 0x7fb63963b040>([<Atom 'C'>, <Atom 'C'>, <Atom 'C'>, <Atom 'C'>, <Atom 'C'>, <Atom 'C'>, <Atom 'C'>, <Atom 'C'>])
 +    where <function is_bicyclic at 0x7fb63963b040> = <module 'rmgpy.data.thermo' from '/home/runner/work/RMG-Py/RMG-Py/rmgpy/data/thermo.py'>.is_bicyclic
 +      where <module 'rmgpy.data.thermo' from '/home/runner/work/RMG-Py/RMG-Py/rmgpy/data/thermo.py'> = <module 'rmgpy.data' from '/home/runner/work/RMG-Py/RMG-Py/rmgpy/data/__init__.py'>.thermo
 +        where <module 'rmgpy.data' from '/home/runner/work/RMG-Py/RMG-Py/rmgpy/data/__init__.py'> = rmgpy.data
FAILED test/rmgpy/reactionTest.py::TestReactionToCantera::test_arrhenius - AssertionError: assert {'rate-constant': {'A': 660000.0, 'b': 1.62, 'Ea': 45354560.0}} == {'rate-constant': {'A': 660000.0000000001, 'b': 1.62, 'Ea': 45354560.0}}
  
  Differing items:
  {'rate-constant': {'A': 660000.0, 'Ea': 45354560.0, 'b': 1.62}} != {'rate-constant': {'A': 660000.0000000001, 'Ea': 45354560.0, 'b': 1.62}}
  
  Full diff:
    {
        'rate-constant': {
  -         'A': 660000.0000000001,
  ?                      ---------
  +         'A': 660000.0,
            'Ea': 45354560.0,
            'b': 1.62,
        },
    }
FAILED test/rmgpy/reactionTest.py::TestReactionToCantera::test_multi_arrhenius - assert 0.001 == 0
 +  where 0.001 = round(0.0009765625, 3)
 +    where 0.0009765625 = abs((5000000000000.0 - 5000000000000.001))
 +      where 5000000000000.0 = <ArrheniusRate at 7fb42083a9b0>.pre_exponential_factor
 +        where <ArrheniusRate at 7fb42083a9b0> = HO2(5) + OH(4) <=> H2O(27) + O2(6)    <Reaction(Arrhenius)>.rate
 +      and   5000000000000.001 = <ArrheniusRate at 7fb42083a3b0>.pre_exponential_factor
 +        where <ArrheniusRate at 7fb42083a3b0> = HO2(5) + OH(4) <=> H2O(27) + O2(6)    <Reaction(Arrhenius)>.rate
FAILED test/rmgpy/reactionTest.py::TestChargeTransferReaction::test_get_rate_coeff - assert 0.0078125 < 1e-06
 +  where 0.0078125 = abs((43870506959778.99 - 43870506959779.0))
== 5 failed, 1970 passed, 38 skipped, 3939699 warnings in 2107.75s (0:35:07) ===
```

jonwzheng:
`test_is_bicyclic1` fails on `is_bicyclic` for the 2nd and 3rd examples, `C1=CC2C=CC1=CC2` and `C1=CC2C=CC=1C=C2`:
image
image

And it was fine for this `C1=CCC2C1=C2`:
image


------------------------------------
`test_deterministic_bicyclic_decomposition` fails on `assert rmgpy.data.thermo.is_bicyclic(polyrings[0])`
for this compound:
image

rwest:
I didn't catch until just now that they changed the API and GetSSSR now actually returns a sequence of rings, not just the count.
https://github.com/rdkit/rdkit/pull/5437  I think made it into 2022.09 release - though I missed it when working on it earlier this year, because their online documentation is still inconsistent about this backwards-incompatible change.

jonwzheng:
> I think that proposed solution would be fine. I can't think of many cases where this would be an issue, anyway.
> 
> My understanding is that this design choice will just change which rings are picked when the decision is arbitrary. The tress were presumably trained using some equally arbitrary decision, so I don't know if they need to be refit (?)
> 
> @jonwzheng do you know who might know this better?

I don't think any current members will know. @mjohnson541 did some of the automatic thermo and kinetic re-fitting, and before that @mliu49 worked on [polycyclic corrections](https://github.com/ReactionMechanismGenerator/RMG-database/blob/main/input/thermo/groups/polycyclic.py). There are some examples here like the correction for barrelene-related groups that seem to have some hints.

PR #2806, opened 2025-06-04
A note about installing developer version from Docker.

rwest:
The last Docker release was 2 years ago. Users might want to use Docker to get a more recent version. This PR updates the installation instructions to tell them how.

The last Conda binary was 4+ years ago, so I hid/removed the instructions recommending that installation method.
As noted in a commit message, we should re-visit these instructions once we've fixed conda builds.

I also found a link to the pdf of the 3.2.0 user guide, so included that link (for people following the instructions to install it), partially addressing #2809.

### DIFF
--- a/documentation/source/users/rmg/faq.rst
+++ b/documentation/source/users/rmg/faq.rst
@@ -28,8 +28,7 @@ Running RMG
 
 #. **How do I run a basic RMG job?**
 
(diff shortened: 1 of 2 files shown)


rwest:
For reviewers' benefit I built the current head of this branch locally and it renders like this: 
[installation.pdf](https://github.com/user-attachments/files/20701986/installation.pdf)

Mari51239:
Thank you for the instructions. I followed them and successfully installed the latest version of RMG.

PR #2808, opened 2025-06-06
Docker build updates

### LINKED ISSUE #2805: Docker build CI not working. Failing to install Julia (or not trying)
### Bug Description

Looking at https://github.com/ReactionMechanismGenerator/RMG-Py/actions/workflows/CI.yml?query=branch%3Amain 
it seems the [build-and-push-docker](https://github.com/ReactionMechanismGenerator/RMG-Py/actions/runs/15437561392/job/43467476904#step:5:9744) step keeps failing?

buildx failed with: ERROR: failed to solve: process "conda run --no-capture-output -n rmg_env /bin/bash -c source install_rms.sh" did not complete successfully: exit code: 1

It looks like the install_rms.sh script is doing its job of telling the user to install julia. 
Presumably it should be installed already.

rwest:
### Motivation or Problem
Docker builds were failing due to several issues with Julia/RMS integration and environment configuration. The build process also lacked flexibility for testing different branches and had poor logging visibility in CI environments.

### Description of Changes
- **Fixed Docker build failures**: Biggest fixes are Install Julia, and compile RMG!
- **Improved environment variable handling**: Added explicit exports for Docker compatibility and set `PYTHON_JULIAPKG_PROJECT` to ensure Julia packages install in the correct location
- **Made RMS branch configurable**: Added `RMS_BRANCH` environment variable with `for_rmg` as default, allowing different RMS branches to be tested
- **Enhanced CI flexibility**: Runs Docker build (but not push) on Pull Requests.  Updated CI to pass current branch names to Docker build args instead of hardcoded values
- **Improved debugging**: Added plain Docker build output and debugging echo statements to aid troubleshooting
- **Updated test example**: Changed to use RMS-enabled example to properly test Julia precompilation
- **Reduced docker image size**: I assume this helped - I cleaned out some caches etc.

### Testing
Continuous integration on GitHub has been used to verify these changes.

Closes #2805

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,6 +28,7 @@ RUN wget -qO- https://install.julialang.org | sh -s -- --yes --default-channel 1
     /root/.juliaup/bin/juliaup add 1.10 && \
     /root/.juliaup/bin/juliaup default 1.10 && \
     /root/.juliaup/bin/juliaup list && \
+    rm -rf /root/.juliaup/downloads /root/.juliaup/tmp
 ENV PATH="/root/.juliaup/bin:$PATH"
 
 # Install conda
--- a/Dockerfile
+++ b/Dockerfile
@@ -89,11 +89,10 @@ ENV PYTHON_JULIAPKG_PROJECT="/miniconda/envs/rmg_env/julia_env"
 RUN source install_rms.sh
 
 # RMG-Py should now be installed and ready - trigger precompilation and test run
-RUN python rmg.py examples/rmg/minimal/input.py
-# delete the results, preserve input.py
-RUN mv examples/rmg/minimal/input.py . && \
-    rm -rf examples/rmg/minimal/* && \
-    mv input.py examples/rmg/minimal/
+RUN python rmg.py examples/rmg/rms_constant_V/input.py
+# delete the results, restore input.py from git
+RUN rm -rf examples/rmg/rms_constant_V/* && \
+    git checkout -- examples/rmg/rms_constant_V/
 
 # when running this image, open an interactive bash terminal inside the conda environment
 RUN echo "conda activate rmg_env" >> ~/.bashrc
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,14 @@ RUN apt-get update && \
     apt-get autoremove -y && \
     apt-get clean -y
 
+
+# Install Julia 1.10 using juliaup
+RUN wget -qO- https://install.julialang.org | sh -s -- --yes --default-channel 1.10 && \
+    /root/.juliaup/bin/juliaup add 1.10 && \
+    /root/.juliaup/bin/juliaup default 1.10 && \
+    /root/.juliaup/bin/juliaup list && \
+ENV PATH="/root/.juliaup/bin:$PATH"
+
 # Install conda
 RUN wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh && \
     bash Miniconda3-latest-Linux-x86_64.sh -b -p /miniconda && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,6 +43,8 @@ ENV PATH="/miniconda/bin:$PATH"
 SHELL ["/bin/bash", "-c"]
 
 # Add build arguments for RMG-Py, RMG-database, and RMS branches.
+# The defaults are set here, but they can be overridden at build time
+# using the --build-arg option, or in the continous integration CI.yml file.
 ARG RMG_Py_Branch=main
 ARG RMG_Database_Branch=main
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -75,8 +75,10 @@ RUN conda clean --all --yes
 ENV RUNNER_CWD=/rmg
 ENV PATH="$RUNNER_CWD/RMG-Py:$PATH"
 
-# 1. Build RMG
-# 2. Install and link Julia dependencies for RMS
+# Build RMG
+RUN make
+
+# Install and link Julia dependencies for RMS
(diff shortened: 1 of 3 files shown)


rwest:
Encountered the exact issue described in [SciMLOperators.jl#263](https://github.com/SciML/SciMLOperators.jl/issues/263) during package precompilation:
`WARNING: Method definition islinear(Any) in module SciMLOperators [...] overwritten in module SciMLBase
ERROR: Method overwriting is not permitted during Module precompilation`
Versions involved:

SciMLOperators v0.3.13
SciMLBase v1.98.1

This is the problematic combination mentioned in the issue - SciMLOperators 0.3.13 added fallbacks that conflict with SciMLBase <= 2.77.2.
Impact: 30 packages failed precompilation (including ReactionMechanismSimulator, DiffEqBase extensions, ModelingToolkit, etc.) but installation continued with fallbacks. Packages may have slower first load times and potential runtime issues.
Workaround needed: Either downgrade SciMLOperators to <0.3.13 or upgrade SciMLBase to >2.77.2, pending the registry restrictions mentioned in the issue comments.
The ⌅ symbols in the package manifest indicate version constraints are preventing automatic resolution of this conflict.
Maybe this is an issue for [ReactionMechanismSimulator.jl ](https://github.com/ReactionMechanismGenerator/reactionmechanismsimulator.jl)

Is that enough to make the Docker build to fail?

JacksonBurns:
> In both cases the core now has 35 species instead of 37. But the run time is crazy. The "reference" was 6 minutes. 3 days ago it was an hour. Today it is 22 minutes. Why so random??

Now _this_ is high performance computing

Dockerfile
```diff
@@ -61,19 +77,24 @@ RUN conda clean --all --yes
...
-    rm -rf examples/rmg/minimal/* && \
-    mv input.py examples/rmg/minimal/
+RUN python rmg.py examples/rmg/rms_constant_V/input.py
+# delete the results, restore input.py from git
+RUN rm -rf examples/rmg/rms_constant_V/* && \
+    git checkout -- examples/rmg/rms_constant_V/
 
 # when running this image, open an interactive bash terminal inside the conda environment
-RUN echo "source activate rmg_env" >~/.bashrc
+RUN echo "conda activate rmg_env" >> ~/.bashrc
```
JacksonBurns, line 99:
```suggestion
RUN echo "source activate rmg_env" >> ~/.bashrc
```

We never run `conda init` so we can't run `conda activate`, we have to use `source activate` to get into the environment. I confirmed this locally - changing this line still drops you in a conda environment on running the image, the version in this PR does not.

JacksonBurns (changes requested):
I tested with `docker buildx build --no-cache --build-arg RMGPy_Branch=docker .` and the build works. When running the image in a container the conda environment does not activate correctly - see my comment that fixes this.

There are a couple opinionated changes we could make as well - please feel free to ignore those.

Dockerfile
rwest: Thanks. Would running `conda init` be a good idea, so that people inside the docker have a conda setup that looks more "normal"?

Dockerfile
```diff
@@ -61,19 +77,24 @@ RUN conda clean --all --yes
...
+# Build RMG
+RUN make
+
+# Install and link Julia dependencies for RMS
 # setting this env variable fixes an issue with Julia precompilation on Windows
 ENV JULIA_CPU_TARGET="x86-64,haswell,skylake,broadwell,znver1,znver2,znver3,cascadelake,icelake-client,cooperlake,generic"
+ENV RMS_BRANCH=${RMS_Branch}
+# Usually this is set automatically, but we're not actually running
+# in an active conda environment when building the Docker so we need to set it manually
+ENV PYTHON_JULIAPKG_PROJECT="/miniconda/envs/rmg_env/julia_env"
```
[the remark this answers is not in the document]
rwest: I think it is needed, because each is run in a separate shell. The `install_rms.sh` script will set the variable inside that shell for installing RMS, and the combination of the `conda env config vars set` lines with the `RUN echo "source activate rmg_env" >> ~/.bashrc` and the `ENTRYPOINT ["/bin/bash", "--login"]` will ensure it's set for end users of the docker image,  but nothing sets it for the `RUN python rmg.py examples/rmg/rms_constant_V/input.py` line.

Dockerfile
JacksonBurns: Good idea, added in [c1792e4](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2808/commits/c1792e4aef7441e24e8c8aeeec2a1f9d89d28b8f) so that now `conda activate rmg_env` works as expected

Dockerfile
```diff
@@ -45,8 +57,12 @@ RUN git clone --single-branch --branch ${RMG_Py_Branch} --depth 1 https://github
     git clone --single-branch --branch ${RMG_Database_Branch} --depth 1 https://github.com/ReactionMechanismGenerator/RMG-database.git
 
 WORKDIR /rmg/RMG-Py
+# Copy environment.yml to ensure Docker cache invalidates if it changes
+COPY environment.yml .
```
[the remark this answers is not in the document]
rwest: I think you're right and am removing this change. (will push soon)

rwest:
Thanks for your review, and for checking that the docker image actually works (something I wasn't able to do just testing via the CI). I adopted or responded to your other changes/questions. I suggest we rebase (with `-i --autosquash`) once #2804 is merged.

JacksonBurns (approved):
@rwest agreed, please rebase when the other PR is merged and set this to automerge

rwest:
I just anticipated (and tried to fix) another possible problem.

JacksonBurns (dismissed):
Latest looks correct to me

PR #2810, opened 2025-06-17
Refactored reaction.py to add get_free_energy_of_non_charge_transfer_reaction for code reuse and circular dependency prevention

LekiaAnonim:
1. Typo correction in _apply_CHE_model

In the _apply_CHE_model, in the ReactionError line, `self.electrons` is used twice instead of `self.protons` and `self.electrons`.

image

2. Add a new method `get_free_energy_of_non_charge_transfer_reaction` 

The `get_reversible_potential` method calculates the potential at equilibrium using the equation.
 
dG = dGrxn - nF(V-V0) ----- (1)

at equilibrium dG = 0, and taking V = 0,
V0 can be set to V0 = dGrxn/nF ---- (2)

This means it is more preferable to calculate the `get_reversible_potential`  with reference to another method `get_free_energy_of_non_charge_transfer_reaction` instead of `_get_free_energy_of_charge_transfer_reaction` which may also be calculated based on the reference potential, if the reference potential is not zero, Eqn (1). This can introduce a circular issue because the `set_reference_potential` is calculated from `get_reversible_potential`.

Where `get_free_energy_of_non_charge_transfer_reaction` is 

```
def get_free_energy_of_non_charge_transfer_reaction(self, T):
 
      dGrxn = 0.0
      for reactant in self.reactants:
          try:
              dGrxn -= reactant.get_free_energy(T)
          except Exception:
              logging.error("Problem with reactant {!r} in reaction {!s}".format(reactant,        self))
              raise
      for product in self.products:
          try:
              dGrxn += product.get_free_energy(T)
          except Exception:
              logging.error("Problem with product {!r} in reaction {!s}".format(reactant, self))
              raise
      return dGrxn
```
 
We can the call the `get_free_energy_of_non_charge_transfer_reaction` in `_get_free_energy_of_charge_transfer_reaction`, and `get_free_energy_of_reaction` to avoid a repetition of the code block in `get_free_energy_of_non_charge_transfer_reaction`

 
image


ssun30:
Hi @LekiaAnonim , thanks for opening this PR. Make sure the title of the PR is concise (by default GitHub uses the full commit message of the latest commit as the title that's why it's so long). You can also make the commits more readable by having a short description in the first line and add detailed explanations after the second line when you edit the commit messages.

LekiaAnonim:
> Hi @LekiaAnonim , thanks for opening this PR. Make sure the title of the PR is concise (by default GitHub uses the full commit message of the latest commit as the title that's why it's so long). You can also make the commits more readable by having a short description in the first line and add detailed explanations after the second line when you edit the commit messages.

Thanks @ssun30, I have an update to the PR title and commit message.

rwest (changes requested):
The typo fixes look good. The "refactoring" I'm not sure about.  Doesn't seem to reduce much code duplication by making a separate method that is only used once. I found it confusing.

I am quite possibly wrong  - I just found the diff confusing.

rwest:
Hi Prosper,
if some or all of this is still worth doing, please could you rebase and update the PR?
Or close this one and open separate PRs if that is more appropriate.
I notice this is from your `main` branch, which is a bit unusual. And it has many merge commits in it.
And look through the review comments on github (whether from me or copilot).
thanks

ssun30:
Hi @LekiaAnonim , I couldn't quite recall which issue this PR is meant to fix, or whether a similar fix has already been implemented. Please rebase this branch on the latest main so we can compare the diffs.

PR #2822, opened 2025-07-07
Ignore more example output files in git.

rwest:
### Motivation or Problem
My git diffs and status were polluted with lots of output files from running examples.

### Description of Changes
Added more templates to the .gitignore file, so git ignores changes to these files.

### Testing
I checked locally that my git status was cleaner.

### Reviewer Tips
This is a small PR and should be quick to review.


rwest:
@JacksonBurns any idea why the Docker build failed?

PR #2828, opened 2025-07-11
Modified install_rms.sh to Use The Correct Backend

ssun30:
Added developer install option for install_rms.sh

### Motivation or Problem
Some of our group members have noticed that they have a Python compatibility problem when installing RMS. It seems that their default system Python version is 3.12 or newer. I've identified the problem to be the `JULIA_CONDAPKG_BACKEND` environment variable, which is set to "Null", meaning it's using whichever Python is currently installed and in the user's PATH, and the user must have already installed any Python packages that they need. 

The intended behavior is for `CondaPkg` to use the currently activated Conda environment (say, rmg_env) and use the Python binaries within that environment, so it ensures compatibility. For that, the correct environment variable value should be ["Current"](https://juliapy.github.io/PythonCall.jl/stable/pythoncall/#pythoncall-config). Furthermore, the `JULIA_CONDAPKG_EXE` variable needs to be specified. 

I'm not sure why our CI didn't catch this problem, but at least a fix is found and some testing has been done on my local computer. I'll need some more testing results from other users later. 

### Description of Changes
1) Changed `CondaPkg` configurations to use the correct backend. 
2) Added an option in `install_rms.sh` to use a developer build of RMS.


JacksonBurns:
Also, the reason we didn't set Current is that Julia would break the environment when allowed to install its own packages into rmg_env. The exact reason is buried somewhere in the comments of the Python 3.9 upgrade, but @rwest might recall. My first thought is that it broke sundials, but I can run through that PR tomorrow and check.

rwest:
Reading 
https://juliapy.github.io/PythonCall.jl/stable/pythoncall/#pythoncall-config
why is "Null" necessarily wrong?
> By setting the CondaPkg backend to Null, it will never install any Conda packages. In this case, PythonCall will use whichever Python is currently installed and in your PATH. You must have already installed any Python packages that you need.

We could just set the path right and install the required packages?

JacksonBurns:
> We could just set the path right and install the required packages?

This is what we currently do, though we let `conda` handle pointing us towards the correct `python` executable. All of the packages needed by RMS are installed by either our `environment.yml` or our `install_rms.sh`.

rwest:
Some of these additions I'm definitely a fan of.  
Eg. I like confirming the active conda environment (I have run into that mistake several times myself, and suspect other issues we've been having are also caused by it; especially since the process of installing JuliaUp has you restart the shell, then you forget to re-activate the right conda environment). 

I like checking things more thoroughly too.

As a middle ground the "dev" install could perhaps be triggered just by setting environment variables before running the install_rms script rather than it asking everybody what type of install they want?
Could some code duplication be reduced there?

And yes we need better documentation. 👍

ssun30:
> I'm not against adding _something_ to RMG to make co-developing RMG and RMS easier, but I'm not sure it needs to be an automated script like this. I believe a better fit would be some documentation for developers pointing out how to follow the same 'spirit' as the `install_rms.sh` script but configure it to use whatever Julia/conda environment/etc. that one wishes.
> 
> IMO an automated script to handle a complex developer install doesn't make a lot of sense. Someone looking to do something this complicated should be able to run the commands on their own, especially with some help from a nice docs page.


> Some of these additions I'm definitely a fan of. Eg. I like confirming the active conda environment (I have run into that mistake several times myself, and suspect other issues we've been having are also caused by it; especially since the process of installing JuliaUp has you restart the shell, then you forget to re-activate the right conda environment).
> 
> I like checking things more thoroughly too.
> 
> As a middle ground the "dev" install could perhaps be triggered just by setting environment variables before running the install_rms script rather than it asking everybody what type of install they want? Could some code duplication be reduced there?
> 
> And yes we need better documentation. 👍

I have made "standard" mode default so it won't be asking the user which mode to use. Now the "developer" mode will only be used if the user use the command `./install_rms.sh developer`. I think this is the middle ground Richard is leaning towards.

rwest:
I've just rebased this onto main, which required resolving a couple of merge conflicts.

I'm curious about why we want the last commit "[Replace return 1 with exit 1](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2828/commits/1b1624af391beadb4ef6f24231bbcb27252ed77b)"  given the commit message in https://github.com/ReactionMechanismGenerator/RMG-Py/commit/c46a861a88961acec5ef211b0a38cfc39df54313
The commit message in the former offers no explanation.

rwest (dismissed):
I now approve, but I made the last dozen or so commits, so someone else should review it. This new script works for me on MacOS with zsh (surprisingly many differences) and works on the Github Actions CI. Would be good if reviewers could try it in other settings too.

documentation/source/users/rmg/installation/anacondaDeveloper.rst
```diff
@@ -129,6 +129,25 @@ Installation by Source Using Anaconda Environment for Unix-based Systems: Linux
     which can involve several steps including restarting your terminal or shell.
     Run the script again, until it finishes installing RMS and all of its dependencies, and reports that ReactionMechanismSimulator is installed.
 
+    **Advanced Usage**:
+    The ``install_rms.sh`` script has options, that can be set as environment variables or variables to the script.
+    By default, it installs the latest release of the ``for_rmg`` branch of RMS, from https://github.com/ReactionMechanismGenerator/ReactionMechanismSimulator.jl/branches, which is recommended for most users.
```
JacksonBurns:
```suggestion
    By default, it installs the latest release of the ``for_rmg`` branch of RMS, from `GitHub <https://github.com/ReactionMechanismGenerator/ReactionMechanismSimulator.jl/branches>`_, which is recommended for most users.
```

JacksonBurns (dismissed):
Just make this a hyperlink, then it's good. Thanks for putting this together @ssun30 and @rwest this make the installation much easier :+1:

PR #2829, opened 2025-07-14
Update Jupyter Notebooks to call RMS using juliacall

mjohnson541:
I've updated the jupyter notebooks to call RMS using juliacall. 

The good news seems to be that at least in these examples unexpectedly juliacall doesn't seem to have the issues that necessitated writing any separate python functions to use RMS functionality. Flux diagrams don't seem to require a new function as was necessary for pyjulia. However, there are more nuances about how we feed in dictionaries and lists into RMS functions, most of the time it seems like just calling to_julia(obj) gets you the right object to feed in, although in the threadedsensitivities call that was insufficient and I needed to convert the string keys into symbols.

### DIFF
--- a/test/rmgpy/rmg/mainTest.py
+++ b/test/rmgpy/rmg/mainTest.py
@@ -119,6 +119,11 @@ class TestMain:
         """Test that edge seed mechanisms are created in the correct database locations."""
         assert os.path.exists(self.seedKinetics)
 
+    def test_rmg_rms_mechanism_files_creation(self):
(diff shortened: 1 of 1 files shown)


jonwzheng:
Thanks for working on this Matt!

I'm not sure why, but running the `RMG` simulation didn't generate an `rms/chem[x].rms` file. Did you need to modify your input files?

jonwzheng (approved):
LGTM and verified that the notebooks run.

PR #2831, opened 2025-07-14
Test rmg_new RMS branch to replace for_rmg

mjohnson541:


mjohnson541:
@rwest Yes, that was my plan, I needed to test it would work on RMG before doing that. for_rmg is entirely diverged from RMS now. It isn't on main/stable so I'm not sure how we would maintain history other than just saving a "old_for_rmg" branch of sort. I think eventually we would like RMG to be on RMS versions rather than branches, however, current latest is missing some of the changes on for_rmg. 

For now I propose: 
1) We save the current for_rmg as old_for_rmg
2) We move for_rmg to RMS's current main
3) At the next RMS release we move RMG to that release

mjohnson541:
Implemented, closing.

PR #2834, opened 2025-07-18
Dockerfile: use miniforge instead of miniconda to avoid `defaults`

JacksonBurns:
### Motivation or Problem
Current docker image uses `miniconda` which attempts to use the `defaults` channel. We are already avoiding this in the CLI with miniforge.

### Description of Changes
Docker image now builds from `conda-forge/miniforge` which is the same version of ubuntu that we were using but with `miniforge` already installed.

### Testing
I've run this Docker manually and it builds

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -89,6 +89,6 @@ RUN rm -rf examples/rmg/rms_constant_V/* && \
     git checkout -- examples/rmg/rms_constant_V/
 
 # when running this image, open an interactive bash terminal inside the rmg_env conda environment
-RUN sed -i '/\. \/opt\/conda\/etc\/profile.d\/conda.sh && conda activate base/c\. /opt/conda/etc/profile.d/conda.sh && conda activate rmg_env' ~/.bashrc
+RUN sed -i 's/conda activate base/conda activate rmg_env/' ~/.bashrc
 ENTRYPOINT ["/bin/bash", "--login"]
 


rwest (commented):
Seems reasonable. 
Looking at their recipe for that image ([here](https://github.com/conda-forge/miniforge-images/blob/master/ubuntu/Dockerfile))
one difference is they add this to the .bashrc:
`echo ". ${CONDA_DIR}/etc/profile.d/conda.sh && conda activate base" >> ~/.bashrc`
and we later add
`echo "conda activate rmg_env" >> ~/.bashrc`
It's presumably OK to activate rmg_env after already activating base? but would it be cleaner to remove the `conda activate base` line from our `.bashrc`?


Secondly - do we still need the `RUN conda init` line?

Dockerfile
```diff
@@ -98,3 +92,4 @@ RUN rm -rf examples/rmg/rms_constant_V/* && \
 RUN conda init
 RUN echo "conda activate rmg_env" >> ~/.bashrc
```
rwest, line 93:
```suggestion
RUN sed -i 's/^conda activate base$/conda activate rmg_env/' ~/.bashrc
```

JacksonBurns:
@rwest  I was able to drop the `init` and I added a (slightly modified version of) your `sed` command to handle the activation more elegantly. Confirmed that it works on my machine

Dockerfile
```diff
@@ -94,7 +88,7 @@ RUN python rmg.py examples/rmg/rms_constant_V/input.py
 RUN rm -rf examples/rmg/rms_constant_V/* && \
     git checkout -- examples/rmg/rms_constant_V/
 
-# when running this image, open an interactive bash terminal inside the conda environment
-RUN conda init
-RUN echo "conda activate rmg_env" >> ~/.bashrc
+# when running this image, open an interactive bash terminal inside the rmg_env conda environment
+RUN sed -i '/\. \/opt\/conda\/etc\/profile.d\/conda.sh && conda activate base/c\. /opt/conda/etc/profile.d/conda.sh && conda activate rmg_env' ~/.bashrc
```
rwest, line 92: Is it only the `conda activate base` part of the line that's changing?
```suggestion
RUN sed -i 's/conda activate base/conda activate rmg_env/' ~/.bashrc
```
(my previous suggestion would have failed because I included a `^` which only matches at the start of a line)

JacksonBurns:
Ah I just saw this: https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2836#issuecomment-3090634060

I restarted the regression test, perhaps it will pass

rwest (approved):
Looks good to me. 
Some other PRs are failing the build-docker CI step; hopefully this helps them.

PR #2836, opened 2025-07-18
Try echem branch of RMS

rwest:
An incompatibility with the current `for_rmg` branch of RMS is causing RMG regression tests to fail.
See https://github.com/ReactionMechanismGenerator/ReactionMechanismSimulator.jl/issues/275

This is a test to see if the echem branch works.
(though it currently fails to pass the RMS CI tests)


rwest:
This worked, so I have reset [`for_rmg`](https://github.com/ReactionMechanismGenerator/ReactionMechanismSimulator.jl/tree/for_rmg) branch of RMS to be at the mostly working `echem` branch. 
This should get RMG regression tests to run again.

Issue #2744, opened 2024-12-09
Updates to draw.py are causing a crash for charge separated bidentates


kirkbadger18:
When I try to run RMG in cases where charge separated bidentates show up, the recent updates to draw.py cause the following error:

![image](https://github.com/user-attachments/assets/aa8216f4-f427-4cb1-95fd-10e58215abb9)

I have isolated the issue by running the following input file in rmg-py main and rmg-database main with just the addition of one reaction in a reaction library. This one reaction contains a charge separated bidentate: XOXNO. 

The dictionary looks like:
![image](https://github.com/user-attachments/assets/549f2791-4269-4551-a807-426f52d22b2e)
And the reaction looks like:
![image](https://github.com/user-attachments/assets/eb7992d5-0288-42e8-a9ca-7690f614a80d)

Here are the files needed to regenerate this error:
[generate_error.zip](https://github.com/user-attachments/files/18068733/generate_error.zip)

I think it has to do with this check in draw.py:
![image](https://github.com/user-attachments/assets/24e34842-3950-44dd-bf0d-a0ba75965460)

I tried deleting this check and it fixed the error for me, but I'm sure these lines must be here for a reason, and maybe deleting them would cause issues for someone else.

kirkbadger18:
@rwest, this is the draw.py issue we discussed.

rwest:
On a new branch, please can you add a (failing) unit test in [test/rmgpy/molecule/drawTest.py](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/main/test/rmgpy/molecule/drawTest.py)
(eg. like    `test_draw_bidentate_adsorbate` )
and open a pull request from that branch?
That will help us figure out what is crashing `_find_straight_chain_backbone`.

I wasn't clear from your description what your modification is doing, but from the stack trace it looks like the problem was with using our own coordinate generation, so your change is probably making it use RDKit instead.
It looks like the test was previously saying if any of the atoms have a charge then don't use RDKit.
I think introduced in https://github.com/ReactionMechanismGenerator/RMG-Py/commit/ad8001893de0e2383d957baa960061dafabf5c5c with the note "Draw molecule using own algorithm if one of the molecule's atoms is charged. This has more likelihood of not failing on rdkit due to incorrect valence."
Perhaps things have changed in RDKit in the intervening 10 years (!!) and it doesn't complain or fail any more.
[...]

PR #2838, opened 2025-07-24
Fix drawing errors for charged species by moving to `rdkit` geometry generation

kirkbadger18:
RMG crashes at the moment when it tries to draw bidentates with charge separation. This issue is discussed in issue #2744. Here I am adding a test that should be passed for any fix to draw.py.

### DIFF
--- a/rmgpy/molecule/draw.py
+++ b/rmgpy/molecule/draw.py
@@ -155,7 +155,7 @@ class MoleculeDrawer(object):
         self.surface = None
         self.cr = None
 
-    def draw(self, molecule, file_format, target=None):
+    def draw(self, molecule, file_format, target=None, use_rdkit=True):
         """
         Draw the given `molecule` using the given image `file_format` - pdf, svg, ps, or
         png. If `path` is given, the drawing is saved to that location on disk. The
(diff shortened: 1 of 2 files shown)


jonwzheng:
I went ahead and made `RDKit` the default drawing algorithm.

`RDKit` appears to be superior in many cases for ions, but has some blind spots with surfaces. It correctly renders the test case in #2744:

Image

I wrote some other test cases which are depicted below.
---

Cation With RDKit:
image

old RMG algorithm:
image

---

Anion With RDKit:
image

old RMG algorithm:
image

---

Zwitterion With RDKit (both algorithms do a good job of this):
image

---

Cation bidentate With RDKit:
image

old RMG algorithm: **fails**

---

Anion binded to surface
image

jonwzheng:
I realized that `test_draw_hydrogen_bond_adsorbate` will fail on this branch, I think because of an issue in hydrogen bond representation.
But it will pass on #2796, which makes some changes to the RDKit code to make things work. 
Maybe we can move these changes to there?

rwest (approved):
Approving review based on looking at the diffs, and discussion. I've not actually tried it out.

I like Jon's suggestion:
>  Perhaps there can be improvements to coerce the orientation of the molecules if surface atoms are present (e.g. if surface atom z-index is high, then flip over)

But we could just open a (low priority) issue to fix that later if you want to press ahead with merging this.

JacksonBurns:
@jonwzheng can you open a new issue to track the above?

PR #2839, opened 2025-07-25
Avoid Accidental `conda` Uploads

JacksonBurns:
The current `conda_build` action is overwriting the currently uploaded version of RMG 3.3.0 every time it is run (including on pushes to `main`, PRs, etc.) because - for some absolutely STUPID reason - if you pass the `--user` and `--token` arguments to `conda build` it will upload the built binary REGARDLESS of if automatic uploads are enabled.

This slightly shuffles around the logic to avoid this.


jonwzheng (approved):
Nice catch, lgtm

PR #2844, opened 2025-08-20
Add yaml input file support

vincthnngyn:
### Motivation or Problem
- YAML files are more readable and less error-prone than Python files
- Better separation of configuration from code
- Easier validation and tooling support

### Description of Changes
- Added `YAMLInputReader` class in `yaml_input_reader.py`
- Added auto-detection of file format based on extension
- Modified `input.py` to support both formats
- Added example YAML input files

### Testing
- Ran the butane example input until no errors popped up
- will run an input that needs to process solvents so I can trouble shoot in same manner

### Reviewer Tips
- send me ANY and ALL issues please!!


vincthnngyn:
Understood understood!

jonwzheng:
Not stale, would be nice to get this in

PR #2852, opened 2025-10-15
Expanding the unit testing of ring decomposition features.

rwest:
### Motivation or Problem
There's some debate in reviewing #2796 as to whether certain changes showing in the regression tests indicate that it is  breaking things or improving things. 
So I have made a couple of unit tests, that should pass.

Spoiler alert: they pass on `main` and not yet on #2796. 

### Description of Changes
One is just adding a few bicyclic molecules to the `test_is_bicyclic1` test.

Another is adding a new test_deterministic_bicyclic_decomposition, that addresses issue #2562.
This one currently "skips" on main about 1/3 of the time, because we have a nondeterministic behavior, and can get one of two different ring decompositions.  But the other parts of the test all pass.

See the commit messages and diffs for more details.

### DIFF
--- a/test/rmgpy/data/thermoTest.py
+++ b/test/rmgpy/data/thermoTest.py
@@ -2272,6 +2272,32 @@ class TestMolecularManipulationInvolvedInThermoEstimation:
         expected_aromatic_bond_num_in_bicyclics = [0, 0, 0]
         assert aromatic_bond_num_in_bicyclics == expected_aromatic_bond_num_in_bicyclics
 
+    def test_deterministic_bicyclic_decomposition(self):
+        """
+        Test that the decomposition of a polyring into bicyclics and then into single rings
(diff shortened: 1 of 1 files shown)


JacksonBurns:
Which of these fail on the new SSSR approach? Are we failing to perceive bicycles, and/or failing to reproducibly decompose them?

Based on the contents of the regression test you commented [here](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2796#issuecomment-3394195616) it seems like the new SSSR correctly perceives bicycles, just decomposes them differently than `main` (in that one instance the regression test is comparing against).

rwest:
It's annoying regression tests come back "failing" with different entropies (I guess non deterministic symmetry numbers? Maybe we can finally fix this too!). But at least the enthalpies (and thermo comments) are unchanged.

JacksonBurns (approved):

PR #2853, opened 2025-10-19
Added completedNetworks option to prevent exploring PDep networks that are complete.

rwest:
### Motivation or Problem
Sometimes you have a pressure-dependent network that has been studied in great detail and put in a seed mechanism or reaction library, (such as CH2O2) and you don't want RMG to add reactions to the network and your model, because doing so will make it worse. 
Any net reactions that are in your seed mechanism won't be changed by RMG's estimates, but currently if RMG finds a new reaction then it will add it.   If the original authors did a thorough job and still left out a reaction, it's because they think it is slow, and adding RMG's estimate will probably make your model worse.
@Nora-Khalil has been running into this problem (it's not just hypothetical), although I'm not sure if we have an issue open for it yet.

### Description of Changes

This PR adds an option that lets you specify in the input file that you don't want RMG to expand certain networks further. 
You specify them with a line in your pressureDependence block of your input file like

`   completedNetworks = ['CH2O2'],`

This can be a list of things, specified by chemical formula of the potential energy surface that has been explored.

### Testing

I added a line to an example input file, for testing, but this is poor example (preventing ethane from reacting in the ethane oxidation example!) and we should change it before shipping this, so we're not distributing bad examples.
But for now, it demonstrates the new feature.

### Reviewer Tips
See if it works as intended, in a better example than the one I gave. Something where the PDep network _is_ complete in the seed mechanism.

### Up for debate
I wasn't sure where in the input file this should go. 
The list of networks is stored in the `rmg.reaction_model` object, not the `rmg.pressure_dependence`, but I think it only makes sense when you have pressure dependence on, so I put it in the pressureDependence block of the input file. I also considered the `generatedSpeciesConstraints` block. Or another block.  But maybe we're rewriting all the input files in yaml soon anyway...

We also considered in off-line discussion whether this should in fact live in the reaction library or seed mechanism, and you get networks marked as "complete" according to the seed mechanism you use. But that would involve editing the database also, and so for this testing of the concept I didn't try that. But would that be better in the long run?

### DIFF
--- a/test/rmgpy/rmg/inputTest.py
+++ b/test/rmgpy/rmg/inputTest.py
@@ -31,6 +31,7 @@ from unittest.mock import patch
 
 import rmgpy.rmg.input as inp
 from rmgpy.rmg.main import RMG
+from rmgpy.rmg.model import CoreEdgeReactionModel
 from rmgpy.ml.estimator import ADMONITION
 
 import pytest
@@ -384,6 +385,7 @@ class TestInputPressureDependence:
         """This method is run before every test in this class"""
         global rmg
         # Reset the completed networks set before each test
+        rmg.reaction_model = CoreEdgeReactionModel()
         rmg.reaction_model.completed_pdep_networks = set()
 
     def test_completed_networks_single(self):


Nora-Khalil:
Tested with 7 seven halocarbon models. Added `completedNetworks = ['CH2O2']` to pressureDependence block of my input files. RMG logs have: 

```
Added CH2O2 to list of completed PDep networks that will not be further explored.
...
Not adding reaction <some reaction> to unimolecular networks because the network for CH2O2 is marked as completed.
```
Whereas before, networks with CH2O2 were included in my models, CH2O2 is now left in the edge and no network is built that includes it.

PR #2856, opened 2025-11-13
Modify node regularization in kinetic families. 

Nora-Khalil:
### Motivation or Problem
Noticed that the nodes of some families were not being regularized. For example, in groups.py of the Retroene family, the following entry (below) specifies that the \*5 atom can be a fluorine, but the entire training set of the family includes zero fluorinated reactions. 

```
entry(
    index = 93,
    label = "Root_N-4R!H->C",
    group = 
"""
1 *3 R!H                       u0 r0 {2,S} {3,S}
2 *4 R!H                       u0 r0 {1,S} {4,S}
3 *2 C                         u0 {1,S} {5,[D,T,B]}
4 *5 [I,P,Br,Cl,N,Si,S,F,Li,O] u0 r0 {2,S} {6,S}
5 *1 R!H                       u0 {3,[D,T,B]}
6 *6 [H,Li]                    u0 r0 {4,S}
""",
    kinetics = None,
)

```

This indicates that node regularization was not carried out on this node. My understanding of node regularization is that it is performed to limit how "general" a node is and ensures that the node fits it's reaction data "tightly." Successful regularization on this node would mean that this node should not explicitly specify atoms like I, P, Br, Cl, Si, S, F, or Li if they are not included in the training reaction set. There are numerous other nodes in this family that are not regularized, and this is also occurring in other families. This is a problem because, if you were descending a fluorinated reaction down this tree, it would fall to a deeper depth down the tree to a node with fewer training reactions compared to the Root, but those fewer training reactions are not at all representative of the fluorinated reaction that we want to estimate a rate for. I could see this being an issue for other chemistries involving Br, Cl, Si, S, Li, etc. 

After some digging, I think regularization is being skipped because: 
1)  compliment nodes are missing the regularization dimensions that provide information on how the node should be regularized (this information is only passed to the extension node and not its compliment). I fix this by passing the info to the compliment. However, even when I pass the compliment the regularization info, if that compliment is further split via an atomExt extension as the tree is further extended, you run into issue 2. 
2) when a parent node is split via an atomExt extension (which splits the parent node into an extension node and compliment node), the regularization information of the parent node is overwritten with the regularization info of the new extension node (which resets the regularization information to an empty list). 

### Description of Changes
Different approach to passing the regularization information to each node. After an extension is selected, if that extension is of the type `atomExt` (i.e. changing an atom's atomtype R -> C) that splits the training reactions of the parent node into an extension node and complement node, then regularization info is passed to the extension node AND the compliment node. Regularization info of compliment is found by analyzing all of the reactions that fit the complimentary node and determining which atomtypes are in those reactions. After this information is found, regularization info is also passed to the parent node (and this information is no longer an empty list). 

### Testing
Simplest testing is with Retroene family. 

1) load instance of database
2) clean Retroene tree
3) generate tree 
4) Identify nodes that are not regularized (will have atomtype `[Si,Li,S,N,P,F,Br,I,Cl,C]`)
5) regularize family
6) re-check identified nodes from step 4. If things work, these nodes should have more specific atomtypes than before.
7) check entire tree for isomorphism.

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -2960,10 +2960,17 @@ class KineticsFamily(Database):
                     out_exts[-1].append(exts[i])  # this extension splits reactions (optimization dim)
                     if typ == 'atomExt':
                         reg_dict[(typ, indc)][0].extend(grp2.atoms[indc[0]].atomtype)
+                        #still pass in the regularization data to the grp2. However, this doesn't take care of the grpc
+                        #reg_dict[(typ, indc)][1].extend(grp2.atoms[indc[0]].atomtype)
+                        #now take care of the compliment: 
+
                     elif typ == 'elExt':
                         reg_dict[(typ, indc)][0].extend(grp2.atoms[indc[0]].radical_electrons)
+                        #reg_dict[(typ, indc)][1].extend(grp2.atoms[indc[0]].radical_electrons)
                     elif typ == 'bondExt':
                         reg_dict[(typ, indc)][0].extend(grp2.get_bond(grp2.atoms[indc[0]], grp2.atoms[indc[1]]).order)
+                        #reg_dict[(typ, indc)][1].extend(grp2.get_bond(grp2.atoms[indc[0]], grp2.atoms[indc[1]]).order)
+
 
                 elif boo:  # this extension matches all reactions (regularization dim)
                     if typ == 'intNewBondExt' or typ == 'extNewBondExt':
@@ -2990,10 +2997,12 @@ class KineticsFamily(Database):
                 reg_val = reg_dict[(typr, indcr)]
 
                 if first_time and parent.children == []:
-                    # parent
+                    
+                    #parent
                     if typr != 'intNewBondExt' and typr != 'extNewBondExt':  # these dimensions should be regularized
                         if typr == 'atomExt':
-                            grp.atoms[indcr[0]].reg_dim_atm = list(reg_val)
+                            pass 
+                            #grp.atoms[indcr[0]].reg_dim_atm = list(reg_val)
                         elif typr == 'elExt':
                             grp.atoms[indcr[0]].reg_dim_u = list(reg_val)
                         elif typr == 'ringExt':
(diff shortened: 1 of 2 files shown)


Nora-Khalil:
As a sanity check to see if this would work on a family other than Retroene, I ran this fix on 1+2_Cycloaddition family. Tree is generated and regularized without errors and passes final isomorphism check. Here's a file that shares the nodes that were unregularized, and the resulting group after regularization with my fix: [1+2_Cycloaddition.txt](https://github.com/user-attachments/files/23551344/1%2B2_Cycloaddition.txt)

Nora-Khalil:
Another comment, mostly to keep track of my thinking. There is definitely a more sophisticated way to fix this issue that doesn't involve reading in atomtypes directly from the training reactions of that node. It's probably better to keep with the architecture that Matt already has. 

Similar to the fix with the leaf nodes Matt mentioned, I see now that the problem should be solved if the problematic node then acts as a parent node and generates a further extension that is non-splitting on the atom of interest (even if that isn't the extension that is ultimately selected to extend the tree). This would update the reg_dict to include regularization information on the atom of interest in the parent (our problematic node). 

i.e. our problematic node 
```
Root_N-4R!H->O 
1 *3 R!H                       u0 {2,S} {3,[S,D]}
2 *4 R!H                       u0 {1,S} {4,[S,D]}
3 *2 R!H                       u0 {1,[S,D]} {5,[D,T,B]}
4 *5 [Si,Li,S,N,P,F,Br,I,Cl,C] u0 {2,[S,D]} {6,S}
5 *1 R!H                       u0 {3,[D,T,B]}
6 *6 [H,Li]                    u0 {4,S}
```
can be passed regularization info for \*5 if a non-splitting extension could be found on that atom. However, some debugging is showing me that attempts to extend this node doesn't propose any non-splitting extensions on *5 (only on *2 and *6). Hence why there's no regularization information for that atom in the group, and why it isn't regularized. I'm still figuring out how exactly `get_extension_edge` works, so there may be something I'm still missing...

Going to focus now on figuring out 1) whether or not there should be a non-splitting extension on *5 and 2) if there shouldn't be, how can we still pass regularization information to *5?

Nora-Khalil:
To save time, above commit was tested by regenerating tree of Retroene family instead of a bulky family. This family only has ~70 training reactions and doesn't require the cascade algorithm, but we can build the tree with cascading if we lower max_batch_size to 20 when generating tree. Confirmed that nodes in Retroene family are regularized when cascading is not used, but regularization is not performed on non-leaf nodes when cascading is used on this family.

rwest:
@Nora-Khalil should this be done?

Nora-Khalil:
I think so, this probably should not be closed. I can pick this back up after defense : )

PR #2866, opened 2025-12-05
Update Physical Constants and Corresponding Unit Tests

### LINKED ISSUE #2435: Update CODATA Constants
#1627 attempted to update the CODATA list of constants but was abandoned. This is worth doing, but should now be done with the 2022 release.

ttpekkan:
### Motivation or Problem
RMG currently uses outdated values for many physical constants. Many of these constants were given exact values when they were last updated, so now we have an opportunity to fix the values of these once and for all. This likely will have little to no difference in practice but is the correct thing to do. 

### Description of Changes
I changed the physical constants to the values reported here: [https://arxiv.org/pdf/2409.03787](url) For atomic weights, I used the most recent values I found from IUPAC: [https://iupac.qmul.ac.uk/AtWt/](url) Note that for some important elements like H and O it's not possible to give an unambiguous isotope-corrected atomic weight as it depends on the environment. I just went with the average value given in the table. Also, converting the dimensionless atomic weights to mol/kg requires multiplication by the molar mass constant (1.0000000010531e-3), which is no longer exactly 1.0e-3 due to how the physical constant were updated. I did this conversion, and although the correct thing to do, does not make a difference in practice. This also slightly affects the atomic mass unit as it is defined as the molar mass constant divided by Avogadro's constant. I did not include unstable elements. These can be added as needed, but I don't think RMG currently supports using them anyway. 

### Testing
I ran make–test, and as expected, it threw quite a few errors because in the unit tests some RMG output values are compared with pre-defined values. The RMG output values changed a tiny amount because of the physical-constant and atomic-weight updates, but enough for the unit tests to fail. I simply updated the pre-set values in the unit tests to those that RMG outputs with the updated physical constant and atomic weights. In a few cases I adjusted the tolerances for approximate equality comparisons to make the unit tests pass. Most of the errors appear to come from the change to the Hartree --> kJ/mol conversion factor. 

### Reviewer Tips
I guess what mostly needs to be done is to check that the physical constants do not have typos.


JacksonBurns:
#2920 should resolve the known sources of non-determinism. Once that's merged, let's rebase this PR and see what actual changes are happening.

JacksonBurns (commented):
So despite the fact that these _should_ be uniformly better estimates, I think we will shelve this PR for the time being. The tree estimators are fit to the old values, and this seems to cause some not-insignificant changes in the regression tests. Once regular refitting of estimators is possible this is worth merging almost as-is, but until then we will close this.

PR #2867, opened 2025-12-11
Correct constant-species index mapping in SimpleReactor

alongd:
### Motivation or Problem

`SimpleReactor.get_const_spc_indices()` was possibly effectively broken:

- It reused the same loop variable name in nested loops.
- It compared `spc.label == spc`, i.e. a string to a species object.

As a result, `const_spc_indices` was never populated and species listed in `const_spc_names` were perhaps not actually treated as constant.

### Description of Changes

In `SimpleReactor.get_const_spc_indices()`:

- Use different variable names for `const_spc_names` and `core_species`.
- Compare the constant-species name (string) against `spc.label` instead of `spc` itself.
- Initialize `const_spc_indices` once and append core-species indices when labels match.
- Return early if `const_spc_names` is `None`.

### Testing

A small unit test was added.

### Reviewer Tips

In a small reactor run with a constant species, check that its amount stays fixed over time.

### DIFF
--- a/test/rmgpy/solver/simpleTest.py
+++ b/test/rmgpy/solver/simpleTest.py
@@ -759,3 +759,29 @@ class SimpleReactorTest:
         # order: Ar, N2, O2, H, CH3, CH4
         for i in range(len(simulated_mole_fracs)):
             assert round(abs(simulated_mole_fracs[i] - expected_mole_fracs[i]), 6) == 0
+
+    def test_get_const_spc_indices(self):
+        """
+        Test that const_spc_names are correctly mapped to core species indices.
+        """
+        # Dummy species with labels that should match const_spc_names
+        a = Species(smiles='C', label="CH4")
+        b = Species(smiles='[OH]', label="OH")
+        core_species = [a, b]
+
+        T = 1000.0  # K
+        P = 1.0e5   # Pa
+
+        rxn_system = SimpleReactor(
+            T,
+            P,
+            initial_mole_fractions={a: 0.5, b: 0.5},
+            n_sims=1,
+            termination=[],
+            const_spc_names=["CH4"])
+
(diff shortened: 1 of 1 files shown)


alongd:
@JacksonBurns, any ideas why the Conda Build fails here?

rmgpy/solver/simple.pyx
```diff
@@ -169,16 +169,19 @@ cdef class SimpleReactor(ReactionSystem):
...
-            if self.const_spc_indices is None:  # initialize once the list if constant SPC declared
-                self.const_spc_indices = []
-            for spc in core_species: #Need to identify the species object corresponding to the the string written in the input file
-                if spc.label == spc:
-                    self.const_spc_indices.append(core_species.index(spc)) 
+        if self.const_spc_names is None:
+            return
+        if self.const_spc_indices is None:
+            self.const_spc_indices = []
+        for name in self.const_spc_names:
```
JacksonBurns:
```suggestion
        if self.const_spc_indices is None:
            self.const_spc_indices = []
        else:
            return  # this has already been run, results are cached
        for name in self.const_spc_names:
```

rwest:
Oh boy, it's inconsistent ring detection causing the thermo estimate to change for 
image

What's going on with the ring decomposer PR? (I've lost track). Or what was used to generate the "reference" run thermo?

JacksonBurns:
That PR (#2796) has not been merged yet, so the ring decomposition results in this PR are just particularly bad. :shrug:

#2796 was bogged down by a CI issue [which has since been fixed](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2796/#issuecomment-3710880305), so I just updated that branch and will see where it's at. If memory serves, there was one unit test failing, in somewhat of a surprising way.

PR #2606, opened 2024-02-07
Added a species constraint for the max number of rings fused together

alongd:
### Motivation or Problem
Sometimes users know they are not interested in ring growth in their model, yet RMG spends significant time on suggesting species with complex fused rings. Using the maximum heavy atoms constraints isn't always helpful since other species with the same Mw could be relevant for the model.

### Description of Changes
Added a `maximumFusedRings` key to the constraints dict.
A new method was implemented in Molecule along with a test.

### Testing
A unit test was added.
FWIW, the feature was also tested locally by generating a model that previously gave many fused ring structures and verifying that they are absent now.

### Reviewer Tips
Note that a previously non-implemented constraint was removed from the docs
I saw that this branch fails online due to a Julia installation, I think. Looks unrelated to this feature.

PR #2868, opened 2025-12-12
Added a species constraint for the max number of rings fused together (revived)

alongd:
### Motivation or Problem
(reviving abandoned PR #2606)
Sometimes users know they are not interested in ring growth in their model, yet RMG spends significant time on suggesting species with complex fused rings. Using the maximum heavy atoms constraints isn't always helpful since other species with the same Mw could be relevant for the model.

### Description of Changes
Added a `maximumFusedRings` key to the constraints dict.
A new method was implemented in Molecule along with a test.

### Testing
A unit test was added.
FWIW, the feature was also tested locally by generating a model that previously gave many fused ring structures and verifying that they are absent now.

### Reviewer Tips
Note that a previously non-implemented constraint was removed from the docs

### DIFF
--- a/arkane/ess/gaussian.py
+++ b/arkane/ess/gaussian.py
@@ -123,6 +123,24 @@ class GaussianLog(ESSAdapter):
 
         return n_atoms
 
+    def load_route_section(self):
+        """
+        Return the route section of the Gaussian log file as a string. The route
+        section is the line in the Gaussian input/output file that starts with "#"
+        If there are multiple lines that start with # it returns the first.
+        """
+        route = ''
+        with open(self.path, 'r') as f:
+            line = f.readline()
+            while line != '':
+                line = line.strip()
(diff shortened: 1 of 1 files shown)


documentation/source/users/rmg/input.rst
```diff
@@ -1030,6 +1031,9 @@ from either the input file, seed mechanisms, or reaction libraries to bypass the
 Note that this should be done with caution, since the constraints will still apply to subsequent
 products that form.
 
+``maximumFusedRings`` is the maximum number of rings in any fused ring system in the species
```
mjohnson541: The language is tricky here. I had to think about the wording for this comment. I don't think it's clear here the distinction between the maximum number of groups of inter-fused rings allowed in the molecule (which I don't think this is constraining) and the maximum number of SSSR rings in each set of inter-fused rings.

PR #2869, opened 2026-01-05
Update CI to use new macOS runner images

jonwzheng:
As of December, the macOS 13 runner image is deprecated (https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/) 

This PR just switches `macos-13` with `macos-15-intel` so that we still have a test for the Intel architecture.


JacksonBurns (approved):
Thanks @jonwzheng , should be a drop-in replacement

PR #2800, opened 2025-05-30
Yaml writer addition

rwest:
This pull request adds generation, comparison, and testing of Cantera YAML files. The main changes include refactoring the Chemkin-to-Cantera conversion flow, supporting multiple Cantera YAML writers, comparing the outputs of different generation methods. 
For now it implements two different Cantera YAML Writers, because they were developed in parallel and it was not immediately obvious which to choose. 

**Key changes:**

### Cantera YAML generation and comparison

* Refactored the Chemkin-to-Cantera conversion logic into a new method `generate_cantera_files_from_chemkin`, which now returns the output file path and saves files in a new `cantera_from_ck` directory. The original method name and references have been updated throughout the codebase. [[1]](diffhunk://#diff-80ae0bfd0f3343f811a4399c3f615ca153e05be0141c6b0839b8959230fb5ac6L1805-R1833) [[2]](diffhunk://#diff-80ae0bfd0f3343f811a4399c3f615ca153e05be0141c6b0839b8959230fb5ac6R1851) [[3]](diffhunk://#diff-a6377cca9995ba5db5fe9bd643d3056a4dcd5048c66a76bd076ba7258e16c6c1L1065-R1065)
* Added support for two new direct Cantera YAML writers, `CanteraWriter1` and `CanteraWriter2`, and registered them alongside the existing RMS writer. [[1]](diffhunk://#diff-80ae0bfd0f3343f811a4399c3f615ca153e05be0141c6b0839b8959230fb5ac6R82-R86) [[2]](diffhunk://#diff-80ae0bfd0f3343f811a4399c3f615ca153e05be0141c6b0839b8959230fb5ac6L773-R777)
* Enhanced the main execution flow to compare the Cantera YAML files generated by direct writers and by Chemkin conversion, producing comparison reports for both methods.

### YAML/RMS writer refactor

* Renamed `rmgpy/yml.py` to `rmgpy/yaml_rms.py` and updated function names from `write_yml`/`convert_chemkin_to_yml` to `write_rms`/`convert_chemkin_to_rms` for clarity and consistency. [[1]](diffhunk://#diff-6b16ff7dcc7ca4420661b9a5ed6abec18d00d24ae506f02a2b72d5deb0fb3b11L52-R60) [[2]](diffhunk://#diff-6b16ff7dcc7ca4420661b9a5ed6abec18d00d24ae506f02a2b72d5deb0fb3b11L282-R282)

### Motivation or Problem
[...]

PR #2870, opened 2026-01-09
Add CanteraWriter

alongd:
### Motivation or Problem
Currently RMG outputs a Cantera YAML file only after it converges, using the external ck2yaml script.
Sometimes the script fails, and sometimes if RMG doesn't converge, a Cantera YAML model might still be useful as an intermediate step to assess the development status.

### Description of Changes
A CanteraWriter module was added

### Testing
Tests were added

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -26,6 +26,7 @@ Performance-critical code uses Cython (`.pyx` files) with declaration files (`.p
 - Use `cpdef` for methods callable from both Python and Cython
 - Use `cimport` for Cython-level imports (e.g., `cimport rmgpy.constants as constants`)
 - Register new Cython modules in `setup.py` `ext_modules` list
+- Remember to re-compile (e.g. run `make`) after modifying any `.pyx` or `.pxd` files, or if there seem to be weird bugs in them.
 
 ## Development Commands
 ```bash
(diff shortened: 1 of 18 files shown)


rwest (commented):
This looks great - thanks for doing it Alon. We've been wishing for it for a long time (see https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2321 from 2022, then #2770 and #2800).

I made a few comments, from just reading the diff on GitHub. I have not actually tried running this, but if you say it works I believe you :)

I would like to know what unit testing we're doing and how exhaustive. 
Can we round-trip to/from Cantera and back?

I don't want striving to for perfection to prevent this ever happening (see #2321!) but a "wish list" we can revisit would be nice.

I'm also curious about speed. I think we could probably speed this up a lot, fairly easily (templating, caching. Making dicts each time and using yaml libraries will be relatively slow). But again, that could come after it's definitely working and tested and implemented, and profiled.

alongd:
Thanks @rwest! Supper appreciated!

We need to decide whether to continue with this approach (and I'll address the above comments), or whether we still want to use/support all the `to_cantera()` functionalities currently embedded in RMG's objects, which might mean delaying the implementation further.
We definitely don't want duplication (the current state of this PR). If the `to_cantera` methods don't play additional roles, I may suggest that if we can show the round trip conversion of this module works as Richard noted, and that it's relatively robust and fast enough, then we may continue and consider deprecating the other functionalities. I'll check and proceed cautiously.

Issue #2865, opened 2025-12-04
Sensitivity Analysis Import Issue


calvinp0:
### Bug Description
When attempting to run the sensitivity analysis, I and another colleague receive the issue 

```python
Traceback (most recent call last):
  File "/home/calvin/code/RMG-Py/scripts/simulate.py", line 38, in <module>
    from rmgpy.tools.simulate import run_simulation
  File "/home/calvin/code/RMG-Py/rmgpy/tools/simulate.py", line 35, in <module>
    from rmgpy.kinetics.diffusionLimited import diffusion_limiter
  File "/home/calvin/code/RMG-Py/rmgpy/kinetics/diffusionLimited.py", line 37, in <module>
    from rmgpy.species import Species
  File "rmgpy/species.py", line 57, in init rmgpy.species
  File "/home/calvin/code/RMG-Py/rmgpy/pdep/__init__.py", line 32, in <module>
    from rmgpy.pdep.network import Network
  File "/home/calvin/code/RMG-Py/rmgpy/pdep/network.py", line 41, in <module>
    from rmgpy.reaction import Reaction
  File "rmgpy/reaction.py", line 59, in init rmgpy.reaction
ImportError: cannot import name diffusion_limiter
```

The current fix I do is to move the import of Species from the top of `diffusionLimited.py` file, and make it a local import in the `get_diffusion_limit` function.

### How To Reproduce
Run an RMG model, then do the sensitivity analysis. I will attach the `input.py` I used, but I think maybe `superminimal` should be fine to show the issue.

### Expected Behavior
Run RMG, then afterwards, edit the input and provide the `sensitivity` and `sensitivityThreshold` blocks, then run this line `python simulate.py input.py chem.inp species_dictionary.txt`

### Installation Information
Describe your installation method and system information.
 - OS (include version if known): (e.g. Windows 10, Ubuntu 18, macOS 10.13, etc.)
 - Installation method: (e.g. installation from binary or source, with or without anaconda)
 - RMG version information:
   
   - RMG-Py: d1379a70ee99399370d443b2edba5c1db5c9bc6c (the git describe --tags didnt output a name but it is the latest)
   - RMG-database: 3.30

### Additional Context
If you run the input.py I attach, there is another error that `  TypeError: Cannot convert 'complex' with non-zero imaginary component to 'double' (this most likely comes from the '**'
  operator; use 'cython.cpow(True)' to return 'nan' instead of a complex number).`
But I can open an additional

[...]

PR #2880, opened 2026-02-04
Imports Species class locally for diffusion_limited_rate function 

calvinp0:
Moves the import of the `Species` class to be local to the diffusion_limited_rate` function. This avoids a circular import issue as mentioned here: https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2865

This is a quick fix for the issue when a user attempts to use the sensitivity analysis


rwest (approved):
Pity this wasn't caught by any functional or unit tests....
And I wonder if the culprit was adding the Fragment method.
Anyway, this looks like a reasonable patch.  Thanks for finding and fixing, Calvin!

PR #2891, opened 2026-03-03
Specify initial core species using external dictionary

sevyharris:
### Motivation or Problem
Sometimes you want to start RMG with a list of core species that's too long to easily fit inside an input file. This PR allows you to specify an external species dictionary from which to load all your starting core species.

### Description of Changes
This adds the `coreSpeciesFile` option to input files, where you can specify an external species dictionary from which to read more species.

Then the block you'd put in the input file looks something like this:
```
coreSpeciesFile(
    species_dictionary_file="species_dictionary.txt"
)
```

### Testing
I ran this with a small species dictionary and the superminimal input file and it added in the additional species.
[input.txt](https://github.com/user-attachments/files/25722553/input.txt)
[species_dictionary.txt](https://github.com/user-attachments/files/25722554/species_dictionary.txt)

Then I commented out the coreSpeciesFile option and it didn't add the species in species_dictionary.txt

I built the new documentation on my local computer and it looks okay.

### Reviewer Tips
¯\_(ツ)_/¯

### DIFF
--- a/documentation/source/users/rmg/input.rst
+++ b/documentation/source/users/rmg/input.rst
@@ -322,6 +322,14 @@ The following is an example of a typical species item, based on methane using SM
 			"""
 	)
 
+For long lists of initial core species, you can specify a coreSpeciesList, shown below::
+
+    coreSpeciesList(
+        species_dictionary_file="path/to/your/species_dictionary.txt"
+    )
+
+This specifies the path for an RMG-formatted :ref:`species dictionary <the_chemkin_folder>` from which additional species will be loaded. Note that RMG will complain if any species in the input file is duplicated in the species dictionary.
+
 .. _forbidden_structures:
 
 Forbidden Structures
--- a/documentation/source/users/rmg/input.rst
+++ b/documentation/source/users/rmg/input.rst
@@ -328,7 +328,7 @@ For long lists of initial core species, you can specify a coreSpeciesList, shown
         species_dictionary_file="path/to/your/species_dictionary.txt"
     )
 
-This specifies the path for an RMG-formatted :ref:`species dictionary <the_chemkin_folder>` from which additional species will be loaded. Note that RMG will complain if any species in the input file is duplicated in the species dictionary.
+This specifies the path for an RMG-formatted :ref:`species dictionary <the_chemkin_folder>` from which additional species will be loaded. Note that RMG will complain if any species in the input file is duplicated in the species dictionary. All species are assumed to be reactive.
 
 .. _forbidden_structures:
 
--- a/rmgpy/rmg/input.py
+++ b/rmgpy/rmg/input.py
@@ -40,6 +40,7 @@ from rmgpy.data.surface import MetalDatabase
 from rmgpy.data.vaporLiquidMassTransfer import (
     liquidVolumetricMassTransferCoefficientPowerLaw,
 )
+from rmgpy.chemkin import load_species_dictionary
 from rmgpy.exceptions import DatabaseError, InputError
 from rmgpy.molecule import Molecule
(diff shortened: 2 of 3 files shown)


rmgpy/rmg/input.py
```diff
@@ -190,6 +191,16 @@ def convert_binding_energies(binding_energies):
     return new_dict
 
 
+def core_species_file(species_dictionary_file):
+    # all species here are assumed to be reactive
+    rmg.species_core_file = species_dictionary_file
```
rwest, line 196: We save it here, but for any purpose?

What happens when one tries to save an RMG input file? (Is that still a thing?)

rwest (commented):
Nice idea. And implementation looks OK. 

Should we aim to be able to save an RMG input file?
Is the source (once saved) used for anything? If not, do we need to save it? 
Is there anything we can usefully unit test?
Should we add to one of the examples?

Nora-Khalil:
Not that anyone asked for my review 👀 ...

But I was able to use this PR to create an RMG job that pulls core species from an external dictionary with 147 species defined. RMG job started fine, and I can see that my first chemkin file indeed has all of the species defined in my external dictionary. I also believe that if a common species is defined in a seed AND the external dictionary, it is only added in once from the dictionary as desired. 

Very helpful addition, @sevyharris!

rmgpy/rmg/input.py
Nora-Khalil: Just tested. So you can load an RMG object with the input file fine, which is good, and we can confirm that the RMG object will have an attribute `species_core_file` that prints the path to the dictionary file. 

If you then call `save_input` on that RMG object, an input file is saved that has all the core species (including those coming from the dictionary) listed in the traditional format i.e. many blocks of 

```
species(
    label = label_here,
    reactive = True,
    structure = adjacencyList(
adjacency_list_here)
```

There is no coreSpeciesFile() block in the saved input file. 

I personally think it's a good thing that the input.py file lists everything in species blocks instead of including a coreSpeciesFile() block; it gives the user an easy way to generate a "traditional-looking" input.py file, and then its a single stand-alone file too which is nice.

Nora-Khalil (dismissed):
looks good to me : ) Approving 😎

rmgpy/rmg/input.py
sevyharris: I'm not sure if it's necessary to save the species_core_file to the rmg object. I'll check and delete if it's not used.

And, yes, I would also prefer for the species in a species_core_file to also get saved directly in the input.py file if you write it.

sevyharris: The species_core_file is not used at all, so I deleted the line saving it to the rmg object.

documentation/source/users/rmg/input.rst
```diff
@@ -322,6 +322,14 @@ The following is an example of a typical species item, based on methane using SM
 			"""
 	)
 
+For long lists of initial core species, you can specify a coreSpeciesList, shown below::
+
+    coreSpeciesList(
+        species_dictionary_file="path/to/your/species_dictionary.txt"
+    )
+
+This specifies the path for an RMG-formatted :ref:`species dictionary <the_chemkin_folder>` from which additional species will be loaded. Note that RMG will complain if any species in the input file is duplicated in the species dictionary.
```
JacksonBurns:
```suggestion
This specifies the path for an RMG-formatted :ref:`species dictionary <the_chemkin_folder>` from which additional species will be loaded. Note that RMG will complain if any species in the input file is duplicated in the species dictionary. All species are assumed to be reactive.
```

JacksonBurns (approved):
I concur with @Nora-Khalil

PR #2909, opened 2026-03-27
Added the auto-database feature

alongd:
Adds automatic library and kinetics family selection to RMG. Users can now write `thermoLibraries='auto'` (and same for reaction libraries, transport, seeds, and kinetics families) in their input file, and RMG will pick the right libraries based on what species and reactor conditions are in the input (and based on the correctness of [recommended_libraries.yml](https://github.com/ReactionMechanismGenerator/RMG-database/pull/712)).
                                                                                                                                                      
  The selection logic detects elements (N, S, O, halogens, Li), reactor type (gas/liquid/surface), and temperature to trigger the appropriate         
  chemistry sets defined in a new `recommended_libraries.yml` file in RMG-database. Kinetics families are similarly auto-selected from the existing
  recommended.py sets.                                                                                                                                
                                                            
  Key design choices:
  - 'auto' is opt-in, not the default, so existing input files work unchanged (backwards compatible)
  - Users can mix manual and auto: ['myLib', 'auto'] gives myLib higher priority                                                                      
  - PAH formation libraries (70+ kinetics libs) are only auto-included for pure C/H pyrolysis; oxygenated systems need the explicit <PAH_libs> keyword
   to opt in                                                                                                                                          
  - Families support ['!H_Abstraction', 'auto'] to exclude specific families from the auto set                                                        
                                                                                                                                                      
 A Preview notebook at ipython/auto_library_selection.ipynb

We should first merge the db branch https://github.com/ReactionMechanismGenerator/RMG-database/pull/712

PR adapted from the existing implementation in T3 ([libraries.yml](https://github.com/ReactionMechanismGenerator/T3/blob/main/data/libraries.yml) and [code](https://github.com/ReactionMechanismGenerator/T3/blob/migration/t3/utils/libraries.py))

### DIFF
--- a/ipython/auto_library_selection.ipynb
+++ b/ipython/auto_library_selection.ipynb
@@ -1,0 +1,526 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "uak48ywvas",
+   "metadata": {},
+   "source": [
+    "# Auto Library & Family Selection Preview\n",
+    "\n",
+    "Preview what RMG's `'auto'` mode would select for thermo libraries, kinetics libraries,\n",
+    "transport libraries, seed mechanisms, and kinetics families — given an RMG input file.\n",
+    "\n",
+    "Just set `input_file_path` below and run all cells."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "5vx9v43udjl",
+   "metadata": {
+    "execution": {
+     "iopub.execute_input": "2026-04-25T04:32:49.811977Z",
+     "iopub.status.busy": "2026-04-25T04:32:49.811860Z",
+     "iopub.status.idle": "2026-04-25T04:32:49.816013Z",
+     "shell.execute_reply": "2026-04-25T04:32:49.815710Z"
+    }
+   },
+   "outputs": [],
+   "source": [
+    "input_file_path = '../examples/rmg/superminimal_auto/input.py'"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "id": "vyt4k7q9as",
+   "metadata": {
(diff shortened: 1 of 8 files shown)


ipython/auto_library_selection.ipynb
```diff
@@ -0,0 +1,103 @@
...
+   "execution_count": null
+  },
+  {
+   "metadata": {},
+   "cell_type": "code",
+   "source": "",
+   "id": "7dba6e3be37a396b",
+   "outputs": [],
+   "execution_count": null
+  }
```
JacksonBurns, line 503: Can delete this empty cell

ipython/auto_library_selection.ipynb
JacksonBurns, line 1: I would save an executed version of this notebook in RMG-Py just so users know what to expect from the output. This may warrant adding another example, which could just be the superminimal example with `'auto'`

rmgpy/data/auto_database.py
```diff
@@ -0,0 +1,603 @@
...
+        solvent: solvent name string, or None for gas phase.
+
+    Returns:
+        ChemistryProfile with detected characteristics.
+    """
+    profile = ChemistryProfile()
+
+    for spc in initial_species:
+        if not spc.reactive:
+            continue
```
JacksonBurns, line 128: Since we are (correctly) ignoring non-reactive species, I would add in the documentation that this is the case. A user might be confused if they use nitrogen as a bath gas and the nitrogen library is not used (at least, I would be, at first).

JacksonBurns (dismissed):
I think this PR is ready to go. I have left some largely inconsequential comments.

alongd:
Thanks, @JacksonBurns! All comments were addressed, I added an example file.
Changes to Docs and iPyNB were squashed

PR #2914, opened 2026-04-05
Added Arkane xTB ESS adapter

alongd:
### Motivation or Problem
Allowing Arkane to work with `gfn2` in xTB as a new ESS adapter

### Description of Changes
Added an ESS adapter with all relevant parsers for xTB

### Testing
Added tests for parsing xTB

### DIFF
--- a/arkane/ess/xtb.py
+++ b/arkane/ess/xtb.py
@@ -1,0 +1,347 @@
+#!/usr/bin/env python3
+
+###############################################################################
+#                                                                             #
+# RMG - Reaction Mechanism Generator                                          #
+#                                                                             #
+# Copyright (c) 2002-2023 Prof. William H. Green (whgreen@mit.edu),           #
+# Prof. Richard H. West (r.west@neu.edu) and the RMG Team (rmg_dev@mit.edu)   #
+#                                                                             #
+# Permission is hereby granted, free of charge, to any person obtaining a     #
+# copy of this software and associated documentation files (the 'Software'),  #
+# to deal in the Software without restriction, including without limitation   #
+# the rights to use, copy, modify, merge, publish, distribute, sublicense,    #
+# and/or sell copies of the Software, and to permit persons to whom the       #
+# Software is furnished to do so, subject to the following conditions:        #
+#                                                                             #
+# The above copyright notice and this permission notice shall be included in  #
+# all copies or substantial portions of the Software.                         #
+#                                                                             #
+# THE SOFTWARE IS PROVIDED 'AS IS', WITHOUT WARRANTY OF ANY KIND, EXPRESS OR  #
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,    #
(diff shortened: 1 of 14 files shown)


alongd:
@JacksonBurns, one of the Conda Builds fails, I think it's not related to this branch specifically. Could you please take a look?

JacksonBurns:
@alongd the Python 3.11 conda build seems to be taking so long that it gets cancelled... I recall back when working on the PR to spee dup the CI that 3.9 was faster. I will add that onto this PR and see if it fixes the problem

sevyharris:
This is a great! I'm able to read in frequencies, energies, and geometries for xtb calculations.

One question I have is whether we want the ability to run full Arkane statmech/thermo jobs with the results? I assumed this was your intention, but maybe adding this ess adapter has other uses.

If we do want to run statmech/thermo jobs, I think we'd have to make the following additions:
1. [statmech.py](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/main/arkane/statmech.py) would need special handling of XTBLogs to construct the conformer since the frequency doesn't contain the geometry and statmech.py defaults to the frequency file to figure out things like symmetry.

2. We'd need to add xtb to the list of valid software names in arkane/modelchem.py
https://github.com/ReactionMechanismGenerator/RMG-Py/blob/9cc2cfdedd6a7090d92782011d8f84c6292bce09/arkane/modelchem.py#L279

3. xtb atom energy corrections? https://github.com/ReactionMechanismGenerator/RMG-database/blob/main/input/quantum_corrections/data.py



##
If you're okay with me pushing to this branch, I'm happy to make these additions. It could also go in a separate PR.

If you just wanted to be able to read the logfiles into Arkane objects in memory, let me know and I'll approve once I finish checking just a few more things.

alongd:
@sevyharris, sure, please feel free to push to this branch :)
Thanks!

arkane/ess/xtb.py
```diff
@@ -0,0 +1,347 @@
...
+            _optical_isomers, _symmetry, _ = self.get_symmetry_properties()
+            if optical_isomers is None:
+                optical_isomers = _optical_isomers
+            if symmetry is None:
+                symmetry = _symmetry
+
+        if spin_multiplicity == 0:
+            spin_multiplicity = self._load_spin_multiplicity()
+
+        unscaled_frequencies = self._load_frequencies()
```
sevyharris, line 299: I wasn't able to figure out how to call xtb such that it outputted a single logfile with frequencies and geometries together so that I could call load_conformer() using a single file and get all the modes.

If there is a way to get geometries and frequencies in the same output file, please let me know and I'll undo the workaround I added to statmech.py because it feels kinda messy piecing together a conformer object from separate geometry and frequency files. But maybe that's okay.

sevyharris (approved):
I've tested reading in the geometry, energies, frequencies, ZPE, and it all looks good to me.

I've also run some example thermo calculations using the xTB atom energy corrections proposed here https://github.com/ReactionMechanismGenerator/RMG-database/pull/721 and the overall xtb thermo calculations look good to me.

If you think the changes to statmech look good, then I'd say feel free to merge this in.

alongd:
Thanks, @sevyharris!

PR #2917, opened 2026-04-06
Blowers-Masel fitting: changes and refactoring

rwest:
### Motivation or Problem

In recent years, @Nora-Khalil had been having issues with re-training some kinetics trees with halocarbon data, and @davidfarinajr had made various tweaks before that.  Along the way we collected various changes to the Blowers-Masel fitting code. Some of these changes date back to 2021, but somehow only ever existed on David and Nora's topic branches.

Here they are, collected and rebased, with some extra refactoring along the way.

The Blowers-Masel (BM) fitting in `ArrheniusBM` and `ArrheniusChargeTransferBM` had several issues: the optimizer used poor initial guesses (hard-coded `lnA=1, n=1, E0=w0/10`) that frequently led to non-converging or physically unreasonable fits; negative `E0` values encountered during optimization caused undefined/incorrect behavior; Additionally, fitted BM parameters with `abs(n) > 5` are usually artifacts of reactions that don't follow BM enthalpy-dependence, but these were previously trusted.

### Description of Changes

Behavioral changes:
-   **Reject untrustworthy BM fits** where `abs(n) > 5.0` in `_make_rule` (kinetics family), as these typically indicate reactions that don't exhibit BM enthalpy-dependence.
-   **Negative `E0` handling** in `get_activation_energy`: instead of crashing or producing nonsensical values, returns `E0 + max(dHrxn, 0)` when `E0 < 0` --- a physically continuous and monotonic choice at the `E0=0` boundary.  This is what I chose:
BMsketch

Small behavioral changes:
-   **Better initial guesses** for the `curve_fit` optimizer: `lnA`, `n`, and `E0` are now seeded from the mean Arrhenius parameters of the training reactions (using a BEP relation with `alpha=0.25` for `E0`), rather than fixed values.
-   **Enforced `w0 >= 2*E0`** before optimization, since the BM expression requires this constraint; `w0` is updated adaptively if the _fitted_ `E0` exceeds `w0`.
-   **Unit handling fix**: `w0` and `E0` are now stored internally in kJ/mol.

Performance improvements:
-   **Performance**: replaced `np.log(T**n * np.exp(-Ea/(R*T)))` with `np.log(T)*n - Ea/(R*T)` in the objective function.
-   `dHrxn` is now computed once per reaction rather than redundantly inside the temperature loop. 

Code simplification refactoring:
-   **Minor refactor** of `_make_rule` in `rmgpy/data/kinetics/family.py`: removed an always-true `if True:` block and reduced indentation; no behavioral change.
-   **Simplified `Vp` formula** in `get_activation_energy` and `fit_to_reactions` for both `ArrheniusBM` and `ArrheniusChargeTransferBM`: simplified `2*w0*(2*w0+2*E0)/(2*w0-2*E0)` → `2*w0*(w0+E0)/(w0-E0)` and the denominator `(2*w0)^2` → `4*w0*w0` (algebraically equivalent, avoids redundant factors of 2).

### Testing

So far, not much besides the standard test suites.

### Reviewer Tips

Note there is code duplication between `ArrheniusBM` and `ArrheniusChargeTransferBM` --- changes (should) appear twice throughout.

### Open question

How should we evaluate different ways of doing fitting? 
How do we know what's best?

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -4596,13 +4596,16 @@ def _make_rule(rr):
     for i, rxn in enumerate(rxns):
         rxn.rank = ranks[i]
     rxns = np.array(rxns)
-    rs = np.array([r for r in rxns if type(r.kinetics) != KineticsModel])
+    rs = np.array([r for r in rxns if type(r.kinetics) is not KineticsModel]) # KineticsModel is the base class with no data.
     n = len(rs)
-    if n > 0 and isinstance(rs[0].kinetics, Marcus):
+    if n == 0:
+        return None
+
+    if isinstance(rs[0].kinetics, Marcus):
         kin = average_kinetics([r.kinetics for r in rs])
         return kin
     data_mean = np.mean(np.log([r.kinetics.get_rate_coefficient(Tref) for r in rs]))
-    if n > 0:
+    if n > 0: # not needed. at this point n should always be > 0
         if isinstance(rs[0].kinetics, Arrhenius):
             arr = ArrheniusBM
         else:
@@ -4772,7 +4775,7 @@ def average_kinetics(kinetics_list):
             beta=(beta,"1/m"),
             wr=(wr * 0.001, "kJ/mol"),
             wp=(wp * 0.001, "kJ/mol"),
-            comment="Averaged from {} reactions.".format(len(kinetics_list)),
+            comment=f"Averaged from {len(kinetics_list)} rate expressions.",
             )
     elif isinstance(kinetics, SurfaceChargeTransfer):
         averaged_kinetics = SurfaceChargeTransfer(
@@ -4782,6 +4785,7 @@ def average_kinetics(kinetics_list):
             alpha=alpha,
             V0=(V0,'V'),
             Ea=(Ea * 0.001, "kJ/mol"),
+            comment=f"Averaged from {len(kinetics_list)} rate expressions.",
             )
     elif isinstance(kinetics, ArrheniusChargeTransfer):
         averaged_kinetics = ArrheniusChargeTransfer(
@@ -4791,6 +4795,7 @@ def average_kinetics(kinetics_list):
             alpha=alpha,
             V0=(V0,'V'),
             Ea=(Ea * 0.001, "kJ/mol"),
+            comment=f"Averaged from {len(kinetics_list)} rate expressions.",
             )
     else:
         averaged_kinetics = Arrhenius(
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -4605,73 +4605,72 @@ def _make_rule(rr):
         kin = average_kinetics([r.kinetics for r in rs])
         return kin
     data_mean = np.mean(np.log([r.kinetics.get_rate_coefficient(Tref) for r in rs]))
-    if n > 0: # not needed. at this point n should always be > 0
-        if isinstance(rs[0].kinetics, Arrhenius):
-            arr = ArrheniusBM
+
+    if isinstance(rs[0].kinetics, Arrhenius):
+        arr = ArrheniusBM
+    else:
+        arr = ArrheniusChargeTransferBM
+    if n > 1:
+        kin = arr().fit_to_reactions(rs, recipe=recipe)
+    if n == 1 or kin.E0.value_si < 0.0:
+        kin = average_kinetics([r.kinetics for r in rs])
+        #kin.comment = "Only one reaction or Arrhenius BM fit bad. Instead averaged from {} reactions.".format(n)
+        if n == 1:
+            kin.uncertainty = RateUncertainty(mu=0.0, var=(np.log(fmax) / 2.0) ** 2, N=1, Tref=Tref, data_mean=data_mean, correlation=label)
         else:
-            arr = ArrheniusChargeTransferBM
-        if n > 1:
-            kin = arr().fit_to_reactions(rs, recipe=recipe)
-        if n == 1 or kin.E0.value_si < 0.0:
-            kin = average_kinetics([r.kinetics for r in rs])
-            #kin.comment = "Only one reaction or Arrhenius BM fit bad. Instead averaged from {} reactions.".format(n)
-            if n == 1:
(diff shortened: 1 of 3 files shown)


rmgpy/kinetics/arrhenius.pyx
```diff
@@ -573,16 +573,20 @@ cdef class ArrheniusBM(KineticsModel):
...
-            w0 = self._w0.value_si
-            Vp = 2 * w0 * (2 * w0 + 2 * E0) / (2 * w0 - 2 * E0)
-            return (w0 + dHrxn / 2.0) * (Vp - 2 * w0 + dHrxn) ** 2 / (Vp ** 2 - (2 * w0) ** 2 + dHrxn ** 2)
+            Vp = 2 * w0 * (w0 + E0) / (w0 - E0)
+            Ea = (w0 + dHrxn / 2.0) * (Vp - 2 * w0 + dHrxn) * (Vp - 2 * w0 + dHrxn) / (Vp * Vp - (4 * w0 * w0) + dHrxn * dHrxn)
+            if (Ea < 0.0) or (dHrxn < -4 * E0):
+                Ea = 0.0
+            elif (Ea < dHrxn) or (dHrxn > 4 * E0):
+                Ea = dHrxn
+        return Ea
```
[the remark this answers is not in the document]
rwest: BMsketch

rmgpy/kinetics/arrhenius.pyx
```diff
@@ -642,23 +647,40 @@ cdef class ArrheniusBM(KineticsModel):
             def kfcn(xs, lnA, n, E0):
                 T = xs[:,0]
                 dHrxn = xs[:,1]
-                Vp = 2 * w0 * (2 * w0 + 2 * E0) / (2 * w0 - 2 * E0)
-                Ea = (w0 + dHrxn / 2.0) * (Vp - 2 * w0 + dHrxn) * (Vp - 2 * w0 + dHrxn) / (Vp * Vp - (2 * w0) * (2 * w0) + dHrxn * dHrxn)
+                Vp = 2 * w0 * (w0 + E0) / (w0 - E0)
+                Ea = (w0 + dHrxn / 2.0) * (Vp - 2 * w0 + dHrxn) * (Vp - 2 * w0 + dHrxn) / (Vp * Vp - (4 * w0 * w0) + dHrxn * dHrxn)
```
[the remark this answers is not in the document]
rwest: Good catch, see reply below.

rmgpy/kinetics/arrhenius.pyx
```diff
@@ -667,10 +689,18 @@ cdef class ArrheniusBM(KineticsModel):
...
             while keep_trying:
                 keep_trying = False
                 try:
-                    params = curve_fit(kfcn, xdata, ydata, sigma=sigmas, p0=[1.0, 1.0, w0 / 10.0], xtol=xtol, ftol=ftol)
+                    params = curve_fit(kfcn, xdata, ydata, sigma=sigmas, p0=[lnA, n, E0], xtol=xtol, ftol=ftol)
+                    lnA, n, E0 = params[0].tolist()
+                    if abs(E0/self.w0.value_si) > 1 and attempts < 5:
+                        keep_trying = True
+                        if attempts > 0:
+                            self.w0.value_si *= 1.25
```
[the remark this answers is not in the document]
rwest: Good catch, I'll fix this, but you missed the other instance of the same thing, on line 1706.   Don't trust copilot too much!

rwest:
I addressed co-pilot's comments. (Some via fixup commits that I have since rebased).

rmgpy/data/kinetics/family.py
```diff
@@ -4596,79 +4596,89 @@ def _make_rule(rr):
...
         else:
-            arr = ArrheniusChargeTransferBM
-        if n > 1:
-            kin = arr().fit_to_reactions(rs, recipe=recipe)
-        if n == 1 or kin.E0.value_si < 0.0:
-            kin = average_kinetics([r.kinetics for r in rs])
-            #kin.comment = "Only one reaction or Arrhenius BM fit bad. Instead averaged from {} reactions.".format(n)
-            if n == 1:
-                kin.uncertainty = RateUncertainty(mu=0.0, var=(np.log(fmax) / 2.0) ** 2, N=1, Tref=Tref, data_mean=data_mean, correlation=label)
+            if kin.E0.value_si < 0.0:
```
Nora-Khalil, line 4622: We can enter on line 4615 (in green code of this PR) if n>1 but satisfies either E0 is negative or n>5. In this case, our kin type is ArrheniusBM. On line 4617, kin is converted to an Arrhenius type in `average_kinetics` if it is not of the type SurfaceChargeTransfer, ArrheniusChargeTransfer, Marcus. 

So we could have a ArrheniusBM -> Arrhenius, which errors out at line 4622 with `AttributeError: 'rmgpy.kinetics.arrhenius.Arrhenius' object has no attribute 'E0'`

Nora-Khalil:
Question: Does the updated BM fitting introduced in this PR change the error in nodes of the families? 

Performed K-fold cross validation on a few families (K=0, folds = length of training reactions, essentially leave-one-out) on families retrained with BM rules from main RMG-Py (commit 0c4fa99b27a058891d029424d91a1b628706baff) and families retrained with BM rules from this PR. 

For most of the families tested, errors remain similar. For 1,3-sigmatropic-rearrangement, looks like there was some redistribution in errors. 

error_plot

Nora-Khalil:
An update: after discussion with @rwest, it was found that the `cross_validate` method in family.py differs from the way it's fits the node when "leaving-one-out". 

To do: will have to keep the cross_validate method and the new BM fitting algorithm proposed in this PR consistent.

mjohnson541 (commented):
These changes look good to me.

rwest:
Thanks for reviewing, @mjohnson541 .
Do you think my last commit, changing the `cross_validate` method to match the `_make_rule` method, makes sense?

mjohnson541:
Yes, those changes look fine to me.

Nora-Khalil (approved):

rmgpy/data/kinetics/family.py
rwest: Just checking  - this is the problem you resolved in a subsequent commit?

Nora-Khalil: yes

Nora-Khalil: unless you are still encountering it with my commit applied...then no

PR #2918, opened 2026-04-07
Cov dep example notebook

sevyharris:
### Motivation or Problem
It would be convenient to have a notebook to quickly test/demo RMG's use of coverage-dependent thermo libraries. 

### Description of Changes
This adds an example notebook for running RMG with coverage dependent thermo. It then compares the results to the baseline of no coverage dependence. It depends on having an example library merged into RMG-database: https://github.com/ReactionMechanismGenerator/RMG-database/pull/714

### Testing
This uses the same data as the coverage-dependent thermo example in Cantera: https://cantera.org/dev/examples/python/thermo/coverage_dependent_surf.html
At the moment, the enthalpy matches, but the entropy does not, so it still needs some debugging.


rwest:
I like the notebook. Does it still need work? At least it needs the RMG-database PR to be merged.
I have rebased it onto the latest #2646.
I also re-ran with 
```
jupyter nbconvert --execute --inplace --to notebook  coverage_dependent_thermo.ipynb --ClearMetadataPreprocessor.enabled=True 
```
I couldn't change the base target for this PR to be the other PR, so this one only has 1 commit instead of 50, but maybe if we just merge it manually, or merge it later, it'll sort itself out.

sevyharris:
I think it's ready now that I've explained the entropy differences, but let me know if you have more comments.

bjkreitz (dismissed):
This looks good to me. I have nothing to add.

bjkreitz (approved):
Everything looks good

PR #2920, opened 2026-04-09
RDKit `GetSSSR` Non-determinism Fix

### LINKED ISSUE #2912: RDKit `GetSSSR` Nondeterminism
### Bug Description

You can see in the regression tests that we are getting non deterministic decompositions of the fused ring systems:

Image

### How To Reproduce

Just making documentation changes in a PR will show this. Could also probably run locally, but I haven't tried this yet.

### Expected Behavior

Both ring decompositions are valid, but we just want to always get the same one.

### Installation Information

Referring to the CI in particular here - if running locally, I'll update this.

### Additional Context

We traded one non-determinism for another in #2796 

RDKit is deterministic in the `GetSSSR` step, but that step is based on the atom identifiers, which are not sorted deterministically when converting from RMG to RDKit (I think). This is some AI-assisted code which I think might solve the problem, and I will (eventually) open a PR to test:

```python
def get_smallest_set_of_smallest_rings(self, symmetrized=False):
        """
        Returns the smallest set of smallest rings (SSSR) as a list of lists of Atom objects, optionally with symmetrization.

        Uses RDKit's built-in ring perception (GetSSSR) by default.
        Note that this is not the same as the Symmetrized SSSR (GetSymmSSSR) with symmetrized=True.
        The symmetrized SSSR is at least as large as the SSSR for a molecule. 
        In certain highly-symmetric cases (e.g. cubane), the symmetrized SSSR 
        can be a bit larger (i.e. the number of symmetrized rings is >= NumBonds-NumAtoms+1).
        It is usually more chemically meaningful, and is less random/arbitrary than the SSSR,
        though RMG uses SSSR for historical reasons.
        """
        if symmetrized:
            if self._symm_sssr is not None:
                return list(self._symm_sssr)
        else:
            if self._sssr is not None:
                return list(self._sssr)

        # RDKit does not support electron
        if self.is_electron():
            return []
        
        from rdkit import Chem
        
        sssr = []
        # Get the symmetric SSSR using RDKit
        rdkit_mol = self.to_rdkit_mol(remove_h=False,
                                      sanitize=False,
                                      return_mapping=False,
                                      save_order=True,
                                      ignore_bond_orders=True)

        if symmetrized:
            ring_info = Chem.GetSymmSSSR(rdkit_mol)
        else:
            # Force deterministic SSSR by canonicalizing the atom numbering first.
            # This prevents RDKit's arbitrary graph traversal from selecting 
            # different sets of equivalent rings across different runs/platforms.
            ranks = list(Chem.CanonicalRankAtoms(rdkit_mol, breakTies=True))
            rank_to_idx = {rank: idx for idx, rank in enumerate(ranks)}
            
            # new_order maps the new atom index to the original atom index
            new_order = [rank_to_idx[i] for i in range(rdkit_mol.GetNumAtoms())]
            
            canonical_mol = Chem.RenumberAtoms(rdkit_mol, new_order)
            canonical_rings = Chem.GetSSSR(canonical_mol)
            
            # Map the resulting ring indices back to the original rdkit_mol numbering
            ring_info = []
            for ring in canonical_rings:
                orig_ring = tuple([new_order[idx] for idx in ring])
                ring_info.append(orig_ring)

        for ring in ring_info:
            atom_ring = [self.atoms[idx] for idx in ring]
            sorted_ring = self.sort_cyclic_vertices(atom_ring)
            sssr.append(sorted_ring)
            
        if symmetrized:
            self._symm_sssr = tuple(sssr)
        else:
            self._sssr = tuple(sssr)
            
        return sssr
```

JacksonBurns:
Goal of this PR is to resolve: https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2912

The RDKit `GetSSSR` algorithm _is_ deterministic for the same input (unlike the previous RMG version) __but__ the process of converting the RMG molecule into RDKit does _not_ guarantee the same ordering of atoms, which results in non-determinism. This PR forces the RDKit molecule to have a consistent ordering, and should (hopefully) give the same results each time.

Easiest way to test this: just need to run the regression tests multiple times on the same reference results (so, on the same day, without any changes to `main`). The results should be different from the `main` results, but in the same way.

### DIFF
--- a/rmgpy/molecule/molecule.py
+++ b/rmgpy/molecule/molecule.py
@@ -2681,31 +2681,21 @@ class Molecule(Graph):
                                       return_mapping=False,
                                       save_order=True,
                                       ignore_bond_orders=True)
-
+        rdkit_mol.UpdatePropertyCache(strict=False)
+        ranks = list(Chem.CanonicalRankAtoms(rdkit_mol, breakTies=True))
+        rank_to_idx = {rank: idx for idx, rank in enumerate(ranks)}
+        new_order = [rank_to_idx[i] for i in range(rdkit_mol.GetNumAtoms())]
+        canonical_mol = Chem.RenumberAtoms(rdkit_mol, new_order)
         if symmetrized:
-            ring_info = Chem.GetSymmSSSR(rdkit_mol)
+            ring_info = Chem.GetSymmSSSR(canonical_mol)
         else:
-            # Force deterministic SSSR by canonicalizing the atom numbering first.
-            # This prevents RDKit's arbitrary graph traversal from selecting 
-            # different sets of equivalent rings across different runs/platforms.
-            ranks = list(Chem.CanonicalRankAtoms(rdkit_mol, breakTies=True))
-            rank_to_idx = {rank: idx for idx, rank in enumerate(ranks)}
-            
-            # new_order maps the new atom index to the original atom index
-            new_order = [rank_to_idx[i] for i in range(rdkit_mol.GetNumAtoms())]
-            
-            canonical_mol = Chem.RenumberAtoms(rdkit_mol, new_order)
-            canonical_mol.UpdatePropertyCache(strict=False)
-            canonical_rings = Chem.GetSSSR(canonical_mol)
-            
-            # Map the resulting ring indices back to the original rdkit_mol numbering
-            ring_info = []
-            for ring in canonical_rings:
-                orig_ring = tuple([new_order[idx] for idx in ring])
-                ring_info.append(orig_ring)
+            ring_info = Chem.GetSSSR(canonical_mol)
 
         for ring in ring_info:
-            atom_ring = [self.atoms[idx] for idx in ring]
+            # Map the new canonical indices back to the original RMG atom indices
+            original_idx_ring = [new_order[idx] for idx in ring]
+            atom_ring = [self.atoms[idx] for idx in original_idx_ring]
+            
             sorted_ring = self.sort_cyclic_vertices(atom_ring)
             sssr.append(sorted_ring)
         if symmetrized:
--- a/rmgpy/molecule/molecule.py
+++ b/rmgpy/molecule/molecule.py
@@ -2685,7 +2685,24 @@ class Molecule(Graph):
         if symmetrized:
             ring_info = Chem.GetSymmSSSR(rdkit_mol)
         else:
-            ring_info = Chem.GetSSSR(rdkit_mol)
+            # Force deterministic SSSR by canonicalizing the atom numbering first.
+            # This prevents RDKit's arbitrary graph traversal from selecting 
+            # different sets of equivalent rings across different runs/platforms.
+            ranks = list(Chem.CanonicalRankAtoms(rdkit_mol, breakTies=True))
+            rank_to_idx = {rank: idx for idx, rank in enumerate(ranks)}
+            
+            # new_order maps the new atom index to the original atom index
+            new_order = [rank_to_idx[i] for i in range(rdkit_mol.GetNumAtoms())]
+            
+            canonical_mol = Chem.RenumberAtoms(rdkit_mol, new_order)
+            canonical_rings = Chem.GetSSSR(canonical_mol)
+            
+            # Map the resulting ring indices back to the original rdkit_mol numbering
+            ring_info = []
+            for ring in canonical_rings:
+                orig_ring = tuple([new_order[idx] for idx in ring])
+                ring_info.append(orig_ring)
+
         for ring in ring_info:
             atom_ring = [self.atoms[idx] for idx in ring]
             sorted_ring = self.sort_cyclic_vertices(atom_ring)
--- a/rmgpy/molecule/molecule.py
+++ b/rmgpy/molecule/molecule.py
@@ -2695,6 +2695,7 @@ class Molecule(Graph):
             new_order = [rank_to_idx[i] for i in range(rdkit_mol.GetNumAtoms())]
             
             canonical_mol = Chem.RenumberAtoms(rdkit_mol, new_order)
+            canonical_mol.UpdatePropertyCache(strict=False)
             canonical_rings = Chem.GetSSSR(canonical_mol)
             
             # Map the resulting ring indices back to the original rdkit_mol numbering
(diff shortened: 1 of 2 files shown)


JacksonBurns:
Ok, the numbers generated by the code in _this_ PR here: https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2920#issuecomment-4240355754

Should hopefully match those from the next comment from the bot about the regression tests... hopefully

JacksonBurns:
Ok, this fixes the non-determinism issue. @rwest please review.

To check the results, I pulled the overlapping species and reactions from the two regression tests on this PR and the regression results from here: https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2931

You can see that we are now consistently coming up with the same ring decompositions. The order of the printed string is different in the estimates, but the estimate is the same.

Species | Main Run 1 | Main Run 2 | PR Run 1 (This PR) | PR Run 2 (This PR)
-- | -- | -- | -- | --
C1=CC2C=CC1=CC2 | ... + ring(1,3-...) + ring(1,4-...) | ... + ring(1,3-...) + ring(1,3-...) | ... + ring(1,4-...) + ring(1,3-...) | ... + ring(1,3-...) + ring(1,4-...)
[CH]=CC1C=C2C=CC1C=C2 | ... + ring(1,3-...) + ring(1,4-...) + rad... | ... + ring(1,3-...) + ring(1,3-...) + rad... | ... + ring(1,4-...) + ring(1,3-...) + rad... | ... + ring(1,3-...) + ring(1,4-...) + rad...
C=CC1C=C2C=CC1C=C2 | ... + ring(1,3-...) + ring(1,4-...) | ... + ring(1,3-...) + ring(1,3-...) | ... + ring(1,4-...) + ring(1,3-...) | ... + ring(1,3-...) + ring(1,4-...)

Activation energy estimations are fixed as well:

Reaction | Main Run 1 | Main Run 2 | PR Run 1 (This PR) | PR Run 2 (This PR)
-- | -- | -- | -- | --
[c]1ccccc1(3) + C1=CC2C=C[C]1C=C2(49) <=> C1#CC=CC=C1(10) + C1=CC2C=CC1=CC2(83) | 4.743 | 6.009 | 4.743 | 4.743
[CH]1C2=CC=CC12(8) + C1=CC2C=C[C]1C=C2(49) <=> C1=CC2=CC2C=1(26) + C1=CC2C=CC1=CC2(83) | 35.170 | 38.554 | 35.170 | 35.170
C1=CC2C=C[C]1C=C2(49) + [CH]1C2=CC=CC1C=C2(48) <=> C1=CC2=CC(=C1)C=C2(60) + C1=CC2C=CC1=CC2(83) | 86.692 | 90.142 | 86.692 | 86.692
C1=CC2C=C[C]1C=C2(49) + [CH]1C2=CC=CC1C=C2(48) <=> C1=CC2=CC(C=1)C=C2(64) + C1=CC2C=CC1=CC2(83) | 16.860 | 20.310 | 16.860 | 16.860
C1=CC2C=C[C]1C=C2(49) + C1=CC2C=C[C]1C=C2(49) <=> C1=CC2C=CC=1C=C2(79) + C1=CC2C=CC1=CC2(83) | 33.226 | 2.336 | 1.036 | 1.036
C1=CC2C=C[C]1C=C2(49) + C1=CC2C=C[C]1C=C2(49) <=> C1=CC2=CC=C1C=C2(81) + C1=CC2C=CC1=CC2(83) | 96.732 | 100.182 | 96.732 | 96.732

As are the kinetics estimates (this table is $log_{10}k(1000K)$:

Reaction | Main Run 1 | Main Run 2 | PR Run 1 (This PR) | PR Run 2 (This PR)
-- | -- | -- | -- | --
[c]1ccccc1(3) + C1=CC2C=C[C]1C=C2(49) <=> C1#CC=CC=C1(10) + C1=CC2C=CC1=CC2(83) | 4.39 | 4.67 | 4.67 | 4.67
[CH]1C2=CC=CC12(8) + C1=CC2C=C[C]1C=C2(49) <=> C1=CC2=CC2C=1(26) + C1=CC2C=CC1=CC2(83) | -2.29 | -1.55 | -1.55 | -1.55
C1=CC2C=C[C]1C=C2(49) + [CH]1C2=CC=CC1C=C2(48) <=> C1=CC2=CC(=C1)C=C2(60) + C1=CC2C=CC1=CC2(83) | -13.15 | -12.40 | -12.40 | -12.40
C1=CC2C=C[C]1C=C2(49) + [CH]1C2=CC=CC1C=C2(48) <=> C1=CC2=CC(C=1)C=C2(64) + C1=CC2C=CC1=CC2(83) | 1.70 | 2.46 | 2.46 | 2.46
C1=CC2C=C[C]1C=C2(49) + C1=CC2C=C[C]1C=C2(49) <=> C1=CC2C=CC=1C=C2(79) + C1=CC2C=CC1=CC2(83) | 6.11 | -0.64 | 6.39 | 6.39
C1=CC2C=C[C]1C=C2(49) + C1=CC2C=C[C]1C=C2(49) <=> C1=CC2=CC=C1C=C2(81) + C1=CC2C=CC1=CC2(83) | -15.35 | -14.59 | -14.59 | -14.59

jonwzheng (commented):
This looks good to me in principle, based on the test results and code changes. 

So we understand for the future: what specifically is causing the RMG-to-RDKit pipeline to be non-deterministic? I'm surprised to see there's non-determinism even in just the conversion process - maybe we should address that root cause so this doesn't crop up in the future

JacksonBurns:
Thanks @jonwzheng ! 

The nondeterminism in RMG is because of the way we deal with atoms. In RMG, when we try and sort a list of atoms, atoms of the same element will look identical (e.g., just multiple instances of the 'C' atom), leading to non- determinism. This fixes that because atoms are sorted based on their local environment, which can always be done deterministically (for a reason is still admit that I don't fully understand, yet).

This is only a pain point when interchanging with rdkit on order specific functions, which is why things like SMILES conversion are fine. 

The comprehensive fix would be to rewrite RMG's molecule class to store the structure as an rdkit molecule rather than a custom,  rmg molecule. But for now, when we only need a few things from rdkit, I think we can stick with these little patches.

jonwzheng (approved):

PR #2929, opened 2026-04-17
Update charge for Val5, and N

kirkbadger18:
I am creating four families of reactions for manipulating bonds with charge separation across them. These are the four families I am adding in RMG-database PR [#719](https://github.com/ReactionMechanismGenerator/RMG-database/pull/719):

famfig

This requires a Val5 atom type (but N is the current intended case) having a positive charge, and another R!H having a negative charge. I suppose Val6 could have this positive charge too, but doesn't seem chemically relevant to me. As part of the reaction template the Val5 must decrement charge, and the negative charge atom must increment charge. Currently increment and decrement charge are implemented for R!H, but not Val5, or N (Or any of the more specific types). Here I have implemented increment and decrement charge for Val5 and N. 

Are there any test cases I should be thinking about, and implementing?


sevyharris:
Hey @kirkbadger18 do you have any example reactions that match the Surface_Abstraction_Charge_Separation family? (or really any of the new families) I didn't see any in the training folder and it would be helpful to have a few smiles/adjacency list examples for testing/debugging purposes.

As for the specific tests to consider, you could try adding one of these new families as a family test like the example here: https://github.com/ReactionMechanismGenerator/RMG-Py/blob/9a5490a51f7f7083be5e73445ec2bbb03a9d2ba2/test/rmgpy/data/kinetics/familyTest.py#L588 

I think running RMG on some big, complicated nitrogen systems is also a good way to test this because it's likely to come up with weird examples of families/different parts of RMG not playing nicely together.

sevyharris:
Also, I tried running a NH3 combustion mechanism on this branch and the database one https://github.com/ReactionMechanismGenerator/RMG-database/pull/719 and got this error:

[input.txt](https://github.com/user-attachments/files/26879600/input.txt)


```
Error: Error when loading reaction family '/home/moon/rmg/RMG-database/input/kinetics/families/Surface_Abstraction_Charge_Separation'
  File "/home/moon/rmg/RMG-Py/rmg.py", line 4, in <module>
    __main__.main()
  File "/home/moon/rmg/RMG-Py/rmgpy/__main__.py", line 102, in main
    rmg.execute(**kwargs)
  File "/home/moon/rmg/RMG-Py/rmgpy/rmg/main.py", line 802, in execute
    requires_rms = self.initialize(**kwargs)
  File "/home/moon/rmg/RMG-Py/rmgpy/rmg/main.py", line 569, in initialize
    self.load_database()
  File "/home/moon/rmg/RMG-Py/rmgpy/rmg/main.py", line 402, in load_database
    self.database.load(
  File "/home/moon/rmg/RMG-Py/rmgpy/data/rmg.py", line 100, in load
    self.load_kinetics(os.path.join(path, 'kinetics'),
  File "/home/moon/rmg/RMG-Py/rmgpy/data/rmg.py", line 171, in load_kinetics
    self.kinetics.load(path,
  File "/home/moon/rmg/RMG-Py/rmgpy/data/kinetics/database.py", line 125, in load
    self.load_families(os.path.join(path, 'families'), families, depositories)
  File "/home/moon/rmg/RMG-Py/rmgpy/data/kinetics/database.py", line 233, in load_families
    family.load(family_path, self.local_context, self.global_context, depository_labels=depositories)
  File "/home/moon/rmg/RMG-Py/rmgpy/data/kinetics/family.py", line 689, in load
    self.forward_template.products = self.generate_product_template(self.forward_template.reactants)
  File "/home/moon/rmg/RMG-Py/rmgpy/data/kinetics/family.py", line 1042, in generate_product_template
    product_structure = self.apply_recipe(reactant_structure, forward=True, unique=False)
  File "/home/moon/rmg/RMG-Py/rmgpy/data/kinetics/family.py", line 1368, in apply_recipe
    self.forward_recipe.apply_forward(reactant_structure, unique)
  File "/home/moon/rmg/RMG-Py/rmgpy/data/kinetics/family.py", line 530, in apply_forward
    return self._apply(struct, True, unique)
  File "/home/moon/rmg/RMG-Py/rmgpy/data/kinetics/family.py", line 498, in _apply
    atom.apply_action(['LOSE_CHARGE', label, 1])
  File "rmgpy/molecule/group.py", line 465, in rmgpy.molecule.group.GroupAtom.apply_action
  File "rmgpy/molecule/group.py", line 486, in rmgpy.molecule.group.GroupAtom.apply_action
  File "rmgpy/molecule/group.py", line 397, in rmgpy.molecule.group.GroupAtom._lose_charge
rmgpy.exceptions.ActionError: Unable to update GroupAtom due to LOSE_CHARGE action: Unknown atom type produced from set "[<AtomType "Val5">]".
```

If you give me an example reaction, I can help debug it.

kirkbadger18:
@sevyharris This mechanism runs fine for me, I am guessing that you forgot to checkout the rmgpy branch here. Here is what I get:
[ammonia.zip](https://github.com/user-attachments/files/26982861/ammonia.zip)

I modified the convergence thresholds to make sure some of these new reactions showed up. 

Also, I will look into the test you mentioned and try to emulate it.

sevyharris:
> @sevyharris This mechanism runs fine for me, I am guessing that you forgot to checkout the rmgpy branch here. Here is what I get: [ammonia.zip](https://github.com/user-attachments/files/26982861/ammonia.zip)
> 
> I modified the convergence thresholds to make sure some of these new reactions showed up.
> 
> Also, I will look into the test you mentioned and try to emulate it.

I forgot to build! I didn't realize atomtype is cythonized because it's .py not .pyx. Anyways, my simple example runs fine now. I'll try running it on something big and weird.

kirkbadger18:
Ahh, that always gets me too. I have added a test case for one of these reaction families. Should I add one for each of the 4 families if this test looks good?

kirkbadger18:
Maybe another good test would be to make sure given XNO and XO, RMG is able to identify those as products of a surface dissociation charge separation reaction. @sevyharris do you know if there are any tests I could emulate to do this or if this would be easy to figure out?

sevyharris:
> @sevyharris This mechanism runs fine for me, I am guessing that you forgot to checkout the rmgpy branch here. Here is what I get: [ammonia.zip](https://github.com/user-attachments/files/26982861/ammonia.zip)
> 
> I modified the convergence thresholds to make sure some of these new reactions showed up.
> 
> Also, I will look into the test you mentioned and try to emulate it.

I think one is probably good. Adding extra families means you have to load more from the database, which can unnecessarily lengthen the unit tests. There is a downside to being too thorough.

sevyharris:
> Maybe another good test would be to make sure given XNO and XO, RMG is able to identify those as products of a surface dissociation charge separation reaction. @sevyharris do you know if there are any tests I could emulate to do this or if this would be easy to figure out?

I think if you follow the procedure [here](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/01d447cad11018fd11456360c9feb686d43484ec/test/rmgpy/data/kinetics/familyTest.py#L735-L750), that'll get you what you want:

- start with molecules that don't have atom label 
- try to apply the labels with `family.add_atom_labels_for_reaction(labeled_rxn)`
- apply the recipe with `family.apply_recipe([m.molecule[0] for m in labeled_rxn.reactants])`
- check on the products or reactants depending on the direction you're going

I think `add_atom_labels_for_reaction` does most of the heavy lifting for identifying species as part of that family.

kirkbadger18:
@sevyharris I have added a test now which checks that given unlabeled reactants that they are able to be labeled and reacted to form the correct products. I also test that the reverse template can be applied properly.

sevyharris:
Sorry for the delay on my review. The tests look good.

I tried running a weird NH3 + CH4 + O2 mechanism and it didn't crash, but I was only able to generate a single charge separation reaction out of 1000+ reactions. (and the ammonia demo only got 2 others). I'll try a few more variations to make sure I hit all four families with my demos.

I'm seeing a bunch of `Error: Could not update atomtypes for this molecule:` errors for charge separated species in the RMG log and I'm not sure what's going on there.

For example:
```
Error: Could not update atomtypes for this molecule:
multiplicity 1
1 *1 O u1 p0 c+2 {2,T}
2 *1 N u1 p0 c0 {1,T} {3,S}
3    H u0 p0 c0 {2,S}

Error: Could not update atomtypes for this molecule:
1 *1 O u0 p1 c+1 {2,T}
2 *1 N u0 p1 c-1 {1,T} {3,S}
3    H u0 p0 c0 {2,S}
```

It isn't crashing, but I'd like to look at why this is happening to see if it matters.

Anyway, I think this PR's almost ready and I'll try to finish my review tomorrow.

sevyharris:
Turns out the AtomType error I was seeing is an expected part of RMG trying out different families. I don't love seeing "Errors" logged as part of the normal RMG operation, but that has nothing to do with this PR.

I've run examples with all reaction families, and RMG is able to reconstruct the atom labeling for each family, however, I also came across this weird bug where RMG generates a mechanism with this reaction listed as a Surface_Dissociation_Charge_Single 
surface_dissociation_charge_single


```
! Reaction index: Chemkin #961; RMG #2824
! Template reaction: Surface_Dissociation_Charge_Single
! Flux pairs: C2H4NX2(154), HX(9); C2H4NX2(154), C2H3NX2(650); 
! Exact match found for rate rule [Combined;VacantSite]
! Euclidian distance = 0
! family: Surface_Dissociation_Charge_Single
X(4)+C2H4NX2(154)<=>HX(9)+C2H3NX2(650)              9.880000e+21 0.000     4.508    
DUPLICATE
```

But then it can't apply the atom labeling when I try to construct the reaction from scratch:
```
X = rmgpy.species.Species(smiles="*")

C2H4NX2 = rmgpy.species.Species().from_adjacency_list("""
1 N u0 p1 c0 {3,S} {7,S} {9,S}
2 C u0 p0 c0 {3,S} {4,S} {5,S} {6,S}
3 C u0 p0 c0 {1,S} {2,S} {8,D}
4 H u0 p0 c0 {2,S}
5 H u0 p0 c0 {2,S}
6 H u0 p0 c0 {2,S}
7 H u0 p0 c0 {1,S}
8 X u0 p0 c0 {3,D}
9 X u0 p0 c0 {1,S}""")


HX = rmgpy.species.Species(smiles='[H]*')
C2H3NX2 = rmgpy.species.Species().from_adjacency_list("""
1 N u0 p1 c0 {3,D} {8,S}
2 C u0 p0 c0 {3,S} {4,S} {5,S} {6,S}
3 C u0 p0 c0 {1,D} {2,S} {7,S}
4 H u0 p0 c0 {2,S}
5 H u0 p0 c0 {2,S}
6 H u0 p0 c0 {2,S}
7 X u0 p0 c0 {3,S}
8 X u0 p0 c0 {1,S}""")


# C2H3NX2.generate_resonance_structures()
# C2H4NX2.generate_resonance_structures()

my_reaction = rmgpy.reaction.Reaction()
my_reaction.reactants = [
    X,
    C2H4NX2,
]
my_reaction.products = [
    HX,
    C2H3NX2,
]
family = 'Surface_Dissociation_Charge_Single'
display(my_reaction)
database.kinetics.families[family].add_atom_labels_for_reaction(my_reaction)
```

[input.txt](https://github.com/user-attachments/files/27186318/input.txt)
[chem_annotated-surface.inp.txt](https://github.com/user-attachments/files/27186316/chem_annotated-surface.inp.txt)
[species_dictionary.txt](https://github.com/user-attachments/files/27186353/species_dictionary.txt)


I'll approve once I get to the bottom of this.

sevyharris:
### The Problem
The actual Surface_Dissociation_Charge_Single reaction RMG finds is:
actual_surface_dissociation_charge_single

But it's getting recorded as:
surface_dissociation_charge_single

Because when RMG goes to check if it already has the species in the model, the isomorphism check sees the charge separated molecule `[C-]([NH+]=[Pt])([Pt])C` as equivalent to the non-charge-separated version `CC(=[Pt])N[Pt]`

https://github.com/ReactionMechanismGenerator/RMG-Py/blob/3b4b7b1a20eefd20996cb455a67015ee338bb638/rmgpy/rmg/model.py#L286 

**RMG adds the kinetics, but fails to add the charge-separated species to the core to go on to react with other things.**


### How to proceed
I'm not entirely sure what the ideal behavior is here. It feels wrong that the charge separated species doesn't get its own definition if it gets its own family. Also, as noted above, you can't apply the atom labels with the family/adjacency lists RMG records. At the same time, RMG has been treating molecules with different electron configurations as identical for a long time and I would be hesitant to change the `strict=False` behavior in isomorphism check to account for charge because speed is so important there.

It looks like this is only a problem when RMG finds a charge-separated version of a species that's already in the core, or conversely if it finds the charge-separated species first it won't add the non-separated version. The danger here is that RMG will miss a bunch of reactions because it only keeps the first version of the species it finds.

So @kirkbadger18, if you feel strongly that charge separated species need their own definition in the species_dictionary, then I'd maybe add a check that requires strict isomorphism for molecules with charge. Something like this in model.py's check_for_existing_species:
```
    def check_for_existing_species(self, molecule):
        """
        Check to see if an existing species contains the same
        :class:`molecule.Molecule` as `molecule`. Comparison is done using
        isomorphism without consideration of electrons. Therefore, resonance
        structures of a species will all match each other.

        Returns the matched species if found and `None` otherwise.
        """

        # First check cache and return if species is found
        strictness = False
        if molecule.has_charge():  # Treat charged species with strict isomorphism to avoid matching charge-separated species with their neutral versions
            strictness = True
        for i, spec in enumerate(self.species_cache):
            if spec is not None and spec.is_isomorphic(molecule, strict=strictness):
                self.species_cache.pop(i)
                self.species_cache.insert(0, spec)
                return spec

        # If not found in cache, check all species with matching formula
        formula = molecule.get_formula()
        try:
            species_list = self.species_dict[formula]
        except KeyError:
            pass
        else:
            for spec in species_list:
                if spec.is_isomorphic(molecule, strict=strictness):
                    self.species_cache.pop()
                    self.species_cache.insert(0, spec)
                    return spec

        # At this point we can conclude that the species is new
        return None
```

But I think the updated Val5 charge is probably fine to merge in as is.

sevyharris (approved):

kirkbadger18:
@sevyharris Thank you for looking into this! I think I also saw this happening a while back. I was under the impression that RMG searched for resonant structures and searched for templates matching each one. If that is the case only one species needs to be in the core as they all will share the same thermochemistry. If that is not the case, then yes this could be an issue. Even if it is working, it is a little stange to see the template matching where at first glance it looks wrong before considering resonance. 
@bjkreitz What do you think?

sevyharris:
> @sevyharris Thank you for looking into this! I think I also saw this happening a while back. I was under the impression that RMG searched for resonant structures and searched for templates matching each one. If that is the case only one species needs to be in the core as they all will share the same thermochemistry. If that is not the case, then yes this could be an issue. Even if it is working, it is a little stange to see the template matching where at first glance it looks wrong before considering resonance. @bjkreitz What do you think?

Just wanted to clarify that this is an example where the charge-separated species is NOT a resonance structure of the non-separated version. I don't know if it's supposed to be, but I think you're right that RMG does usually search for resonant structures when trying to match templates. The non-resonance would explain why I wasn't able to recreate the reaction from the non-charge-separated species.

kirkbadger18:
These two species look like resonant structures to me:
image
image
Although though there is no real reason to use the charge separated version of this species in my mind.

bjkreitz:
RMG only looks for resonance structures if we tell it which resonance structures to explore. So it would be necessary to add new pathways and templates for species with charge separation. If it finds resonance structures, it will apply all possible templates to all possible resonance structures to find the reactions.

sevyharris:
Can I press merge on this PR? It sounds like the problem I found isn't much of an issue and can be dealt with down the line with future updates to the resonance structures.

kirkbadger18:
@sevyharris I have rebased RMG-database PR [#719: New surface nitrogen families](https://github.com/ReactionMechanismGenerator/RMG-database/pull/719), and have run make test-all locally using this branch here along with the branch corresponding to the New surface nitrogen families PR, and I get:
image

I think there are some issues building RMG with Github actions at the moment, so merging this may not immediately get RMG-database PR [#719: New surface nitrogen families] to pass.

sevyharris:
Hey @kirkbadger18, Jackson and I merged a partial fix to the regression tests like an hour ago.

It doesn't get all the way to sending the email notification, but you can at least check the Github Action logs to see that it's running them to completion. I'll rebase this one more time to include that update, and if it runs the regression tests, I'll press merge.

sevyharris:
It's running all the regression tests to completion. The regression test logs look comparable to the version we just merged into main for the regression test partial fix. I'm gonna say this is good enough. Hopefully we'll have the regression tests all the way fixed soon 🤞

sevyharris:
My mistake, I thought automerge would merge this despite the regression test failure. I may or may not have permission to force push to main, but I've never done that before, and we now have a [fix for the regression tests](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2944) anyways. That should get merged in a couple hours, and then we can do this one after that.

PR #2939, opened 2026-05-01
Add SIDT adsorption correction predictors

mjohnson541:
This enables use of SIDTs stored in the RMG-database to predict adsorption corrections. The SIDTs are added in https://github.com/ReactionMechanismGenerator/RMG-database/pull/723 trained on Pt111 adsorption corrections data that @kirkbadger18 sent me. 

Without a framework for computing independent errors on the current RMG adsorption groups it is difficult to definitively measure how much of an improvement this should be. However, the leave-one-out errors on the SIDTs are similar to the training errors on the RMG adsorption groups, so this should almost certainly be a significant improvement.

### DIFF
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -860,6 +860,8 @@ class ThermoDatabase(object):
         self.libraries = {}
         self.surface = {}
         self.groups = {}
+        self.sidts = {}
+        self.sidt_taggings_and_decompositions = {}
         self.adsorption_groups = "adsorptionPt111"
         self.library_order = []
         self.local_context = {
@@ -886,6 +888,7 @@ class ThermoDatabase(object):
             'depository': self.depository,
             'libraries': self.libraries,
             'groups': self.groups,
+            'sidts': self.sidts,
             'library_order': self.library_order,
             'surface' : self.surface,
         }
@@ -898,6 +901,7 @@ class ThermoDatabase(object):
         self.depository = d['depository']
         self.libraries = d['libraries']
         self.groups = d['groups']
+        self.sidts = d['sidts']
         self.library_order = d['library_order']
         self.surface = d['surface']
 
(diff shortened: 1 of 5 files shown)


rmgpy/data/thermo.py
```diff
@@ -40,6 +40,9 @@
 
 import numpy as np
 
+from pysidt import read_nodes,MultiTargetSingleEvalSubgraphIsomorphicDecisionTree
```
JacksonBurns:
```suggestion
from pysidt import read_nodes, MultiTargetSingleEvalSubgraphIsomorphicDecisionTree
```

JacksonBurns (commented):
Unit tests would be a good addition - something to check if we can load and interact with the SIDTs. Form of the implementation looks good, though.

mjohnson541:
I've added some database tests to test that the SIDTs are loaded and that they work properly. They look a little bit out of place, I chose the test structure I did because each SIDT can have different tagging/decomposition and test molecules, but I'm open to better ideas if anyone has any.

JacksonBurns (dismissed):
These tests are a great starting point, and enough for the scope of this initial PR adding the feature. I will update the branch and set to merge

rwest (dismissed):
Not actually tried it, but read through the diff and it looks ok. Will you update user instructions at some point?

kirkbadger18:
I have run RMG for a stoichiometric exhaust mix containing: carbon monoxide, ethane, ethylene, methane, water, carbon dioxide, hydrogen, and nitric oxide. From an RMG-Py prospective this is working and I approve this! You have to add a line to the input file to get it to work, so I think making this clear in documentation would be good. Here is the mechanism I ran with output:
[input.py](https://github.com/user-attachments/files/27529780/input.py)
[output.html](https://github.com/user-attachments/files/27529782/output.html)

kirkbadger18 (dismissed):
Looks good, I approve! Look to my previous comment for more detail.

Issue #2936, opened 2026-04-28
External bond extensions in tree generation process causing split violations


Nora-Khalil:
### Motivation or Problem
Currently retraining the R_Recombination kinetic tree. Tree generation has been running for 8 days and is stuck in a loop trying to recover from a split violation at a specific node. 

The issue is a node with 2 reactions: 
CH2F + CH2F <=> C2H4F2
<img width="234" height="22" alt="Screenshot 2026-04-28 at 1 17 45 PM" src="https://github.com/user-attachments/assets/1e8c2a31-9fc9-4efb-83e7-117ba24c2c62" />

CF + CF <=> C2F2
<img width="209" height="22" alt="Screenshot 2026-04-28 at 1 18 13 PM" src="https://github.com/user-attachments/assets/8e5b168e-a9b0-4cb5-836d-e6ca2c9e8d4e" />

The node group definition is:
```
1 * C   u1 r0 {3,[S,B,D,T,Q]}
2 * R   u1 {4,[S,B,D,T,Q]}
3   F   ux {1,[S,B,D,T,Q]}
4   R!H ux {2,[S,B,D,T,Q]}
```

The algorithm that extends the node is unable to suggest a node extension that both 1) splits the reactions of the node and 2) fits at least one of the reactions in the node (i.e the criteria evaluated by `family.eval_ext()`) 

We could look at this node and manually suggest a child node that does split the reactions here, i.e.: 
```
1 * C   u1 r0 {3,[S,B,D,T,Q]}
2 * R   u1 {4,[S,B,D,T,Q]} {5,[S,B,D,T,Q]}
3   F   ux {1,[S,B,D,T,Q]}
4   R!H ux {2,[S,B,D,T,Q]}
5   H   u0 {2,[S,B,D,T,Q]} 
```

In fact, I have created a script that recreates this node causing the split violation and manually suggests the extension above, and the node is successfully extended without a split violation. 

However, RMG does not propose an external bond extension like this because it looks like `specify_external_new_bond_extensions` in RMG-Py/rmgpy/molecule/group.py will only ever "generate extensions for the creation of a bond (of undefined order) between an atom and a new atom that is not H".

What was the motivation for only suggesting external bond extensions to R!H atoms? If bond extensions could include H atoms, this split violation error would be resolved. 

### Testing
To test, I did the following: 

1. Load an instance of the database
2. Clean the R_Recomb tree so the tree only has the Root node
[...]

PR #2941, opened 2026-05-05
iter_item_cap passed to make_tree_nodes even if cascade algo isn't used

Nora-Khalil:
### Motivation or Problem
I have added new halocarbon/PFAS training reactions to the R_Recombination kinetic family and am currently retraining. 

I was previously running into a split violation during tree generation (described here in https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2936). I have gotten past this split violation road block via a quick [fix](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/leaf_node_max/rmgpy/data/kinetics/family.py#L3082) that allows for a node to be a leaf node if it cannot be further extended and if the number of training reactions in that node do not exceed a user specified `leaf_node_max` (default set to 2). Now tree generation can proceed rather than stay stuck in the infinite loop that caught the split violation. 

I'm now facing another issue where tree generation is taking forever because `grp.get_extensions` has found bond formation extensions that match all the training reactions in the node and are flagged to be further expanded, which explodes the amount of extensions to be evaluated. I see that this issue was addressed in commit https://github.com/ReactionMechanismGenerator/RMG-Py/commit/b6ae40d0bb5d9de6b5e58342fca4e784de071b95, where an iter_item cap will force it to give up the split if it runs into this issue **but only when the cascade algorithm is used**. 

The R_Recomb family now has 370 training reactions (my new training reactions included), so I wouldn't have thought to decrease max_batch size (usually 800) to force this family retraining through the cascade algorithm. Can we just pass in the `extension_iter_item_cap` parameter even when not using cascade algorithm? 

### Description of Changes
Passing in `extension_iter_item_cap` to `make_tree_nodes` regardless of whether we need to use batches or not in retraining. 

### Testing
Using `generate_tree`, retrained the R_Recombination tree in 85 seconds with this fix (and following the leaf node [fix](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/leaf_node_max/rmgpy/data/kinetics/family.py#L3082)). 

### Reviewer Tips
Something for the reviewer (mostly @mjohnson541 ) to think about: It's possible that maybe this issue only comes up because of the fact that I'm bypassing the split violation with the leaf_node_max fix... so perhaps these two fixes should be paired up? or both discarded if electron pair extensions will be added into RMG soon (and thus no need for the leaf_node_max fix)? 

Thank you!

### DIFF
--- a/rmgpy/data/kinetics/family.py
+++ b/rmgpy/data/kinetics/family.py
@@ -3294,7 +3294,7 @@ class KineticsFamily(Database):
             template_rxn_map = self.get_reaction_matches(rxns=rxns, thermo_database=thermo_database, remove_degeneracy=True,
                                                          fix_labels=True, exact_matches_only=True, get_reverse=True)
             self.make_tree_nodes(template_rxn_map=template_rxn_map, obj=obj, T=T, nprocs=nprocs - 1, depth=0,
-                                 min_splitable_entry_num=min_splitable_entry_num, min_rxns_to_spawn=min_rxns_to_spawn,extension_iter_max=extension_iter_max)
+                                 min_splitable_entry_num=min_splitable_entry_num, min_rxns_to_spawn=min_rxns_to_spawn,extension_iter_max=extension_iter_max, extension_iter_item_cap=extension_iter_item_cap)
         else:
             def rxnkey(rxn):
                 c = 0


mjohnson541 (approved):
Looks good to me!

rwest (approved):

PR #2944, opened 2026-05-06
Fix regression tests attempt 2: Work around failed conda activation in regression test

sevyharris:
### Motivation or Problem
The regression tests are still failing, possibly because the rmg_env_without_rms fails to activate silently. 

### Description of Changes
This uses the conda run command to force it to use a particular conda env

TODO: If it turns out the conda activate command is failing, then it probably makes sense to delete it.

### Testing
We'll see if this fixes the tests.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -276,6 +276,9 @@ jobs:
           REFERENCE: stable_regression_results
         run: |
           conda activate rmg_env_without_rms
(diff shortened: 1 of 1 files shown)


JacksonBurns:
Second time is the charm (hopefully) thanks @sevyharris !

sevyharris:
It looks like that last one worked. I'm going to reformat it as a single commit with a few more comments

JacksonBurns:
CI passed, merging, thanks a million @sevyharris

JacksonBurns (approved):

PR #2949, opened 2026-05-13
Fix thermo tests

bjkreitz:
### Motivation or Problem
The recent updates to the adsorption correction changed the naming convention of the species from e.g. C=*R2 to C=XR2. We replaced * with X because working with * in Linux can cause issues. This change broke the unit tests for the adsorption corrections in `TestThermoDatabase.py` since it expected the old notation. 

### Description of Changes
I have updated the `test_parse_thermo_comments` and `test_adsorbate_thermo_raises_error` to work with the new notation.

### DIFF
--- a/test/rmgpy/data/thermoTest.py
+++ b/test/rmgpy/data/thermoTest.py
@@ -1249,26 +1249,26 @@ multiplicity 2
         )
         groups = self.database.groups["adsorptionPt111"].entries
         # save a few things we're about to change
-        _nstarparent = groups["N*"].parent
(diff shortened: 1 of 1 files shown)


JacksonBurns (approved):
Thanks @bjkreitz ! Was wondering why the CI failed this morning, appreciate the quick fix!

sevyharris (approved):
Thanks for fixing this! I didn't realize or forgot that the database CI doesn't run RMG-Py's unit tests and probably should've checked before merging.

PR #2955, opened 2026-05-15
Fixes to Chemkin file writing and reading

### LINKED ISSUE #2961: wrong spacing for chemkin spacing default temperature range
### Bug Description
The spacing for the default thermo temperature range in a chemkin file is different for the surface one than for the gas one:

#### Surface code
https://github.com/ReactionMechanismGenerator/RMG-Py/blob/3b6dd7988d8d22016293c6e7d4e6a66ae36c9f92/rmgpy/chemkin.pyx#L2201

#### Gas code
https://github.com/ReactionMechanismGenerator/RMG-Py/blob/3b6dd7988d8d22016293c6e7d4e6a66ae36c9f92/rmgpy/chemkin.pyx#L2134

This prevents the default temperature ranges from getting read in.

### How To Reproduce
Generate an RMG surface mechanism and then try to read in the chemkin file.

rwest:
### Motivation or Problem
Loading any CHEMKIN file written by save_chemkin_surface_file (or by a pre-2016 save_chemkin_file) produces a noisy "Thermo file has no default temperature ranges" warning during parsing. The THERM ALL default-range header was written one column too wide, so the strict fixed-width parser failed on every load. A similar fix was made in 37dbd53ae4b329aa2357c1d1b3657971a982a9d3

**Update:** I later found myself fixing how Chemkin writes species that have isotopes (not that I ever use this). That commit seemed more relevant to this PR, so I added it here.

### Description of Changes
- save_chemkin_surface_file now writes the default-range header with 3 leading spaces, matching the cols 1-10 / 11-20 / 21-30 spec
- read_thermo_block slices widened from 9 chars to 10 ([0:9] → [0:10], etc.) to match the documented column layout. The old slices silently truncated column 10/20/30 — e.g. `'      1000'` would have been misread as 100. 
- 26 committed fixtures regenerated to use the corrected 3-space header.
- New unit test (see below)
- Element class now correctly use the symbol for .chemkin_name
- Species using isotopes now have correct composition in chemkin files.

### Testing
- New `test_read_thermo_block_temperature_header_columns` in chemkinTest.py parses a header that packs values to both edges of each 10-column slot. Confirmed it fails against the old slices and passes with the fix.
- test/rmgpy/tools/mergemodelsTest.py::test_merge_different_models (which originally surfaced the warning) now runs without the warning.
- new test for isotopic species

### Reviewer Tips
Cython rebuild required (make build) since chemkin.pyx changed


rmgpy/chemkin.pyx
```diff
@@ -1588,8 +1588,13 @@ def write_thermo_entry(species, element_counts=None, verbose=True):
...
 
     # Determine the number of each type of element in the molecule
+    # Need to use the element's chemkin name, not the element symbol, because of isotopes.
+    # so we can't just use molecule[0].get_element_count().
     if element_counts is None:
-        element_counts = get_element_count(species.molecule[0])
+        element_counts = {}
+        for atom in species.molecule[0].atoms:
+            element = atom.element.chemkin_name
+            element_counts[element] = element_counts.get(element, 0) + 1
```
JacksonBurns: Add `from collections import defaultdict` to the top of the file and then you can make this a bit easier to read:

```suggestion
        element_counts = defaultdict(int)
        for atom in species.molecule[0].atoms:
            element_counts[atom.element.chemkin_name] += 1
```

test/rmgpy/chemkinTest.py
```diff
@@ -144,6 +146,26 @@ def test_read_thermo_entry_no_temperature_range(self):
...
+        assert len(header) == 30
+        f = io.StringIO("THERM ALL\n" + header + "\nEND\n")
+
+        read_thermo_block(f, species_dict={})
+
+        mock_logging.info.assert_any_call(
+            "Thermo file has default temperature range 150.0 to 1000.0 and 1000.0 to 9999.0"
+        )
+        for call in mock_logging.warning.call_args_list:
+            assert "badly formatted" not in call.args[0]
```
JacksonBurns, line 167: Can you add the opposite test where you expect `"badly formatted"` to be logged?

JacksonBurns (changes requested):

rmgpy/chemkin.pyx
rwest: yes but... with a likely small performance penalty (once cythonized). I was about to push another optimization, but this code is tweaked on another PR so might just drop it and minimize changes here to avoid conflicts


Codex local compiled microbenchmarks in the RMG env:

6 atoms:
dict.get       293 nsec/call
defaultdict    630 nsec/call

test/rmgpy/chemkinTest.py
rwest: done

JacksonBurns (approved):

PR #2958, opened 2026-05-19
Optimizations (largely Cython)

rwest:
### Motivation or Problem
Trying to speed things up.

### Description of Changes
While working on the vdW PR, I found myself tweaking some Cython declarations.
That then led to broader changes, which I then branched off, into this PR.
I then asked Claude to help, armed with some profiling data and Cython annotations.

### Co-Pilot's autogenerated summary of the PR:
(lightly edited, to delete some of the worst BS)

This pull request is focused on standardizing the use of `edges` and `vertices` instead of `bonds` and `atoms` in molecule and group classes. It also improves performance and code clarity in some filtration and resonance-related functions. The most significant changes are grouped below.

### API Standardization and Refactoring

[...]

PR #2960, opened 2026-05-21
Fix cython declarations and some optimizations in molecule.py

rwest:
### Motivation or Problem
I noticed more Cython inefficiencies in molecule.py. Once I started fixing, I got carried away.
I thought I'd just add these commits to #2958, but I was just too late and it's already been merged.
So here's a new PR.

### Description of Changes
Mostly just adding or correcting Cython declarations.
Some of this guided by the annotated HTML files that you can make Cython generate, that show how the code is being translated into C.
For a couple of methods that I think we call a lot, I expanded into explicit loops and tried to do a little more optimization.

### Testing
Looked it over. Ran unit tests locally (the ones I guessed to be relevant). Will wait for CI

### DIFF
--- a/rmgpy/molecule/molecule.py
+++ b/rmgpy/molecule/molecule.py
@@ -1290,7 +1290,7 @@ class Molecule(Graph):
         Placing hydrogens last during sorting ensures that functions with hydrogen
         removal work properly.
         """
-        cython.declare(vertex=Vertex, a=Atom, index=int)
+        cython.declare(vertex=Vertex, index=int)
         for vertex in self.vertices:
             if vertex.sorting_label < 0:
                 self.update_connectivity_values()
@@ -1325,7 +1325,7 @@ class Molecule(Graph):
         """
         Return the molecular formula for the molecule.
         """
-        cython.declare(atom=Atom, symbol=str, element_dict=dict, keys=list,
+        cython.declare(atom=Atom, symbol=str, element_dict=dict, keys=list, key=str,
                        formula=str, count=int)
 
         # Count the number of each element in the molecule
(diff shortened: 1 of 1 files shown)


rmgpy/molecule/molecule.py
```diff
@@ -604,6 +604,7 @@ def apply_action(self, action):
         required parameters. The available actions can be found
         :ref:`here <reaction-recipe-actions>`.
         """
+        cython.declare(act=str, i=cython.Py_ssize_t, n=cython.Py_ssize_t)
```
JacksonBurns, line 607: I don't really know what `Py_ssize_t` is but I assume it is a special type for generators (like the outputs from `range` here... so I will just trust you on it.

rmgpy/molecule/molecule.py
```diff
@@ -612,17 +613,23 @@ def apply_action(self, action):
...
-        elif action[0].upper() == 'LOSE_PAIR':
-            for i in range(abs(action[2])): self.decrement_lone_pairs()
+            n = action[2]
+            for i in range(n): self.decrement_charge()
+        elif act == 'GAIN_PAIR':
+            n = action[2]
+            for i in range(n): self.increment_lone_pairs()
+        elif act == 'LOSE_PAIR':
+            n = action[2]
+            for i in range(n): self.decrement_lone_pairs()
```
JacksonBurns, line 632: Can we just consolidate `n = action[2]` to before the first `if` statement?

JacksonBurns (commented):
Looks broadly good. Will leave this as a comment rather than an approval in case you want to add more to it, but please consider this my approval of the current changes.

rmgpy/molecule/molecule.py
rwest: I thought of that. Yes, it would be less code to read, but actually more code to execute, so I left it the ugly way.

PR #2968, opened 2026-05-29
Add uq assign_intermediate_uncertainties tests

sevyharris:
### Motivation or Problem (not part of RMG 4.0)
I'm going to make some changes to rmgpy/tools/uncertainty.py, so I want to expand the test coverage first to help catch more bugs and prevent myself from breaking things.

### Description of Changes
- Adds tests for uncertainty's assign_intermediate_uncertainties functionality to make sure we don't break this later on
- Extra uncertainty test to make sure we reference the correct library entries when using radical or adsorption correction

### Testing
- This PR does not include any updated functionality and is purely extra testing capability. For what it's worth, the `assign_intermediate_uncertainties` testing mostly makes use of the `assign_parameter_uncertainties` already in place

### DIFF
--- a/test/rmgpy/tools/uncertaintyTest.py
+++ b/test/rmgpy/tools/uncertaintyTest.py
@@ -237,6 +237,54 @@ class TestUncertainty:
         assert np.isclose(thermo_covariance, expected_uncorrelated_thermo_uncertainties, rtol=1e-4).all()
         assert np.isclose(kinetic_covariance, expected_uncorrelated_kinetic_uncertainties, rtol=1e-4).all()
 
+    def test_source_correlations(self):
+        # Check some examples of different species containing the same sources
+
+        # ------------------------------------------------------------------------------
+        # Make sure CH3 (Library + Radical) has a library index/value in common with CH4
+        i_CH4 = rmgpy.tools.uncertainty.get_i_thing(rmgpy.species.Species(smiles='C'), self.uncertainty.species_list)
+        assert i_CH4 >= 0
+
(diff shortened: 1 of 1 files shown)


test/rmgpy/tools/uncertaintyTest.py
```diff
@@ -169,16 +169,119 @@ def test_uncertainty_assignment(self):
...
+
+        src1 = self.uncertainty.species_sources_dict[self.uncertainty.species_list[i_CH4]]  # CH4
+        src2 = self.uncertainty.species_sources_dict[self.uncertainty.species_list[i_CH3X]]  # CH3X
+
+        assert 'Library' in src1
+        assert 'Library' in src2
+        assert 'ADS' in src2
+        assert 'GAV' in src2
+        assert src1['Library'] == src2['Library']  # make sure they refer to the same library source
+        self.uncertainty.species_list.pop()  # remove the extra species so it doesn't affect other tests
```
rwest: Will this get skipped if the test aborts (fails) early? Would that make other tests seem to fail that shouldn't? Would a teardown method or an exception clause protect this?

rwest (commented):
Good; thanks. I approve if you reply to or address my comment.

test/rmgpy/tools/uncertaintyTest.py
sevyharris: Thanks, I've added a try/finally block to protect against that situation.

JacksonBurns (approved):
Final comment addressed, and lgtm, so merging!

PR #2972, opened 2026-06-01
modify max_heavy_atoms constraint

bjkreitz:
### Motivation or Problem
Applying the `maximumHeavyAtoms` constraint for adsorbates currently includes the number of surface sites. For example, setting `maximumHeavyAtoms=4` would lead to cases where a bidentate species with 3 carbon atoms (e.g., XCH2XCHCH3) is excluded while the a monodentate species with 3 carbon atoms is included (e.g., XCH2CH2CH3). Therefore, I think we should not count the number of surface sites as heavy atoms. We have the constraint `maximumSurfaceSites` to adjust the maximum denticity for an adsorbate. 

### Description of Changes
This was a very quick fix and I just modified the 'max_heavy_atoms' function in `constraints.py`

### Testing & Review
I generating a mechanism with this constraint to confirm that it works as intended. Reviewing should be quick. I suggest to test this in one of your existing input files to confirm that it works.

### DIFF
--- a/rmgpy/constraints.py
+++ b/rmgpy/constraints.py
@@ -113,7 +113,7 @@ def fails_species_constraints(species):
 
     max_heavy_atoms = species_constraints.get('maximumHeavyAtoms', -1)
     if max_heavy_atoms != -1:
-        heavy_atoms = struct.get_num_atoms() - struct.get_num_atoms('H')
+        heavy_atoms = struct.get_num_atoms() - struct.get_num_atoms('H') - struct.get_num_atoms('X')
         if heavy_atoms > max_heavy_atoms:
             return f"Exceeded maximumHeavyAtoms: {heavy_atoms} > {max_heavy_atoms}"
 


kirkbadger18:
I have also seen this and agree with this. I will run a quick test to confirm.

JacksonBurns (approved):
This is a good change, thanks.

kirkbadger18 (approved):
This fixes the issue, and I do not see any new issues.

I ran this input file for the NOx-containing exhaust gas conversion:
[input.py](https://github.com/user-attachments/files/28471822/input.py)
I set max heavy atoms to 2, and max surface sites to 2, and as expected the main branch does not allow bidentate ethylene or any of the important bidentate hydrocarbons. The new constraints branch fixes this.
Here is the output from the main branch:

[output.html](https://github.com/user-attachments/files/28471867/output.html)
And here is the output from this branch:

[output.html](https://github.com/user-attachments/files/28471879/output.html)

PR #2974, opened 2026-06-01
Improvements to constraints handling.

rwest:
### Motivation or Problem

Recent PR #2972 changed handling of maximumHeavyAtoms constraint (to exclude surface sites). While reviewing it, I made some changes. But I pushed them too late and it's already merged, so here's a new PR. 
Since the species constraints are enforced on every proposed product of every hypothetical reaction, it happens a lot, so I thought some speed tweaks might be worth it. Also added unit tests.

### Description of Changes

- **Add `get_num_heavy_atoms` to `Molecule` and `Fragment`.** Returns the number of atoms that are neither Hydrogen (`'H'`) nor a surface site (`'X'`). The `Molecule` version is cythonized; the `Fragment` override additionally skips `CuttingLabel` vertices.
- **Use `get_num_heavy_atoms` in `fails_species_constraints`.** Replaces the inline `maximumHeavyAtoms` computation with the new method. Behavior is unchanged, but the intent is clearer and the logic is shared with `Fragment`, and it should be faster.
- **Avoid calling `fails_species_constraints` twice on rejection.** Each call site (`family.py`, `main.py`, two in `model.py`) now stores the single return value and branches on it, instead of calling once for the condition and again for the `reason` string. This halves the constraint-checking work on the failing path with no change in behavior.

### Testing

- Added unit tests for `get_num_heavy_atoms` on both `Molecule` and `Fragment` (`moleculeTest.py`, `fragmentTest.py`).
- Extended `test_heavy_constraint` in `constraintsTest.py` to assert heavy-atom counts directly and to cover an adsorbate, confirming that hydrogens and surface sites (`'X'`) are excluded.
- Ran full test suite.

### Reviewer Tips

- The `Molecule.get_num_heavy_atoms` change are cythonized, so the reviewer should rebuild (`make build`) before running tests.

### DIFF
--- a/rmgpy/constraints.py
+++ b/rmgpy/constraints.py
@@ -113,9 +113,8 @@ def fails_species_constraints(species):
 
     max_heavy_atoms = species_constraints.get('maximumHeavyAtoms', -1)
     if max_heavy_atoms != -1:
-        heavy_atoms = struct.get_num_atoms() - struct.get_num_atoms('H') - struct.get_num_atoms('X')
-        if heavy_atoms > max_heavy_atoms:
-            return f"Exceeded maximumHeavyAtoms: {heavy_atoms} > {max_heavy_atoms}"
+        if struct.get_num_heavy_atoms() > max_heavy_atoms:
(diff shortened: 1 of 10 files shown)


JacksonBurns (approved):
Thanks @rwest looks good!

PR #2975, opened 2026-06-02
Fix covdep

### LINKED ISSUE #2973: RMG crashes when using coverage-dependent thermo
### Bug Description
There are some issues with the coverage-dependent thermochemistry feature that occurred while working on the case study for the RMG paper. 

1) Mechanism generation took forever when including coverage effects for some species. Sometimes I had to wait for up to 30-60 minutes for the next expansion step to occur. 
2) RMG just freezes during the generation, likely due to solver crashes.  

I have a hunch about the reason for both issues and an idea for solutions. However, I'm not entirely sure if this is the best solution. Here is the problem: 

Consider a reaction where a product (e.g., XCCH3) is subjected to a coverage-dependent enthalpy of 15 kcal/mol. If the coverage of XCCH3 is 1, the enthalpy of formation is increased (less negative) by 15 kcal/mol, and the heat of reaction is increased by 15 kcal/mol as well (made more endothermic). Since activation barriers are defined with respect to the forward direction, this can lead to a submerged barrier at high coverages if the change in the enthalpy of formation of the product is larger than the sum of the reaction enthalpy and the activation barrier. The following picture illustrates this problem

Image

We have encountered this problem before and had to adjust the mechanisms by reversing the reaction direction. In this way, the coverage effects make the reaction enthalpy increasingly exothermic, preventing submerged barriers, but the activation barrier for the reverse reaction increases at higher reactant coverages. It is still an open scientific question whether this is reasonable. Of course, this becomes a headache if you have a reaction where the reactant and the product are subjected to coverage dependencies. 

### How To Reproduce
Here is an example input file that I have used for the mechanism generation with coverage effects 

[input.py](https://github.com/user-attachments/files/28472445/input.py)

This requires the following modified thermochemistry library to work

[surfaceThermoPt111_covdep.py](https://github.com/user-attachments/files/28472477/surfaceThermoPt111_covdep.py)

### Proposed solution
I can think of two solutions to fix this problem. Both are kind of hacky, and I think there is still a lot of research to be performed in order to actually figure this out. @cfgoldsmith and Chenge are working on studying this. So hopefully in the future we have some better approaches. For the short term, I think we can either: 

1) Flip the reactions as we have done before. 
2) Do not flip the reaction, but raise the activation barrier in the forward direction to at least match the endothermicity of the reaction at a coverage of 1 ML of the coverage-dependent species. 

I think both create some problems. Approach 2 will significantly overestimate the barrier at low coverages, while approach 1 is challenging if there are multiple species with coverage effects. I started to work on approach 1 and will open a draft PR soon.

bjkreitz:
### Motivation or Problem
This PR addresses issue #2973. RMG was really slow when generating mechanisms that include coverage-dependent thermo.  One problem was that the coverage dependence could cause submerged barriers in the reverse direction, which is discussed in depth in #2973. However, this was not the main reason for the slowness. RMG was extremely slow because it needed a few million time steps to achieve convergence.  

### Description of Changes
I added a functionality that enables changing the direction of a reaction, where a product is coverage dependent. The parameters of this new forward direction are determined from the thermochemistry. 
The main reason why RMG needed a few million time steps to converge was that the Jacobian used the uncorrected reverse rate constant instead of the coverage-dependent kr. Therefore, RMG took many very small time steps. Adjusting the `jacobian` routine did the trick. 

### Testing
I tested this by generating a mechanism for the ethanol steam reforming. 

[input.py](https://github.com/user-attachments/files/28507102/input.py)

Without the changes, the mechanism generation took about 1 hour, but with the updated `jacobian` routine it now takes about 1 minute. 

### Reviewer Tips
I have some more testing and clean-up to do, but I wanted to already share the approach to get some initial feedback. There are further ways to speed things up. Currently, the `compute_thermo_coverage_corrections` is the bottleneck. It can be accelerated by converting it to a `cdef` function and by adjusting the routine. I'll keep working on it.

### DIFF
--- a/rmgpy/reaction.pxd
+++ b/rmgpy/reaction.pxd
@@ -155,7 +155,5 @@ cdef class Reaction:
 
     cpdef get_mean_sigma_and_epsilon(self, bint reverse=?)
 
-    cpdef Reaction get_reverse_reaction(self)
-
 cpdef bint same_species_lists(list list1, list list2, bint check_identical=?, bint only_check_label=?,
                               bint generate_initial_map=?, bint strict=?, bint save_order=?) except -2
--- a/rmgpy/reaction.pxd
+++ b/rmgpy/reaction.pxd
@@ -155,7 +155,7 @@ cdef class Reaction:
 
     cpdef get_mean_sigma_and_epsilon(self, bint reverse=?)
 
-    cpdef get_reverse_reaction(self)
+    cpdef Reaction get_reverse_reaction(self)
 
 cpdef bint same_species_lists(list list1, list list2, bint check_identical=?, bint only_check_label=?,
                               bint generate_initial_map=?, bint strict=?, bint save_order=?) except -2
--- a/rmgpy/reaction.pxd
+++ b/rmgpy/reaction.pxd
@@ -143,7 +143,7 @@ cdef class Reaction:
 
     cpdef generate_pairs(self)
 
-    cpdef copy(self)
+    cpdef Reaction copy(self)
 
     cpdef ensure_species(self, bint reactant_resonance=?, bint product_resonance=?, bint save_order=?)
 
--- a/rmgpy/reaction.pxd
+++ b/rmgpy/reaction.pxd
@@ -155,5 +155,7 @@ cdef class Reaction:
 
     cpdef get_mean_sigma_and_epsilon(self, bint reverse=?)
 
+    cpdef get_reverse_reaction(self)
+
 cpdef bint same_species_lists(list list1, list list2, bint check_identical=?, bint only_check_label=?,
                               bint generate_initial_map=?, bint strict=?, bint save_order=?) except -2
--- a/rmgpy/reaction.py
+++ b/rmgpy/reaction.py
@@ -1842,10 +1842,19 @@ class Reaction:
         raise NotImplementedError("generate_high_p_limit_kinetics is not implemented for all Reaction subclasses.")
 
     def get_reverse_reaction(self):
+        """
+        Return a new :class:`Reaction` with the reactants and products swapped and
+        the kinetics re-parameterized in the reverse direction (via
+        :meth:`generate_reverse_rate_coefficient`). The original reaction is left
+        unmodified.
 
+        Raises :class:`ReactionError` if the reaction has no kinetics, since the
+        reverse rate coefficient cannot otherwise be derived.
+        """
         cython.declare(rev_reaction=Reaction)
 
-        assert self.kinetics is not None
+        if self.kinetics is None:
+            raise ReactionError("Cannot generate the reverse of reaction {0!r}: no kinetics defined.".format(self))
         rev_reaction = self.copy()
         rev_reaction.reactants = self.products[:]
         rev_reaction.products = self.reactants[:]
--- a/rmgpy/reaction.py
+++ b/rmgpy/reaction.py
@@ -1841,6 +1841,15 @@ class Reaction:
         """
         raise NotImplementedError("generate_high_p_limit_kinetics is not implemented for all Reaction subclasses.")
 
+    def get_reverse_reaction(self):
+        assert self.kinetics is not None
+        rev_reaction = deepcopy(self)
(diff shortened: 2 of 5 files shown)


rmgpy/rmg/model.py
```diff
@@ -631,6 +632,18 @@ def make_new_reaction(self, forward, check_existing=True, generate_thermo=True,
                 # we need to make sure the barrier is positive.
                 forward.fix_barrier_height(force_positive=True,solvent="")
 
+            # If any product has thermo_coverage_dependence, flip the reaction
+            # and parameterize in the reverse direction
+            if (any(getattr(spc.thermo, 'thermo_coverage_dependence', None) and spc.thermo.thermo_coverage_dependence for
```
rwest: Should we also check that there isn't a coverage-dependent reactant? (i.e. if we have both reactants and products that are coverage-dependent, we shouldn't flip?)

rmgpy/rmg/model.py
```diff
@@ -631,6 +632,18 @@ def make_new_reaction(self, forward, check_existing=True, generate_thermo=True,
...
+                spc in forward.products) and not isinstance(forward.kinetics, StickingCoefficient)):
+                forward = Reaction.get_reverse_reaction(forward)
+                forward.reactants = [self.make_new_species(reactant, generate_thermo=generate_thermo)[0] for reactant in
```
rwest: is this needed to generate the thermo? some other side-effect? (perhaps add a comment as to why this is happening?)

rmgpy/solver/surface.pyx
```diff
@@ -417,6 +417,24 @@ cdef class SurfaceReactor(ReactionSystem):
                 bimolecular_threshold_rate_constant,
                 trimolecular_threshold_rate_constant)
 
+    def compute_thermo_coverage_corrections(self, coverages):
```
rwest: We should probably cythonize this thoroughly (update the surface.pxd and add declarations in the method)

rmgpy/reaction.pxd
```diff
@@ -155,5 +155,7 @@ cdef class Reaction:
 
     cpdef get_mean_sigma_and_epsilon(self, bint reverse=?)
 
+    cpdef get_reverse_reaction(self)
```
rwest: Could we declare the return type as Reaction?
```suggestion
    cpdef Reaction get_reverse_reaction(self)
```

(if so, we should do so for the `copy` method on line 146, while we're at it)

rmgpy/reaction.py
```diff
@@ -1841,6 +1841,17 @@ def generate_high_p_limit_kinetics(self):
         """
         raise NotImplementedError("generate_high_p_limit_kinetics is not implemented for all Reaction subclasses.")
 
+    def get_reverse_reaction(self):
+
+        cython.declare(rev_reaction=Reaction)
+
+        assert self.kinetics is not None
```
rwest: Do we want this assertion?
What about wrap the `rev_reaction.kinetics = self.generate_reverse_rate_coefficient()` with an `if self.kinetics is not None:`  instead?

rwest (commented):
A good start. A few minor comments on the changes so far. 

Some unit tests would be good.
Including one that would have caught the Jacobian was wrong!? (compare it to a finite difference? or a known state? making a test like that is the type of thing where I'd try using Claude Code or similar)

rwest:
@bjkreitz thanks for your PR. Can you see if my additional commits make sense to you? (and still work)

bjkreitz:
Thanks @rwest! I ran into some issues when adding more species with coverage-dependent thermo using the `get_reverse_reaction` function. Turns out, we actually don't need it, so I cleaned up the code a bit. I can also confirm that RMG still works and that it discovers the same mechanism with all these changes.

rwest:
Your latest changes look good. I think we should append to the reaction (or kinetics) comment about why it was flipped.

bjkreitz:
Is this what you had in mind?

rwest (approved):
Looks good to me. Thanks Bjarne.

I tweaked the comment code, refactored another bit, and rebased (squashing some fixup commits).

PR #2978, opened 2026-06-04
Reduce "error" logging in regression testing (only WARN and above)

rwest:
### Motivation or Problem
Previously, pysidt, pulled in transitively via rmgpy.data.thermo, would configure a logger at import time, which put all INFO messages into stderr, which would then end up showing as errors in the pull request annotations.

### Description of Changes
Now, only WARNINGs and ERRORs should be reported, to stderr.
The INFO should, however, still go to stdout.

### Testing
Let's see if it works (await the Regression testing annotation on this PR)

### Other context
I have opened https://github.com/zadorlab/PySIDT/issues/44 on PySIDT

### DIFF
--- a/rmgpy/tools/observablesregression.py
+++ b/rmgpy/tools/observablesregression.py
@@ -291,7 +291,6 @@ class ObservablesTestCase(object):
         conditions_broken = []
         variables_failed = []
 
-        print('{0} Comparison'.format(self))
         # Check the species profile observables
         if 'species' in self.observables:
             if self.old_sim.surface_species_list:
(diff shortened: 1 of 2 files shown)


JacksonBurns:
@rwest based on the github actions comment this seems to have worked - ready for review and merge?

rwest:
I hope the last (most recent) commit is the last (final) commit.

rmgpy/tools/observablesregression.py
```diff
@@ -291,7 +291,7 @@ def compare(self, tol, plot=False):
         conditions_broken = []
         variables_failed = []
 
-        print('{0} Comparison'.format(self))
+        #print('{0} Comparison'.format(self))
```
JacksonBurns:
```suggestion
```

Deleting rather than commenting out

JacksonBurns (dismissed):
LGTM, will wait to see the output though

JacksonBurns (dismissed):

PR #2982, opened 2026-06-10
Require atom style objects to have the same labels to match

mjohnson541:


rwest:
Claude's analysis:

> 
> The intent is good: ensuring that **labeled atoms in reaction templates** match correctly. But the implementation is too broad. The solution is **conditional label checking**:
> 
> ```python
> # In Atom.equivalent() and is_equivalent():
> # Only require label match if EITHER atom has a label
> if (self.label or other.label) and self.label != other.label:
>     return False
> ```
> This way:
> 
> - ✅ Labeled atoms (`*1`, `*2`, etc.) match only with the same label
> - ✅ Unlabeled atoms (`''`) can still act as wildcards
> - ✅ Database lookups work correctly
> - ✅ Reaction templates work as intended
> 
> The current commit conflates two different use cases:
> 
> 1. **Exact matching**: When both atoms have labels, they must match exactly
> 2. **Pattern matching**: When an atom has no label, it matches any atom
> 
> A better approach would be to apply the strict check only when needed for reaction-specific logic (like template matching), not in the generic `equivalent()` method used throughout the codebase.
> 

----

But I still don't know what problem it's trying to fix.

mjohnson541:
I'm starting to run into a lot of cases where it's much faster and more convenient to have the label check as part of the isomorphism rather than trying to construct the label map beforehand and feed it in, which we only have implemented for some of the isomorphism algorithms. 

I mostly created this just to see how much would break from that modification. I went down a rabbit hole adding a `check_labels` option, but something went wrong with it. For the sake of urgency I'm using a special branch of molecule instead for now. 

What Claude is proposing here doesn't make any sense because the only labels we have that are "null" are `''` so if `not (self.label or other.label)` then `self.label == '' == other.label` anyway. So the change it proposes shouldn't change output.

PR #2983, opened 2026-06-22
Make Regression Tests Pass by Default

JacksonBurns:
This PR is two things: (1) an important effort towards fixing reproducibility issues that crop up in RMG and its regression tests and (2) an experiment for me in using AI to fix bugs.

All of the code changes in this PR were made by Qwen 3.6 27B in opencode. I gave the model a script that ran a given regression test twice and then used `scripts/checkModel.py` on the output to ensure reproducibility, then asked the model to fix any non-reproducible regression tests.

## Description

Our regression tests 'fail' by default - in reality, RMG just doesn't always make consistent estimates of thermochemistry (and consequently kinetics). This PR resolves it.

### `sulfur`

The sulfur regression test typically fails with:

```
The original model has 1 reactions that the tested model does not have. ❌
rxn: O(4) + SO2(15) (+N2) <=> SO3(16) (+N2) origin: primarySulfurLibrary
The tested model has 1 reactions that the original model does not have. ❌
rxn: O(4) + SO2(15) (+N2) <=> SO3(16) (+N2) origin: primarySulfurLibrary
```

Obviously this is not actually a failure, we just aren't checking equality correctly. This was resolved in https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2983/commits/8b52c36fbb6e87e97ece3572181f68aabe8710b5 by checking for isomorphism rather than equality (there may be a better way to do this, read the extended description of that commit to understand why it was done this way).

### `aromatics`

The test typically fails with:

```
Non-identical thermo! ❌
original: [CH]1C2=CC3C1C=CC23
tested: [CH]1C2=CC3C1C=CC23
```

which is due to a non-deterministic choice in ring decomposition. PR https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2920 partially resolved this, and the remaining issue was fixed here: https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2983/commits/9156e759af290dd22d219ce022950bc8b9d85d33 - all that was required was changing `list` to `sorted` (`sorted` returns a list) to ensure that lists of component cycles are always in the same order.

### `liquid_oxidation` and `RMS_CSTR_liquid_oxidation`

Both of these often failed, typically with missing reactions and species in the edge (and sometimes core, too):

```
Test model has 1588 reactions. ❌
The original model has 5 reactions that the tested model does not have. ❌
rxn: CC(C[CH]COO)OO(115) <=> [OH](22) + CC(CCC=O)OO(116) origin: intra_H_migration
rxn: CC(C[CH]COO)OO(115) <=> [OH](22) + CC(=O)CCCOO(112) origin: intra_H_migration
rxn: CC(CC(C)OO)O[O](90) + CC(CCCOO)O[O](108) <=> oxygen(1) + CC([O])CC(C)OO(110) + CC([O])CCCOO(123) origin: Peroxyl_Disproportionation
rxn: CC(CC(C)OO)O[O](90) + CC(CCCOO)O[O](108) <=> oxygen(1) + CC(=O)CC(C)OO(100) + CC(O)CCCOO(152) origin: Peroxyl_Termination
rxn: CC(CC(C)OO)O[O](90) + CC(CCCOO)O[O](108) <=> oxygen(1) + CC(=O)CCCOO(112) + CC(O)CC(C)OO(143) origin: Peroxyl_Termination
The tested model has 2 reactions that the original model does not have. ❌
rxn: CC(C[CH]COO)OO(118) <=> CC(CC[CH]OO)OO(133) origin: intra_H_migration
rxn: CC(C[CH]COO)OO(118) <=> C[C](CCCOO)OO(132) origin: intra_H_migration
```

This was resolved in the remaining commits, which are a bit messy. From what I understand it is _also_ related to sorting, but now with regard to the order in which species are added to the model. Would appreciate some further insight on this one.

### DIFF
--- a/rmgpy/data/base.py
+++ b/rmgpy/data/base.py
@@ -1066,11 +1066,10 @@ class Database(object):
             else:
                 return root
         else:
-            # Multiple children match - sort by node label for deterministic selection
-            next_node.sort(key=lambda n: n.label)
-            # logging.warning('For {0}, a node {1} with overlapping children {2} was encountered '
-            #                 'in tree with top level nodes {3}. Assuming the first match is the '
-            #                 'better one.'.format(structure, root, next, self.top))
+            # Multiple children match - pick the first in tree order.
+            # The tree is constructed deterministically, so this is deterministic.
+            # (Sorting by label can pick the wrong node when siblings overlap,
+            # e.g. Nitrites sorts before Nitro but Nitro is the intended match.)
             return self.descend_tree(structure, atoms, next_node[0], strict)
 
     def are_siblings(self, node, node_other):
--- a/rmgpy/data/base.py
+++ b/rmgpy/data/base.py
@@ -1066,6 +1066,8 @@ class Database(object):
             else:
                 return root
         else:
+            # Multiple children match - sort by node label for deterministic selection
+            next_node.sort(key=lambda n: n.label)
             # logging.warning('For {0}, a node {1} with overlapping children {2} was encountered '
             #                 'in tree with top level nodes {3}. Assuming the first match is the '
             #                 'better one.'.format(structure, root, next, self.top))
--- a/rmgpy/data/base.py
+++ b/rmgpy/data/base.py
@@ -1058,19 +1058,14 @@ class Database(object):
             if self.match_node_to_structure(child, structure, atoms, strict):
                 next_node.append(child)
 
-        if len(next_node) == 1:
+        if len(next_node) != 0:
+            # Multiple children match - pick the first in tree order.
             return self.descend_tree(structure, atoms, next_node[0], strict)
-        elif len(next_node) == 0:
+        else:
             if len(root.children) > 0 and root.children[-1].label.startswith('Others-'):
                 return root.children[-1]
             else:
                 return root
-        else:
-            # Multiple children match - pick the first in tree order.
-            # The tree is constructed deterministically, so this is deterministic.
-            # (Sorting by label can pick the wrong node when siblings overlap,
-            # e.g. Nitrites sorts before Nitro but Nitro is the intended match.)
-            return self.descend_tree(structure, atoms, next_node[0], strict)
 
     def are_siblings(self, node, node_other):
         """
--- a/rmgpy/data/base.py
+++ b/rmgpy/data/base.py
@@ -1059,7 +1059,7 @@ class Database(object):
                 next_node.append(child)
 
         if len(next_node) != 0:
-            # Multiple children match - pick the first in tree order.
+            # at least one child matches - pick the first in tree order.
             return self.descend_tree(structure, atoms, next_node[0], strict)
         else:
             if len(root.children) > 0 and root.children[-1].label.startswith('Others-'):
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -317,16 +317,6 @@ def common_atoms(cycle1, cycle2):
     return set1.intersection(set2)
 
 
-def combine_cycles(cycle1, cycle2):
-    """
-    INPUT: two cycles with type: list of atoms
-    OUTPUT: a combined cycle with type: list of atoms
-    """
-    set1 = set(cycle1)
-    set2 = set(cycle2)
-    return sorted(set1.union(set2))
-
-
 def is_aromatic_ring(submol):
     """
     This method takes a monoring submol (Molecule initialized with a list of atoms containing just 
--- a/rmgpy/data/thermo.py
+++ b/rmgpy/data/thermo.py
@@ -324,7 +324,7 @@ def combine_cycles(cycle1, cycle2):
     """
     set1 = set(cycle1)
     set2 = set(cycle2)
-    return list(set1.union(set2))
+    return sorted(set1.union(set2))
 
 
 def is_aromatic_ring(submol):
--- a/rmgpy/molecule/graph.pyx
+++ b/rmgpy/molecule/graph.pyx
@@ -490,7 +490,7 @@ cdef class Graph(object):
         self.update_connectivity_values()
 
         # Build sort keys with optional chemical tiebreaker (sorting_key attribute)
-        # to ensure deterministic ordering when connectivity values are identical.
+        # for when connectivity values are identical.
         # Sort (key, index) pairs, then use the resulting index order to reorder.
         paired = []
         for index, vertex in enumerate(self.vertices):
--- a/rmgpy/molecule/graph.pyx
+++ b/rmgpy/molecule/graph.pyx
@@ -498,7 +498,7 @@ cdef class Graph(object):
         for index, vertex in enumerate(self.vertices):
             conn = get_vertex_connectivity_value(vertex)
             sort_key = getattr(vertex, 'sorting_key', None)
-            paired.append(((conn, sort_key if sort_key is not None else ()), index))
+            paired.append(((conn, sort_key if sort_key is not None else (conn,)), index))
 
         paired.sort()
         ordered = [self.vertices[idx] for _, idx in paired]
--- a/rmgpy/molecule/graph.pyx
+++ b/rmgpy/molecule/graph.pyx
@@ -491,14 +491,23 @@ cdef class Graph(object):
 
         # Build sort keys with optional chemical tiebreaker (sorting_key attribute)
         # to ensure deterministic ordering when connectivity values are identical.
-        # Index is included as a final tiebreaker to avoid comparing vertex
-        # objects directly (GroupAtom has no __lt__).
         # Sort (key, index) pairs, then use the resulting index order to reorder.
         paired = []
         for index, vertex in enumerate(self.vertices):
             conn = get_vertex_connectivity_value(vertex)
             sort_key = getattr(vertex, 'sorting_key', None)
-            paired.append(((conn, sort_key if sort_key is not None else (conn,)), index))
+            paired.append(
+                (
+                    # the sorting key is a tuple of keys which are consumed in order.
+                    # we first attempt to sort by conn (always defined).
+                    # in the case of a tie, sort by the contents of the second element:
+                    #  - sort_key, which is a tuple of sortable things defined elsewhere, if it is defined
+                    #  - empty tuple, if sort_key is not defined
+                    # the latter results in a tie, which we 'break' by just not changing the order (stable sort)
+                    (conn, (sort_key if sort_key is not None else tuple())),
+                    index
+                )
+            )
 
         paired.sort()
         ordered = [self.vertices[idx] for _, idx in paired]
--- a/rmgpy/molecule/graph.pyx
+++ b/rmgpy/molecule/graph.pyx
@@ -469,7 +469,10 @@ cdef class Graph(object):
     cpdef sort_vertices(self, bint save_order=False):
         """
         Sort the vertices in the graph. This can make certain operations, e.g.
-        the isomorphism functions, much more efficient.
+        the isomorphism functions, much more efficient.  Vertices that have a
+        ``sorting_key`` attribute (e.g. :class:`Atom`) will use it as a chemical
+        tiebreaker for vertices with identical connectivity values, ensuring
+        deterministic ordering.
         """
         cdef Vertex vertex
         cdef int index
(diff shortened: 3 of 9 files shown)


JacksonBurns:
Move moleucle atom sorting changes into graph, if possible? and needed?

JacksonBurns:
~~I can't get the RMS_CSTR_liquid_oxidation to fail locally~~ my tests indicate that the other reg test fixes took care of this as a consequence

JacksonBurns:
@rwest I've taken a stab at getting the regression tests to pass by default, with the help of AI this go around. I think this actually fixes it, to boot. Please take a look when you have some time.

JacksonBurns:
@KnathanM could you take a look at this when you get a chance? It fixes the non-determinism in the regression tests, and I think it might be useful for your ML-based thermo estimator work since it addresses the root cause of why thermo estimates vary between runs. Much appreciated!

rmgpy/molecule/graph.pyx
```diff
@@ -485,7 +488,21 @@ cdef class Graph(object):
...
+        # Build sort keys with optional chemical tiebreaker (sorting_key attribute)
+        # to ensure deterministic ordering when connectivity values are identical.
+        # Index is included as a final tiebreaker to avoid comparing vertex
+        # objects directly (GroupAtom has no __lt__).
+        # Sort (key, index) pairs, then use the resulting index order to reorder.
+        paired = []
+        for index, vertex in enumerate(self.vertices):
+            conn = get_vertex_connectivity_value(vertex)
+            sort_key = getattr(vertex, 'sorting_key', None)
+            paired.append(((conn, sort_key if sort_key is not None else ()), index))
```
KnathanM: Why is this not `paired.append(((conn, sort_key if sort_key is not None else (conn,)), index))`? In other words, the current code uses `get_vertex_connectivity_value` as the primary sorting key. Your change makes it so that if you are sorting something without sorting_key, then the sorting is only done by the index. Is that intentional?

rmgpy/molecule/graph.pyx
```diff
@@ -485,7 +488,21 @@ cdef class Graph(object):
         # If we need to sort then let's also update the connecitivities so
         # we're sure they are right, since the sorting labels depend on them
         self.update_connectivity_values()
-        self.vertices.sort(key=get_vertex_connectivity_value)
+
+        # Build sort keys with optional chemical tiebreaker (sorting_key attribute)
+        # to ensure deterministic ordering when connectivity values are identical.
+        # Index is included as a final tiebreaker to avoid comparing vertex
```
KnathanM: Do you need a final tiebreaker? I thought python's list.sort() was stable, meaning if there is a tie, the order remains the same, which is also the effect of your sorting by index.

rmgpy/data/thermo.py
```diff
@@ -324,7 +324,7 @@ def combine_cycles(cycle1, cycle2):
     """
     set1 = set(cycle1)
     set2 = set(cycle2)
-    return list(set1.union(set2))
+    return sorted(set1.union(set2))
```
KnathanM: Where in the code is `combine_cycles` used? I only see it in the tests and there its output immediately gets converted to a set.

rmgpy/data/base.py
```diff
@@ -1066,9 +1066,10 @@ def descend_tree(self, structure, atoms, root=None, strict=False):
             else:
                 return root
         else:
-            # logging.warning('For {0}, a node {1} with overlapping children {2} was encountered '
-            #                 'in tree with top level nodes {3}. Assuming the first match is the '
-            #                 'better one.'.format(structure, root, next, self.top))
+            # Multiple children match - pick the first in tree order.
+            # The tree is constructed deterministically, so this is deterministic.
+            # (Sorting by label can pick the wrong node when siblings overlap,
+            # e.g. Nitrites sorts before Nitro but Nitro is the intended match.)
```
KnathanM: The logging line was commented out 12 year ago e8c7f1d, so is probably fine to remove.

What does the "sorting by label" part of the new comment refer to?

rmgpy/data/base.py
```diff
@@ -1066,9 +1066,10 @@ def descend_tree(self, structure, atoms, root=None, strict=False):
             else:
                 return root
         else:
-            # logging.warning('For {0}, a node {1} with overlapping children {2} was encountered '
-            #                 'in tree with top level nodes {3}. Assuming the first match is the '
-            #                 'better one.'.format(structure, root, next, self.top))
+            # Multiple children match - pick the first in tree order.
+            # The tree is constructed deterministically, so this is deterministic.
```
KnathanM: I'm assuming this comment is correct, that tree construction is deterministic.

rmgpy/reaction.py
```diff
@@ -615,7 +615,12 @@ def is_isomorphic(self, other, either_direction=True, check_identical=False, che
                                                     save_order=save_order)
 
         # Compare specific_collider to specific_collider
-        collider_match = (self.specific_collider == other.specific_collider)
+        if self.specific_collider is None:
+            collider_match = other.specific_collider is None
+        elif other.specific_collider is None:
+            collider_match = False
+        else:
+            collider_match = self.specific_collider.is_isomorphic(other.specific_collider)
```
KnathanM, line 623: The commit comment says this is to fix: `When two models are loaded from separate Chemkin files, the specific_collider (N2) species are distinct object instances despite being chemically identical,
causing == to always return False.` Maybe this should use `is_identical` instead of `is_isomorphic` as that is closer to the original equality check while still allowing the objects to not be the same in memory.

test/regression/RMS_CSTR_liquid_oxidation/input.py
```diff
@@ -53,12 +53,12 @@
     toleranceMoveToCore=0.01,
     toleranceKeepInEdge=0.001,
     toleranceInterruptSimulation=1e8,
-    maximumEdgeSpecies=10000,
+    maximumEdgeSpecies=300,
```
KnathanM, line 56: Are the changes to this file still required after your other changes, or could this be reverted? Changing from 10000 to 300 seems a little drastic to me. Can you comment on this change? (and the similar change to `maxNumSpecies`?

KnathanM (commented):
I haven't run the tests myself to check if they are deterministic, but we should probably manually do that before merging this. For now, please see my comments.

test/regression/RMS_CSTR_liquid_oxidation/input.py
JacksonBurns: This had nothing to do with reproducibility and everything to do with runtime. This test takes 10+ minutes on `main` and only a few here.

This is more in like with the size of the reg tests, as well, so I think it is still a scientifically valid regression test.

rmgpy/reaction.py
JacksonBurns: Good point, will change online and let the CI run

JacksonBurns: 4743a3d935c5acc025f02ab1637f514bc3f3fc5c

rmgpy/data/base.py
JacksonBurns: Yes the tree is constructed from fixed 'instructions' in the database that do not involve any tie breaking.

The _original_ construction of those trees, i.e. when they are trained, is non-deterministic. I know for a fact that this is partly because of species ring assignments -- which could warrant retraining. There may also be some tie-breaking involved in generating (fitting) the trees, but I don't know enough about it to say for sure.

rmgpy/data/base.py
JacksonBurns: I will actually remove the outdated lines. During an earlier attempt at fixing the non-determinism, the LLM made an ill-guided change here, the details of which I don't recall.

I also noticed when making the above fix that there is some awkward logic here. I am changing that as well.

JacksonBurns: d13bac73e6e3c40a1a02e6f71994c1fd957e9244

rmgpy/data/thermo.py
JacksonBurns: Huh, never thought to check that. I looked at my local copy of RMG-Py, -database, and -Website and found nothing.

I am going to just take it out... 8f1c2ddea05761ff6cce2e0574ed4f5113a60194

rmgpy/data/base.py
```diff
@@ -1058,18 +1058,14 @@ def descend_tree(self, structure, atoms, root=None, strict=False):
             if self.match_node_to_structure(child, structure, atoms, strict):
                 next_node.append(child)
 
-        if len(next_node) == 1:
+        if len(next_node) != 0:
+            # Multiple children match - pick the first in tree order.
```
KnathanM: Nit: The comment doesn't quite line up with the code. We enter this block even if there is only one child match.

rmgpy/molecule/graph.pyx
JacksonBurns: this answer is based on my understanding of the AI's code. i am not running it (yet). 

yes `list.sort()` is stable. however, it still has to call `__lt__` on each element. according to the comment, some objects that we encounter in the `self.vertices` list do not implement this method, which causes an error. by including the `index` as a fallback we avoid this.

rmgpy/molecule/graph.pyx
JacksonBurns: I think you are correct. I will make this change, and then we should test it.

JacksonBurns:
Note to self: I have opencode session floating around where this was fixed, and there should be a script either in that session or on that machine that checks for reproducibility.

JacksonBurns:
@KnathanM following up on our discussion today, I have (1) reverted the `(conn,)` change (2) updated the comments and (3) found the script for testing reproducibility:

```bash
#!/bin/bash
# 
# usage: bash /tmp/check_repro.sh <test_name> (e.g., sulfur, liquid_oxidation, aromatics, superminimal)
#
# execute in the RMG-Py directory
#
TEST=$1

# Run 1
python rmg.py "test/regression/${TEST}/input.py" &> /tmp/repro_debug_${TEST}.txt
cp "test/regression/${TEST}/chemkin/chem_annotated.inp" "/tmp/${TEST}_ca_1.inp"
cp "test/regression/${TEST}/chemkin/species_dictionary.txt" "/tmp/${TEST}_sd_1.txt"
cp "test/regression/${TEST}/chemkin/chem_edge_annotated.inp" "/tmp/${TEST}_cea_1.inp"
cp "test/regression/${TEST}/chemkin/species_edge_dictionary.txt" "/tmp/${TEST}_sed_1.txt"

# Run 2
python rmg.py "test/regression/${TEST}/input.py" &>> /tmp/repro_debug_${TEST}.txt
cp "test/regression/${TEST}/chemkin/chem_annotated.inp" "/tmp/${TEST}_ca_2.inp"
cp "test/regression/${TEST}/chemkin/species_dictionary.txt" "/tmp/${TEST}_sd_2.txt"
cp "test/regression/${TEST}/chemkin/chem_edge_annotated.inp" "/tmp/${TEST}_cea_2.inp"
cp "test/regression/${TEST}/chemkin/species_edge_dictionary.txt" "/tmp/${TEST}_sed_2.txt"

# Compare core
echo "=== ${TEST} core ==="
python scripts/checkModels.py "${TEST}-core" \
    "/tmp/${TEST}_ca_1.inp" "/tmp/${TEST}_sd_1.txt" \
    "/tmp/${TEST}_ca_2.inp" "/tmp/${TEST}_sd_2.txt"

# Compare edge
echo "=== ${TEST} edge ==="
python scripts/checkModels.py "${TEST}-edge" \
    "/tmp/${TEST}_cea_1.inp" "/tmp/${TEST}_sed_1.txt" \
    "/tmp/${TEST}_cea_2.inp" "/tmp/${TEST}_sed_2.txt"
```

This will emit `$1-core.log` and `$1-edge.log` - I ran the four mentioned tests and got green checks in both files, for each.

In the course of debugging I also found that the change from `is_isomorphic` to `is_identical` was failing - not sure why, but the current version fixes the regression test.

Let me know how your testing goes!

rmgpy/molecule/graph.pyx
```diff
@@ -485,7 +488,30 @@ cdef class Graph(object):
         # If we need to sort then let's also update the connecitivities so
         # we're sure they are right, since the sorting labels depend on them
         self.update_connectivity_values()
-        self.vertices.sort(key=get_vertex_connectivity_value)
+
+        # Build sort keys with optional chemical tiebreaker (sorting_key attribute)
+        # to ensure deterministic ordering when connectivity values are identical.
```
KnathanM: When connectivity values are identical, isn't the ordering still deterministic because the `self.vertices.sort` is stable? I think adding `vertex.sorting_key` to the keys for sorting is a good idea to help the isomorphism functions be more efficient (because then the sort is less reliant on the original ordering), but I'm not sure why it would be needed for deterministic sorting. If you find that it is needed, then I am fine leaving the comment as is.

KnathanM (commented):
Two questions on the comments. Otherwise, I have run the tests locally (except for the surface tests) several times and they look deterministic to me. I think the code is ready to merge. @rwest, can you take a look at this PR? 

I'll also note the change to `RMS_CSTR_liquid_oxidation` is motivated by wanting the test to run faster, see Jackson's comment about that.

rmgpy/data/base.py
JacksonBurns: good catch, fixed in: aed0c0e173457a8d6624c87a4a118b6fc7f643c2

rmgpy/molecule/graph.pyx
JacksonBurns: you are correct, I have done this in 5afad12d720899d1068a4caa769be85254c82ac6

KnathanM (approved):

JacksonBurns:
@rwest we have reviewed this PR and confirmed that it does solve the non-determinism in ring decomposition. locally executing the regression tests now passes (see [this comment](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2983#issuecomment-5224294966) for instructions to do so yourself) - could you review as well?

PR #2990, opened 2026-07-28
fix(arkane): make PDep sensitivity meaningful for ILT-based path reactions

alongd:
Arkane's PDep sensitivity analysis perturbs `conformer.E0`. For path reactions whose k(E) comes from an inverse Laplace transform instead of RRKM, that does nothing at all — the sensitivity coefficients come back as exactly zero.

Since `E0(TS) = Σ E0(reactants) + Ea` by construction, this perturbs `Ea` by the same amount alongside E0, which reaches both the ILT convolution and the renormalization target.

A hand-authored modeless TS with a real E0 keeps the old behavior; applied to every path reaction sharing the TS object, matching the E0 perturbation's reach; and skipped for RRKM reactions, where it would double-count.

### DIFF
--- a/arkane/sensitivity.py
+++ b/arkane/sensitivity.py
@@ -47,13 +47,6 @@ from rmgpy.exceptions import InvalidMicrocanonicalRateError, ModifiedStrongColli
 
 ################################################################################
 
-# Tolerance (in J/mol) used when checking whether a TransitionState's E0 was synthesized as
-# E0(TS) = sum(E0(reactants)) + Ea (as RMG and Arkane's own ILT path do), as opposed to being an
-# independent, hand-authored value. 1 kJ/mol is comfortably below the floating-point noise floor
-# introduced by unit conversions and energy corrections along that construction path, while still
-# being small enough to catch a genuinely independent E0.
-SYNTHETIC_TS_E0_TOLERANCE = 1000.0  # J/mol (~1 kJ/mol)
-
 
 class KineticsSensitivity(object):
     """
@@ -224,14 +217,14 @@ class KineticsSensitivity(object):
             ax[i][0].set_yticks(y_pos)
             ax[i][0].set_yticklabels(labels)
             ax[i][0].invert_yaxis()  # labels read top-to-bottom
-            ax[i][0].set_xlabel(r'Sensitivity: $\frac{\partial\:\ln{k}}{\partial\:E0}$, ($\frac{J}{mol}$)')
+            ax[i][0].set_xlabel(r'Sensitivity: $\frac{\partial\:\ln{k}}{\partial\:E0}$, ($\frac{mol}{J}$)')
             ax[i][0].set_title('Forward, {0}'.format(condition))
             ax[i][0].set_xlim([min_sa, max_sa])
             ax[i][1].barh(y_pos, r_values, align='center', color='blue')
             ax[i][1].set_yticks(y_pos)
             ax[i][1].set_yticklabels(labels)
             ax[i][1].invert_yaxis()  # labels read top-to-bottom
-            ax[i][1].set_xlabel(r'Sensitivity: $\frac{\partial\:\ln{k}}{\partial\:E0}$, ($\frac{J}{mol}$)')
+            ax[i][1].set_xlabel(r'Sensitivity: $\frac{\partial\:\ln{k}}{\partial\:E0}$, ($\frac{mol}{J}$)')
             ax[i][1].set_title('Reverse, {0}'.format(condition))
             ax[i][1].set_xlim([min_sa, max_sa])
             plt.ticklabel_format(style='sci', axis='x', scilimits=(0, 0))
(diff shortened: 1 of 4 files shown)


arkane/sensitivity.py
```diff
@@ -404,11 +480,25 @@ def save(self, wells, transition_states):
         path = os.path.join(self.output_directory, filename)
         sa_data = dict()
         sa_data['structures'] = dict()
+        # Record which TS rows carry the combined-E0+Ea (ILT) semantics rather than a plain
+        # dln(r)/dE0(TS), so a downstream consumer of the YAML can tell them apart without having to
+        # re-derive can_tst() per reaction. Labels match the '(TS) <label>' keys used below.
+        ilt_ts_labels = []
+        for rxn in self.job.network.path_reactions:
+            if rxn.transition_state is not None and not rxn.can_tst():
```
calvinp0: Whilst this gate is `not rxn.can_tst()`, `perturb()` at lime 436 also requires `isinstance(kinetics,Arrhenius)` before it actually shifts to `Ea`. When they disagree, the YAML lists a ts under `ilt_trainsition_stats` and the header promises a combined `E0`+`Ea` derivative, whilst its `Ea` is never touched.

arkane/sensitivity.py
```diff
@@ -404,11 +480,25 @@ def save(self, wells, transition_states):
...
+                ts_label = '(TS) ' + rxn.transition_state.label
+                if ts_label not in ilt_ts_labels:
+                    ilt_ts_labels.append(ts_label)
+        sa_data['metadata'] = {'ilt_transition_states': ilt_ts_labels}
```
calvinp0: `sa_coefficients.yml` previously had on `structures` plus reaction strings. Anything looping over th etop-level keys now hits a dict it cannot parse.

arkane/sensitivity.py
```diff
@@ -474,7 +564,11 @@ def plot(self, wells, transition_states):
                 axis.set_yticks(y_pos)
                 axis.set_yticklabels(labels)
                 axis.invert_yaxis()  # labels read top-to-bottom
-                axis.set_xlabel(r'Sensitivity: $\frac{\partial\:\ln{k}}{\partial\:E0}$, ($\frac{J}{mol}$)')
+                # A generic label: for wells and RRKM-based TS rows this is dln(k)/dE0, but for a
+                # TS row belonging to an ILT-based path reaction it is a derivative along the
+                # coordinate that raises E0 and Ea together, not a plain dln(k)/dE0(TS) (see save()).
+                # The coefficient has units of mol/J (the reciprocal of the perturbed energy).
+                axis.set_xlabel(r'Sensitivity coefficient ($\frac{mol}{J}$)')
```
calvinp0, line 640: two different derivatives now share one bar chart and the label was made generic so nothing in the figure distinguishes them. also check `KineticSensitivity.plot()` for the wrong label units that you corrected here.

arkane/sensitivity.py
```diff
@@ -380,12 +390,78 @@ def perturb(self, entry, unperturb=False):
...
+                    kinetics = rxn.kinetics if rxn.network_kinetics is None else rxn.network_kinetics
+                    if isinstance(kinetics, Arrhenius):
+                        if id(kinetics) in perturbed_kinetics:
+                            # Several of this TS's path reactions share one Arrhenius object: it is a
+                            # single barrier, already shifted once to match the single E0 shift above,
+                            # so do not shift it again (that would move Ea by 2x the E0 shift).
+                            continue
+                        perturbed_kinetics.add(id(kinetics))
+                        kinetics.Ea = quantity.Energy(kinetics.Ea.value_si + perturbation, 'J/mol')
+                    else:
```
calvinp0: two things:
1. this will fire twice due to lack of aliasing guard like you have for `if not unperturb`
2. also this isn't reachable in a real run - `apply_inverse_laplace_transform_method` has a Cython typed `Arrhenius kinetics` paramter, so a non-arrhenius ILT raises a `TypeError` in `network.py` before sensitivity runs.

arkane/sensitivity.py
```diff
@@ -380,12 +390,78 @@ def perturb(self, entry, unperturb=False):
...
+                # kinetics of an ILT path reaction owned by a *different* TS, that reaction's rate has
+                # been collaterally perturbed and its sensitivity row is contaminated. This is an
+                # unsupported input (distinct reactions should not share one mutable kinetics object),
+                # so surface it loudly rather than silently reporting a corrupted coefficient.
+                for rxn in self.job.network.path_reactions:
+                    if (rxn.transition_state is not None and rxn.transition_state is not entry
+                            and not rxn.can_tst()):
+                        other_kinetics = rxn.kinetics if rxn.network_kinetics is None else rxn.network_kinetics
+                        if id(other_kinetics) in perturbed_kinetics:
+                            logging.warning(
```
calvinp0, line 466: when shared kinetics is detected, the contaminated coefficicents are still written to the table and the yaml with no marker.

arkane/sensitivity.py
alongd: Fixed. `save()`'s `ilt_transition_states` gate now also requires `isinstance(kinetics, Arrhenius)`, matching what `perturb()` actually does — a TS is marked ILT only when its Ea was really shifted, so the header/metadata no longer promise the combined E0+Ea semantics for a row whose Ea was untouched. (In the reachable universe this is a no-op, since non-Arrhenius ILT can't occur — see the thread on line 444 — but it makes the defensive branch honest.) The ILT/contaminated classification is now a single helper, `_classify_ilt_transition_states()`, shared by `save()` and `plot()`, so the table, YAML and figure can't disagree.

arkane/sensitivity.py
alongd: `structures` was already a top-level non-reaction dict, so a consumer blindly iterating top-level keys as reactions was already broken before this PR (reaction keys are distinguishable by containing `" <=> "`). To remove any regression for the common case, `metadata` is now only written when it carries content, so an all-RRKM network's YAML keeps the pre-existing shape (`structures` + reaction strings) byte-for-byte.

arkane/sensitivity.py
alongd: Fixed both. `PDepSensitivity.plot()` now annotates the TS tick labels — `(E0+Ea)` for ILT rows and `(E0+Ea, contaminated)` for shared-kinetics rows — so the two derivative types on the shared axis are distinguishable. And `KineticsSensitivity.plot()`'s inverted units are corrected to `mol/J` on both the forward and reverse xlabels.

arkane/sensitivity.py
alongd: 1. Fixed — the warning is now guarded by `not unperturb`, so it fires once per perturbation instead of on both the perturb and unperturb passes. Added `test_unsupported_kinetics_warning_not_duplicated_on_unperturb`, confirmed red against the unguarded branch first.

2. Confirmed. `apply_inverse_laplace_transform_method` is Cython-typed `Arrhenius kinetics`, and the unperturbed base run in `__init__` executes before any perturbation, so a non-Arrhenius ILT reaction raises `TypeError` there first — the branch is unreachable in a real run. Keeping it as documented defensive-only code (tests pin the warning for `MultiArrhenius`/`Chebyshev`), now with an inline comment stating the unreachability.

arkane/sensitivity.py
alongd: Fixed — contamination is now surfaced in the output, not just the log. Contaminated TS rows are flagged with `(!)` in the text table (with an explanatory legend) and listed under `metadata.contaminated_transition_states` in `sa_coefficients.yml`. Detection is static (an Arrhenius object mapped to more than one distinct TS among ILT path reactions), via the shared `_classify_ilt_transition_states()` helper, with unit tests for both the ILT gate and contamination detection.

alongd:
@calvinp0 addressed all five in `6ce1564` (amended into the tip commit, force-pushed):

- **line 488 / metadata gate** — `save()` now marks a TS ILT only when its kinetics is Arrhenius, matching `perturb()`; no more over-promise.
- **line 492 / top-level key** — `metadata` is only written when non-empty, so all-RRKM YAML is byte-identical to before (`structures` was already a top-level non-reaction key).
- **line 571 / plot** — TS tick labels annotated `(E0+Ea)` / `(E0+Ea, contaminated)`; `KineticsSensitivity.plot()` units fixed to `mol/J`.
- **line 444 / double-fire + unreachable** — warning guarded by `not unperturb` (fires once); the non-Arrhenius branch is confirmed unreachable and kept as documented defensive code.
- **line 460 / contamination** — contaminated rows now marked `(!)` in the table and listed under `metadata.contaminated_transition_states` in the YAML.

Classification is factored into one helper shared by `save()`/`plot()`. Added unit tests (incl. a double-fire regression confirmed red first); `sensitivityTest.py` (16) and `arkanePdepTest.py` (2) pass locally.

PR #2992, opened 2026-07-30
Reject zero rate coefficients and improve collision limit checks

calvinp0:
### Motivation or Problem
An RMG job I ran on an ammonia oxidation model crashed during the final model check:

```python
Making seed mechanism...
Performing final model checks...
/home/supervisor/Code/RMG-Py/rmgpy/rmg/main.py:1488: RuntimeWarning: divide by zero encountered in log
  violator_list = rxn.check_collision_limit_violation(t_min=self.Tmin, t_max=self.Tmax, p_min=self.Pmin, p_max=self.Pmax)
  File "/home/supervisor/Code/RMG-Py/rmg.py", line 4, in <module>
    __main__.main()
  File "/home/supervisor/Code/RMG-Py/rmgpy/__main__.py", line 102, in main
    rmg.execute(**kwargs)
  File "/home/supervisor/Code/RMG-Py/rmgpy/rmg/main.py", line 1244, in execute
    self.check_model()
  File "/home/supervisor/Code/RMG-Py/rmgpy/rmg/main.py", line 1488, in check_model
    violator_list = rxn.check_collision_limit_violation(t_min=self.Tmin, t_max=self.Tmax, p_min=self.Pmin, p_max=self.Pmax)
  File "rmgpy/reaction.py", line 1704, in rmgpy.reaction.Reaction.check_collision_limit_violation
  File "rmgpy/reaction.py", line 1737, in rmgpy.reaction.Reaction.check_collision_limit_violation
  File "rmgpy/kinetics/arrhenius.pyx", line 129, in rmgpy.kinetics.arrhenius.Arrhenius.get_rate_coefficient
  File "rmgpy/kinetics/arrhenius.pyx", line 139, in rmgpy.kinetics.arrhenius.Arrhenius.get_rate_coefficient
TypeError: Cannot convert 'complex' with non-zero imaginary component to 'double' (this most likely comes from the '**' operator; use 'cython.cpow(True)' to return 'nan' instead of a complex number).
```
It appears the issue chain was:

1. `generate_reverse_rate_coefficient()` buils a `Tlist/klist` and fits an Arrhenius to it
2. One forward rate coefficient in that list underflows to exactly `0.0` at the low end of the range
3. `Arrhenius.it_to_data` validates its input against `inf`, `NaN` and mixed signs but not against zero. Zero passes every check and reaches the least-squares step, which is performed in log space, hence `divide by zero encountered in log`
4. `lstsq` returns `NaN` for `A`, `n` and `Ea`. It does not raise. The degenerate expression is returned to the caller as though the fit succeeded.
5. Evaluating that expression raises the `TypeError` above, far from the cause

### Description of Changes
1. Reject zero rate coefficients when fitting Arrhenius in log space
2. Keep collisioon limit checks alive when a rate cannot be evaluated. `check_collision_limit_violation` now evaluates each direction independently and skips a condition it cannot evaluate, with a warning, instead of propogating out of `check_model` and discarding an otherwise-complete run. 
3. Derive the reverse rate derivative from `Keq` instead of `kb/kf`. `compute_rate_derivative` formed the reverse contribution as `(kb/kf)*[products]`. Every reactor builds `kb` as `kf/Keq` so when `kf` underflows to exactly zero that ratio is `0.0/0.0`, which in C evaluates to `NaN` without raising and spreads through the sensitivity matrix. Since `kb≡kf/Keq`, the ratio is identically `1/Keq` so the stored `Keq` is used directly. This is the same value whenever `kf > 0` and stays correct when `kf` underflows, where the true limist is `[products]/Keq` rather than zero.
4. Checks the reverse collision limit when reactants lack transport data. `calculate_col_limit` raises `ValueError` when a species in the requested direction has no transport data; in the forward direction this was handled with `continue`, but then skipps the reverse direction check for that condition even though it depends only on the products
5. Fix argument documentation and tables markup in `canteramodel` - authored by @rwest 

### Testing
7 unit tests. 

·	test/rmgpy/kinetics/arrheniusTest.py::TestArrhenius::test_fit_to_data_with_zero_rate
·	test/rmgpy/reactionTest.py::TestCollisionLimitViolation - 4 tests
·	test/rmgpy/solver/simpleTest.py::SimpleReactorTest::test_compute_rate_derivative_kf_underflow
·	test/rmgpy/rmg/mainTest.py::TestCheckModelCollisionLimits::test_failing_reaction_does_not_abort_the_check

### DIFF
--- a/rmgpy/kinetics/arrhenius.pyx
+++ b/rmgpy/kinetics/arrhenius.pyx
@@ -158,6 +158,8 @@ cdef class Arrhenius(KineticsModel):
         import scipy.stats
         if not all(np.isfinite(klist)):
             raise  ValueError("Rates must all be finite, not inf or NaN")
+        if any(klist==0):
+            raise ValueError("Rates must all be nonzero; a zero rate coefficient cannot be fit in log space.")
         if any(klist<0):
             if not all(klist<0):
                 raise ValueError("Rates must all be positive or all be negative.")
@@ -1377,6 +1379,8 @@ cdef class ArrheniusChargeTransfer(KineticsModel):
         import scipy.stats
         if not all(np.isfinite(klist)):
             raise  ValueError("Rates must all be finite, not inf or NaN")
+        if any(klist==0):
+            raise ValueError("Rates must all be nonzero; a zero rate coefficient cannot be fit in log space.")
         if any(klist<0):
             if not all(klist<0):
                 raise ValueError("Rates must all be positive or all be negative.")
--- a/rmgpy/reaction.py
+++ b/rmgpy/reaction.py
@@ -1142,8 +1142,9 @@ class Reaction:
         klist = np.zeros_like(Tlist)
         for i in range(len(Tlist)):
             klist[i] = kf.get_rate_coefficient(Tlist[i]) / self.get_equilibrium_constant(Tlist[i])
+        Tfit, kfit = _drop_zero_rate_samples(Tlist, klist)
         kr = Arrhenius()
-        kr.fit_to_data(Tlist, klist, reverse_units, kf.T0.value_si)
+        kr.fit_to_data(Tfit, kfit, reverse_units, kf.T0.value_si)
         kr.solute = kf.solute
         return kr
 
@@ -1166,8 +1167,9 @@ class Reaction:
         klist = np.zeros_like(Tlist)
         for i in range(len(Tlist)):
             klist[i] = kf.get_rate_coefficient(Tlist[i]) / self.get_equilibrium_constant(Tlist[i])
+        Tfit, kfit = _drop_zero_rate_samples(Tlist, klist)
         kr = SurfaceArrhenius()
-        kr.fit_to_data(Tlist, klist, reverse_units, kf.T0.value_si)
+        kr.fit_to_data(Tfit, kfit, reverse_units, kf.T0.value_si)
         kr.solute = kf.solute
         return kr
 
@@ -1193,8 +1195,9 @@ class Reaction:
             klist[i] = \
                 self.get_surface_rate_coefficient(Tlist[i], surface_site_density=surface_site_density) / \
                 self.get_equilibrium_constant(Tlist[i], surface_site_density=surface_site_density)
+        Tfit, kfit = _drop_zero_rate_samples(Tlist, klist)
         kr = SurfaceArrhenius()
-        kr.fit_to_data(Tlist, klist, reverse_units, kf.T0.value_si)
+        kr.fit_to_data(Tfit, kfit, reverse_units, kf.T0.value_si)
         kr.solute = kf.solute
         return kr
 
@@ -1219,8 +1222,9 @@ class Reaction:
         klist = np.zeros_like(Tlist)
         for i in range(len(Tlist)):
             klist[i] = kf.get_rate_coefficient(Tlist[i],V0) / self.get_equilibrium_constant(Tlist[i],V0)
+        Tfit, kfit = _drop_zero_rate_samples(Tlist, klist)
         kr = SurfaceChargeTransfer(alpha=kf.alpha.value, electrons=-1*self.electrons, V0=(V0,'V'))
-        kr.fit_to_data(Tlist, klist, reverse_units, kf.T0.value_si)
+        kr.fit_to_data(Tfit, kfit, reverse_units, kf.T0.value_si)
         kr.solute = kf.solute
         return kr
 
@@ -1243,8 +1247,9 @@ class Reaction:
         klist = np.zeros_like(Tlist)
         for i in range(len(Tlist)):
             klist[i] = kf.get_rate_coefficient(Tlist[i],V0) / self.get_equilibrium_constant(Tlist[i],V0)
+        Tfit, kfit = _drop_zero_rate_samples(Tlist, klist)
         kr = ArrheniusChargeTransfer(alpha=kf.alpha.value, electrons=-1*self.electrons, V0=(V0,'V'))
-        kr.fit_to_data(Tlist, klist, reverse_units, kf.T0.value_si)
+        kr.fit_to_data(Tfit, kfit, reverse_units, kf.T0.value_si)
         kr.solute = kf.solute
         return kr
 
@@ -1737,7 +1742,9 @@ class Reaction:
         """
         Warn if a core reaction violates the collision limit rate in either the forward or reverse direction
         at the relevant extreme T/P conditions. Assuming a monotonic behaviour of the kinetics.
-        Returns a list with the reaction object and the direction in which the violation was detected.
+        Returns ``(violator_list, skipped)``, where `violator_list` holds the reaction object and the
+        direction in which each violation was detected, and `skipped` counts the direction/condition
+        pairs that could not be evaluated.
         """
         conditions = [[t_min, p_min]]
         if t_min != t_max:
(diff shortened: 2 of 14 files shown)


rmgpy/kinetics/arrhenius.pyx
```diff
@@ -158,6 +158,8 @@ cdef class Arrhenius(KineticsModel):
         import scipy.stats
         if not all(np.isfinite(klist)):
             raise  ValueError("Rates must all be finite, not inf or NaN")
+        if any(klist==0):
```
alongd: I'd rather we filtered the bad samples where klist is built than raise here. The five reverse_*_rate helpers (reaction.py:1127, 1150, 1174, 1201, 1227) each build `klist[i]`.

rmgpy/kinetics/arrhenius.pyx
```diff
@@ -1377,6 +1379,8 @@ cdef class ArrheniusChargeTransfer(KineticsModel):
...
```
alongd: The same hole is left open in two places: `SurfaceChargeTransfer.fit_to_data` and `StickingCoefficient.fit_to_data`. Can you add this raise there as well?

rmgpy/reaction.py
```diff
@@ -1748,37 +1748,50 @@ def check_collision_limit_violation(self, t_min, t_max, p_min, p_max):
...
-                    collision_limit_f.append(self.calculate_coll_limit(temp=condition[0], reverse=False))
+                    limit_f = self.calculate_coll_limit(temp=temp, reverse=False)
                 except ValueError:
-                    continue
+                    pass
                 else:
-                    kf_list.append(self.get_rate_coefficient(condition[0], condition[1]))
+                    try:
+                        kf = self.get_rate_coefficient(temp, pressure)
+                    except (ReactionError, KineticsError, TypeError, ValueError,
```
alongd: I'd remove TypeError and ValueError from here, they catch programmatic bugs, not scientific bugs. If you saw them here in the wild, then add a separate exception?

rmgpy/reaction.py
```diff
@@ -1748,37 +1748,50 @@ def check_collision_limit_violation(self, t_min, t_max, p_min, p_max):
...
-                    condition = '{0} K, {1:.1f} bar'.format(conditions[i][0], conditions[i][1] / 1e5)
-                    violator_list.append([self, 'forward', ratio, condition])
-        if len(self.products) >= 2:
-            for i, k in enumerate(kr_list):
-                if k > collision_limit_r[i]:
-                    ratio = k / collision_limit_r[i]
-                    condition = '{0} K, {1:.1f} bar'.format(conditions[i][0], conditions[i][1] / 1e5)
-                    violator_list.append([self, 'reverse', ratio, condition])
+                    try:
+                        kr = self.generate_reverse_rate_coefficient().get_rate_coefficient(temp, pressure)
```
alongd: generate_reverse_rate_coefficient() is inside the condition loop, so it gets rebuilt up to 4x per reaction. The new ValueError makes the same failure now recomputed and re-logged once per condition. Put it above?

rmgpy/reaction.py
```diff
@@ -1748,37 +1748,50 @@ def check_collision_limit_violation(self, t_min, t_max, p_min, p_max):
...
+                    except (ReactionError, KineticsError, TypeError, ValueError,
+                            ZeroDivisionError, OverflowError) as err:
+                        logging.warning(
+                            "Skipping reverse collision limit check for reaction %s at %.1f K, %.3g Pa "
+                            "because reverse rate evaluation failed: %s",
+                            self, temp, pressure, err,
+                        )
+                    else:
+                        reverse_checks.append((kr, limit_r, condition))
+        for direction, checks in (('forward', forward_checks), ('reverse', reverse_checks)):
```
alongd: If every condition gets skipped, check_model still prints "No collision rate violators found in the model's core." (main.py:1695), which isn't true. Could you return a skipped count alongside violator_list so the summary can say how much was actually checked?

rmgpy/solver/base.pyx
```diff
@@ -1355,12 +1355,14 @@ cdef class ReactionSystem(DASx):
             else:  # three reactants
                 fderiv = C[ir[j, 0]] * C[ir[j, 1]] * C[ir[j, 2]]
 
-            if ip[j, 1] == -1:  # only one reactant
-                rderiv = kr[j] / kf[j] * C[ip[j, 0]]
-            elif ip[j, 2] == -1:  # only two reactants
-                rderiv = kr[j] / kf[j] * C[ip[j, 0]] * C[ip[j, 1]]
-            else:  # three reactants
-                rderiv = kr[j] / kf[j] * C[ip[j, 0]] * C[ip[j, 1]] * C[ip[j, 2]]
+            if not (Keq[j] > 0.0):
```
alongd: For a reversible reaction Keq can't be 0 (reaction.py:863 raises), so the only thing that trips this branch is NaN thermo. Worth a comment saying so.
The bigger thing: the irreversible sentinel isn't consistent. simple.pyx:277 sets Keq = np.inf, which passes the check and gives C/inf = 0.0. liquid.pyx:168, surface.pyx:315 and mbSampled.pyx:239 have no else, so Keq stays at the 0.0 from np.zeros_like and trips the check instead. Could we add else: self.kb[j] = 0.0; self.Keq[j] = np.inf to those three so they agree by design?

alongd (commented):
Thanks, @calvinp0! Added a few comments

rmgpy/kinetics/arrhenius.pyx
calvinp0: added a `_drop_zero_rate_samples`
https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2992/changes/77b465f76ea111f6dcef8c9a76703e173fc1ad56#diff-5764fce17aa420267e6d9f5e756cafe5aea5afe6a8b3345ca126f391a2f67d23R1876

rmgpy/kinetics/arrhenius.pyx
calvinp0: added

calvinp0: also, `StickingCoefficient.fit_to_data` has no `isfinite` or sign validation at all unlike the other three, so `inf`/`NaN` rates reach the log fit too. i can make it a separate PR? or fold it into this?

rmgpy/reaction.py
calvinp0: Agreed - narrowed to (ReactionError, KineticsError, ZeroDivisionError, OverflowError)`

rmgpy/reaction.py
calvinp0: true, fixed

rmgpy/reaction.py
calvinp0: true. it now returns `(violator_list, skipped)` and `check_model` tracks skipped check and skipped reactions separately, reporting when either is non-zero. but know this changed cpdef return type from `list` to `tuple`

rmgpy/solver/base.pyx
calvinp0: done - kb = 0.0; Keq = np.inf added to liquid, surface and mbSampled and kept the guard rather than assuming `Keq > 0`

rmgpy/reaction.py
calvinp0: I want to raise though, with this comment and the previous one to extend the zero rate check - together if that check ever fires, we will skip the whole reaction with a traceback rather than one condition with a warning. and the proceeding comments were about preserving. i can instead raise Kinetics Error from that guard instead of Value Error?

rmgpy/reaction.py
calvinp0: since its a public API on the defintion change, people will need to re-compile again but i guess that is expected for every RMG update.

PR #2995, opened 2026-08-17
Fix resonance save order

calvinp0:
### Motivation or Problem
`_generate_resonance_structures(save_order=True)` did not preserve atom oder and at
`keep_isomorphic=True` it also returned the wrong structures, discarding the aromatic ones.

Two separate causes:

1. `_generate_resonance_structures` accepted `save_order` and honoured it in its own  isomorphism checks, but dispatched every generation algorithm as `method(molecule)`. `generate_clar_structures` and `generate_aromatic_resonance_structure` therefore always ran with their own `save_order=False` default, so requesting `save_order=True` still reordered the atoms of any aromatic species. This affected both `keep_isomorphic` settings.
2.  The charge filtration heuristics in `filtration.py` identify atoms by `Vertex.sorting_label`, which the isomorphism machinery leaves either unset or holding a stale permutation that no longer matches the vertex order. Under `save_order` that left the structure list in a mixed state, and since `stabilize_charges_by_proximity` only counts a charged pair when `atom2.sorting_label > atom1.sorting_label`, it measured a real distance for the few labelled structures and zero for the rest, then popped everything above the minimum - which was exactly the aromatic ones.

Measured on `origin/main`, `keep_isomorphic=True`, counting distinct isomorphism classes:

| species | `save_order=False` | `save_order=True` |
|---|---|---|
| 1-nitronaphthalene | 4 distinct, 3 aromatic | **2 distinct, 1 aromatic** |
| 4-nitrophenoxy radical | 5 distinct, 1 aromatic | **4 distinct, 0 aromatic** |

For nitro-substituted naphthalene radicals filtration discarded every candidate and `filter_structures` raised `ResonanceError` outright.

### Description of Changes

- `_generate_resonance_structures` binds `save_order` with `functools.partial` at the two call sites that dispatch a `save_order` algorithm, so the dispatch loop no longer needs to know which algorithm it is calling.
- `charge_filtration`, `find_unique_sites_in_charged_list` and `stabilize_charges_by_proximity` now identify atoms by position in `mol.vertices` instead of `sorting_label`. The label comparison in `stabilize_charges_by_proximity` only existed to visit each charged pair once, so I replace it by `itertools.combinations`, with a guard for the disconnected case where `find_shortest_path` returns `None`.
 
### Testing

Added new tests:
- a polycyclic case (1-naphthol) for atom order preservation, which reaches  `generate_clar_structures` - the existing tests 1 and 2 cover an acyclic radical and a monocyclic aromatic and passed while `save_order` was being dropped;
- a non-polycyclic case covering `generate_aromatic_resonance_structure`, the other dispatch site;
- `save_order` does not change which structures come back for 1-nitronaphthalene, at `keep_isomorphic=True` (4 distinct, 3 aromatic) and on the default path (4 structures, 3 reactive, matching the resonance structure page on the website);
- `test_charge_filtration_independent_of_atom_order` in `filtrationTest.py`, which fails without the filtration fix;
- `test_charge_filtration_of_disconnected_ions`, covering salts at both `save_order` settings, which raises `ResonanceError` without the `distances[1]` correction.

### DIFF
--- a/rmgpy/molecule/filtration.py
+++ b/rmgpy/molecule/filtration.py
@@ -249,7 +249,7 @@ def charge_filtration(filtered_list, charge_span_list):
                 for atom2, bond in atom.edges.items():
                     index2 = indices[id(atom2)]
                     # check if bond is multiple, store only from one side (atom1 < atom2) for consistency
-                    if index2 > index1 and bond.is_double() or bond.is_triple():
+                    if index2 > index1 and (bond.is_double() or bond.is_triple()):
                         mul_bond_sorting_list.append((index1, index2))
         # Find unique radical and multiple bond sites in charged_list and append to unique_charged_list:
         unique_charged_list = []
--- a/rmgpy/molecule/filtration.py
+++ b/rmgpy/molecule/filtration.py
@@ -43,6 +43,7 @@ The rules this module follows are (by order of importance):
 which is quite like http://www.chem.ucla.edu/~harding/IGOC/R/resonance_contributor_preference_rules.html)
 """
 
+import itertools
 import logging
 
 from rmgpy.exceptions import ResonanceError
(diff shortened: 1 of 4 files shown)


mjohnson541 (approved):
LGTM

PR #3006, opened 2026-09-02
Further modifications to the CO2RR example

ssun30:
### Motivation or Problem
One more round of validation run required changes to some parameters to re-produce the mechanism in the RMG 4.0 paper. 

This also requires a [companion PR](https://github.com/ReactionMechanismGenerator/RMG-database/pull/746) where the HCOOH and CH4 species were incorrectly imported into our stat thermo workflow. 

### Description of Changes

CO2 concentration is now changed to 33.0 mol/m^3 (saturation concentration in water). 

The surface to volume ratio was changed to reflect the correct boundary layer thickness (100 μm) used in our reactor setup.

### DIFF
--- a/examples/rmg/CO2RR/input.py
+++ b/examples/rmg/CO2RR/input.py
@@ -7,16 +7,16 @@
 
 # Data sources
 database(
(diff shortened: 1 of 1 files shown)


rwest (approved):
Looking at the diff, I approve. But can you also share or describe the results?

JacksonBurns (approved):
confirming that these are reflected in the manuscript's SI

